PR #45, opened 2026-05-04
chore: add Nix flake for reproducible dev environment

baseballyama:
## Summary

Add a `flake.nix` that pins the full toolchain (Rust stable via fenix, Node 22, pnpm, `wasm-pack`) so every contributor — and eventually CI — runs against the same compiler builds.

- `nix develop` drops you into a shell with everything ready.
- direnv users get auto-activation via `.envrc` (`use flake`).
- Toolchain components: cargo, clippy, rust-analyzer, rust-src, rustc, rustfmt, plus the `wasm32-unknown-unknown` rust-std for the optional `wasm` feature.
- macOS: pulls in `SystemConfiguration` / `CoreServices` frameworks for native crates.

## Follow-up

`flake.lock` is **not** included in this commit — Nix isn't available in the environment that authored this branch. The first contributor with Nix should:

```bash
nix develop                # creates flake.lock
git add flake.lock
git commit -m "chore: lock Nix flake inputs"
```

Once that's in, every contributor and CI run uses byte-for-byte the same toolchain.

## Test plan

- [ ] Run `nix develop` on Linux / macOS — shell starts, `rustc --version` / `node --version` / `pnpm --version` all print.
- [ ] Inside the shell: `cargo build --release` succeeds.
- [ ] Inside the shell: `pnpm install && pnpm run generate-fixtures && cargo test` matches the existing CI baseline.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,8 @@ language-tools
 compat-targets/.cache/
 compat-targets/*/.compat-meta.json
 compat-targets/*/.rsvelte/
+
+# Nix / direnv
+.direnv/
+result
+result-*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,34 @@ debug a failure. For the higher-level project goals, see [`AGENTS.md`](AGENTS.md
 
 ## Environment setup
 
-### Native (recommended for daily work)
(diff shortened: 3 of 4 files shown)


baseballyama:
Closing per docs/wave-2-3-handover.md guidance: this PR can't be merged as-is because flake.lock was never generated, so every CI job fails at the dependency-resolution step. The branch is preserved (not deleting) so whoever picks up Nix support can re-open after running 'nix develop' locally to materialise flake.lock, commit it, push, and re-run CI.

PR #201, opened 2026-05-18
perf(transform/client): AST-based prop member-expression mutations

baseballyama:
## Summary

Phase A continuation, sibling to #198 (prop compound + simple). Migrates the member-mutation branch in `state_transforms.rs::transform_prop_assignments` (lines 2501+) to an OXC-AST visitor.

Only bindable props get this wrap — non-bindable props just get read transforms elsewhere. The text version's `if !non_bindable_prop_vars.contains(var)` guard becomes a single early-return in the visitor.

| Source | Replacement |
|---|---|
| `prop.foo = x` | `prop(prop().foo = x, true)` |
| `prop().foo = x` | `prop(prop().foo = x, true)` (idempotent) |
| `prop[i] = x` | `prop(prop()[i] = x, true)` |
| `prop.foo += x` | `prop(prop().foo += x, true)` |

Both `prop.foo` (bare Identifier root) and `prop().foo` (CallExpression root with no args) collapse to the same output via a Root enum.

## Idempotency

A naive AST visitor would re-wrap on the next pass because the inner `prop()` is still prop-rooted. The visitor instead detects the wrap shape via `visit_call_expression`: `prop(<assignment>, true)` with bindable prop callee triggers a skip marker on the inner assignment's span. `visit_assignment_expression` then bails on that span.

The text loop below also has a `before.ends_with(\"prop(prop().\")` guard as backup.

## Test plan

- [x] 22 unit tests
- [x] cargo fmt + cargo clippy --all-targets --all-features -D warnings clean
- [x] cargo build --release -D warnings green
- [ ] Compatibility Report on CI

### DIFF
--- a/src/compiler/phases/3_transform/client/mod.rs
+++ b/src/compiler/phases/3_transform/client/mod.rs
@@ -14,6 +14,7 @@ mod effect_rune_ast;
 mod expression_utils;
 mod formatting;
 mod prop_assign_ast;
+mod prop_member_mutate_ast;
 mod props_transforms;
 mod reactive_transforms;
 mod reactive_update_ast;
--- a/src/compiler/phases/3_transform/client/prop_member_mutate_ast.rs
+++ b/src/compiler/phases/3_transform/client/prop_member_mutate_ast.rs
@@ -59,7 +59,6 @@ use oxc_ast_visit::walk;
 use oxc_parser::Parser;
 use oxc_span::SourceType;
 use oxc_span::Span;
-use oxc_syntax::operator::AssignmentOperator;
 
 thread_local! {
     static MODULE_PROP_MEMBER_MUTATE_ALLOC: RefCell<Allocator> = RefCell::new(Allocator::default());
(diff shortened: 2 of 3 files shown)


baseballyama:
Closing — runtime-legacy fixtures regressed (3 of 1198: after-render-triggers-update, mutation-correct-return-value, action-receives-element-mounted). Will revisit once I can reproduce locally with fixture generation. The text-based transform_prop_assignments member-mutation branch remains untouched on main.

baseballyama:
Reopened with text-loop bypass fix (4d0291a). Root cause: the AST member-mutation helper emits the same wrap shape as the text member-mutation loop. The text loop's in-loop idempotency guard only works WITHIN a single text-loop iteration; when the AST helper runs first and the text loop starts fresh on the rewritten text, the guard fails and double-wraps. Fix: skip the text member-mutation branch entirely when the AST helper has rewritten (it's a complete replacement, not a pre-pass). Local runtime-legacy: 1201/1201 pass.

PR #281, opened 2026-05-24
docs(npm): rewrite published-package READMEs from a user perspective

baseballyama:
## Summary

- The README files for the published `@rsvelte` packages (`@rsvelte/svelte-check`, `@rsvelte/svelte2tsx`, `@rsvelte/vite-plugin-svelte-native`) described internal release plumbing — "private: true until X is wired up", changesets anchor notes, distribution-channel decisions. They were written when the packages were still in-repo only and are not useful to anyone who lands on the npm page now that the packages actually publish.
- Rewrite the three published-package READMEs as user-facing docs: install, quick start, API / CLI reference, supported platforms, compatibility status, and pointers to the right package for the user's situation (e.g. SvelteKit users want the JS-side Vite plugin fork, not the native bindings directly).
- Add minimal "this is a platform binary, install the loader package instead" READMEs to the per-triple `svelte-check-*` and `vite-plugin-svelte-native-*` packages, which previously shipped with no README at all and showed up on npm as bare placeholders.

`@rsvelte/compiler`'s npm README comes from `pkg/README.md` (regenerated by `wasm-pack` from the repo-root `README.md` on each build) and is already user-facing, so no change is needed there. `npm/compiler/README.md` itself is a workspace-internal anchor that never ships to npm — left as-is.

## Test plan

- [ ] Confirm the three rewritten READMEs render readably on GitHub.
- [ ] Spot-check that the documented CLI flags for `@rsvelte/svelte-check` match `src/bin/svelte_check.rs`.
- [ ] Spot-check that the documented `@rsvelte/svelte2tsx` API matches `npm/svelte2tsx/index.d.ts`.
- [ ] Spot-check that the documented `@rsvelte/vite-plugin-svelte-native` exports match `npm/vite-plugin-svelte-native/index.d.ts`.
- [ ] After the next release, verify the new READMEs appear on npmjs.com for each package.

### DIFF
--- a/docs/static/benchmark-results.json
+++ b/docs/static/benchmark-results.json
@@ -1,75 +1,82 @@
 {
-  "generatedAt": "2026-05-21T01:51:35.233Z",
-  "commitSha": "2bf88d6",
+  "generatedAt": "2026-05-21T04:54:28.795Z",
+  "commitSha": "fa30f1e",
+  "runner": {
+    "label": "local",
+    "os": "darwin",
+    "arch": "arm64",
+    "cpus": 10,
+    "cpuModel": "Apple M1 Pro"
+  },
   "testFilesCount": 3637,
   "javascript": {
-    "durationMs": 2397.758622000001,
-    "throughputFilesPerSec": 1516.8332486137956
+    "durationMs": 2567.3257223333335,
+    "throughputFilesPerSec": 1416.649226999714
   },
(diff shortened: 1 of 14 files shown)


baseballyama:
Rebased onto main as #282 (the original branch was based off chore/local-benchmark-refresh which made the merge dirty).

PR #327, opened 2026-05-25
feat(capi): GitHub Releases pipeline + prebuilt-binary docs

baseballyama:
## Summary

Follow-up to #297 — wires up actual distribution for the C ABI. Tags of the form `capi-vX.Y.Z` now trigger a release workflow that builds per-OS/arch archives and attaches them to a [GitHub Release](https://github.com/baseballyama/rsvelte/releases).

- **`.github/workflows/release-capi.yml`** — 5-triple build matrix + release job that creates / updates the Release.
- **`crates/rsvelte_capi/Cargo.toml`** — crates.io metadata polish; `publish = false` until the parent crate is published.
- **README + crates/rsvelte_capi/README** — "Download prebuilt binaries" sections, SHA-256 verification snippet, on-disk layout, and a "Cutting a release" walkthrough.

## What downstream users get

```bash
VERSION=0.1.0
TRIPLE=darwin-arm64   # or darwin-x64, linux-x64-gnu, linux-arm64-gnu, win32-x64-msvc

curl -L "https://github.com/baseballyama/rsvelte/releases/download/capi-v${VERSION}/rsvelte_capi-${VERSION}-${TRIPLE}.tar.gz" | tar -xz

# Archive layout:
#   rsvelte_capi-<ver>-<triple>/
#     ├── include/rsvelte.h
#     ├── lib/{librsvelte_capi.dylib|.so | rsvelte_capi.dll + .dll.lib}
#     ├── lib/{librsvelte_capi.a | rsvelte_capi.lib}    (static)
#     ├── README.md
#     ├── LICENSE
#     ├── VERSION
#     └── COMMIT
```

A combined `SHA256SUMS` is attached to every release for verification.

## How releases are cut

1. Bump `crates/rsvelte_capi/Cargo.toml`'s `version`, merge.
2. `git tag capi-v0.1.1 && git push origin capi-v0.1.1`.
3. Workflow builds the matrix, refuses to ship if the tag and Cargo.toml version disagree, then creates the Release with all archives + checksums + an auto-generated download table.

`workflow_dispatch` is enabled for dry-runs against any branch (no Release is created).

## Why not crates.io (yet)

`rsvelte_capi` path-depends on `svelte-compiler-rust`, which is not on crates.io. Publishing the child would require publishing the parent first (a separate decision: it's a long-term namespace + versioning commitment). For now:

- Non-Rust users: GitHub Releases (this PR).
- Rust users: git dep —

  ```toml
  [dependencies]
  rsvelte_capi = { git = "https://github.com/baseballyama/rsvelte", tag = "capi-v0.1.0" }
  ```

The crate's `publish = false` is an explicit gate so `cargo publish` can't fire accidentally before the parent is ready.

## Test plan

- [x] `cargo build -p rsvelte_capi --release` — clean
- [x] Workflow YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Local packaging simulation: built `rsvelte_capi-0.1.0-darwin-arm64.tar.gz` with the expected layout (`include/`, `lib/`, `README.md`, `LICENSE`, `VERSION`, `COMMIT`), 40 MB, sha256 verified
- [ ] First real `capi-v…` tag push will exercise the full matrix end-to-end. The workflow uses the same `actions/upload-artifact@v4` + `gh release` machinery as the existing svelte-check / vps-native release pipeline, so the toolchain quirks are known to work.

## Out of scope

- crates.io publish (blocked on parent crate availability).
- Homebrew tap, apt repo, vcpkg manifest (each is a separate PR with its own maintenance footprint; GitHub Releases covers the universal install path first).

### DIFF
--- a/.github/workflows/release-capi.yml
+++ b/.github/workflows/release-capi.yml
@@ -1,0 +1,243 @@
+name: Release rsvelte_capi
+
+# Cut a versioned release of the C ABI bindings.
+#
+#  - Tag the worktree as `capi-v<MAJOR.MINOR.PATCH>` (must match the
+#    `version` in `crates/rsvelte_capi/Cargo.toml`) and push the tag.
+#    The push event triggers this workflow; on success it creates a
+#    GitHub Release with the per-OS/arch archives attached.
+#  - `workflow_dispatch` lets you do a dry-run end-to-end without
+#    actually publishing — pick a "version" input (defaults to "test")
+#    and run; the build matrix produces artifacts you can inspect.
+#
+# The matrix mirrors the existing svelte-check / vps-native release
+# pipelines (.github/workflows/release.yml) so the toolchain quirks
+# are already known to work: aarch64-apple-darwin natively, intel mac
+# cross-compiled from arm64 runner, linux arm64 with the gcc cross-
+# linker, windows-msvc native.
+
+on:
+  push:
+    tags:
+      - 'capi-v*.*.*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to use for the dry-run archive names (no Release is created)'
+        required: false
+        default: 'test'
+
+permissions:
+  contents: write   # gh release create / upload
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    name: build (${{ matrix.triple }})
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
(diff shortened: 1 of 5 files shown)


baseballyama:
The 3 failing checks (Test ubuntu/macos, Code coverage) all fail on the same `test_compile_if_block_with_derived_server` in `src/compiler/mod.rs:1553`, which is pre-existing on `main` — see run https://github.com/baseballyama/rsvelte/actions/runs/26378792847 (chore(release): version packages, #321) which never touched my files. This PR's diff is workflows + README + Cargo.toml metadata only, no `src/` change. Merging once `capi (windows)` reports green.

PR #336, opened 2026-05-25
fix(transform/server): shadow arrow/fn param positions in derived-read wrap

baseballyama:
## Summary

\`compute_shadow_ranges\` correctly added a body/expression-scope shadow range when it detected an arrow function or a regular function-body parameter list, but the **param identifier's own byte position** lived *before* that range started — between the opening \`(\` and \`=>\` / \`{\`. \`wrap_derived_reads_in_script_inner_with_shadow\` therefore saw the identifier at its declaration site as a normal read and appended \`()\` to it whenever the param happened to share a name with a module-level \`\$derived\` binding:

\`\`\`
RADII.find((radius) => radius.name === effectiveRadius)
                ^^^^^^      ^^^^^^
                param       body read
\`\`\`

In the SSR output the body read correctly stays a property access on the param (it's inside the shadow range), but the param identifier itself got rewritten:

\`\`\`
.find((radius()) => radius.name === effectiveRadius())
       ^^^^^^^^                    ^^^^^^^^^^^^^^^^^^
       WRONG: a derived-style call  OK: real derived
       in a param-list position
\`\`\`

…which fails to parse: \`Expected \`,\` or \`)\` but found \`=>\`\`.

## Fix

In the arrow / function-body branch of \`compute_shadow_ranges\`, also push a single-name-wide shadow range for each param identifier that matches a known derived name. The wrap pass then sees the param position as shadowed and leaves it alone.

Also wires \`wrap_derived_reads_for_template\` to compute per-expression shadow ranges instead of using the bare wrapper, so template-side interpolations get the same protection.

## Background

Surfaced by ecosystem-ci on shadcn-svelte. Real-world cases:
- shadcn-svelte/docs \`design-system-provider.svelte\` — \`RADII.find((radius) => radius.name === effectiveRadius)\` where \`radius\` is itself a module-level \`\$derived\`.
- layerchart \`components/Arc.svelte\` — \`function getOuterRadius(outerRadius, chartRadius) { … }\` where \`outerRadius\` is a \`\$derived\`.

## Verified locally

- \`cargo build --release --features napi --lib\` ✅
- \`cargo test --release --lib\` ✅ (1152 passed)
- Minimal repro: emits valid \`(radius) => radius.name === effectiveRadius()\` (body read still rewritten, param identifier untouched).


baseballyama:
Closing in favor of focused PRs. The parser-test skips are moved to a smaller PR; the wrap-derived-reads improvements need a more careful repro setup (svelte submodule init + the async-derived-title-update fixture).

PR #407, opened 2026-05-26
Escape generated preprocess attribute values

mustafa0x:
## Summary
- Escape string values returned in preprocessor tag attributes before reconstructing `<script>`/`<style>` tags.
- Add a regression test covering ampersands, quotes, and angle brackets.

## Why
Preprocessor-returned attributes were interpolated directly inside double quotes. Values containing `"`, `&`, `<`, or `>` could produce invalid markup or change the reconstructed tag meaning.

## Validation
- `cargo fmt`
- `cargo test test_stringify_tag_attributes_escapes_values`


baseballyama (approved):
Thank you!

Issue #433, opened 2026-05-28
Harden parse/transform arena APIs against UB (correctness review C-001..C-004)


baseballyama:
## Background

The correctness review on #431 flags four related findings (C-001..C-004) about arena memory-safety:

- **C-001:** `ParseArena::get_js_node` / `get_js_children` return `&` references via `&self`, while `alloc_js_node` / `alloc_js_child` also take `&self`. Safe Rust can hold a returned `&JsNode` and then call `alloc_js_node` enough times to force `Vec` reallocation. The SAFETY comment claims the borrow checker prevents this, but two `&self` borrows can coexist — there is no compile-time guarantee.
- **C-002:** `ParseArena::get_js_node_mut(&self) -> &mut JsNode` and `get_js_children_mut(&self) -> &mut [JsNode]` are safe `pub` methods that manufacture `&mut` from `&self` using `UnsafeCell`. A safe caller can hold a `&JsNode` from `get_js_node` *and* a `&mut JsNode` for the same id at the same time.
- **C-003:** `try_get_serialize_arena() -> Option<&'static ParseArena>` and `resolve_js_node_for_serialize(id) -> &'static JsNode` are safe public APIs that fabricate a `'static` lifetime from a thread-local raw pointer. A safe caller can leak the `'static` reference past the guard's lifetime.
- **C-004:** Transform-side `JsArena` (`src/compiler/phases/3_transform/js_ast/arena.rs`) has the same shape: `alloc_expr(&self)` + `get_expr(&self)` + `get_expr_mut(&self)` + `take_expr(&self)` + `reset(&self)`. Initial capacity (256 expr / 64 stmt) does not prevent reallocation on real components.

All four are the same class of issue: arena allocation and reference borrowing both go through `&self`, so the borrow checker cannot enforce non-aliasing or no-reallocation invariants. The current code happens to work because builder patterns don't actually hold references across allocations, but the API is unsound by Rust's rules and a future caller can trigger UB without writing `unsafe`.

## Proposed fix

Pick one of these designs (the same one for both arenas):

1. Allocation requires `&mut self`, reads require `&self` — the borrow checker then prevents the bug structurally. This forces callers to thread `&mut arena` through builders.
2. Keep handle-only public APIs and expose access via closures: `with_js_node(id, |node| ...)` / `with_js_children(range, |slice| ...)`. The closure cannot smuggle the reference out, and we can disallow allocation inside the closure with a guard.
[...]

PR #483, opened 2026-05-28
Harden arena reference storage

mustafa0x:
## Summary
- Store parse and transform arena nodes behind stable boxed allocations so safe shared references survive later handle-table growth.
- Convert mutable/destructive transform arena operations to `unsafe` and update the narrow internal call sites.
- Replace fabricated `'static` serialization resolver helpers with closure-scoped access to the active parse arena.
- Add regression tests that hold arena references/slices while forcing later allocations.

## Issue
Addresses #433.

## Validation
- `cargo fmt`
- `git diff --check`
- `cargo test arena`
- `cargo test --lib`
- `cargo clippy --all-targets --all-features -- -D warnings`


baseballyama (approved):
Thank you!

PR #823, opened 2026-06-08
feat(playground): add lint as a dedicated tool next to format

baseballyama:
Promote the linter from a Compiler output tab to its own runnable tool in the playground tool switcher, placed right after **format** (the `fmt` tool's display label is renamed to `format`).

- New `lint` tool (runnable, wasm) listing diagnostics with a live finding count.
- `fmt` tool label → `format` (id unchanged).
- Removed the lint tab from the Compiler output; dropped `'lint'` from the `OutputTab` union.

Verified: `oxlint` clean on .ts; svelte-check shows no new errors from these changes (only the repo's pre-existing type-noise); the rsvelte_lint wasm (compiler + svelte2tsx + lint) builds and the playground type-checks against it.

### DIFF
--- a/apps/playground/src/lib/compiler.ts
+++ b/apps/playground/src/lib/compiler.ts
@@ -88,7 +88,7 @@ export function lint(source: string, filename = 'Component.svelte'): LintDiagnos
 }
 
 export type CompileMode = 'client' | 'server';
-export type OutputTab = 'result' | 'js' | 'css' | 'ast' | 'lint';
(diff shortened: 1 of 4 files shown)


baseballyama:
Note: this PR also includes a one-line `fix(fmt)` removing an unused import in `crates/rsvelte_formatter/src/markup.rs` — `main` was red (`-Dwarnings` fails the unused import), which blocked this PR's CI. The fix unblocks both.

PR #1079, opened 2026-06-17
refactor(phase3): print SSR scripts with rsvelte_esrap, delete oxc_codegen string post-passes

baseballyama:
**First integration of the `rsvelte_esrap` printer into Phase-3 — directly eliminating string surgery, per the project goal.**

The server script normalizer (`normalize_script_with_oxc`) ran `oxc_codegen` followed by a ~700-line tail of string fix-ups whose *only* purpose was to undo ways oxc_codegen diverges from esrap (the printer the official Svelte compiler uses):

- quote restoration (oxc → single quotes)
- numeric-literal restoration (oxc minifies `1000`→`1e3`)
- control-char re-escaping (oxc unescapes `\t`)
- trailing-comment reattachment (oxc drops `x; // c`)
- dangling-comment protection (oxc drops `// c` before `}`)
- long-destructure splitting (oxc doesn't break at esrap's 60-col threshold)

`rsvelte_esrap` **is** that printer (hardened this milestone to ~99% byte-exact on 5,536 real components, 72/72 on the snapshot oracle), so it emits esrap's layout directly. The normalizer now parses the already-transformed script and prints it with `rsvelte_esrap::print`, and all six post-passes + their helpers are deleted.

Only the genuinely-orthogonal fix-ups remain:
- the `;;` (`$inspect`-removal) placeholder dance — both oxc *and* esrap drop empty statements;
- the `$$C$$` in-call-comment decode — that smuggling happens upstream in the export-let lowering, not in codegen.

### Impact
**Net −707 lines** of string-surgery code removed from Phase-3 SSR.

### Verification
- `cargo test` SSR (17), snapshot/compiler_fixtures (16), runtime (19) — all green.
- Corpus-compat CI (6,400+ entries, byte-equal after oxfmt normalization) gates this PR.


baseballyama:
### Corpus CI: at parity with main (no new regressions)

The corpus job reports **1 failure** — `compiler-errors/samples/const-tag-snippet-invalid-reference-1` (`js-mismatch`). This is **pre-existing on main**: the latest `main` corpus run (commit `4cb38dff`) fails on the *exact same single file* with the identical diff. It is an orthogonal snippet/`{@const}` placement quirk on a pathological *invalid-input* sample (duplicate `failed` snippet names), reproducible on `main` locally too — unrelated to this PR's SSR-script normalization change.

All **18 other checks pass** (Clippy, Format, every Test shard, runtime, thread-safety, Compatibility Report, capi, Formatter/Lint parity, fmt corpus, …). This PR introduces **zero new corpus failures** and is at parity with `main`. The pre-existing const-tag-snippet divergence is tracked as separate tech debt.

PR #1103, opened 2026-06-21
fix(fmt): formatter-parity corpus 295→0 (<pre> nested-inline byte-parity)

baseballyama:
Closes the final 2 formatter-parity corpus failures, bringing the unified corpus (sveltejs/svelte + svelte.dev + bits-ui/flowbite/melt/shadcn, ~9,715 components) to **0 known failures, 0 regressions**, with 22 documented principled exclusions.

The 2 cases (`code-viewer.svelte`, `theme-customizer-code.svelte`) are deeply-nested inline elements inside `<pre><code>`. They are handled by extending the **existing string-based `<pre>` verbatim re-indent subsystem** (`reformat_pre_inner` + `collect_pre_*`/`fix_pre_*`), adding `collapse_text_only_spans`, `fix_pre_packed_span_siblings`, and `fix_pre_overflow_close_suffix` (fast-path-guarded; 0 regressions across the 44+ passing `<pre>` files).

**Honest quality note:** `<pre>` content is whitespace-verbatim, so prettier itself prints it via `literalline`/raw rather than Doc-IR element layout, and rsvelte mirrors that with string-level re-indentation — this is the documented exception to the "no string-surgery for layout" rule (which governs ordinary Doc-IR element layout). A faithful Doc-IR `isPreTagContent`/`printPre` printer that would retire these bespoke span helpers is tracked as a dedicated, benchmark-gated follow-up in `docs/corpus-fmt-remaining-work.md` (four independent attempts confirmed the string-based pre path is the only one that reaches parity without regressing the existing `<pre>` corpus).

### DIFF
--- a/.changeset/fmt-corpus-pre-nested-inline.md
+++ b/.changeset/fmt-corpus-pre-nested-inline.md
@@ -1,0 +1,14 @@
+---
+"@rsvelte/fmt": patch
+---
+
+Reach 0 known failures on the formatter-parity corpus (was 2): byte-parity for
+deeply-nested inline elements inside `<pre><code>` (`code-viewer`,
+`theme-customizer-code`). Extends the existing string-based `<pre>` verbatim
+re-indent subsystem (`reformat_pre_inner` and helpers) to collapse text-only
+`<span>`s and correctly pack/unpack/overflow-split sibling spans at the print
+width, with fast-path guards and 0 regressions across the passing `<pre>`
+corpus. (`<pre>` content is whitespace-verbatim — handled by string-level
+re-indentation by design, the documented exception to the Doc-IR element-layout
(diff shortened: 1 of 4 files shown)


baseballyama:
Abandoning: this +531-line change is string-surgery in the <pre> path AND, validated on Linux CI (the authoritative oracle), it only fixes code-viewer — theme-customizer-code still fails there because the Linux oxfmt oracle attr-wraps `<ColorIndicator color={value} />` inside <pre> while the macOS oracle keeps it inline (a cross-platform oracle difference that masks it from local iteration). Poor quality/coverage trade. main stays at the clean 295→2; the final 2 need the Doc-IR isPreTagContent/printPre port, iterated via Linux CI (documented in docs/corpus-fmt-remaining-work.md).

PR #1184, opened 2026-06-25
feat(fmt): format .json/.jsonc/.json5 in-process via oxc_formatter_json

baseballyama:
## What

JSON was the last common file type still delegated to an `oxfmt` subprocess. Since `oxfmt` formats JSON with **`oxc_formatter_json`** — a published oxc crate at the same rev this repo already pins — formatting it in-process is **byte-identical**, and a standalone JSON file now formats **instantly on save** like `.ts`/`.js`/`.svelte` (after #1177/#1179/#1180).

| single `.json` (format-on-save) | time |
|---|---|
| oxfmt subprocess (before) | ~0.35s |
| in-process (after) | **~0ms** |

## How

- **`rsvelte_formatter::format_json_source`** wraps `oxc_formatter_json::format`, picking the variant by extension (`json`/`jsonc`/`json5`). `Expand::Auto` (= Prettier `objectWrap: preserve`) and default bracket-spacing match oxfmt.
- **`run_native_json`** mirrors `run_native_js`: formats each file in parallel, falling back to a single `oxfmt` invocation for the cases that would diverge —
  - **`package.json`** (oxfmt's `sortPackageJson` key-ordering pass isn't in oxc),
  - files an `.oxfmtrc` override touches,
  - base `printWidth` > 320 (native max),
  - parse errors,
  
  so coverage never regresses. Gated by the existing `--no-native-js` escape hatch.
- The oxfmt directory delegation excludes `!**/*.json{,c,5}`; `package.json` is re-delegated as explicit paths by the native-JSON fallback.

## Parity

**243/243 byte-identical** (`oxc_formatter_json` vs `oxfmt`) across flyle's non-`package.json` JSON corpus (incl. comment-bearing `tsconfig.json`). The integrated `rsvelte-fmt --check .` introduces no new diffs — the only JSON it flags are files `oxfmt` would also reformat. cli tests cover the native path (`native_json_formatted_in_process`) and `package.json` delegation (`package_json_delegated_to_oxfmt`).

## Scope note

This speeds up **single-file / format-on-save** JSON (the common case). A full-repo run isn't faster: `package.json` (which must stay on oxfmt) is delegated as a separate concurrent oxfmt call, so oxfmt's total work doesn't drop enough to win there. That's an accepted trade — the value is instant JSON on save.

### DIFF
--- a/.changeset/fmt-native-json.md
+++ b/.changeset/fmt-native-json.md
@@ -1,0 +1,16 @@
+---
+"@rsvelte/fmt": patch
+---
+
+fmt: format `.json` / `.jsonc` / `.json5` in-process via `oxc_formatter_json`
+instead of delegating them to an `oxfmt` subprocess. It's the same engine `oxfmt`
+uses for JSON, so the output is byte-identical (verified 243/243 on a real-world
+corpus) while skipping the per-invocation `oxfmt` startup — a standalone JSON
+file now formats instantly on save, like `.ts`/`.js`/`.svelte` already do.
+
+`package.json` keeps going to `oxfmt`: it additionally runs through
+`sortPackageJson` (a key-ordering pass that lives in oxfmt, not oxc), so
+formatting it natively would diverge. Files matched by an `.oxfmtrc` override, or
+any JSON when the base `printWidth` exceeds the native max (320), also fall back
+to `oxfmt`, as do parse errors — so coverage never regresses. The native JSON
+path is gated by the same `--no-native-js` escape hatch.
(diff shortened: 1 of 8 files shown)


baseballyama:
Rebased onto `main` after #1182 bumped the oxc rev to `7537c58`. `oxc_formatter_json` is now pinned to the same rev (kept in sync with `oxc_formatter`), and JSON byte-parity was **re-verified at 7537c58**: 243/243 identical vs oxfmt on flyle's non-`package.json` corpus at the project's `printWidth: 100`.

Note: there is no JSON byte-parity gate in CI yet (the corpus track covers Svelte markup only), so a future oxc-rev bump could silently shift JSON output away from oxfmt. Worth adding a JSON oracle to the fmt-corpus track as a follow-up.

PR #1263, opened 2026-06-26
fix(compiler/csr): lower component bind write to legacy `$:` var via $.set

### LINKED ISSUE #1228: compiler(CSR): bind write lowered to plain assignment instead of `$.set(...)` (reactivity loss)
## Track
Compile (CSR) — awesome-svelte compat corpus (PR #1219)

## Summary
A bind-target write is lowered to a **plain assignment** `path = $$value;` instead of the reactive `$.set(path, $$value);`. The bound state therefore loses reactivity (writes don't notify subscribers).

## Repro
`smelte/src/routes/_layout.svelte` (and `svelte-calendar/src/lib/components/calendar/DayPicker.svelte`).

## Official vs rsvelte (client)
```diff
-          $.set(path, $$value);
+          path = $$value;
```

## Expected
Writes to a reactive (`$state` / mutable source) bind target must go through `$.set(...)`.

baseballyama:
## What

Fixes #1228 — a `bind:` on a **component** whose target is a legacy `$:` reactive declaration was lowered to a **plain assignment** in the setter, dropping reactivity.

```diff
  set selected($$value) {
-   path = $$value;       // ❌ reactivity lost — subscribers never notified
+   $.set(path, $$value); // ✅
  }
```

The getter was already correct (`$.get(path)`).

## Root cause

In `process_bind_directive` (client `shared/component.rs`), the `is_state_binding` predicate that selects the `$.get`/`$.set` accessor form matched only `is_state_source(...) || Derived` and **omitted `LegacyReactive`** (the kind given to `$:` reactive declarations like `$: path = page.path`).

Upstream builds the component-bind setter by visiting `expr = $$value` through the normal assignment visitor, which applies `transform.assign`. That transform is installed for state sources, `derived`, **and `legacy_reactive`** bindings (`add_state_transformers` gates on exactly those three) — all of which lower the write to `$.set(...)`. The getter happened to stay correct because the non-state branch falls back to the read-transformed expression (`$.get(path)`), masking the asymmetry.

The fix adds `LegacyReactive` to the predicate, restoring parity with upstream and with the element-bind path (which already routes through `apply_transforms_to_expression` and handled this correctly).

## Repro

- smelte `src/routes/_layout.svelte` — `<Tabs bind:selected={path} />` with `$: path = $page.path`
- svelte-calendar `DayPicker.svelte` — `bind:index={monthIndex}` with `$: monthIndex = …`

Both now byte-match the official compiler (CSR) via `scripts/compat-corpus/one.mjs`. Element binds and runes-mode `$state` binds were already correct and remain so.

## Tests

- New regression test `legacy_reactive_component_bind_uses_set` in `component_bind_getter_setter.rs`.
- Full `compatibility_report` (15/15) and `runtime` (19/19) suites pass — 0 failures, no regressions.

### DIFF
--- a/.changeset/fix-component-bind-legacy-reactive.md
+++ b/.changeset/fix-component-bind-legacy-reactive.md
@@ -1,0 +1,13 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler/csr): lower a component `bind:` write to a legacy `$:` variable via `$.set`
+
+A `bind:` on a component whose target is a legacy `$:` reactive declaration was
+lowered to a plain assignment in the generated setter (`path = $$value`), dropping
+reactivity so subscribers were never notified. The `is_state_binding` predicate
+that selects the `$.get`/`$.set` accessor form omitted the `LegacyReactive`
+binding kind; adding it restores parity with upstream (whose `transform.assign`
+runs for state, `derived`, and `legacy_reactive` bindings) and with the
+element-bind path. Fixes #1228.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/component.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/component.rs
@@ -1486,17 +1486,21 @@ fn process_bind_directive(
     // Check if this is a store member expression (e.g., $store.value)
     let is_store_member = is_store_member_expression(&bind.expression, context);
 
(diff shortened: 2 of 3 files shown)


baseballyama:
Closing as a duplicate — the same fix for #1228 already landed on `main` via #1262 (commit e06d43db, `fix(compiler): lower legacy-reactive component bind writes through $.set`). `origin/main` already has the identical `is_state_source || Derived || LegacyReactive` predicate and a regression test, so this PR is redundant.

PR #1531, opened 2026-07-19
fix(formatter): restore collapse child-breaking passes retired as dead in #1505

baseballyama:
## Problem

The CI **Formatter parity** gate (`fmt-parity` in `corpus-compat.yml`) has been red on every PR since #1505 (`56724fd1`, "refactor(formatter): retire dead collapse passes 1.6-1.95 + children-port sweep") landed. #1503/#1504 were green — #1505 is the only formatter change since, and it merged while the parity gate was already failing.

`node scripts/compat-corpus/fmt-verify.mjs` reported **11 NEW failures (not in baseline)** → exit 1.

## Root cause

#1505 deleted collapse passes 1.6, 1.7, 1.8, 1.9, 1.95 and the final children-port sweep on the premise they were dead code. **They were not dead** — each is load-bearing for real corpus components. Removing them regressed 11 components that had matched the `oxfmt(svelte: true)` oracle byte-for-byte, all in the same family: *the element should break its children / open tag onto their own lines, but rsvelte kept them inline*.

Bisecting the restored passes against the 11 regressions (env-gating each pass individually) shows every pass is required by at least one regression:

| pass | regressions it fixes |
|------|----------------------|
| 1.6 `try_collapse` sweep | card-air, calendar-test, ListGroup |
| 1.7 `hug_mixed` non-ws-prefix | flowbite Microsoft SVG icon (`<clipPath>`/`<rect>` hug) |
| 1.8 block-break non-ws-prefix | svelte print `formatting` (Component block child `<div>`) |
| 1.9 break-inline-open-tag | TextDecoration, Underline, html-tag-script-2 |
| children-port sweep | svelte-maplibre radio labels ×2 |
| (combined) | shadcn docs-sidebar |

Because every pass is needed at whole-pass granularity, the passes cannot be partially kept without leaving regressions. This restores `collapse.rs` verbatim to its pre-#1505 state (`56724fd1~1`), returning the formatter to the last green parity state.

## The 11 regressions (all now fixed)

- `bits-ui/docs/.../homepage/card-air.svelte`
- `bits-ui/tests/.../calendar/calendar-test.svelte`
- `flowbite-svelte/.../icons/Microsoft.svelte`
- `flowbite-svelte/.../forms/radio/ListGroup.svelte`
- `flowbite-svelte/.../typography/text/TextDecoration.svelte`
- `flowbite-svelte/.../typography/text/Underline.svelte`
- `shadcn-svelte/docs/.../docs-sidebar.svelte`
- `svelte-maplibre/.../examples/3d_terrain/+page.svelte`
- `svelte-maplibre/.../examples/overlapping_layer_events/+page.svelte`
- `svelte/.../tests/print/samples/formatting/input.svelte`
- `svelte/.../tests/runtime-browser/samples/html-tag-script-2/main.svelte`

## Baseline

The 5 known-failures that #1505's deletion had incidentally *fixed* (layercake, layerchart, 3× svelte-ux) revert to being known failures — they were a genuine over-break tradeoff of the removed passes (3+ distinct over-break root causes), not separable from the regression fix without new per-shape guards that belong to a dedicated formatter-layout change. The ratchet baseline therefore stays at its checked-in 59; **no baseline shrink, and the gate is honestly green (not padded).**

## Verification

- `node scripts/compat-corpus/fmt-verify.mjs` → **11480 included, 11399 matched, 59 known failures, 0 regressions, exit 0** (reproduces the last-green formatter state exactly across the full corpus).
- `cargo test -p rsvelte_formatter --release` and `cargo test -p rsvelte_fmt --release` green.
- `cargo fmt` (no diff) + `cargo clippy --all-targets --all-features -- -D warnings` clean.

https://claude.ai/code/session_01H9Xs5NonZj1k1pcuamEfj7

### DIFF
--- a/.changeset/fix-fmt-restore-collapse-child-breaking-passes.md
+++ b/.changeset/fix-fmt-restore-collapse-child-breaking-passes.md
@@ -1,0 +1,36 @@
+---
+"@rsvelte/fmt": patch
+---
+
+fix(formatter): restore the collapse child-breaking passes retired as "dead" in #1505
+
+#1505 removed collapse passes 1.6–1.95 and the final children-port sweep on the
+premise that they were dead code. They were not: every one of those passes is
+load-bearing for real corpus components, so their removal regressed the
+formatter-parity corpus — 11 components that had matched the `oxfmt(svelte: true)`
+oracle byte-for-byte began diverging, all in the same family of "the element
+should break its children / open tag onto their own lines but rsvelte keeps them
+inline" (e.g. `<button …> Add Day </button>` staying on one line, a `<Span>` open
+tag not breaking on an overflowing prose line, an SVG `<clipPath>`/`<rect>` pair
+not hugging, a Component block child `<div>` not breaking, a `<script></script>`
+sibling of an `{@html}` not breaking).
+
+Each retired pass is required by at least one of those regressions:
+
+- 1.6 `try_collapse` sweep — inline pure-text elements revealed by pass-1
+  restructuring (card-air / calendar-test / ListGroup).
+- 1.7 `hug_mixed` non-ws-prefix sweep — SVG child hug (flowbite Microsoft icon).
+- 1.8 block-break non-ws-prefix — Component block child (svelte print `formatting`).
+- 1.9 break-inline-open-tag — overflowing inline/component open tags
+  (TextDecoration / Underline / html-tag-script-2).
+- final children-port sweep — faithful prettier-plugin-svelte layout for its
+  gated shapes (svelte-maplibre radio labels).
+
+Restoring `collapse.rs` to its pre-#1505 state returns the formatter to the last
+green parity state and re-fixes all 11 regressions. This intentionally reverts
+#1505 in full: the passes cannot be separated at whole-pass granularity (each is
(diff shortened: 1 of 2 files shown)


baseballyama:
### Note on the CodSpeed check (expected, intended tradeoff)

CodSpeed reports a formatter perf regression (~22% on the `rsvelte_formatter` benches). This is **the direct, intended cost of the fix, not a defect**:

- #1505 deleted collapse passes 1.6–1.95 + the children-port sweep. Its speed-up came precisely from *not doing that work* — which is exactly why it regressed 11 formatter-parity corpus components (elements stopped breaking their children/open tags).
- Restoring those passes to fix the parity regression necessarily re-adds that work, so the formatter benches return to their pre-#1505 timings. Correctness (byte-for-byte oracle parity) cannot be had without doing the layout work these passes perform.

The perf could later be recovered by re-implementing the same layout behavior more cheaply (e.g. fewer full re-parses between passes), but that is a separate optimization effort with its own correctness risk, out of scope for this regression hotfix. Suggest acknowledging the CodSpeed regression (main is not branch-protected, so it does not gate merge) and tracking a perf follow-up.

PR #1598, opened 2026-07-20
fix(corpus): resolve bare identifiers via scope in template-chunk is_defined (svelte-table Example2)

baseballyama:
## What

Burns down one client entry from the compiler-output corpus known-failures ratchet: **`svelte-table/example/Example2.svelte`** (client), taking the client list `10 → 9`.

## Root cause

Upstream `scope.evaluate` resolves an identifier through the scope even on the **built (transformed)** template AST. A non-reactive binding that survives transformation as a bare identifier (reactive reads become `$.get(x)` instead) still evaluates to its binding's initial value.

rsvelte's `is_js_expr_defined` (the transformed-value branch of the template-chunk `?? ''` decision) had **no binding context**, so bare identifiers nested inside a transformed `${cond ? iconAsc : iconDesc}` were always treated as not-defined. The whole conditional was then judged not-defined and wrongly wrapped in `?? ''`, where the official compiler emits it bare:

```diff
-`${$.get(refSortBy) === "id" ? ($.get(refSortOrder) > 0 ? iconAsc : iconDesc) : ""} ID`
+`${($.get(refSortBy) === "id" ? ($.get(refSortOrder) > 0 ? iconAsc : iconDesc) : "") ?? ""} ID`
```

## Fix (upstream-faithful)

- Thread `ComponentContext` into `is_js_expr_defined` and resolve `JsExpr::Identifier` through a new shared `identifier_is_defined` helper — the same resolution the original-expression path (`is_expression_defined_json`) already used, so a binding resolves identically whether it appears at the top level or nested in a built expression (mirrors `scope.evaluate` on the transformed AST).
- Recognize a non-updated `let`/`var`/`const x = <primitive literal>` (e.g. the legacy `let iconAsc = "↑"`) as defined, mirroring upstream's `!binding.updated && binding.initial !== null && !is_prop` (a `null` literal stays not-defined).

No new heuristics — the logic mirrors `phases/scope.js`'s `Evaluation`/`is_defined`.

## Verification

- `svelte-table/example/Example2.svelte` is now `astEquivalent` for **both** client and server (only comment-position diffs remain, which the corpus comparison absorbs).
- Byte-exact suites all green with the change: `compiler_fixtures`, `runtime` (1208 legacy + 999 runes + hydration + browser), `ssr`, `css`, `real_world`.
- `cargo fmt` + `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` clean.

## Not addressed (remain in ratchet)

The other 9 client entries + 1 server entry are untouched. Notably `SpatialMenuNavTest` was scoped out — it has **two** independent structural diffs (a `$.set(highlighted, id, true)` proxy-arg and a separate `${cols ?? ''}` divergence), so it needs both fixed to clear.

https://claude.ai/code/session_014EizyABnBLJmcSeQYmGwFi

### DIFF
--- a/compat/corpus/known-failures.client.json
+++ b/compat/corpus/known-failures.client.json
@@ -4,7 +4,6 @@
 	"svelte-form-builder/src/lib/Form/PropertyPanel/PropertyPanelComponentSpecific/PropertyPanelChoiceCheckboxRadioSpecific.svelte",
 	"svelte-form-builder/src/lib/Form/PropertyPanel/PropertyPanelComponentSpecific/Table/TableColumnSpecific.svelte",
 	"svelte-sonner/src/lib/Toaster.svelte",
-	"svelte-table/example/Example2.svelte",
 	"svelte-table/src/SvelteTable.svelte",
 	"svelte-ux/packages/svelte-ux/src/lib/components/AppLayout.svelte",
 	"svelte-ux/packages/svelte-ux/src/lib/components/Button.svelte",
--- a/compat/corpus/known-failures.md
+++ b/compat/corpus/known-failures.md
@@ -14,7 +14,7 @@ compiler-behaviour gap; attempts to fix them are regression-prone against the
 8000+ passing entries (several have been reverted after wide blowups), so they are
 accepted until an upstream-faithful port lands.
 
-## Client (`known-failures.client.json`, 10 entries)
+## Client (`known-failures.client.json`, 9 entries)
 
 - **`layercake/.../LayerCake.svelte`** — store-getter declaration **ordering**
   (`const $X = () => $.store_get(...)`) differs, plus a derived store called as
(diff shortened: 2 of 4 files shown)


baseballyama:
Closing: this fix causes 2 confirmed corpus regressions that cannot be cleanly narrowed while preserving the Example2 win.

## Cause (verified by bisect)

The Example2 fix required `is_js_expr_defined` to resolve bare identifiers through the scope (change **a**) *plus* a Literal branch in the shared resolver (change **b**). Bisecting:

- **Pre-change** (both reverted): `svelte-table/Example2` fails; `layerchart/GeoRaster/tiles-globe` and `layerchart/GeoProjection/satellite` **pass**.
- **Post-change**: Example2 passes; both layerchart ids **regress** (`astEquivalent=false`, client).
- **Remove only (b), keep (a)**: layerchart still regresses → the regression is from **(a)**, the identifier-resolution change — which Example2 also depends on.

The concrete regression: `<GeoRaster {sourceProjection} …>` where `const sourceProjection = $derived(…)` flips from the getter form `get sourceProjection() { return $.get(sourceProjection); }` to the plain `sourceProjection: $.get(sourceProjection)`, because `expression_has_reactive_state` returns `false` for the derived post-change.

## Why not narrowable

The mechanism is **untraceable through the code**: `sourceProjection`'s built value is `$.get(sourceProjection)` (a `CallExpression`, not an `Identifier`), so change (a)'s `is_js_expr_defined` Identifier arm cannot affect its `is_defined` result — and the getter decision (`result.has_state`) never consults `is_defined` at all (`has_state = expression_has_reactive_state || has_call || has_blockers`, none of which call the changed functions). Yet (a) provably flips it. This is a subtle/broad interaction of exactly the kind `compat/corpus/known-failures.md` documents as previously "reverted after wide blowups," and it undermines confidence that any local narrowing wouldn't leave other undetected regressions across the 8000-entry corpus.

Example2 is 1 of 10 client known-failures; it is not worth landing a change with an unexplained regression mechanism. Leaving all 10 client / 1 server entries in the ratchet unchanged.

PR #1652, opened 2026-07-21
fix(parse): preserve TypeScript assertion expressions in the parse AST

### LINKED ISSUE #1645: AST fidelity: `x as const` / TS assertion expressions are stripped from the AST (svelte/compiler keeps TSAsExpression)
## Summary

Parsing `pattern={'chips' as const}` in a `<script lang="ts">` component:

- **svelte/compiler** emits `ExpressionTag.expression = TSAsExpression { expression: Literal 'chips', typeAnnotation: TSTypeReference const }`
- **@rsvelte/compiler 0.8.1** emits `ExpressionTag.expression = Literal 'chips'` — the `TSAsExpression` wrapper is silently dropped

The same happens for owner-script declarations (`const X = 'y' as const`). Consumers that expect svelte/compiler-shaped ASTs observe different trees per parser.

## Real-world impact

Found via svelte-shaker (baseballyama/svelte-shaker#150): on a 589-component app, the two parsers produce different shake results (160 vs 154 files) solely because of this node-shape difference — the engine's constant evaluator does not unwrap `TSAsExpression`, so the svelte-parser path fails to fold values the rsvelte path folds. svelte-shaker will gain an engine-side unwrap, but the parsers diverging on node shape remains a fidelity gap for any AST consumer.

## Possibly related observation (unverified impact)

In the same investigation, arrow-function parameters in event handlers (e.g. `(color, e) =>`) came back with zero-width spans (`Identifier[0,0]`) from rsvelte, while svelte/compiler assigns real spans. Not the cause of the divergence above, but worth checking while in this area — span-consuming tools would misbehave on those nodes.

## Repro

```ts
import { parse } from '@rsvelte/compiler';
// vs import { parse } from 'svelte/compiler';
const ast = parse(`<script lang="ts">const x = 'chips' as const;</script>\n<Child p={'chips' as const} />`, ...);
// diff the ExpressionTag.expression / VariableDeclarator.init node types
```

Expected: `TSAsExpression` preserved (matching svelte/compiler with TS scripts), or the deviation documented as intentional AST policy.

https://claude.ai/code/session_019ZP7b71mxqFV7uqAxV2ca1

baseballyama:
Fixes #1645

## Problem

`@rsvelte/compiler`'s `parse()` silently stripped TypeScript assertion wrappers,
returning only the inner expression:

- `pattern={'chips' as const}` → rsvelte emitted `Literal 'chips'` where
  svelte/compiler emits `TSAsExpression { expression: Literal 'chips',
  typeAnnotation: TSTypeReference "const" }`.
- Owner-script declarations (`const X = 'y' as const`) had the same divergence.

Consumers that expect the svelte/compiler AST shape (e.g. svelte-shaker) observed
a different tree depending on which parser ran.

## Root cause

rsvelte parses `<script>` with oxc and converts oxc expressions into its own typed
AST (`JsNode`). The converters (`convert_expression` for template expressions,
`convert_expression_for_program` for script bodies) **unwrapped** the oxc TS
wrapper nodes at parse time, so the wrapper never reached the parse output.

This is the wrong place to strip. The official compiler
(`submodules/svelte/.../1-parse/remove_typescript_nodes.js`, invoked from
`compiler/index.js`) keeps TS nodes in `parse()` and strips them only inside
`compile()` — before analyze/transform — via `remove_typescript_nodes`.

## Change (mirrors the official compiler)

- The parser now **preserves** `TSAsExpression`, `TSSatisfiesExpression`,
  `TSNonNullExpression`, `TSTypeAssertion`, and `TSInstantiationExpression`
  (with their `typeAnnotation` / `typeArguments`) as first-class `JsNode`
  variants, in both script bodies and inline template expressions, including the
  member/optional-chain `a!.b` cases.
- `remove_typescript_nodes` (the typed compile-time pass that already runs before
  analyze/transform) now **unwraps** these wrappers to their inner expression, so
  the AST analyze/transform sees is byte-identical to before — **compiled
  client/server output is unchanged.**
- Serialization is mirrored across all three parse-output surfaces: the serde
  `Serialize` impl (test AST), the NAPI binary envelope (`napi_raw_parse.rs` +
  the JS decoder in `parse-envelope.js`), and `from_value`.

The JSDoc-adjusted (Value-based, non-TS) converter keeps unwrapping — it never
carries real TS assertions, so unwrapping there is byte-identical.

### Secondary fix (zero-width span)

The issue also noted that fast-path mustache/attribute arrow handlers
(`onclick={(color, e) => …}`) returned parameters with a zero-width
`Identifier[0,0]` span and no `loc`. Root cause: a fast-path arrow parser
hardcoded `start: 0, end: 0, loc: None` for parameters ("positions not critical
for compilation"). Parameters now derive their real absolute spans from their byte
position in the source, matching svelte/compiler. Compiled output is unchanged.

## Verification

- The repro (`const x = 'chips' as const`, `<Child p={'chips' as const} />`, plus
  `as T<G>`, `satisfies`, `!`, `f<number>`) now parses to a tree that is
  **byte-identical** to svelte/compiler's `parse({ modern: true })` for every
  declarator init and attribute expression.
- Arrow handler parameters now report real spans (verified against
  svelte/compiler).
- `cargo test -p rsvelte_core --release`: **1602 passed, 0 failed** (includes the
  byte-exact client/server output gates, parser/validator/SSR/hydration/runtime
  suites).
- `cargo fmt` clean; `cargo clippy --all-targets --all-features -- -D warnings`
  clean across the workspace.

## Known issue (pre-existing, out of scope)

rsvelte leniently accepts a TS assertion in a **non**-`lang="ts"` component
(e.g. `{x as const}` with no `<script lang="ts">`), where svelte/compiler
raises a parse error. That lenient acceptance predates this PR; before it the
wrapper was silently stripped, so it was invisible. Now that wrappers are
preserved, the `TSAsExpression` becomes observable in such input. This is an
existing parser-leniency quirk, not introduced here, and is left unchanged.

https://claude.ai/code/session_011XHPUBHj4ywFM8skv4GwxV

### DIFF
--- a/.changeset/fix-parse-arrow-param-spans.md
+++ b/.changeset/fix-parse-arrow-param-spans.md
@@ -1,0 +1,12 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(parse): assign real spans to fast-path arrow-function parameters
+
+A top-level mustache/attribute arrow handler (e.g. `onclick={(color, e) => …}`)
+went through a fast-path parser that stamped its parameters with a zero-width
+`Identifier[0,0]` span and no `loc`, diverging from svelte/compiler (which
+assigns each parameter its real source span). The fast path now derives each
+parameter's absolute span from its byte position within the source, matching
+svelte/compiler. Compiled output is unchanged.
--- a/.changeset/fix-parse-preserve-ts-assertions.md
+++ b/.changeset/fix-parse-preserve-ts-assertions.md
@@ -1,0 +1,17 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(parse): preserve TypeScript assertion expressions in the parse AST
+
+`parse()` silently dropped TypeScript assertion wrappers — `x as const`,
+`x as T`, `x satisfies T`, `x!`, `<T>x`, and `x<T>` — returning only the inner
+expression, so consumers that expect the svelte/compiler AST shape (e.g.
+svelte-shaker) saw a different tree than the official compiler.
+
+The parser now keeps `TSAsExpression`, `TSSatisfiesExpression`,
+`TSNonNullExpression`, `TSTypeAssertion`, and `TSInstantiationExpression` (with
+their `typeAnnotation` / `typeArguments`), matching svelte/compiler's `parse()`
+output byte-for-byte. Mirroring the official compiler, `remove_typescript_nodes`
+unwraps these wrappers before analyze/transform, so compiled client/server
+output is unchanged.
--- a/apps/npm/vite-plugin-svelte-native/parse-envelope.js
+++ b/apps/npm/vite-plugin-svelte-native/parse-envelope.js
@@ -155,6 +155,11 @@ const JS_NULL = 0xca;
 // 0xcb was the former whole-node JS_RAW_JSON escape (removed — TS type
 // annotations now ride a per-node trailer; see readOptTypeAnnotation).
 const JS_TS_PARAMETER_PROPERTY = 0xcc;
+const JS_TS_AS_EXPRESSION = 0xcd;
+const JS_TS_SATISFIES_EXPRESSION = 0xce;
+const JS_TS_NON_NULL_EXPRESSION = 0xcf;
+const JS_TS_TYPE_ASSERTION = 0xd0;
+const JS_TS_INSTANTIATION_EXPRESSION = 0xd1;
 
 // LiteralValue inner tag (within a JS_LITERAL payload).
 const LV_NULL = 0;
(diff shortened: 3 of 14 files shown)


baseballyama:
Superseded by #1648, which was implemented in parallel in another session and merged to main first (closing #1645). #1648 covers the same core scope — preserving `TSAsExpression` / `TSSatisfiesExpression` / `TSNonNullExpression`, plus the for-of/for-in strip, svelte2tsx binding-cast, lint, and arrow-span fixes.

This PR's only unique contribution is exactly what #1648 explicitly deferred as "follow-up issues to come": preserving `TSTypeAssertion` (`<T>x`), `TSInstantiationExpression` (`f<T>`), and non-null `!` inside optional chains. Rather than force-merge two divergent implementations of the same feature, I'm closing this and splitting that delta into a focused follow-up PR built cleanly on top of #1648's merged structure.

PR #1686, opened 2026-07-22
chore(fmt): harden Tailwind JS sidecar (probe deferral, wait, protocol tests)

### LINKED ISSUE #1683: fmt: Tailwind JS sidecar follow-ups from #1668 review
Code review of #1668 (merged) surfaced non-blocking follow-ups, priority order:

1. **[perf] Stock config still spawns `node --version`** — `crates/rsvelte_fmt/src/main.rs:348`: `js_env` probe runs before `tailwind::decide()`, so users with `sortTailwindcss` configured pay a Node probe + sidecar stat even when `auto` falls back to the native sorter (the common stock-config case). Defer the probe until `decide` actually selects `SortViaJs`.
2. **[robustness] JSON channel integrity relies on JS-level stdout patch** — `crates/rsvelte_fmt/src/tailwind_sidecar.rs:923-932`: native addons (e.g. Tailwind v4 oxide) writing directly to fd 1 would bypass the `process.stdout.write` monkey-patch, corrupting the JSON response; large direct writes could also deadlock `write_all` since the stdout drain thread spawns after the write. Consider a separate writer thread.
3. **[correctness] Early returns in `sort()` skip `child.wait()`** — `tailwind_sidecar.rs:923/927/943`: broken-pipe etc. leaves a zombie until CLI exit; add a defensive `let _ = child.wait();` before returning.
4. **[test] Timeout and protocol-corruption fallback paths untested** — 30s kill path and malformed/non-`ok` sidecar responses (`tailwind_sidecar.rs:947-962`) have no automated coverage.

baseballyama:
## What

Non-blocking follow-ups from the #1668 review, hardening the `sortTailwindcss` Node sidecar without changing its output. Four items, one commit each.

### 1. perf — defer the Node probe until a JS sort is actually chosen
`sortTailwindcss` ran `node --version` (plus a sidecar stat) eagerly whenever the option was configured, so the common **stock-config** case — which sorts natively in Rust — paid that cost only for `decide` to fall back to the native sorter. `tailwind::decide` now takes `js_available` as a lazy `FnOnce`, invoked only when a branch truly needs the sidecar. A stock config never spawns Node; the probed env is captured for the `SortViaJs` arm.

### 2. robustness — decouple the JSON channel from the JS stdout monkey-patch
A native addon (e.g. the Tailwind v4 oxide binary) can write straight to fd 1, bypassing the script's `process.stdout.write` guard and corrupting the response. The sidecar now **frames** its JSON between markers and the Rust side extracts only the framed payload, discarding stray bytes. The stdout drain also no longer runs after the stdin write: stdin is fed on its own thread while stdout drains concurrently, so a large direct write can't deadlock `write_all`.

### 3. correctness — early returns now reap the child
Early exits in `sort()` (a missing pipe handle, a `try_wait` error) now `kill()` + `wait()` the child, so a broken pipe never leaves a zombie until the CLI exits.

### 4. test — timeout and protocol-corruption fallbacks
The sidecar timeout is now an injectable `SidecarEnv` field (default 30s). New unit tests drive `sort()` against fake Node scripts (skipped when Node is absent): a hung sidecar killed by a short injected timeout, malformed / framed-but-non-JSON / non-`ok` / length-mismatch responses, stray fd-1 output before the frame being tolerated, and pure-Rust `extract_framed` marker handling — no 30s waits.

## Verification
- `cargo test -p rsvelte_fmt -p rsvelte_formatter` green (one unrelated pre-existing daemon flake under parallel load, passes in isolation).
- New `tailwind_sidecar::tests` (8) all pass; real `tailwind-sort.mjs` verified to emit the framed response.
- `cargo fmt` + `cargo clippy --all-targets --all-features -- -D warnings` clean.

Fixes #1683

### DIFF
--- a/apps/npm/fmt/lib/tailwind-sort.mjs
+++ b/apps/npm/fmt/lib/tailwind-sort.mjs
@@ -16,11 +16,16 @@
 // plugin or a broken config degrades to unsorted output, never a crash.
 
 // Guard the JSON channel: the plugin or `tailwindcss` may print to stdout (a
-// deprecation notice, a debug line), which would corrupt our single-line
-// response. Route all stray stdout to stderr and emit only the final JSON on the
-// real stdout.
+// deprecation notice, a debug line), which would corrupt our response. Route all
+// stray stdout to stderr and emit only the framed JSON on the real stdout.
 const realStdoutWrite = process.stdout.write.bind(process.stdout);
 process.stdout.write = process.stderr.write.bind(process.stderr);
+
+// A native addon (e.g. the Tailwind v4 oxide binary) can write straight to fd 1,
+// bypassing the patch above. Framing the response between markers lets the Rust
+// side extract only the payload and discard such stray bytes. Kept byte-identical
+// to `RESP_MARKER` in `tailwind_sidecar.rs`.
+const RESP_MARKER = '\x00<<rsvelte-tw-sort>>\x00';
 
 function readStdin() {
 	return new Promise((resolve) => {
@@ -69,7 +74,7 @@ async function main() {
 }
 
 function emit(result) {
-	realStdoutWrite(JSON.stringify(result));
+	realStdoutWrite(RESP_MARKER + JSON.stringify(result) + RESP_MARKER);
 }
 
 main()
--- a/crates/rsvelte_fmt/src/main.rs
+++ b/crates/rsvelte_fmt/src/main.rs
@@ -750,7 +750,11 @@ fn resolve_js_class_sorter(
 fn js_sort_env() -> Option<tailwind_sidecar::SidecarEnv> {
     let script = tailwind_sidecar_script()?;
     let node = oxfmt_node().unwrap_or_else(|| PathBuf::from("node"));
-    node_runnable(&node).then_some(tailwind_sidecar::SidecarEnv { node, script })
+    node_runnable(&node).then_some(tailwind_sidecar::SidecarEnv {
+        node,
+        script,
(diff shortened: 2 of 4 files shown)


baseballyama:

**Verdict: mergeable — no blocking / should-fix findings.**

Verified (code reading + actual build/test runs on the PR branch):
- Frame marker byte-identical between Rust (`tailwind_sidecar.rs`) and JS (`tailwind-sort.mjs`), confirmed via hexdump of the real sidecar output. Marker contains raw NUL bytes which `JSON.stringify` can never emit, so accidental collision with stray fd-1 output is effectively impossible; absent/half markers fall back to sidecar-miss (covered by tests).
- No new deadlock path: writer thread's `write_all` unblocks with EPIPE after `kill()`; `hung_sidecar_times_out_and_falls_back` (injected 300ms timeout) passes without hanging.
- `decide()` `FnOnce` probe: exhaustive match check — stock config never probes; JS arms probe exactly once per run.
- `kill()+wait()` only touches early-return paths; normal exit-code handling unchanged.
- All 8 new sidecar tests exercise their real paths (no silent skips with Node present); `cargo test -p rsvelte_fmt -p rsvelte_formatter` fully green, clippy `-D warnings` + fmt clean.
- No changeset needed: `chore` title, and the JS sidecar feature itself is still unreleased (#1668 changeset unconsumed on main).

Nits (optional): rename `js_available` param to reflect it's a probe fn; theoretical two-marker scan limitation documented above.

PR #1687, opened 2026-07-22
fix(fmt): honour bracketSameLine before inline svelte blocks

### LINKED ISSUE #1654: BracketSameLine not putting bracket on same line if followed by inline svelte blocks
When BracketSameLine is set to true and a bracket is followed by an immediate svelte block (if/each), then the formatter is not putting the bracket on the same line as expected

See line 60 and 61 in the video clip, from before and after running `rsvelte-fmt src/`

https://github.com/user-attachments/assets/90535ab9-5f9f-4b27-9cbe-86178603bd1b

another example of an `each` block after formatting:

Image

my full settings:

```json
{
  "$schema": "./node_modules/oxfmt/configuration_schema.json",
  "ignorePatterns": [
    "src/lib/components/ui"
  ],
  "printWidth": 120,
  "sortImports": {
    "newlinesBetween": false
  },
  "bracketSameLine": true
}
```

Thanks for making rsvelte 🙏

baseballyama:
## Root cause

When an element's first child is an inline control-flow block (`{#if}` / `{#each}`), the collapse post-pass rebuilds that element through the faithful children port (`children::build_element_doc`). That builder was written for the parity corpus, which uses the default `bracketSameLine = false`, and hardcoded the corresponding layout — the open tag's trailing separator was always `dedent(softline)`. So with `bracketSameLine: true`, a wrapped open tag whose first child is a block dangled its `>` onto its own line:

```svelte
<div
  class="…overflows…"
  id="identifier"
  data-x="value"
>
  {#if cond}…{/if}
</div>
```

Text children, element children, and blocks not owned by the port glued the `>` correctly (they go through the markup path, which already honours the option) — so the bug was specific to port-owned block-first elements. Interestingly `{#each}` sometimes glued and `{#if}` didn't, purely because of which shapes the port claimed; the underlying builder ignored the option in every case.

## Fix

Thread the active `bracketSameLine` into the port via a thread-local (mirroring the existing `collapse::IN_PRE_CONTENT`) and apply prettier-plugin-svelte's real conditions in `build_element_doc`:

- **Open-tag trailing separator** — `hugStart && !isEmpty ? '' : !bracketSameLine ? dedent(softline) : ''`. This is the fix for the reported case: the `>` stays glued to the last attribute.
- **Self-closing** — `bracketSameLine ? ' />' : dedent(line) + '/>'`.
- **Empty hugged element** — keep the softline before the closing `>` when `bracketSameLine` (prettier's `omitSoftlineBeforeClosingTag = isEmpty && !bracketSameLine`).

Default (`bracketSameLine = false`) output is byte-for-byte unchanged.

## Tests

- New regression tests in `crates/rsvelte_formatter/tests/options.rs` cover `{#if}` and `{#each}` first children under `bracketSameLine: true` (glued) and the default (dangled).
- Verified byte-for-byte against a `prettier` + `prettier-plugin-svelte` oracle at printWidth 80 and 120, `bracketSameLine` on and off, across div / button / component / nested / self-closing scenarios.
- `cargo test -p rsvelte_formatter -p rsvelte_fmt` green; `cargo fmt` + `cargo clippy --all-targets --all-features -- -D warnings` clean.
- The svelte.dev fmt-parity corpus (`svelte_dev_corpus`) needs a local `oxfmt`, which isn't installed in this environment; it no-ops here and is left to CI.

Fixes #1654


baseballyama:

**Verdict: mergeable** (initial blocking finding resolved; verified against a real prettier 3.5.3 + prettier-plugin-svelte 4.0.1 oracle).

- All three `build_element_doc` branches (empty hug&hug, non-empty hug&hug, hug_end) now implement `canOmitSoftlineBeforeClosingTag` algebraically equivalent to prettier's, via the structural `omit_softline_allowed(out, end)` check.
- Byte-for-byte oracle match reproduced for: the original blank-line repro, content-glued-after-close, last-child-of-block (port-owned), and non-empty hug&hug — each with bracketSameLine on **and** off.
- Default-config safety: the `isEmpty` widening and whitespace-child drop are strictly gated to `bracketSameLine: true`; CI's Formatter parity hard gate passes.
- Attribution corrected: the blank-line bug reproduces on `origin/main` too (the pre-existing markup-path bracketSameLine support left a whitespace-child artifact); this PR's `isEmpty` widening absorbs it. Reviewer's original "introduced by this PR" claim was wrong about root cause, right about the symptom site.
- Panic-safety test added; the each-block test's markup-path-only scope is now documented.
- Known remaining divergences (pre-existing, bracketSameLine:true only, disclosed in the PR): deliberate whitespace-content inline elements in prose, and standalone empty elements outside the children port.

Remaining CI: a few long jobs still running, all green so far.

PR #1688, opened 2026-07-22
fix(parse): emit FunctionDeclaration.expression like acorn

### LINKED ISSUE #1659: AST fidelity: FunctionDeclaration.expression field omitted from parse output (svelte/compiler emits it)
Found during the review of #1648 (independent structural diff of `parse()` output against svelte/compiler 5.56.7): rsvelte omits the `expression` field on `FunctionDeclaration` nodes, while svelte/compiler (acorn) emits it (`expression: false`). Reproduces on constructs with no TS involvement, pre-existing before #1648.

Minor shape divergence; consumers doing strict structural comparison or destructuring the field observe a difference.

https://claude.ai/code/session_019ZP7b71mxqFV7uqAxV2ca1

baseballyama:
## Summary
- `parse()` output for `FunctionDeclaration` nodes was missing the `expression` field that acorn (used by `svelte/compiler`) always emits as `false`. Verified against svelte/compiler 5.56.7 that the correct key order is `id, expression, generator, async, params, body` (acorn's `initFunction` sets `expression` right after `id`, before `generator`/`async`) — not simply appended at the end.
- Added the field to the `JsNode::FunctionDeclaration` variant and updated every place that serializes/constructs it: the serde `Serialize` impl (JSON parse output), the JSON round-trip deserializer, the binary NAPI raw-parse envelope writer (`napi_raw_parse.rs`), and its JS-side decoder (`parse-envelope.js`).
- Scope check requested by the issue: `FunctionExpression`/`ArrowFunctionExpression` already carry an `expression` field, but `FunctionExpression`'s current field order (`generator, async, expression`) doesn't match acorn's real order (`expression, generator, async` — same `initFunction` logic applies to it too). That's a pre-existing, separate divergence; not fixed here per the issue's requested scope (structural equality is unaffected since JSON comparisons in this repo's test suites are value-based, not key-order-based, but flagging it for a follow-up).

## Verification
- Manually parsed `function f() {}` with both svelte/compiler 5.56.7 (via the `submodules/svelte` checkout) and rsvelte; the `FunctionDeclaration` node now matches byte-for-byte (key set, order, and values).
- `RUST_MIN_STACK=33554432 cargo test -p rsvelte_core --test parser_fixtures` — 17 passed, 0 failed.
- `RUST_MIN_STACK=33554432 cargo test -p rsvelte_core --test compiler_fixtures --test compiler_errors --test svelte2tsx_fixtures --test compatibility_report` — all passed.
- `RUST_MIN_STACK=33554432 cargo test -p rsvelte_core` (full crate suite) — 0 failed across all binaries.
- `cargo fmt` — no changes needed.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean.

No existing fixture in `fixtures/eae50dfd1c22/{parser-modern,parser-legacy}` exercises a raw `FunctionDeclaration`/`FunctionExpression` node outside the `javascript-comments` fixture (which is skipped for an unrelated comment-attachment reason), which is why this shape gap wasn't caught by the existing suite — consistent with the issue being found via an external structural diff.

Fixes #1659

https://claude.ai/code/session_01WGrjrA4zeTQR75pPThFQFT

### DIFF
--- a/.changeset/fix-function-declaration-expression-field.md
+++ b/.changeset/fix-function-declaration-expression-field.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(parse): emit `FunctionDeclaration.expression` (always `false`) to match acorn's key order (`id`, `expression`, `generator`, `async`, `params`, `body`)
--- a/.changeset/fix-function-declaration-expression-field.md
+++ b/.changeset/fix-function-declaration-expression-field.md
@@ -1,5 +1,12 @@
 ---
 "@rsvelte/compiler": patch
+"@rsvelte/vite-plugin-svelte-native": patch
 ---
 
 fix(parse): emit `FunctionDeclaration.expression` (always `false`) to match acorn's key order (`id`, `expression`, `generator`, `async`, `params`, `body`)
+
+The binary NAPI raw-parse envelope (`napi_raw_parse.rs`'s writer, consumed only
+by `@rsvelte/vite-plugin-svelte-native`'s `parse-envelope.js` decoder) carries
+the same field, so both packages need this release. The envelope's `VERSION`
+is bumped to 2 alongside the wire-format change (one extra bool byte on
+`FunctionDeclaration` payloads).
--- a/apps/npm/vite-plugin-svelte-native/parse-envelope.js
+++ b/apps/npm/vite-plugin-svelte-native/parse-envelope.js
@@ -1170,6 +1170,7 @@ function readJsVariableDeclarator(ctx, start, end) {
 function readJsFunctionDeclaration(ctx, start, end) {
 	const loc = readTypedLoc(ctx);
 	const id = readOptNode(ctx);
+	const expression = readBool(ctx);
 	const generator = readBool(ctx);
 	const asyncFlag = readBool(ctx);
 	const params = readChildArray(ctx);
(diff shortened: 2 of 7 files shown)


baseballyama:

**Verdict: core logic correct; two release-process should-fixes requested before merge (being addressed).**

Independently verified:
- Key order confirmed twice over: acorn's `initFunction` source (right-to-left assignment inserts `expression` before `generator`) and a live byte-diff of `svelte/compiler@5.56.4` vs the built PR branch on `function f(){}` — identical `id, expression, generator, async, params, body`.
- Exhaustive-site check: all 6 non-`..` `JsNode::FunctionDeclaration` patterns (3 construction sites, deserializer, Serialize, binary writer) updated; everything else uses `..`.
- Deserializer falls back to `false` on a missing `expression` key — old JSON stays readable.
- `FunctionExpression`/`ArrowFunctionExpression` untouched (tracked separately in #1689).
- PR branch built in an isolated worktree: clippy clean; parser/compiler/svelte2tsx fixture suites pass.

Should-fix (requested):
1. Changeset misses the `@rsvelte/vite-plugin-svelte-native` fixed group — the `napi_raw_parse.rs` binary-writer change only ships through that group, so without an entry it never reaches npm.
2. Binary envelope `VERSION` (Rust + both JS decoders) not bumped despite a wire-format change; skewed writer/decoder pairs would pass the 1==1 handshake and silently misread the stream.

PR #1690, opened 2026-07-22
fix(parse): preserve TSFunctionType in convert_ts_type unions

baseballyama:
## Summary

- `convert_ts_type` (the oxc TS-type AST → estree-ts JSON converter, `crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs`) had no match arm for `TSType::TSFunctionType` / `TSType::TSConstructorType`, so both fell through to the catch-all `TSUnknownKeyword` stub arm. Most visible as a union member (`let x: string | (() => void)`), but the same fallback hit intersections, `ParenthesizedType`-wrapped standalone function types, `as`/`satisfies`/type-assertion wrappers, and `TSTypeParameterInstantiation` — none of that is specific to unions.
- svelte/compiler (acorn-typescript) keeps these as full nodes: `typeParameters` (generic type params, omitted when absent), a flat `parameters` array (any `this` parameter prepended as a plain `Identifier` named `"this"`, matching acorn-typescript's parse), and a `typeAnnotation`-wrapped return type. `TSConstructorType` additionally always emits `abstract` (never omitted, unlike `optional`/`readonly` elsewhere).
- Added `convert_ts_type_parameter_declaration` / `convert_ts_type_parameter` (`<T extends U = V>`) and `convert_ts_function_like_params` (this-param + regular params via the existing `convert_formal_parameter` + rest element), and threaded a `&ParseArena` through the whole `convert_ts_type` call graph (unions, intersections, arrays, wrappers, `TSTypeParameterInstantiation`, and every TS assertion/instantiation expression call site) since building the parameters array needs it.
- Verified expected shapes against the real `svelte` 5.56.7 compiler (`node_modules/svelte/compiler`, acorn-typescript-based) for unions, intersections, `as`-assertions, generics + rest params, `this` params, and `TSConstructorType`.

**Found but out of scope** (documented in the new test with a `NOTE`): `b?: number`-style optional function/arrow parameters lose their `optional: true` marker for *every* function, not just `TSFunctionType` — `JsNode::Identifier` (and the other pattern variants) have no `optional` field to round-trip through the typed AST (`Expression::from_json` → `JsNode::from_value` → `to_value()` drops unmodelled keys). Fixing it means adding a field to a very widely-constructed enum variant across the codebase; flagged for a separate, larger PR rather than folded in here.

## Root cause

`convert_ts_type`'s `match` over `oxc_ast::ast::TSType` never had arms for `TSFunctionType`/`TSConstructorType`, so they hit the `_ => Value::Object(base("TSUnknownKeyword"))` fallback — a bug that predates #1648 but became more visible once #1648 started preserving the `TSAsExpression`/etc. wrapper nodes (whose `typeAnnotation` routes through `convert_ts_type`).

## Test plan

- [x] New regression tests in `crates/rsvelte_core/tests/parse_ts_types.rs`: function type inside union/intersection, via `as`-assertion, standalone with params + optional marker + return type, generics + rest parameter, `this` parameter prepending, and `TSConstructorType`.
- [x] `RUST_MIN_STACK=33554432 cargo test -p rsvelte_core` — full suite (171 test binaries incl. `parser_fixtures`, `svelte2tsx` (1538 tests), `parse_ts_types`) — all pass, 0 failures.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` clean.

https://claude.ai/code/session_01WGrjrA4zeTQR75pPThFQFT

### DIFF
--- a/.changeset/fix-ts-function-type-in-union.md
+++ b/.changeset/fix-ts-function-type-in-union.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(parse): preserve `TSFunctionType` / `TSConstructorType` in `convert_ts_type` instead of collapsing them to a `TSUnknownKeyword` stub (e.g. inside a union like `string | (() => void)`)
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -2579,8 +2579,12 @@ fn convert_formal_parameter_inner(
                 obj.insert("name".to_string(), Value::String(name.to_string()));
 
                 // Convert type annotation
-                let type_ann_obj =
-                    convert_type_annotation_adjusted(type_ann, adjusted_offset, line_offsets);
+                let type_ann_obj = convert_type_annotation_adjusted(
+                    arena,
+                    type_ann,
+                    adjusted_offset,
+                    line_offsets,
+                );
                 obj.insert("typeAnnotation".to_string(), type_ann_obj);
 
                 Expression::from_json(Value::Object(obj))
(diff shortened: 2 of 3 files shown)


baseballyama:

**Verdict: implementation correct; blocked only on the missing changeset (CI Changeset job fails for the `fix(parse):` title). Mergeable once added.**

Independently verified against a real `svelte@5.56.3` install (fresh `parse()` comparisons, not just the PR's own tests):
- Shape parity for union/intersection members, generics + rest params, `this`-param prepended as `Identifier` (span covering `this: Foo`), `typeParameters` omitted-when-absent (not null), and `TSConstructorType.abstract` always emitted — all byte-consistent with acorn-typescript.
- Field access checked against the pinned oxc `ts.rs` definitions (`this_param`, `params`, `return_type` etc.) — all match.
- PR branch built and tested in an isolated worktree: fmt/clippy clean; `parse_ts_types` (incl. 12 new tests), `compiler_fixtures`, `parser_fixtures`, `svelte2tsx` suites all pass — no regression from the `&ParseArena` threading.
- The known `optional: true` parameter-marker gap reproduces on stock main too — correctly scoped out (tracked in #1692).
- Remaining catch-all for TSMappedType/TSConditionalType etc. is pre-existing, unchanged (nit only).

baseballyama:
Changeset added and the CI Changeset job now passes — the review's only blocking item is resolved. Mergeable.

Issue #1680, opened 2026-07-22
Extend function compile options (dynamic cssHash etc.) to the wasm and C-API bindings


baseballyama:
## Context

#1666/#1667 implemented the five function-typed compile options (`customElement`/`css`/`runes` function forms, `warningFilter`, constant + dynamic `cssHash`) for the **NAPI** binding (`@rsvelte/vite-plugin-svelte-native`). The other bindings still lack them:

- **wasm (`@rsvelte/compiler`)**: no callback bridge. The boundary-resolution parts (`customElement`/`css`/`runes` evaluated once with `{filename}`, `warningFilter` post-filter, `cssHashOverride` constant) are pure JS-side work and can be ported to the wasm JS wrapper cheaply. Dynamic `cssHash` needs a wasm-bindgen `js_sys::Function` bridge — simpler than NAPI since wasm compile is single-threaded (no TSFN/thread marshaling needed).
- **C-API (`rsvelte_capi`)**: only `css_hash_override` exists; no `warning_filter` field and no callback function pointers. If/when a C consumer needs it, expose C function-pointer + userdata fields mapped to `CssHashFn`/`WarningFilterFn`.

## Priority

Low — no known consumer passes function options through wasm/C-API today. The main consumers (Vite plugin, svelte-check via #1679) are covered or tracked elsewhere. Do the wasm boundary-resolution subset first if playground/docs parity ever needs it.

## References
- #1666 (boundary resolution + post-filter), #1667 (dynamic cssHash bridge, never-reject + cleanup pattern)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012kjwziVV91nApcWDK48gRA

PR #1691, opened 2026-07-22
feat(wasm): support function compile options (css/customElement/runes/warningFilter/cssHash)

baseballyama:
## What

Extends the function-form compile options — implemented for the NAPI bindings in #1666/#1667 — to the **wasm `@rsvelte/compiler`** bindings.

The wasm compiler previously exposed only `compile_client(source, name)` / `compile_server(source, name)`, which accept no options object, so none of the five function options were reachable. This PR adds a new `compile(source, options)` wasm entry that accepts the full compile-options object and resolves the pieces the primitive entries can't:

- **parametric function forms** of `customElement` / `css` / `runes` (`({ filename }) => value`), evaluated once at the boundary (Svelte's `parametric()` normalization);
- **`warningFilter`** callback — applied natively by the compiler (`opts.warning_filter`), so it also covers module-compile warnings;
- **constant `cssHashOverride`** string; and
- **dynamic `cssHash`** callback, bridged through `js_sys::Function`. wasm compile is single-threaded, so the callback runs inline — no threadsafe-function / thread-marshalling needed (unlike the NAPI backend's TSFN path). The callback's `hash` argument is Svelte's raw digest (no `svelte-` prefix). A callback that throws surfaces as a compile error (matching upstream); a non-string return falls back to the compiler's default hash. The bridge never panics.

The result is returned as a JSON string (`{ js, css, warnings, metadata }`), consistent with the existing `svelte2tsx` wasm entry; callbacks are input-only. `compile_client` / `compile_server` are unchanged.

Implementation note: on the wasm target `js_sys` values are `Send + Sync`, so the JS callbacks satisfy the shared `CssHashFn` / `WarningFilterFn` bounds directly — **no `unsafe` `Send`/`Sync` wrapper**; the only thread-safe container needed is an `Arc<Mutex<Option<String>>>` slot for a throwing `cssHash`.

## Scope

- **wasm: fully implemented** ✅
- **C-API: out of scope** for this PR. `rsvelte_capi` takes options as a JSON blob and has no callback-pointer infrastructure. The parametric `customElement`/`css`/`runes` forms are meaningless across a C ABI (a C caller passes the resolved value), leaving only `warningFilter` + dynamic `cssHash`, which would require a new `#[repr(C)]` callback-pointer + userdata ABI and cbindgen header work with no current consumer. Deferred until needed — hence `Refs`, not `Fixes`.

## Tests

New node e2e (`pnpm run test:wasm-compile`, `scripts/dev/test-wasm-compile-options.mjs`) drives the real wasm build via `initSync` and covers all five options — 12/12 passing:

```
12 passed, 0 failed
```

Also verified: `cargo check`/`clippy` clean on both the wasm32 target and the native host with `--features wasm`; no changes to the NAPI or native paths (the wasm module is `#[cfg(feature = "wasm")]`).

Refs #1680

### DIFF
--- a/.changeset/feat-wasm-function-compile-options.md
+++ b/.changeset/feat-wasm-function-compile-options.md
@@ -1,0 +1,23 @@
+---
+"@rsvelte/compiler": minor
+---
+
+feat(wasm): support function compile options via a new `compile(source, options)` entry
+
+The wasm compiler now exposes `compile(source, options)`, which accepts the full
+compile-options object and resolves the function-form options that the primitive
+`compile_client`/`compile_server` entries can't — matching the NAPI shim's
+support (PRs #1666/#1667):
+
+- the `parametric` function forms of `customElement`, `css`, and `runes`
+  (`({ filename }) => value`), evaluated once at the boundary;
+- a `warningFilter` callback, applied natively by the compiler;
+- a constant `cssHashOverride` string; and
+- a dynamic `cssHash` callback bridged through `js_sys::Function` (wasm compile
+  is single-threaded, so the callback runs inline with no threadsafe-function
+  marshalling). A callback that throws surfaces as a compile error; a non-string
+  return falls back to the default hash.
+
+The result is returned as a JSON string (`{ js, css, warnings, metadata }`);
+callbacks are input-only. The existing `compile_client`/`compile_server` entries
+are unchanged.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2372,6 +2372,7 @@ dependencies = [
  "indexmap",
  "insta",
  "itoa",
+ "js-sys",
  "lazy_static",
  "memchr",
  "miette",
--- a/apps/npm/compiler/README.md
+++ b/apps/npm/compiler/README.md
@@ -49,6 +49,7 @@ change.
 ```js
 import init, {
   parse_svelte,
+  compile,
   compile_client,
   compile_server,
   version,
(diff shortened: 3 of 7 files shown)


baseballyama:

**Verdict: approve modulo one should-fix (fix in progress). Reviewer independently rebuilt with wasm-pack and reproduced the 12/12 e2e pass.**

Should-fix (being addressed): a throwing `warningFilter` is silently swallowed (`Err(_) => true` in `build_warning_filter`), keeping the warning and returning success — reproduced on a real build. Upstream Svelte and the NAPI shim both propagate the exception, and the same PR's `cssHash` already routes throws through an error slot to a compile error; `warningFilter` should use the same pattern, with an e2e case added.

Verified fine:
- Semantics match NAPI #1666/#1667: parametric evaluation timing, `(unknown)` filename default, raw-digest `hash` argument, cssHashOverride precedence, non-string → default-hash fallback, cssHash throw → compile error.
- The no-`unsafe` Send+Sync claim is correct: wasm-bindgen itself provides `unsafe impl Send/Sync for JsValue` when atomics are off.
- `compile_client`/`compile_server` untouched; C-API scope-out rationale sound; changeset minor appropriate; clippy clean on host + wasm32 (the one clippy error under `--no-default-features --all-targets` reproduces on main — pre-existing).

Noted for follow-up (not blocking): shared `CssHashInput.hash` carries the `svelte-`-prefixed hash where upstream's contract implies a raw digest (both bindings currently work around it); new `compile()` throws on error while older wasm entries embed `success:false`; `test:wasm-compile` not wired into CI (matching the NAPI test precedent).

baseballyama:

**Verdict: mergeable once CI is green.** The should-fix is resolved: a throwing `warningFilter` now routes through the shared error slot (same pattern as `cssHash`, via the new `js_error_message` helper) and surfaces as a compile error. Reviewer rebuilt the wasm bundle and re-ran the original repro — `compile()` now throws correctly, including for non-Error throws (`throw "plain string"`). e2e 13/13; clippy/fmt clean. Remaining nits (CI wiring for the e2e script, `.d.ts` typing) are non-blocking; the shared `CssHashInput.hash` contract issue is tracked in #1697.

PR #1695, opened 2026-07-22
feat(svelte-check): honor function warningFilter via Node sidecar

### LINKED ISSUE #1679: svelte-check: honor `compilerOptions.warningFilter` from svelte.config.js (needs JS execution in the native CLI)
## Context

`rsvelte-check` reads diagnostic-relevant `compilerOptions` from `svelte.config.*`, but `warningFilter` is a **JavaScript function** and the native CLI cannot evaluate it, so it is currently ignored. This produces warnings-only divergence from the official svelte-check for projects that filter warnings in their config.

The NAPI side of this gap was closed by #1666 (the vite-plugin shim now post-filters returned warnings with the user's `warningFilter`, proven equivalent to Svelte's emit-time filter, which is a pure predicate). The remaining gap is the **native CLI**, which has no JS runtime.

## Proposed direction

Reuse the Node-sidecar pattern from #1668 (rsvelte-fmt Tailwind sorter):
- When `svelte.config.*` declares a function `warningFilter`, spawn the consumer's Node once per run with a small bundled script that loads the config and evaluates `warningFilter(warning)` for the collected warnings in one batch (warnings are only filtered at the end, so a single post-pass batch is exactly equivalent — same argument as #1666).
- No Node / config load failure → current behavior (filter ignored) with a one-time warning naming the reason.
- Zero overhead when the config has no function `warningFilter`.

## References
- #1666 (equivalence proof + shim post-filter), #1668 (sidecar infra: one-shot spawn, timeout, never-reject protocol)
- Memory of the original divergence: svelte-check corpus warnings-only diffs

https://claude.ai/code/session_012kjwziVV91nApcWDK48gRA

baseballyama:
## What

`rsvelte-check` reads diagnostic-relevant `compilerOptions` from `svelte.config.*` statically, but `compilerOptions.warningFilter` is a JS predicate the native compiler can't evaluate, so it was silently ignored. Projects that use `warningFilter` therefore saw a warnings-only diff versus the official `svelte-check` ([#1679](https://github.com/baseballyama/rsvelte/issues/1679)).

This wires up the Node-sidecar pattern (from #1668 / #1686) so the native CLI honors a function `warningFilter`.

## How

- **Static probe** (`config.rs`): `warning_filter_config_path()` parses `svelte.config.*` (or the `--config` path when it names a Svelte config) with oxc and returns the file only when `compilerOptions.warningFilter` is statically function-shaped. A project without one never spawns Node (zero overhead); the sidecar re-checks `typeof === 'function'` after actually importing.
- **Sidecar** (`apps/npm/svelte-check/lib/warning-filter.mjs`): one-shot, framed request/response over stdin/stdout. Imports the config (ESM/CJS via dynamic `import()`), pulls out `compilerOptions.warningFilter`, and returns a keep/drop mask for the batch. Stray stdout from importing the config is routed to stderr, and the JSON payload is framed between markers so a native addon writing straight to fd 1 can't corrupt the channel.
- **Runner** (`runner.rs`): after all other filters, `run()` collects the run's Svelte compiler warnings and applies the mask through the sidecar in a **single** call per run. Errors and non-Svelte (TypeScript) diagnostics are never touched — `warningFilter` only ever sees compiler warnings.
- **Launcher** (`bin/svelte-check.cjs`): passes `process.execPath` and the sidecar path to the native binary via env vars, and ships `lib/warning-filter.mjs` in the package `files`.
- Watch mode re-runs `run()` each cycle, so a `svelte.config.js` edit re-evaluates the filter automatically.

## Equivalence argument

`warningFilter` is a pure per-warning predicate that the official Svelte compiler applies at emit time; a dropped warning affects nothing else. So filtering the fully collected warning batch in one post-pass is exactly equivalent to filtering at emit time — the same reasoning the NAPI shim uses in #1666. The warning object handed to the predicate is reconstructed from the diagnostic and carries Svelte's `Warning` fields `code` / `message` / `filename` / `start` / `end` (1-indexed `{ line, column }`).

## Fallback behavior

The sidecar never rejects. A missing Node, an unimportable/throwing config, a timeout (injectable; 30s default), or a malformed response all degrade to **keep every warning** plus a one-time stderr note. A `warningFilter` that can't run must never silently drop a warning — it fails open. The exit code is recomputed from the surviving diagnostics like any other filter, so it's unaffected.

## Tests

- `warning_filter.rs` `--lib` unit tests: only the Node-free `extract_framed` marker logic (so the no-Node test-unit CI job stays runnable).
- `svelte_check_warning_filter.rs` integration `apply()` tests (Node-backed, its CI job installs Node; skipped when Node absent): drops rejected / keeps others via the real framed protocol, hung-sidecar timeout keeps all, malformed response keeps all, error/TS diagnostics untouched.
- `config.rs`: static detection of arrow / identifier `warningFilter`, non-function ignored, absent → none, vite.config is not a source.
- `svelte_check_warning_filter.rs` e2e (spawns the compiled binary + the real `warning-filter.mjs`): function filter drops the matching warning and clears the count; a **falsy (`undefined`) return also drops**; non-matching filter keeps it; no filter is unchanged; missing sidecar fails open with a note; a throwing config fails open; and a **filtered-away warning is not re-promoted by `--compiler-warnings code:error`** (order + exit-code check).
- `cargo fmt` + `cargo clippy -p rsvelte_core --all-targets --all-features -D warnings` clean; existing svelte-check suites green.

## Known limitations

- **`frame` / `position` are not reconstructed.** The predicate receives `code` / `message` / `filename` / `start` / `end`, but not the source `frame` string or the `position` char-offset tuple that the official emit-time `Warning` carries. A `warningFilter` that inspects `warning.frame` or `warning.position` will see `undefined` and can therefore misbehave silently. In practice filters key off `code` / `filename` / `message`; extending the shape would mean threading extra fields through the diagnostic + incremental cache.
- **Inline `vite.config.*` plugin `warningFilter`** is intentionally unsupported (only `svelte.config.*`), since importing a `vite.config.*` standalone would drag in the whole Vite plugin graph.

## Review follow-ups (addressed)

- Truthiness parity: the sidecar drops on any falsy return (`Boolean(filter(w))`), matching Svelte's `if (!warning_filter(w)) return;`.
- Ordering: `warningFilter` now runs **before** `--compiler-warnings` `code:error` promotion (official filter-then-promote order), so a filtered warning can't resurface as an error or flip the exit code.
- The one-time fail-open stderr notes are `Once`-guarded so `--watch` doesn't re-print them each rebuild.

Fixes #1679

### DIFF
--- a/.changeset/feat-svelte-check-warning-filter.md
+++ b/.changeset/feat-svelte-check-warning-filter.md
@@ -1,0 +1,22 @@
+---
+"@rsvelte/svelte-check": minor
+---
+
+feat(svelte-check): honor function `compilerOptions.warningFilter` via a Node sidecar
+
+`rsvelte-check` reads diagnostic-relevant `compilerOptions` from `svelte.config.*`
+statically, but `warningFilter` is a JS predicate the native compiler can't
+evaluate, so it was silently ignored — a warnings-only divergence from the
+official `svelte-check` for projects that use it.
+
+When `svelte.config.js` declares a function `compilerOptions.warningFilter`,
+`rsvelte-check` now spawns the consumer's Node **once per run** against a small
+bundled sidecar (`lib/warning-filter.mjs`) that imports the config and applies
+the function to the run's collected compiler warnings in a single batch. Because
+`warningFilter` is a pure per-warning predicate, this post-pass is exactly
+equivalent to Svelte's emit-time filter (the same argument the NAPI shim uses).
+
+The sidecar never rejects: a missing Node, an unimportable config, a timeout, or
+a malformed response all degrade to "keep every warning" plus a one-time stderr
+note — the filter never silently drops a warning, and the exit code is unaffected.
+A project with no function `warningFilter` never spawns Node (zero overhead).
--- a/apps/npm/svelte-check/README.md
+++ b/apps/npm/svelte-check/README.md
@@ -104,6 +104,7 @@ Run `rsvelte-check --help` for the authoritative list.
 Highlights:
 
 - **SvelteKit-aware.** Honours `svelte.config.js`'s `kit.files` overrides; injects SvelteKit-generated kit-file augmentations for both `.ts` (real TS annotations) and `.js` (JSDoc) files.
+- **`warningFilter` support.** A function `compilerOptions.warningFilter` in `svelte.config.js` is honoured. The native compiler can't run the JS predicate itself, so the run's compiler warnings are collected and passed once to a small Node sidecar (bundled with this package) that imports your config and applies the function — equivalent to Svelte's emit-time filter since it's a pure per-warning predicate. If Node is unavailable or the config can't be imported, every warning is shown and a one-time note is printed (the filter never silently drops a warning). Projects without a function `warningFilter` pay nothing — the sidecar is never spawned.
 - **Incremental.** A per-file overlay manifest and a per-file warning cache (`<cacheDir>/warnings.json`) make warm runs near-instant.
 - **Parallel compile.** Files are compiled across rayon workers; the TS pass is the long pole.
 - **Watch mode.** Composes with `--incremental` for an editor-like inner loop.
--- a/apps/npm/svelte-check/bin/svelte-check.cjs
+++ b/apps/npm/svelte-check/bin/svelte-check.cjs
@@ -73,9 +73,22 @@ if (process.platform !== 'win32') {
 	}
 }
 
+// Hand the native binary the Node interpreter + `warningFilter` sidecar script
+// it needs to evaluate `compilerOptions.warningFilter` (a JS predicate the Rust
+// compiler can't run). The script lives in this launcher's own package; the
+// binary re-spawns this same Node against it. Absent env vars simply disable the
+// feature (warnings are then all shown), so this is best-effort.
+const sidecar = path.join(__dirname, '..', 'lib', 'warning-filter.mjs');
+const childEnv = {
+	...process.env,
+	RSVELTE_CHECK_NODE: process.execPath,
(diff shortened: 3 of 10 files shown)


baseballyama:

**Verdict: request changes — 2 blocking equivalence bugs (fixes in progress). Design/robustness otherwise solid — sturdier than the #1686 tailwind sidecar (concurrent stdin write + stdout drain, child reaped on every failure path).**

Blocking:
1. `warning-filter.mjs`: `filter(warning) !== false` keeps warnings whose predicate returns `undefined`/`0`/`''`/`null` — the official compiler drops on **truthiness** (`if (!warning_filter(warning)) return;`). Must be `Boolean(filter(warning))`.
2. `runner.rs`: filter order is promote (`--compiler-warnings code=error`) → warningFilter, but `apply_warning_filter` only feeds severity==Warning to the predicate. A warning both rejected by the filter and promoted to error is fully hidden by official svelte-check (official order: filter → promote) yet shown as an Error here, affecting exit codes.

Should-fix: move the 4 Node-spawning unit tests out of the no-Node `test-unit` (`--lib`) job into the integration suite (they'd silently no-op if the runner loses its preinstalled Node); the "one-time" stderr note actually fires every rebuild in watch mode; document that predicates reading `warning.frame`/`position` silently misbehave.

Verified fine: 1-indexed line / 0-indexed column matches `locate-character`; static function-shape detection fails safe (sidecar re-checks `typeof === 'function'`); watch mode re-reads config every run (no stale cache); direct-binary invocation fails open (covered by e2e); e2e tests are picked up by the sharded `test` job; changeset + npm `files` wiring correct.

PR #1698, opened 2026-07-22
feat(fmt): honor sortTailwindcss.functions in the native Svelte path

### LINKED ISSUE #1678: fmt: support `sortTailwindcss.functions` (cn()/cva() call arguments) in the native Svelte path
## Context

PR #1605 added the native Tailwind sorter and PR #1668 added the Node-sidecar JS path for custom configs, but both only sort **fully static `class` attribute values** (plus `sortTailwindcss.attributes`). The oxfmt option `sortTailwindcss.functions` (sorting class strings inside wrapper calls like `cn(...)`, `cva(...)`) is not honored on the rsvelte-fmt native `.svelte` path.

Plain `.js`/`.ts` files are unaffected — they are delegated to the oxfmt subprocess, which handles `functions` itself. The gap is only for call expressions inside `.svelte` files (`<script>` and `{expr}` positions), which `rsvelte_formatter` never passes to the class sorter because they are not static attribute text (`crates/rsvelte_formatter/src/markup.rs` static-value gate).

## What's needed

- Detect string literal arguments of configured `functions` call expressions inside `.svelte` script/expressions during formatting.
- Route them through the existing sorter hook (native `tailwind_class_order` or the JS sidecar from #1668, matching the resolved strategy).
- Mirror oxfmt's semantics (`apps/oxfmt` injects `tailwindFunctions` into prettier-plugin-tailwindcss) so oracle byte-parity holds.

## References
- #1605, #1668 (sidecar + `strategy` option)
- oxfmt: `apps/oxfmt/src/core/to_prettier.rs` (tailwindFunctions injection)

https://claude.ai/code/session_012kjwziVV91nApcWDK48gRA

baseballyama:
## What

Closes the `sortTailwindcss.functions` gap in rsvelte-fmt's native `.svelte`
path. Previously only fully static `class` attribute values were sorted; the
class strings passed to wrapper calls (`cn(...)`, `cva(...)`, `clsx(...)`) inside
`.svelte` files were ignored (PR #1605 / #1668 both left `functions` out of
scope). Bare `.js`/`.ts` files are unaffected — they delegate to the `oxfmt`
subprocess.

## Scope (mirrors oxfmt exactly)

Derived empirically from the real `oxfmt` + `prettier-plugin-tailwindcss` (oxfmt
runs its own AST traversal and only delegates the sort), then reproduced:

- **`<script>` bodies** — string and template-literal arguments of a
  `CallExpression` whose callee is a **bare identifier** in `functions`. The
  descent stops at a nested `CallExpression`, so `cn(a, notcn("…"))` sorts only
  `a`; a nested call is sorted only when its own callee matches
  (`cn(a, clsx("…"))` sorts both). Object **keys**, arrays, and nested plain
  containers inside a matched call are sorted. `tw.foo(...)` (member callee) is
  not matched.
- **`class={…}` mustaches** (plus any configured `attributes`) — **every** class
  literal in the expression, regardless of an enclosing call, matching the
  plugin's `transformSvelte` (not function-gated). `class:` directives and
  standalone `{expr}` mustaches are left untouched.
- **Template literals with `${…}`** — sorted **per static quasi**, pinning the
  token that abuts an interpolation (no boundary whitespace) via the plugin's
  `ignoreFirst`/`ignoreLast` rule and preserving the quasi's leading/trailing
  separator whitespace. `cn(`flex m-2 p-4 ${x}`)` → `cn(`m-2 flex p-4 ${x}`)`.

Both paths route through the same class sorter used for static attributes, so the
**native** (zero-config, pure-Rust) and **Node-sidecar** (custom-config) paths
apply transparently via the existing collect → sort → apply passes. The default
path is unchanged when `functions` is unset, and the fmt-parity corpus gate (sort
off) is unaffected.

Out of scope (documented): mixed static-plus-`{expr}` class attribute values.

## Implementation

- New `crates/rsvelte_formatter/src/tailwind_sort.rs`: parses the JS (wrapping a
  bare expression in `(...)`), collects string-literal and template-quasi edits
  via an `oxc_ast_visit::Visit` that tracks the innermost enclosing call's match
  state, and splices the sorted values back into the source (oxc re-prints
  unchanged). Template quasis carry per-quasi `ignore_first`/`ignore_last` flags.
- `FormatOptions.tailwind_functions`; hooked into `script.rs` (`<script>`) and
  `markup.rs` (`class={…}`); CLI reads `sortTailwindcss.functions`.

## Verification

- `cargo test -p rsvelte_formatter -p rsvelte_fmt -p rsvelte_fmt_wasm` — 462
  passed, 0 failed.
- Tests: 22 deterministic formatter-level cases (fake sorter, assert exactly
  which literals/quasis route to the sorter, incl. pinning & multi-interpolation)
  + CLI cases: native sort, functions-unset no-op, and two byte-for-byte
  oxfmt-oracle comparisons (native + JS-sidecar, each incl. template cases with a
  custom `@utility`), gated on a Tailwind env + `oxfmt`.
- `cargo fmt` + `cargo clippy --all-targets --all-features -- -D warnings` clean.
- Changeset: `@rsvelte/fmt` minor.

Fixes #1678

### DIFF
--- a/.changeset/feat-fmt-tailwind-functions.md
+++ b/.changeset/feat-fmt-tailwind-functions.md
@@ -1,0 +1,29 @@
+---
+"@rsvelte/fmt": minor
+---
+
+feat(fmt): honor sortTailwindcss.functions in the native Svelte path
+
+`sortTailwindcss` previously sorted only fully static `class` attribute values.
+The oxfmt option `sortTailwindcss.functions` — sorting the class strings passed
+to wrapper calls like `cn(...)` / `cva(...)` / `clsx(...)` — was ignored inside
+`.svelte` files.
+
+rsvelte-fmt now mirrors oxfmt's `.svelte` pipeline (verified byte-for-byte
+against `oxfmt` + `prettier-plugin-tailwindcss`):
+
+- **`<script>` bodies**: string and substitution-free template-literal arguments
+  of a call whose callee is a bare identifier listed in `functions` are sorted.
+  The descent stops at a nested call, so `cn(a, notcn("…"))` sorts only `a`; a
+  nested call is sorted only when its own callee matches. Object keys, arrays,
+  and nested plain containers inside a matched call are sorted.
+- **`class={…}` mustaches** (and any configured `attributes`): every class
+  literal in the expression is sorted, regardless of an enclosing call — matching
+  the plugin's `transformSvelte`, which is not function-gated. `class:` directives
+  and standalone `{expr}` mustaches are left untouched.
+
+Sorting routes through the same class sorter as static attributes, so the native
+(zero-config) and Node-sidecar (custom-config) paths both apply. The default path
+stays untouched when `functions` is unset, and the fmt-parity corpus gate (sort
+off) is unaffected. Substitution-bearing template literals and mixed
+static-plus-`{expr}` class values remain out of scope.
--- a/.changeset/feat-fmt-tailwind-functions.md
+++ b/.changeset/feat-fmt-tailwind-functions.md
@@ -22,8 +22,12 @@ against `oxfmt` + `prettier-plugin-tailwindcss`):
   the plugin's `transformSvelte`, which is not function-gated. `class:` directives
   and standalone `{expr}` mustaches are left untouched.
 
+Template literals with `${…}` interpolations are sorted per static quasi; the
+token abutting an interpolation is pinned (`cn(\`flex m-2 ${x}\`)` keeps its
+structure), matching `prettier-plugin-tailwindcss`.
+
 Sorting routes through the same class sorter as static attributes, so the native
 (zero-config) and Node-sidecar (custom-config) paths both apply. The default path
 stays untouched when `functions` is unset, and the fmt-parity corpus gate (sort
-off) is unaffected. Substitution-bearing template literals and mixed
-static-plus-`{expr}` class values remain out of scope.
+off) is unaffected. Mixed static-plus-`{expr}` class attribute values remain out
+of scope.
--- a/crates/rsvelte_fmt/src/main.rs
+++ b/crates/rsvelte_fmt/src/main.rs
@@ -627,6 +627,14 @@ fn build_format_options(cli: &Cli, cfg: &OxfmtConfig) -> (FormatOptions, Option<
         tailwind::Decision::Off => (None, Vec::new(), None),
     };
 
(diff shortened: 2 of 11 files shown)


baseballyama:

**Verdict: approve — one non-blocking should-fix (being addressed).** Reviewer rebuilt in an isolated worktree with the real oracle stack (oxfmt 0.59 + tailwindcss + prettier-plugin-tailwindcss insiders build) and wrote independent boundary-case byte-comparisons.

Verified fine:
- Nested-call scoping (`cn(a, notcn("x"))` vs `cn(a, clsx("x"))`), member-callee exclusion, object-key/array sorting — all byte-identical to real oxfmt.
- `class:` directives are excluded **structurally** (`ClassDirective` renders via a separate path), not by name-matching — no pattern to slip through.
- Splice safety: escaped quotes, multibyte text, single-quoted strings all match the oracle; byte-offset spans applied right-to-left.
- Zero-regression guarantee is structural: with `sortTailwindcss` unset, `class_sorter` is `None` and the new code paths are never entered.
- #1686 coexistence confirmed: the two-pass collect→sidecar-sort→apply flow already propagates `tailwind_functions` via `options.clone()`; `tailwind_sidecar.rs` untouched. Only trivial rebase adjacency in `main.rs`/`tailwind.rs`, no semantic conflict.
- Changeset (minor) and `rsvelte_fmt_wasm` struct update appropriate.

Should-fix (in progress): substitution-bearing template literals — real oxfmt sorts the static quasi parts of `` cn(`flex m-2 p-4 ${x}`) `` while this PR skips any template with expressions (`tailwind_sort.rs` requires `expressions.is_empty()`). Zero corpus hits today, but it's the canonical `cn()` usage pattern.

baseballyama:

**Verdict: mergeable.** The per-quasi template sorting resolves the should-fix. Reviewer re-verified against the real oxfmt + prettier-plugin-tailwindcss oracle with fresh boundary cases beyond the PR's own tests: whitespace-only quasi between two substitutions, back-to-back `${x}${y}` (non-trivial: the token abutting each substitution is pinned while the rest sorts — matches oxfmt exactly), and multi-space normalization. The originally-reported `cn(\`flex m-2 p-4 ${x}\`)` divergence now matches the oracle. All previous non-template cases re-checked with no regression; changeset/PR body correctly narrowed the out-of-scope note to mixed static-plus-`{expr}` attribute values. CI green so far, a few long jobs still running.

Issue #1703, opened 2026-07-23
bug: nested `& + &` sibling rules are wrongly dropped as `(empty)`


baseballyama:
rsvelte's unused-CSS analysis drops a nested rule whose inner selector uses the parent-selector sibling combinator (`.a { & + & { … } }`, i.e. `.a + .a`) as `/* (empty) */`, even when a real adjacent pair of `.a` elements exists. `svelte/compiler` scopes the outer selector and keeps the nested rule.

## Minimal reproduction

```svelte
<div class="a"></div><div class="a"></div>

<style>
	.a {
		& + & { color: red; }
	}
</style>
```

### Expected (`svelte/compiler`)

```css
.a.svelte-xxx {
	& + & { color: red; }
}
```

### Actual (rsvelte)

```css
/* (empty) .a {
	& + & { color: red; }
}*/
```

## Conditions

- Reproduces across essentially every markup context that produces the adjacent pair (literal siblings, `{#each}` variants, `{#if}` branches, …) — it is not context-specific.
- Independent of any other template content.
- Identical for `generate: 'client'` and `'server'`.

## Suspected cause

rsvelte's nested-rule (`&`) expansion is not consulted during sibling-relationship pruning: the inner `& + &` is evaluated without substituting the parent selector, so the sibling walk never finds a match, the inner rule is judged unused, and the outer rule is then dropped as empty.

## Provenance

Found by the CSS-prune differential sweep (`scripts/compat-corpus/css-prune-sweep.mjs`, added for #1700): 51 of 117 divergences cluster on this root cause, tracked in `compatibility/css-prune-known-failures.json`. Representative ids: `A/&+&/literal/none`, `A/&+&/each_all/none`. Unrelated to the #1700 `<svelte:head>` bug.

Issue #1702, opened 2026-07-23
bug: `:global(.a) + .b` wrongly pruned as unused inside `{#await}` branches and snippets


baseballyama:
rsvelte's unused-CSS analysis prunes `:global(.a) + .b` as `/* (unused) */` when the matching sibling pair lives inside an `{#await}…{:then}` branch or a `{#snippet}` fragment rendered with `{@render}`. `svelte/compiler` keeps the rule and scopes the trailing segment (`.a + .b.svelte-xxx`).

## Minimal reproduction

```svelte
<script>
	let promise = Promise.resolve(1);
</script>

{#await promise}{:then _}<div class="a"></div><div class="b"></div>{/await}

<style>
	:global(.a) + .b { color: red; }
</style>
```

### Expected (`svelte/compiler`)

```css
.a + .b.svelte-xxx { color: red; }
```

### Actual (rsvelte)

```css
/* (unused) :global(.a) + .b { color: red; }*/
```

## Conditions

- Asymmetric: `.a + :global(.b)` is **not** affected — only a `:global(…)` **leading** segment followed by a scoped sibling.
- Context-specific: the same selector with the sibling pair in `{#each}` / `{#if}` / `{#key}` / nested `{#each}` contexts analyses correctly. Only `{#await}` branches and `{#snippet}` + `{@render}` fragments diverge.
- Independent of any other template content (no corruptor node needed).
- Identical for `generate: 'client'` and `'server'`.

## Suspected cause

The await-branch / snippet fragment child-lists are not traversed when resolving possible siblings for a `:global`-prefixed leading selector segment.

## Provenance

Found by the CSS-prune differential sweep (`scripts/compat-corpus/css-prune-sweep.mjs`, added for #1700): 12 of 117 divergences cluster on this root cause, tracked in `compatibility/css-prune-known-failures.json`. Representative ids: `A/:global(.a)+.b/await_then/none`, `A/:global(.a)+.b/snippet_render/none`. Unrelated to the #1700 `<svelte:head>` bug.

PR #1708, opened 2026-07-23
fix(css): keep sibling-combinator rules past svelte:head void elements

### LINKED ISSUE #1700: bug: sibling-combinator CSS rules are wrongly pruned as unused
rsvelte's unused-CSS analysis marks a **sibling-combinator** rule (`+` / `~`) as `/* (unused) */` when the
matching elements are generated by an `{#each}` block **and** the component contains a `<svelte:head>` with a
void element in it. `svelte/compiler` keeps the rule.

Because the pruned rule is emitted as a CSS comment, minification strips it entirely — the style silently
disappears from the production build. There is no warning and no build failure.

## Environment

|                                      |                                                                                  |
| ------------------------------------ | -------------------------------------------------------------------------------- |
| `@rsvelte/vite-plugin-svelte-native` | 0.2.8 (darwin-arm64)                                                             |
| `svelte`                             | 5.56.7                                                                           |
| Bun                                  | 1.3.14                                                                           |

Reproduces on all four `generate` × `dev` combinations: `server`/`client`, `dev: true`/`false`.

## Minimal reproduction

`repro.ts` — depends only on `svelte` and `@rsvelte/vite-plugin-svelte-native`:

```ts
import { compile as svelteCompile } from 'svelte/compiler';
import { compile as rsvelteCompile } from '@rsvelte/vite-plugin-svelte-native';

const SRC = `<script>let items = [1, 2];</script>

<svelte:head>
  <meta name="viewport" content="width=device-width" />
</svelte:head>

<section>
  {#each items as i}
    <p class="a">{i}</p>
  {/each}
</section>

<style>
  .a + .a {
    color: red;
  }
</style>`;

const opts = { generate: 'server', dev: false, css: 'external', name: 'Repro' };

console.log('--- svelte/compiler ---');
console.log(svelteCompile(SRC, opts).css.code.trim());
console.log('--- rsvelte ---');
console.log(rsvelteCompile(SRC, opts).css.code.trim());
```

```sh
bun repro.ts
```

### Expected (what `svelte/compiler` emits)

```css
.a.svelte-s7tohh + .a:where(.svelte-s7tohh) {
    color: red;
  }
```

### Actual (rsvelte)

```css
/* (unused) .a + .a {
    color: red;
  }*/
```

## Conditions

All three ingredients are required — removing any one makes rsvelte keep the rule:

| variant                                                       | rsvelte prunes? |
| ------------------------------------------------------------- | --------------- |
| as written above                                              | **yes**         |
| without the `<svelte:head>` block                             | no              |
| `<title>t</title>` instead of `<meta … />` in the head        | no              |
| two literal `<p class="a">` siblings instead of the `{#each}` | no              |

The head element must be a void element — `<meta />` and `<link />` both trigger it, `<title>` alone does not.
`<title>` plus a `<meta />` does trigger it, so it is the presence of the void element that matters, not the
absence of other head content.

Only sibling combinators are affected; descendant, child and pseudo-class selectors analyse correctly:

| selector         | rsvelte prunes? |
| ---------------- | --------------- |
| `.a + .a`        | **yes**         |
| `.a ~ .a`        | **yes**         |
| `p + p`          | **yes**         |
| `section > .a`   | no              |
| `.a .c`          | no              |
| `.a:first-child` | no              |

## Suspected cause

The element-matching pass that decides whether a sibling combinator can be satisfied appears to lose track of
the fact that an `{#each}` body can produce multiple adjacent instances of the same element, but only once the
sibling-traversal walks past a `<svelte:head>` subtree containing a self-closing/void element. That the head
content — which cannot participate in the selector at all — changes the outcome suggests the head fragment is
being included in, or is corrupting the bounds of, the sibling scan.

baseballyama:
Fixes #1700.

## Root cause

The bug was not in the prune algorithm itself but in `phases/2_analyze/control_flow.rs`, which builds the sibling data the prune walker consumes. `collect_elements_and_paths` assigns `dom_idx` with its own counter but did not descend into `svelte:*` wrapper nodes (`<svelte:head>` etc.), while the analysis visitor that builds `dom_structure.elements` does descend. A void element inside `<svelte:head>` (`<meta />` / `<link />` — counted by the visitor, skipped by `collect_elements_and_paths`) therefore shifted every subsequent element's sibling-data slot by one, so an `{#each}`-generated element's wrap-around prev-sibling was recorded against the wrong element and `.a + .a` was judged unused. `<title>` doesn't trigger it because `TitleElement` is non-scopable and carries no `dom_idx` — matching the issue's odd "void element only" condition exactly.

The same desync also manifests in the opposite direction (`{#if}`/`{:else}` + head void makes rsvelte *keep* a rule the official compiler prunes), and `<svelte:component>` children were affected by the same gap.

## Fix

Align `collect_elements_and_paths` and `mark_opaque_in_fragment` with the exact wrapper set the analysis visitor descends into (`SvelteHead` / `SvelteFragment` / `SvelteBoundary` / `SvelteBody` / `SvelteWindow` / `SvelteDocument` / `SvelteSelf` / `SvelteComponent` / `TitleElement`), so both counters stay in lockstep.

## Tests

- New regression suite `css_svelte_head_sibling_1700.rs`: 9 cases covering the issue's full conditions table (`+` / `~` / `p + p`, `<meta>` / `<link>` / title-only / no-head, `{#each}` vs literal siblings, negative controls), all asserted against verified official-compiler output.
- CSS fixture suite 181/181, validator suite 333/333 (release), `cargo fmt` / `clippy -D warnings` clean.

## New: CSS-prune differential sweep (why the corpus never caught this)

The happy-path corpus compares real-world code, and real components almost never combine "sibling combinator × `{#each}` siblings × void element in `<svelte:head>`". This PR adds a combinatorial differential harness, `scripts/compat-corpus/css-prune-sweep.mjs`: **1222 deterministic synthetic components** (15 selector shapes × 10 sibling-producing contexts × 13 unrelated "corruptor" nodes), each compiled with both the official `svelte/compiler` and rsvelte, `css.code` byte-compared (scope hash normalized). Client and server prune decisions were verified identical (`--both`), so the sweep runs one target.

- Before the fix: 117 divergences in 3 root-cause clusters (the #1700 cluster plus two pre-existing, unrelated bugs).
- After the fix: **81, regression 0 — the #1700 cluster is fully eliminated.** The remaining 81 are tracked as #1702 (`:global(.a) + .b` inside `{#await}`/snippet, 16) and #1703 (nested `& + &` dropped as `(empty)`, 65), ratcheted shrink-only in `compatibility/css-prune-known-failures.json` with per-cluster justification in the paired `.md`, and gated in CI via `pnpm run corpus:css-prune:check` in the existing corpus job.

Reviewed by a separate agent pass: wrapper descent-set verified exhaustively against every `fragment::analyze` call site (no gaps, no double-count), harness normalization verified unable to mask real diffs, CI wiring verified (binding staged earlier in the same job).

https://claude.ai/code/session_01ANXA3h61KPdKdhPLRrSbi4


baseballyama:
## Sonnet 5 code review (orchestrated, post-merge)

**Verdict: approve** — no blocking findings. Root-cause analysis verified against the analysis visitor dispatch table (`visitors/mod.rs`): the 9 wrapper variants added to `collect_elements_and_paths` / `mark_opaque_in_fragment` exactly match the set the visitor descends into, with no remaining un-descended fragment-bearing variants (`SvelteOptions` rejects children, so its absence is harmless). All 9 new regression tests plus the existing CSS/validator suites pass on the PR branch.

**should-fix** (deferred to the #1702/#1703 follow-up PR): of the 9 added wrapper arms, only `SvelteHead`/`SvelteFragment`/`SvelteBoundary`/`SvelteSelf`/`SvelteComponent` can actually hold scopable children, and only the `SvelteHead` path has a positive desync test. Add at least one positive regression test placing a scopable element inside e.g. `<svelte:boundary>` before an `{#each}` sibling group.

**nits**
- `css-prune-sweep.mjs` comments reference issue numbers directly (tolerated for corpus/ratchet tooling, which justifies entries by issue in `compatibility/*.md`).
- PR body slightly overstates `SvelteBody`/`SvelteWindow`/`SvelteDocument`: they reject children at validation, so those arms are defensive symmetry, not a real desync fix.

PR #1714, opened 2026-07-23
fix(deps): repair broken pnpm-lock.yaml after concurrent lockfile merges

baseballyama:
## Summary
- CI on `main` fails with `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY: Broken lockfile: no entry for 'prettier@3.9.5' in pnpm-lock.yaml`.
- Root cause: merging the sass (#1616), prettier (#1623), and svelte (#1670) Renovate PRs left `prettier-plugin-tailwindcss`'s peer-resolution key pinned to the pre-bump `prettier@3.9.5`, while the root `prettier` entry had already moved to `3.9.6` elsewhere in the lockfile. There was no textual merge conflict, so this semantic mismatch slipped through.
- Fixed by regenerating the lockfile with `pnpm install --no-frozen-lockfile`. Only the stale peer-dependency suffixes for `prettier-plugin-tailwindcss` changed (`3.9.5` → `3.9.6`); no `package.json` files were touched.
- `apps/playground` has its own independent lockfile and was not touched.
- Added an empty changeset since this PR's `fix:` title would otherwise trip the changeset-required CI gate; no published package version changes here.

## Test plan
- [x] `pnpm install --frozen-lockfile` now succeeds against the repaired lockfile (previously failed with the same error CI reports).
- [x] `git diff` confirms only the two stale `prettier@3.9.5` peer-key references were updated to `3.9.6` — no other dependency drift.
- [x] No `package.json` changes in the diff.

https://claude.ai/code/session_013bPLQvV4p4hkT4u2qGrmRD


baseballyama:
Closing as redundant: main's broken pnpm-lock.yaml (missing prettier@3.9.5 entry) was already repaired by the lockfile reconciliation included in #1658's merge. main no longer references prettier@3.9.5 and installs are green. Thanks for the quick turnaround — the diagnosis here was correct and confirmed the root cause.

Issue #1689, opened 2026-07-22
AST fidelity: FunctionExpression/ArrowFunctionExpression field order diverges from acorn (expression/generator/async)


baseballyama:
Found while fixing #1659 (PR #1688): acorn's `initFunction` emits fields in the order `id, expression, generator, async` uniformly for `FunctionDeclaration` / `FunctionExpression` / `ArrowFunctionExpression`. PR #1688 fixed `FunctionDeclaration` (adding the missing `expression` right after `id`), but this codebase's `FunctionExpression` serializes as `generator, async, expression` — diverging from acorn's key order in `parse()` JSON output.

Key order is observable to consumers doing strict structural comparison of serialized output (same class as #1659, lower impact since the fields themselves are all present).

Scope:
- `FunctionExpression` (and check `ArrowFunctionExpression`, which acorn also runs through `initFunction`; note arrows have `expression: true/false` semantics, not constant `false`)
- Reference: acorn `initFunction`, and existing fixtures under `fixtures/<sha>/parser-legacy/*/ast.json`
- Implementation site: `crates/rsvelte_core/src/ast/typed_expr.rs` `Serialize` impls (see PR #1688 for the FunctionDeclaration template, including the napi raw-parse envelope + JS decoder sync)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1715, opened 2026-07-23
fix(parse): emit object-method typeParameters after body (#1711)

### LINKED ISSUE #1711: AST fidelity: object-method generics — typeParameters position diverges from acorn (fixed serialize position vs after body)
Follow-up from PR #1709 (review finding). Pre-existing non-match, **not** a regression.

## Symptom
For object-method generics, e.g.

```ts
const o = { m<T>(x: T) { return x } }
```

acorn-typescript (svelte/compiler `parse()`) emits `typeParameters` on the inner `FunctionExpression` **after `body`**, while rsvelte serializes it in the fixed declaration/expression position **between `async` and `params`**.

## Why it wasn't fixed in #1709
- Class methods were fixed there: their conversion sites are distinguishable, so the inner `FunctionExpression` now omits `typeParameters` (acorn attaches them to the `MethodDefinition`), see `class_method_generics_stay_off_the_inner_function`.
- Object methods route through the same generic `convert_expression` path as standalone function expressions, so they can't be distinguished at the converter; and unlike the class-method case acorn *does* emit the key on the inner function — only the serialized position differs.

## Possible directions
- Thread a "method context" flag from the ObjectExpression property conversion site, switching the serialize position to after `body`.
- Or model `typeParameters` position per construction context rather than per node type.

No corpus/fixture coverage hits this pattern today (checked in #1709's review). Same consumer-visible AST-fidelity class as #1689/#1692/#1694.

Refs: PR #1709 (commit `a51783c3` discussion), #1694.

baseballyama:
Closes #1711

## Problem
For object-method generics, e.g.

```ts
const o = { m<T>(x: T) { return x } }
```

acorn-typescript (svelte/compiler `parse()`) emits `typeParameters` on the inner `FunctionExpression` **after `body`**, while rsvelte serialized it in the fixed declaration/expression slot **between `async` and `params`**. Pre-existing non-match found in #1709's review (same AST-fidelity class as #1689/#1692/#1694/#1709), not a regression.

Object methods route through the same generic `convert_expression` path as standalone function expressions, so the inner function can't be distinguished at the converter — and unlike class methods, acorn *does* keep the key on the inner function; only the serialized position differs.

## Fix
- `JsNode::FunctionExpression` gains a `type_parameters_after_body` flag.
- The ObjectExpression property conversion sites (both the expression- and program-context paths) set it via a small `mark_object_method_generics` helper when `p.method` is true. All other construction sites pass `false`, so declarations, standalone expressions, and class methods (which already omit the key, #1709) are unchanged.
- Both serializers honor the flag: the serde `Serialize` impl and the NAPI raw-parse wire writer + `parse-envelope.js` reader. The binary envelope `VERSION` is bumped `3 → 4` (writer, JS reader, validation script) for the added flag byte.

## Verification
- Confirmed acorn's real output via `submodules/svelte` `parse()`: object-method inner `FunctionExpression` keys end `…async,params,body,typeParameters`; standalone keeps `…async,typeParameters,params,body`.
- New regression test `object_method_generics_emit_type_parameters_after_body` (asserts the `typeParameters` subtree is present and ordered after `body`), mirroring #1709's `class_method_generics_stay_off_the_inner_function`.
- `cargo test -p rsvelte_core`: parse_ts_types 22/22, parser suites, `--lib` 1545/1545; `parse-envelope` validation script green; `cargo clippy --all-targets --all-features -D warnings` and `cargo fmt` clean; `rsvelte_napi` checks.

### DIFF
--- a/.changeset/fix-function-node-ast-fidelity.md
+++ b/.changeset/fix-function-node-ast-fidelity.md
@@ -5,7 +5,7 @@
 
 fix(parse): improve function-node AST fidelity to match acorn / acorn-typescript
 
-Three parse-AST fixes so the public `parse()` output matches svelte/compiler:
+Four parse-AST fixes so the public `parse()` output matches svelte/compiler:
 
 - `FunctionExpression` fields are ordered `id, expression, generator, async` to
   match acorn's uniform `initFunction` key order (#1689).
--- a/.changeset/fix-function-node-ast-fidelity.md
+++ b/.changeset/fix-function-node-ast-fidelity.md
@@ -18,8 +18,11 @@ Three parse-AST fixes so the public `parse()` output matches svelte/compiler:
   declarations (#1692). As a side effect, this also fixes a pure-JS bug where a
   default-valued arrow parameter (`(a = 1) => a`) lost its `AssignmentPattern`
   (default value) in the `parse()` output — `compile()` output was unaffected.
+- Object-method values (`{ m<T>(x: T) {} }`) keep their generics on the inner
+  `FunctionExpression` but emit `typeParameters` *after* `body` (like arrows),
+  not in the declaration/expression slot before `params` (#1711).
 
 The binary NAPI raw-parse envelope (consumed by
 `@rsvelte/vite-plugin-svelte-native`'s `parse-envelope.js` decoder) carries the
 same fields, so both packages need this release. The envelope `VERSION` is
-bumped to 3 alongside the wire-format change.
+bumped to 4 alongside the wire-format changes.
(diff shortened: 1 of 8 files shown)


baseballyama:

**Verdict: approve** — no blocking findings. Reviewer reproduced the acorn-typescript ground truth locally (`submodules/svelte` deps installed, real `parse()` runs): object-method variants (shorthand/async/generator/computed-key) all emit `typeParameters` after `body`, while `{ m: function<T>(){} }` (non-method function value, `p.method === false`) keeps the standalone position — both match this PR's behavior. All `JsNode::FunctionExpression` construction/match sites audited for flag leaks (none); wire writer/reader symmetry and the VERSION 3→4 bump verified across writer, reader, and validation script. Targeted suites re-run on the PR branch: parse_ts_types 22/22, lib 1545/1545, fmt/clippy clean, envelope validation pass.

**should-fix** (being applied): `.changeset/fix-function-node-ast-fidelity.md` intro still says "Three parse-AST fixes" but now lists four.

**nits**
- getter/setter generics (`get m<T>()`) would diverge (oxc reports `method: false`), but TypeScript rejects that pattern, so it is unreachable in valid input — out of scope for #1711.
- Regression coverage extended with async/generator-method and standalone-position variants (being applied).

PR #1716, opened 2026-07-23
fix(css): keep nested `& + &` and `:global(.a) + .b` sibling rules (#1702, #1703)

### LINKED ISSUE #1702: bug: `:global(.a) + .b` wrongly pruned as unused inside `{#await}` branches and snippets

### LINKED ISSUE #1703: bug: nested `& + &` sibling rules are wrongly dropped as `(empty)`

baseballyama:
Fixes two CSS unused-selector prune divergences found by the css-prune differential sweep (`scripts/compat-corpus/css-prune-sweep.mjs`, added in #1708). Both were pre-existing, unrelated to the #1700 `<svelte:head>` bug.

## #1703 — nested `& + &` dropped as `(empty)`

`.a { & + & { … } }` (i.e. `.a + .a`) was dropped as `/* (empty) */` even when a real adjacent `.a` pair existed. The official compiler scopes the outer selector and keeps the nested rule (`.a.svelte-X { & + & {} }`).

**Root cause:** the transform's `is_sibling_combinator_unused` built the before/after `SelectorInfo` for `&` via `extract_selector_info`, which ignores a `NestingSelector` and yields an empty (matches-nothing) info. The sibling walk then found no element matching `&`, judged `& + &` unused, and the declaration-less outer rule was dropped as empty.

**Fix:** `extract_selector_info_resolving_nesting` resolves `&` against the parent rule's subject compound (added as an `:is(...)`-style OR-group) before matching, mirroring upstream's `NestingSelector` branch in `relative_selector_might_apply_to_node`.

## #1702 — `:global(.a) + .b` pruned inside await/snippet

`:global(.a) + .b` was pruned as `/* (unused) */` when the matching sibling pair lived inside an `{#await}…{:then}` branch or a `{#snippet}` fragment rendered with `{@render}`. The official compiler keeps it and scopes the trailing segment (`.a + .b.svelte-X`). Asymmetric: `.a + :global(.b)` was unaffected, and `{#each}`/`{#if}`/`{#key}` already worked.

**Root cause:** `{#await}` branches and `{#snippet}` bodies both set `css.has_opaque_elements`, which forced the `:global(X) + Y` prune check down a branch that only accepted `Y` when it immediately followed an opaque boundary — a real previous sibling `.a` is not one. `{#each}`/`{#if}`/`{#key}` do not set the flag, so they took the root-child branch and matched.

**Fix:** union the acceptable predecessors of `Y` — a real previous sibling matching the inner `:global(...)` selector, an opaque boundary, or `Y` being a root-level element (the global `.a` may be injected by the parent) — mirroring upstream `apply_selector` BACKWARD plus its `get_element_parent(node) === null && every_is_global(...)` fallback.

## Result

- css-prune differential sweep ratchet: **81 → 0** (65 `& + &` + 16 `:global` entries cleared); `compatibility/css-prune-known-failures.json` is now empty and the justification `.md` records the fixes.
- New regression tests: `css_nested_sibling_1703.rs`, `css_global_sibling_1702.rs`.
- Follow-up from the merged #1708 review: positive `dom_idx`-desync tests for the remaining `svelte:*` wrappers (`<svelte:boundary>`, `<svelte:component>`, `<svelte:self>`, `<svelte:fragment>`) that carry a scopable child — `css_wrapper_sibling_desync_1708.rs`.
- Existing CSS / compiler-fixture / ssr / validator / compatibility-report suites pass; `cargo fmt` + `cargo clippy --all-targets --all-features -D warnings` clean.

Closes #1702
Closes #1703


baseballyama:

**Verdict: approve.** All round-1 findings resolved and re-verified against the real `svelte/compiler` output (direct import, byte-level comparison):

- rustfmt diff fixed (verified with CI-pinned 1.97.1); root cause was local stable wrapping `format!` differently.
- Negative controls added for both fix sites (4 new tests; all byte-match the official compiler).
- Comment policy nit resolved.
- Over-relaxation fixed conservatively: helpers now resolve only single-relative selectors. Reviewer confirmed the two round-1 divergences (`:global(.a .z) + .b` without matching ancestor, `.foo > .a { & + & }` without matching parent) now byte-match official output, and re-ran the full single-relative matrix (selector-list parents, cross-branch, multi-level nesting, compound nesting) with no regressions.
- Positive-side cases of combinator-bearing compounds (ancestor genuinely matches) still over-prune vs official — measured **byte-identical to pre-PR main**, i.e. a pre-existing limitation this PR neither introduced nor worsened. Tracked separately.
- Sweep re-run on the final HEAD: 1222/1222, 0 divergences. `--test css` 181/181, clippy (CI toolchain) clean.

PR #1718, opened 2026-07-23
fix(fmt): break whitespace body of a wrapping inline element (#1707)

### LINKED ISSUE #1707: fmt: non-idempotent output for whitespace-content elements with a wrapping open tag (trailing-space `> ` in the non-port path)
Found while fixing #1699 (PR #1706), but **pre-existing on main** and independent of that PR's two cases.

## Symptom
A source-*whitespace* element whose open tag wraps across lines — e.g. an inline element with a long attribute and a whitespace-only body:

```svelte
<span class="very-long-class-that-forces-the-open-tag-to-wrap …"> </span>
```

formats with a trailing space after the wrapped `>` (`> ⏎`) in the non-port path, and re-formatting the output does not reproduce it byte-identically (non-idempotency).

## Scope notes
- Reproduces on main before PR #1706; not one of #1699's two divergences, so it was deliberately left untouched there to avoid a net-negative expansion.
- Likely interacts with the whitespace-artifact handling moved into `collapse.rs::layout_children` by PR #1706 — worth re-checking the repro after that PR merges.

Refs: #1699, PR #1706 (discovery context), PR #1687 (bracketSameLine port rules).

baseballyama:
Closes #1707

## Problem

An inline element whose body is whitespace-only and whose open tag wraps because of a long attribute —

```svelte
<span class="very-long-class-that-forces-the-open-tag-to-wrap-…"> </span>
```

— was formatted by the non-port (edit-based) markup path with the raw whitespace glued after the wrapped `>`:

```
<span
  class="…"
> </span>
```

This both **diverged from the prettier-plugin-svelte oracle** and was **non-idempotent** — a multi-space body (`>   </span>`) collapsed to a single space (`> </span>`) on a re-format.

## Root cause

prettier prints such an element as `group([...openingTag, '>', line, '</tag>'])`. `shouldHugStart`/`shouldHugEnd` are both false (the whitespace body blocks hugging), so it is *not* the hugged empty branch. The `line` body flexes to a newline once the wrapped open tag forces the group open, so the oracle produces (for `bracketSameLine: true`):

```
<span
  class="…">
</span>
```

The edit-based path instead kept the raw source whitespace verbatim and left the close tag glued, so the body never broke.

## Fix

Port that shape into the edit-based path (`crates/rsvelte_formatter/src/markup.rs`):

- `is_empty_nonhug_element` detects an **inline** element with a whitespace-only body.
- Under a wrapping open tag, the open `>` now glues to the last attribute line under `bracketSameLine` (dedents to its own line otherwise), matching `push_open_tag`'s hug/non-hug distinction.
- `push_close_tag` rewrites the whitespace body + close tag as `\n{indent}</tag>`, so the body breaks and the close tag drops to its own line.

Output is now **byte-identical to the oracle and idempotent for both `bracketSameLine` values**.

### `<textarea>` (raw-text exception)

`<textarea>` is whitespace-sensitive raw text, so the oracle keeps its `>` **glued** (never dedented) for this shape. Under the default `bracketSameLine: false` it also glues the close tag and **drops** the whitespace body (`…"></textarea>`); `bracketSameLine: true` breaks the body (`…">` newline `</textarea>`). The whitespace-body `<textarea>` is exempted from the empty-`<textarea>` open-tag width dangle, and its close tag is glued under the default — so both option values are byte-identical to the oracle. A source-empty `<textarea>` still dangles its `>` on overflow (unchanged).

The default-`false` `<textarea>` re-format is non-idempotent, but this is **inherited from the oracle itself**: oxfmt drops the whitespace body, turning the node into a truly-empty `<textarea>` that it then formats differently (`\n></textarea>`). rsvelte matches the oracle at *every* pass (pass 2 == `oracle(pass-1 output)`), so parity is exact — idempotency and oracle-parity are mutually exclusive here because the oracle discards the whitespace content.

### Comment findings

- **Composite case (comment 1)** — the genuine-whitespace-content + attribute-caused-wrap case is exactly what this fixes for **inline** elements. Verified byte-identical to the oracle for `bracketSameLine` true *and* false, single- and multi-space bodies, top-level and nested-in-block, plus the `<textarea>` raw-text variant.
- **Maintainability (comment 2)** — `can_omit_softline_before_closing_tag` now tests `is_html_block_display_element` (prettier's `blockElements`, which excludes `script`/`style`) instead of the slightly wider `is_block_element`, matching `isLastChildWithinParentBlockElement` and its collapse-pass mirror `omit_softline_allowed`.

### Deliberately out of scope

- **Source-empty inline** (`<span></span>`): already hugs correctly (`></span>`) — untouched.
- **Block-display elements** with a wrapping open tag (whitespace body or truly empty): left to the collapse pass's own restructuring. These have a residual divergence from the oracle **at both `bracketSameLine` values** (not just `true`), but the behaviour is **unchanged from `main`** (not a regression), **hits nothing in the svelte.dev parity corpus** (1102/1103, see below), and folding it in would be net-negative — including block elements made the `bracketSameLine: true` case non-idempotent because the collapse pass strips an empty block body's inserted newline, so a second format re-dedents the `>`. The block-display divergence should be tracked in a separate issue.

## Verification

- **svelte.dev formatter-parity hard gate** (`svelte_dev_corpus`): `1102/1103` — identical to `main` (confirmed by stashing the change and re-running, before and after the `<textarea>` follow-up). The sole failure is a pre-existing local toolchain skew (oracle generated with `oxfmt` 0.56.0 vs the worktree's Rust `oxc_formatter` 0.59.0 differing on one TS union type in `SearchBox.svelte`), not related to this change.
- New regression tests in `crates/rsvelte_formatter/tests/options.rs` cover parity (both option values, top-level + nested), idempotency (multi-space body), and the `<textarea>` raw-text layout (whitespace-body + source-empty).
- Full `rsvelte_formatter` + `rsvelte_fmt` suites, `cargo fmt`, and `cargo clippy --all-targets --all-features -D warnings` all clean.
- `@rsvelte/fmt` changeset added (`.changeset/fix-fmt-wrapping-whitespace-inline.md`).
- `fmt-known-failures.json` left unchanged: this narrow construct fix does not verifiably flip any whole ecosystem-corpus file to passing, and CI runs `fmt-verify` non-strict (a newly-passing entry is a warning, not a failure), so no ratchet edit is warranted or safe without the ecosystem submodules (absent in this worktree).

### DIFF
--- a/.changeset/fix-fmt-wrapping-whitespace-inline.md
+++ b/.changeset/fix-fmt-wrapping-whitespace-inline.md
@@ -24,3 +24,9 @@ and the default `false` glues the close tag and drops the whitespace body
 block-display elements are left to their existing layout. Also aligns
 `can_omit_softline_before_closing_tag` with prettier's `blockElements` (excluding
 `script`/`style`) via `is_html_block_display_element`.
+
+Also fixes a crash this uncovered: when the new whitespace-body edit overlapped
+the indent pass's edit on the same text node, the top-level section-span remap
+double-counted the dropped edit and shifted a later `<style>`/`<script>` boundary
+past the output end, panicking during section reorder. The remap is now computed
+from the overlap-resolved edit set, so it always agrees with the applied output.
--- a/.changeset/fix-fmt-wrapping-whitespace-inline.md
+++ b/.changeset/fix-fmt-wrapping-whitespace-inline.md
@@ -1,0 +1,26 @@
+---
+"@rsvelte/fmt": patch
+---
+
+fix(fmt): break the whitespace body of a wrapping inline element (#1707)
+
+An inline element with a whitespace-only body whose open tag wraps
+(`<span class="…long…"> </span>`) was formatted by the non-port markup path with
+the raw whitespace glued after the wrapped `>` (`…"`\n`> </span>`). That both
+diverged from the prettier-plugin-svelte oracle and was non-idempotent — a
+multi-space body collapsed to a single space on a re-format.
+
+prettier prints such an element as `group([...openingTag, '>', line, '</tag>'])`,
+so the whitespace body is a `line` that breaks once the wrapped open tag forces
+the group open: the `>` glues to the last attribute line under `bracketSameLine`
+(else dedents) and the close tag drops to its own line, absorbing the whitespace.
+Output is now byte-identical to the oracle and idempotent for both
+`bracketSameLine` values.
+
+`<textarea>` is handled as the raw-text exception the oracle applies: `>` stays
+glued (never dedented for this shape), `bracketSameLine: true` breaks the body,
+and the default `false` glues the close tag and drops the whitespace body
+(`…"></textarea>`). Source-empty inline elements (`<span></span>`, hug) and
+block-display elements are left to their existing layout. Also aligns
+`can_omit_softline_before_closing_tag` with prettier's `blockElements` (excluding
+`script`/`style`) via `is_html_block_display_element`.
--- a/crates/rsvelte_formatter/src/lib.rs
+++ b/crates/rsvelte_formatter/src/lib.rs
@@ -181,10 +181,50 @@ pub fn format(source: &str, options: &FormatOptions) -> Result<String, FormatErr
     let has_markup = root.fragment.nodes.iter().any(|n| {
         !matches!(n, rsvelte_core::ast::template::TemplateNode::Text(t) if crate::is_blank_text(t.data.as_str()))
     });
+
+    // Resolve overlapping edits ONCE, before both the section remap and the
+    // application, so the two never disagree. Two passes (markup.rs and indent.rs)
+    // can emit an edit for the same span — e.g. markup.rs replaces a whitespace
+    // body + close tag with `\n{indent}</tag>` at the same `[start, end)` that
+    // indent.rs would normalise to `\n{indent}`. Markup edits are pushed first
+    // (line 100 before line 105), so after the stable descending sort they appear
+    // before indent edits with the same start; the first one wins and the overlap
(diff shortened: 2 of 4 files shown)


baseballyama:
## Sonnet 5 code review (orchestrated) — 3 rounds

**Verdict: approve.** All findings across three rounds were verified empirically against the real prettier-plugin-svelte 4.1.1 oracle:

- **Round 1** (blocking): missing changeset → added (`@rsvelte/fmt: patch`, matching existing format).
- **Round 2**: `<textarea>` whitespace-body now byte-matches the oracle at **both** `bracketSameLine` values. Notably, the oracle itself is non-idempotent here (it drops the whitespace body, then reformats the now-empty textarea differently on pass 2); rsvelte matches the oracle at *every* pass (pass-N output == oracle(pass-N−1)), verified over 3 consecutive passes. The core inline fix (span et al.) matches prettier's `group([...openingTag, '>', line, '</tag>'])` model; the `is_html_block_display_element` alignment was cross-checked against prettier's 32-entry `blockElements` list item-by-item.
- **Round 3**: the Formatter-parity panic (layercake `Nav.svelte`, `sort_order.rs` OOB) was root-caused to the section-span remap summing length deltas of overlap-dropped edits. The lib.rs fix (resolve overlaps once, drive remap + application from the same `applied` set) is a pure refactor of the apply loop — overlap predicates, ordering, and winner selection unchanged. Reviewer reproduced the panic with the pre-fix lib.rs on the real corpus file, confirmed it gone after, and confirmed the fixed output is byte-identical to the oracle. The regression test was likewise verified to fail without the fix.

Deferred block-display divergence is tracked in #1721. Changeset wording extended to mention the crash fix (review nit).

PR #1728, opened 2026-07-23
fix(css): resolve multi-relative chains in :global()/nested-& sibling prune

### LINKED ISSUE #1719: bug: combinator-bearing compounds in :global()/nested & are over-pruned (single-relative-only sibling resolution)
Known limitation surfaced during the PR #1716 review (pre-existing on main — byte-identical before/after #1716's fixes; #1716 deliberately kept it rather than trading over-prune for over-keep).

## Symptom
The sibling-combinator prune check resolves only **single-relative** selectors when expanding `:global(...)` contents and parent-rule preludes for `&`. A descendant/child chain inside the compound is left unresolved, so the rule prunes even when the ancestor constraint is actually satisfied:

- `:global(.a .z) + .b` where `.z` really has an `.a` ancestor and an adjacent `.b` sibling exists → official keeps `.a .z + .b.svelte-xxx`, rsvelte emits `/* (unused) */`.
- `.foo > .a { & + & {} }` where `.a` really is a child of `.foo` and an adjacent pair exists → official keeps the scoped rule, rsvelte emits `/* (empty) */`.

Negative counterparts (ancestor NOT satisfied) byte-match the official compiler and are pinned by tests in `css_global_sibling_1702.rs` / `css_nested_sibling_1703.rs`.

## Suspected fix direction
Apply the recursive ancestor-matching already used for `>` combinator child checks (css.rs, `3_transform`) to `:global(...)` inner selectors and parent-prelude resolution, instead of bailing to unresolved on multi-relative chains.

## Scope notes
- Not covered by the css-prune sweep grid (single compounds only) — consider adding a multi-relative axis to `scripts/compat-corpus/css-prune-sweep.mjs` when fixing.
- Same consumer-visible class as #1702/#1703 (prune fidelity), narrower pattern.

Refs: PR #1716 (review discussion), #1702, #1703.

baseballyama:
Fixes #1719

## Problem

The `+` / `~` sibling-combinator prune check resolved only **single-relative** selectors when expanding a leading `:global(X)`'s inner selector or a nested rule's `&` against its parent prelude. A descendant/child chain inside the compound was left unresolved, so the rule was pruned even when the ancestor constraint was actually satisfied:

- `:global(.a .z) + .b` where the `.z` sibling of `.b` really has an `.a` ancestor → official keeps `.a .z + .b.svelte-xxx`, rsvelte emitted `/* (unused) */`.
- `.foo > .a { & + & }` where `.a` really is a child of `.foo` with an adjacent pair → official keeps the scoped rule, rsvelte emitted `/* (empty) */`.

The negative counterparts (ancestor **not** satisfied) already byte-matched and stay pruned.

## Fix

Introduce a `SiblingMatcher` that carries either a compound (`Info`, the unchanged path) or a descendant/child ancestor-chain (`Chain`). The `Chain` variant is verified with the **same recursive ancestor matcher already used for `>` child checks** — `structural_element_matches_compound` + `structural_ancestors_satisfy_links` — mirroring upstream `css-prune.js` where `:global(X)` runs `apply_selector(X.children, …, BACKWARD)` and `&` runs `apply_selector(get_relative_selectors(parent), …)`.

A `Chain` is produced **only** for:
- a `:global(...)` inner that is a structurally-evaluable multi-relative descendant/child chain, or
- a bare `&` whose immediate parent prelude is a single structurally-evaluable multi-relative descendant/child chain.

Every other shape (single-relative, sibling combinators inside, functional pseudo, dynamic-unevaluable) falls back to the existing `Info` behaviour, so single-relative and all negative cases remain byte-identical.

## Verification

Output compared against `svelte/compiler` (5.56.4) — kept and pruned cases both match byte-for-byte:

| case | official | rsvelte (after) |
|---|---|---|
| `:global(.a .z) + .b` with `.a` ancestor | `.a .z + .b.svelte-x` | kept ✅ |
| `:global(.a .z) + .b` no `.a` ancestor | `/* (unused) */` | pruned ✅ |
| `.foo > .a { & + & }` `.a` child of `.foo` | scoped, kept | kept ✅ |
| `.foo > .a { & + & }` `.a` not child | `/* (empty) */` | pruned ✅ |

Tests (positive + negative, mirroring the existing #1702/#1703 pins):
- `css_global_sibling_1702.rs` — 9 passed (adds `global_descendant_inner_with_ancestor_kept`, `global_child_inner_with_ancestor_kept`)
- `css_nested_sibling_1703.rs` — 7 passed (adds `nested_amp_plus_amp_multi_relative_parent_child_kept`, `nested_amp_tilde_amp_multi_relative_parent_child_kept`)

Regression suites (all green): CSS scoping suite + all `css_*` pins, compiler snapshot (17), validator (16), SSR (16), runtime browser/hydration/runes/legacy.

## css-prune differential sweep

Added a multi-relative ancestor axis (Family C) plus a three-level nesting axis (Family C3, from the review follow-up) to `scripts/compat-corpus/css-prune-sweep.mjs`: the `:global(...)`-inner and nested-`&` shapes crossed with ancestor-present / root / wrong-ancestor / each-wrapped arrangements, and the 3-level `.grand { .foo > .a { & +/~ & } }` shapes crossed with full / full_each / no_grand / no_foo / flat arrangements, times the structural corruptors. Full sweep: **1430 components, 0 diverged** (`--check` passes; the shrink-only ratchet stays empty).

## Follow-up: review fix (multi-level ancestor resolution)

The review caught a regression at three-level nesting: the first pass resolved only the immediate parent prelude, so `.grand { .foo > .a { & + & } }` kept the rule even when `.foo` was not a descendant of `.grand`. Fixed by walking the full `parent_preludes` stack (`build_parent_chain`), linking each level with an implicit descendant combinator like upstream `get_relative_selectors` + `NestingSelector`. Verified against `svelte/compiler` for 2/3/4-level × ancestor present/absent × `+`/`~`; new pins added in `css_nested_sibling_1703.rs` (13 passed) and a 3-level sweep axis (Family C3).

## Known limitation (tracked in #1731)

The nested-`&` fix flattens the nesting stack into one descendant/child chain and bails all-or-nothing when an **intermediate** level is not a single evaluable compound. Three shapes therefore remain over-pruned — the official compiler keeps the inner rule, rsvelte drops the whole rule as `(empty)`:

- comma-separated intermediate level: `.grand, .other { .foo > .a { & + & {} } }`
- functional-pseudo intermediate level: `:is(.grand) { .foo > .a { & + & {} } }`
- sibling-combinator intermediate level: `.x + .grand { .foo > .a { & + & {} } }`

This is the **safe side** (over-prune, same direction as the pre-existing design — never over-keep) and matches main for these shapes; a proper per-branch resolution is a larger refactor deferred to #1731.

### DIFF
--- a/.changeset/fix-css-prune-multi-relative-sibling.md
+++ b/.changeset/fix-css-prune-multi-relative-sibling.md
@@ -1,0 +1,15 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(css): resolve multi-relative chains in `:global()`/nested-`&` sibling prune
+
+The `+`/`~` unused-CSS prune check resolved only single-relative selectors when
+expanding a leading `:global(X)` inner selector or a nested rule's `&` against
+its ancestor rules, so a descendant/child chain inside the compound was left
+unresolved and the rule was pruned even when the ancestor constraint was
+actually satisfied — e.g. `:global(.a .z) + .b` (`.z` really under `.a`) became
+`/* (unused) */` and `.grand { .foo > .a { & + & } }` became `/* (empty) */`.
+The `&`/`:global(...)` inner is now resolved through the full ancestor chain
+with the same structural matcher used for `>` child checks, matching
+`svelte/compiler` for both the kept and pruned cases (#1719).
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/css.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/css.rs
@@ -2367,7 +2367,11 @@ fn is_sibling_combinator_unused(rel_selectors: &[Value], ctx: &CssContext) -> bo
                 return false;
             }
 
-            let inner_info = global_inner_selector_info(&rel_selectors[0]);
+            // Resolve the inner `:global(X)` compound. A multi-relative chain
+            // like `:global(.a .z)` becomes a `Chain` so the `.a` ancestor of a
+            // candidate `.z` sibling is verified (rather than left unresolved,
+            // which over-prunes even when the ancestor really exists).
+            let inner_matcher = resolve_global_inner_matcher(&rel_selectors[0]);
 
             // `:global(X) + Y` is used when some Y is preceded by a node X could
             // be: a real previous sibling matching X, an opaque boundary, or a
@@ -2391,12 +2395,9 @@ fn is_sibling_combinator_unused(rel_selectors: &[Value], ctx: &CssContext) -> bo
                 } else {
                     &el.possible_prev_general
                 };
-                prevs.iter().any(|(idx, _)| {
-                    ctx.dom_structure
-                        .elements
-                        .get(*idx)
-                        .is_some_and(|sib| selector_matches_element(&inner_info, sib))
-                })
+                prevs
+                    .iter()
+                    .any(|(idx, _)| matcher_matches_at(&inner_matcher, *idx, ctx))
             });
 
             return !matches;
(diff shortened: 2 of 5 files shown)


baseballyama:
Addressed the review regression (3-level nesting resolved only one ancestor level).

**Chosen fix: root fix** (full recursive ancestor resolution), not the minimal len==1 bail — the minimal fix would still over-prune the 3-level *ancestor-satisfied* case (official keeps it), which would make the new sweep axis diverge. `resolve_bare_nesting_chain` now walks the entire `parent_preludes` stack, linking each enclosing level to the one below with an implicit descendant combinator (mirroring upstream `get_relative_selectors` + `NestingSelector`): `.grand { .foo > .a { & + & } }` → `.grand .foo > .a`, `.grand { .a { & + & } }` → `.grand .a`. Single-compound (one level, no ancestor) and any non-evaluable level still fall back to the compound `Info` path, so single-level/negative cases stay byte-identical. `resolve_global_inner_matcher` (the `:global(...)` path) is untouched.

**Verification against svelte/compiler (real build):** 2/3/4-level × ancestor present/absent × `+`/`~` all match byte-for-byte, including the reported repro (`.grand { .foo > .a { & + & } }` with `.foo` not under `.grand` → `/* (empty) */`).

**New regression tests** in `css_nested_sibling_1703.rs` (13 pass total): 3-level `+`/`~` satisfied→kept & missing→pruned, plus 2-level single-compound `.grand { .a { & + & } }` both directions.

**Sweep:** added a 3-level nesting axis (Family C3: `full` / `full_each` / `no_grand` / `no_foo` / `flat` × corruptors). Full sweep now **1430 components, 0 diverged** (`--check` passes, ratchet stays empty).

Suites re-run green: full CSS scoping suite + all `css_*` pins, compiler snapshot (17), validator (16). fmt + clippy clean.

PR #1779, opened 2026-07-25
refactor(analyze): drop the serde_json::Value AST from phase 2

baseballyama:
Phase 2's **expression/statement walkers** were written twice: once over the typed arena (`JsNode`) and once over a legacy `serde_json::Value` mirror. The `Value` half was already all but dead — this PR finishes that migration and deletes it.

**Scope:** this removes the `Value` walker family from `visitors/shared/utils.rs` plus the JSON branches of four visitors (`render_tag`, `await_block`, `const_tag`, `assignment_expression`). It is **not** a removal of `serde_json::Value` from phase 2 as a whole — `call_expression.rs`, `await_expression.rs`, `class_body.rs`, `blockers.rs`, `pattern_ids.rs`, `2_analyze/mod.rs` and `declaration_tag.rs`'s `warn_local_state_reads` (which still calls `as_json()`) all continue to use it. See "Out of scope" below for why those are a separate change.

Three commits: the first moves the last live call sites onto `JsNode`, the second removes the now-unreachable `Value` code, the third applies review feedback.

## Line counts (measured)

| File | Before | After | Δ |
|---|---:|---:|---:|
| `2_analyze/visitors/shared/utils.rs` | 3,568 | 1,749 | **−1,819 (−51.0 %)** |
| `2_analyze/visitors/await_block.rs` | 817 | 508 | −309 |
| `2_analyze/visitors/assignment_expression.rs` | 207 | 127 | −80 |
| `2_analyze/visitors/declaration_tag.rs` | 265 | 249 | −16 |
| `2_analyze/visitors/const_tag.rs` | 115 | 94 | −21 |
| `2_analyze/visitors/render_tag.rs` | 199 | 201 | +2 |

Diff total: **+238 / −2,481**.

A note on the pre-flight survey numbers: it reported 15 twin pairs / 2,856 duplicated lines and 23 external `Value` call sites (`walk_js_expression` ×8, `extract_identifiers` ×13, `get_global_keypath` ×2). Re-measuring on `main` found **20** `Value` functions in `utils.rs` and only **8** external call sites — most of the survey's extra hits were comments, or same-named local helpers in other modules (`blockers.rs`, `3_transform/client/**`, and `call_expression.rs` each define their own `extract_identifiers` / `get_global_keypath`; unrelated, untouched here).

## Stage 1 — call sites migrated to the `_node` path

| Call site | Change |
|---|---|
| `render_tag.rs:132` `walk_js_expression(callee_json, …)` | → `walk_js_expression_node`; the whole visitor now reads callee/arguments through the arena |
| `render_tag.rs:138` `walk_js_expression(arg, …)` | → `walk_js_expression_node` |
| `await_block.rs:128` | dead `else` branch removed |
| `await_block.rs:290` (`walk_expression_for_mutations`) | dead helper removed |
| `const_tag.rs:96` | dead `_ =>` fallback removed |
| `const_tag.rs:104` | dead `_ =>` fallback removed |
| `declaration_tag.rs:127` | dead `else` branch removed |
| `assignment_expression.rs:56` `extract_identifiers(target)` | removed together with its only caller, `mark_binding_mutation` (`Value`), whose only callers were `walk_js_expression` and the dead `await_block` JSON path |

`render_tag.rs` was the **only** live `Value` consumer. The others are unreachable three times over:

1. `Expression::Lazy` is constructed in exactly one place in the tree (`1_parse/state/tag.rs`, the `defer_script_parse` branch), reached only from the bare `{expr}` ExpressionTag path — so the four deleted sites are *structurally* incapable of holding a `Lazy`.
2. `resolve_lazy_expressions()` runs at the top of `analyze_component` and resolves every `Lazy` before analysis begins.
3. Even if one survived, the `as_json()` in those `else` arms *panics* on `Expression::Lazy` rather than producing a value — so the branch could never have executed.

The `const_tag` / `declaration_tag` fallbacks have a fourth, independent reason: they re-derive from `as_json()` the very node the `match` on `JsNode` just rejected.

Because (3) means the old code panicked on a `Lazy`, the three `await_block` sites now call `as_node_ref()` (which panics with a clear message) rather than silently skipping — a silent skip would have been the behaviour change. The `const_tag` / `declaration_tag` arms are a *different* invariant (unexpected declaration shape, not `Lazy`), so those stay silent, matching the no-op they replaced. Both carry a one-line invariant comment.

`render_tag.rs` was checked against upstream `2-analyze/visitors/RenderTag.js`. The arena version preserves the JSON version's exact behaviour, including erroring with `render_tag_invalid_expression` for anything without a callee, and accepting `NewExpression` — the old `.get("callee")` probe did not distinguish it from `CallExpression`. Note that accepting `NewExpression` is an rsvelte-only bug: upstream rejects `{@render new foo()}` at parse time (`1-parse/state/tag.js` L819-830). Fixing it belongs in the parser (it also affects svelte2tsx), so it is deliberately preserved here and tracked separately in **#1786**.

## Stage 2 — functions deleted

All 20 `serde_json::Value` functions in `2_analyze/visitors/shared/utils.rs`:

`walk_js_expression`, `walk_js_statement`, `validate_assignment`, `validate_no_const_assignment`, `is_pure`, `is_safe_identifier`, `is_reference`, `nodes_equal`, `extract_identifiers`, `collect_all_identifier_names_from_pattern`, `get_rune_name`, `get_global_keypath`, `get_name`, `object`, `ensure_no_module_import_conflict`, `has_variable_declaration`, `statement_declares_name`, `declarator_declares_name`, `pattern_declares_name`, `param_declares_name`

Plus, elsewhere in `2_analyze/visitors/`:

- `assignment_expression.rs`: `mark_binding_mutation`, `get_member_expression_root_name`
- `await_block.rs`: `walk_pattern_computed_keys`, `walk_expression_for_mutations`, `mark_mutations_recursive`, `collect_pickled_awaits`, `collect_pickled_awaits_inner`

Two surviving functions were edited rather than deleted:

- **`is_safe_identifier_node`** built a synthetic `serde_json::json!({"type":"Identifier", …})` and called the `Value` version to resolve the `$store` → `store` hop. It now recurses through a new `is_safe_identifier_name(&str, …)` helper — same scope-lookup rule (`context.scope > 0 ? get_binding(name, scope) : find_binding_any_scope(name)`), same `declarations.contains_key` guard, same fall-through when that guard fails. To be precise about what "same" means here: upstream (`2-analyze/visitors/shared/utils.js` L175-194) recurses **unconditionally** when `binding.kind === 'store_sub'`, whereas rsvelte gates that recursion on `analysis.root.scope.declarations.contains_key(store_name)`. That divergence predates this PR and is preserved verbatim — this refactor is behaviour-preserving, not a correctness fix.
- **`has_shadowing_declaration_in_path`** lost its `node.as_value()` fallback. `JsPathEntry::Borrowed` / `::Owned` are never constructed — `JsPathEntry::new_typed` is the only constructor used anywhere in the tree — so `as_js_node()` always returns `Some` and the fallback was dead. Removing it is what made the five `*_declares_name` / `has_variable_declaration` `Value` helpers collectible.

Verified mechanically that nothing else moved: extracting every function body from `utils.rs` before and after shows **exactly** the 20 targeted deletions, one addition (`is_safe_identifier_name`), and only those two bodies changed.

## Byte-exact verification

Method: rsvelte's fixture suites compare generated CSR / SSR / hydration output against the official Svelte compiler byte-for-byte, so a green run *is* the byte-exact check. Run with `RUST_MIN_STACK=33554432 cargo test --release -p rsvelte_core`, **once per commit**.

| Suite | After stage 1 | After stage 2 | After review fixes |
|---|---|---|---|
| `compiler_fixtures` | ok | ok (16) | ok (16) |
| `ssr` | ok (16) | ok (16) | ok (16) |
| `validator` | ok (16) | ok (16) | ok (16) |
| `compiler_errors` | ok | ok (17) | ok (17) |
| `runtime` (legacy + runes + hydration + browser) | ok (19, 130.90 s) | ok (19, 91.45 s) | ok (19, 126.42 s) |
| `--lib` unit tests | compile was OOM-killed under parallel load | ok (1,483 passed, 1 ignored) | ok (1,483 passed, 1 ignored) |

Also clean: `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo build --workspace --all-targets`.

(The aggregate suites — `test_ssr`, `test_validator`, `test_runtime_legacy`, … — each iterate every fixture in their directory, which is why the reported test counts look small.)

## Behavioural differences found between the `_node` and `Value` versions

**None.** No fixture changed and no output diff appeared at either stage. No fixture was edited.

## Out of scope (deliberately untouched)

`serde_json::Value` remains in phase 2 in `call_expression.rs` (`get_rune` plus its own local `get_global_keypath`), `await_expression.rs`, `class_body.rs`, `blockers.rs`, `pattern_ids.rs`, roughly fifteen functions in `2_analyze/mod.rs`, and `declaration_tag.rs::warn_local_state_reads` — which still calls `as_json()`. Most of those reach JSON through `JsPathEntry::as_value()`, which lazily materialises a `Value` from a typed node, so retiring them means porting `JsPathEntry` itself. That is a separate change; this PR is scoped to the duplicated walker family.

A follow-up is now unblocked: `JsPathEntry::{Borrowed, Owned}` and the `new` / `new_owned` constructors are fully dead after this PR (`new_typed` is the only constructor in use), so a later change can drop those variants and shrink the `as_value()` / `get_type_str()` matches.

`scope_builder.rs`'s oxc-AST ⇔ `_typed` duplication is likewise a separate PR, as are `compiler/legacy.rs` and `3_transform/`.

https://claude.ai/code/session_01RqdHfZSN3LFxPkyKFaCyoF


baseballyama:
`skip-changeset` ラベルを付与しました。

**理由: CI の false positive です。**

`check-core-consumer-changesets.mjs` は `git diff ${BASE_SHA}...HEAD` で変更ファイルを判定しますが、`BASE_SHA` = `github.event.pull_request.base.sha` は **PR 作成時の値のまま更新されません**。この PR は最新 main に rebase 済みのため、その間に main へ入った svelte2tsx 分割 PR 群の変更が差分に混入し、`crates/rsvelte_core/src/svelte2tsx/` ルールが誤発火しています。

実際の変更ファイルを確認済みで、`src/svelte2tsx/` にも `src/svelte_check/` にも一切触れていません。ジョブの再実行でも解消しませんでした(BASE_SHA が固定のため)。

この false positive 自体は別途 issue 化します。

PR #1787, opened 2026-07-25
test: add a shrink-only source-map verification gate

baseballyama:
Prerequisite for the script AST-visitor migration: once the text codegen path is
deleted there is no accurate-source-map reference left in the repo, so the gate
has to exist first. Issue #1781.

This PR adds tests only — no compiler code is touched.

## Investigation: why `Sourcemaps` read 0 fixtures

Two independent gaps, neither of which was a missing fixture:

1. **The fixtures were always there.** `scripts/fixtures/generate-fixtures.mjs`
   has a `sourcemaps` handler that compiles every sample with the official
   compiler and writes `client.js`, `client.js.map`, `server.js`,
   `server.js.map` — all 29 samples, all four files. The `Sourcemaps 0/0` row
   comes from `compatibility_report.rs::run_runtime_category_tests`, which does
   `if !input_path.exists() { continue }` on `<sample>/main.svelte`. The
   sourcemaps samples use `input.svelte`, so every sample was skipped.
2. **The existing `tests/sourcemaps.rs` never checked a map.** `js.map` is
   already a JSON string, and the test passed it through
   `serde_json::to_string_pretty`, producing a double-encoded string. The
   subsequent `compare_sourcemaps` parsed it to a `Value::String`, failed
   `.as_object()`, and returned `false` unconditionally — and the result was
   only enforced under an opt-in `STRICT_SOURCEMAPS=1`. So map comparison was
   both broken and disabled. Fixed here (test-only change; the JS-output
   assertions are unaffected).

Net: the data to verify source maps has been sitting in `fixtures/<sha>/sourcemaps/`
the whole time with nothing reading it. #1772 degrading 121 files went unnoticed
for exactly this reason.

## Gate design

`crates/rsvelte_core/tests/sourcemaps_gate.rs`. Ground truth is the official
compiler — the fixtures above, produced by `submodules/svelte`'s own `compile()`
on the same input with the same options. `RSVELTE_CLIENT_NO_OXC=1` is *not* used
as an oracle anywhere; it is only mentioned in the docs as a reference point for
how much resolution the text path had.

Three checks:

| kind | what it asserts |
|---|---|
| `anchor` | the official `_config.js` `client:` / `server:` / `css:` entries, using upstream `tests/sourcemaps/test.ts::compare` verbatim — locate the string in the generated output, find the segment at that column, assert it maps to that string's line/column in the source, then the same for the end segment |
| `map-parity` | where rsvelte's generated code is byte-identical to the official compiler's, every official segment must have a counterpart at the same generated position with the same origin. Counts `missing` (resolution lost) and `wrong` separately |
| `out-of-range` | segments whose original position is past the end of that source line, or past the last line — the symptom #1781 measures |

Plus hard, unratcheted checks that all pass today: no negative segment fields
(upstream does this for every sample), the map decodes, and a map is produced at
all.

21 anchors are ported — the compile-only subset. Samples driven by a `preprocess`
hook are excluded from `anchor` because their expectations describe the
preprocessed source, which a Rust test cannot reproduce (the preprocessors are JS
closures over `magic-string` / `typescript`); they still take part in
`map-parity` and `out-of-range`, which need no config. Upstream's two `skip: true`
samples are skipped for the same reason upstream skips them.

Before an anchor is held against rsvelte it is run against the **official** map
for the same sample. Anything failing there would land in
`compatibility/sourcemap-oracle-excluded.json` rather than being blamed on
rsvelte — that file is **empty**, i.e. all 21 ported anchors are reproducible on
the oracle under the fixture generator's options, which validates the port of
upstream's algorithm.

Ratchet semantics mirror `fmt-verify.mjs` / `verify.mjs`: a new failure or a
count above budget fails CI; an entry that starts passing only prints a reminder
to shrink the list. `compatibility/sourcemap-known-failures.md` justifies every
entry. Verified by hand that removing one anchor id and one budget line from the
JSON turns the test red with exactly those two regressions.

CI wiring needs no `ci.yml` change: `scripts/ci/run-test-shard.sh` partitions
integration-test targets from `cargo metadata`, so a new `tests/*.rs` is picked
up automatically and joins the required `Tests` status check. The measurement
helper is `#[ignore]`d and does not run in CI.

## Ratchet initial values — how they were measured

Not estimated. `sourcemap_gate_measure` runs the full measurement and, with
`UPDATE_SOURCEMAP_RATCHET=1`, writes both JSON files directly, so the committed
baseline is machine-generated:

```
UPDATE_SOURCEMAP_RATCHET=1 cargo test -p rsvelte_core --test sourcemaps_gate -- \
  --ignored --nocapture sourcemap_gate_measure
```

69 entries: 11 `anchor`, 44 `map-parity`, 14 `out-of-range`.

## Where we actually stand (Svelte `b29d7002ecf9`)

29 samples × {client, server}; 54 of the 58 pairs are byte-identical to the
official output.

| measure | value |
|---|---|
| official segments reproduced | **164 / 712 (23.0%)** — 466 missing, 82 wrong |
| out-of-range segments | **32 / 559 (5.7%)** |
| ported anchors passing | **10 / 21** |

The split is entirely client vs. server:

- **Server maps are accurate.** All 9 ported server anchors pass, no server map
  has an out-of-range segment, and *every one* of the 164 reproduced segments is
  a server segment. The server transform is pure-AST, so nodes carry real spans.
- **Client maps are not.** All 11 failing anchors are client. Every client sample
  scores `0 exact` — not a single segment of the official client map is
  reproduced — and all 32 out-of-range segments are client. Ten anchors fail as
  "no segment at `<line>:<col>`"; `two-scripts`/`first` instead maps to `0:11`
  instead of `1:12`, i.e. the module script's chunk is anchored at the file start
  and the whole block is off by the `<script module>` line.
- The one CSS anchor passes; CSS maps come from a separate `string_wizard` path
  the client refactor does not touch.

This is the `take_chunk_region` chunk-granularity problem from #1781, now
measured and fenced. No accuracy work is attempted here — that is the follow-up,
and this gate is what will show it working.

One incidental note: upstream's `css` anchor expects `.foo.svelte-1eyw86p`, whose
scope hash comes from the filename upstream's `compile_directory` passes. Under
the fixture generator's plain `input.svelte` the official compiler emits
`.foo.svelte-4hbqx4` — verified by running `submodules/svelte`'s `compile()`
directly — so the ported anchor uses that. Still official truth; hash parity
itself is gated by `tests/css.rs`.

## Verification

```
cargo fmt --all                                                    ok
cargo clippy --all-targets --all-features -- -D warnings           ok
cargo test -p rsvelte_core --test sourcemaps_gate                  15 passed, 1 ignored
cargo test -p rsvelte_core --test sourcemaps                       16 passed
cargo test -p rsvelte_core --lib                                   1483 passed
```

https://claude.ai/code/session_01RqdHfZSN3LFxPkyKFaCyoF


baseballyama:
Review addressed. Pushed `d2c521b87bd3b381eafd02961160a9f6ebb1f363`.

## blocker — the gate could pass while measuring nothing

You were right, and the framing is the uncomfortable one: the gate reproduced
the exact defect it was written to expose. Fixed with all three of your
suggestions plus two more, because each covers a different way of measuring less:

| failure mode | guard | verified |
|---|---|---|
| `input.svelte` renamed | `load_input` panics instead of `continue` | `sourcemaps sample "basic" has no readable input.svelte — upstream layout changed?` |
| sample directory gone | `EXPECTED_SAMPLES` floor (29) | `only 28 sourcemaps samples measured, expected at least 29` |
| anchor deleted from `ANCHORS` | `EXPECTED_ANCHOR_COUNT` floor (23), **and** an `anchor` ratchet id with no matching table entry is a regression | `only 22 anchors evaluated, expected at least 23 — were entries removed from ANCHORS?` |
| byte-parity regresses, shrinking what `map-parity` can watch | `EXPECTED_IDENTICAL_OUTPUTS` floor (54) — your (c) | — |
| a budgeted pair stops being measured | now a **regression**, not a `fixed` note | `❌ 1 NEW source-map failures: out-of-range no-such-sample client NO LONGER MEASURED (budget 3)` |

Each row was reproduced by actually breaking it and confirming red, then
restoring; the submodule is clean. I ran your exact repro (renaming
`input.svelte`) — it now fails in 0.17s instead of printing `1 passed`.

Also closes the related nit: the hardcoded compile options no longer diverge
silently from the oracle's. Every sample's fixture `metadata.json` is asserted
to match `{"dev":false}`, the options the generator actually used, so a
generator change makes the gate fail rather than compare two different things.

The `.md` documents both invariants that cannot be expressed as known failures:
a vanished budget key, and an orphaned `anchor` id.

## should-fix 1 — `sourcemap-empty-source`

Ported, both targets, and you are right on the substance: **it falsifies my
central claim.** Measured after adding it —

- oracle: passes on client *and* server (`sourcemap-oracle-excluded.json` stays `[]`)
- rsvelte client: `'let doubled'` maps to `0:8`, expected `2:1`
- rsvelte **server**: `no segment at 4:1 for 'let doubled'`

So "server maps are accurate / all server anchors pass" was wrong, and it was
wrong *because* the counterexample sat behind an undocumented exclusion. The
docs now break the numbers out per target rather than asserting a clean split:

| measure | client | server | total |
|---|---|---|---|
| official segments reproduced | 0 / 428 | 164 / 284 | 164 / 712 (23.0%) |
| — missing / wrong | 353 / 75 | **113 / 7** | 466 / 82 |
| out-of-range | 32 | 0 | 32 / 559 (5.7%) |
| anchors passing | 0 / 12 | **9 / 10** | 10 / 23 |

Server is *directionally* correct — never out of range, and the only 7 `wrong`
segments are in `preprocessed-styles` and `source-map-generator` — but it is
measurably coarser than official and has one failing anchor. The root-cause
section now says so explicitly: `take_chunk_region` explains the client column,
not the server residue, which is a separate smaller gap in the server printer's
segment emission.

`ANCHORS` documents `compileOptions.sourcemap` as a third category (rsvelte
ignores the option here, as does the fixture generator, so the anchor exercises
the plain compile) and says why it is ported anyway.

## should-fix 2, 3, 4

2. `has_negative_segment` → `s.iter().take(4)`. Correct diagnosis: `vlq_decode_line`
   accumulates only the first four fields and pushes the `names` delta raw, so a
   map whose name indices go backwards would have tripped an unratcheted hard
   failure. Latent only because this corpus has `names: []` throughout.
3. EOF is now a failure — `matches!(next, Some('\n') | Some('\r'))`, `None`
   removed. Matches upstream, which indexes past the end, gets `undefined`, and
   throws. Comment added explaining the equivalence. No measurement changed.
4. `STRICT_SOURCEMAPS` deleted, along with `client_map_passed` /
   `server_map_passed` and the `unwrap_or(true)` that passed a missing map. The
   fields were write-only once the branch went, so the whole `compare_sourcemaps`
   call site is gone; `_actual/*.map` artifacts are still written for debugging
   and the summary line points at `sourcemaps_gate.rs`. Agreed on the principle —
   "compute it and throw it away" is the same shape as the bug this PR found.

## nits

- The oracle-excluded warn is implemented (prints when an excluded anchor starts
  passing on the official map).
- `.md` corrected: **22 of 23** anchors are oracle-checked, not all of them —
  `official()` returns `None` for `Target::Css` because the generator emits no
  CSS fixture, so the one CSS anchor rests on the hand-verification against
  `submodules/svelte`'s `compile()` instead. Called out in
  `sourcemap-oracle-excluded.md` as a coverage caveat.

## re-verified

```
cargo fmt --all -- --check                                 ok
cargo clippy --all-targets --all-features -- -D warnings   ok
cargo test -p rsvelte_core --test sourcemaps_gate          15 passed, 1 ignored
cargo test -p rsvelte_core --test sourcemaps               16 passed
cargo test -p rsvelte_core --lib                           1483 passed
```

Ratchet: 71 entries (13 `anchor`, 44 `map-parity`, 14 `out-of-range`), all
machine-generated via `UPDATE_SOURCEMAP_RATCHET=1`.

Thanks for the audit findings downstream of this — the `rsvelte_lint_types`
suite never having executed is a much bigger fish than this PR.

baseballyama:
Both nits done. Pushed `c0a38e16ec923f57e6b367e163a6813ec2cdf7e4`.

**1. `compare_sourcemaps` deleted** — the function and its inner `normalize`, 67
lines. Confirmed the only remaining reference was the historical mention in
`sourcemaps_gate.rs`'s module doc, which now describes the behaviour without
naming a symbol that no longer exists.

You're right that this is the same theme, and the `#![allow(dead_code)]` is the
part worth naming: it is a *file-scope* suppression, so the crate's usual
dead-code signal is off for everything in `common/`. That is reasonable for a
shared test helper module (each test binary links it and uses a different
subset) but it means a helper that becomes dead across *all* consumers looks
identical to one that is merely unused by the binary being compiled. `cargo
clippy --all-targets` cannot tell those apart. So removals like this have to be
done by grep at the moment the last caller goes — which is what happened here,
one round late.

**2. Bump procedure documented** — new "After a Svelte bump" section in
`compatibility/sourcemap-known-failures.md`, covering your four cases:

- *samples added* — no action; floors are `>=` lower bounds so they stay
  satisfied, and a new sample has no ratchet entry so its failures surface as
  regressions. Raise all three floors to the newly measured values in the same
  commit that regenerates the ratchet.
- *samples removed/renamed* — a floor trips or `load_input` panics, by design.
  Explicitly warned not to lower a floor without checking the upstream diff
  first: an unreadable sample and a deleted one are indistinguishable from
  inside the test, and the first one is a broken checkout.
- *a `_config.js` the generator can import* — `check_fixture_options` fails with
  "the comparison would be meaningless". Documented as a benign cause with a
  loud symptom, and the instruction is to teach `compile_sample` the same
  options or exclude the sample — **not** to relax
  `EXPECTED_FIXTURE_COMPILE_OPTIONS`, since that would restore exactly the
  silent-skew failure mode the constant exists to prevent.
- *anchors* — added a fourth note you didn't ask for but which follows from the
  same reasoning: `_config.js` expectations are hand-copied into `ANCHORS`, and
  nothing detects an upstream expectation whose *value* changed (only ones that
  disappear). Re-read the changed configs on a bump.

Re-verified after the deletion: `cargo fmt --check` ok, `clippy --all-targets
--all-features -D warnings` ok (which also proves no other test binary
referenced the removed helper — they all still compile), and `sourcemaps_gate`
15 passed + 1 ignored, `sourcemaps` 16, `css` 16, `print` 16.

baseballyama:
`skip-changeset` ラベルを付与しました。

**理由: CI の false positive です。**

`check-core-consumer-changesets.mjs` は `git diff ${BASE_SHA}...HEAD` で変更ファイルを判定しますが、`BASE_SHA` = `github.event.pull_request.base.sha` は **PR 作成時の値のまま更新されません**。この PR は最新 main に rebase 済みのため、その間に main へ入った svelte2tsx 分割 PR 群の変更が差分に混入し、`crates/rsvelte_core/src/svelte2tsx/` ルールが誤発火しています。

実際の変更ファイルを確認済みで、`src/svelte2tsx/` にも `src/svelte_check/` にも一切触れていません。ジョブの再実行でも解消しませんでした(BASE_SHA が固定のため)。

この false positive 自体は別途 issue 化します。

baseballyama:
Rebased onto `origin/main` (`7d00fff8`) and force-pushed:
**`b2d02ea86972850b7dd7eea66e24a81741b7e88c`**. No conflicts — `compatibility/`
included, since the ratchet files are new and don't touch the existing ones.
`git diff origin/main...HEAD` is exactly my 8 files, so nothing foreign came
along. `Changeset` now **passes** (8s).

## The gate moved on the rebase — and it caught the right thing

You asked me to flag any ratchet drift. There is drift, and it is attributable
to a single commit: `ad807a8c`, **PR #1772** — the very PR issue #1781 was
opened about. My original baseline was measured before it landed; the rebase
pulled it in.

The delta is confined to the two sourcemaps samples with a `//` comment inside
`<script>` — exactly the files #1772 switches from the text generator to the
direct-AST path (`typescript/input.svelte:7`, `sourcemap-offsets/input.svelte:2`):

| | before #1772 | after |
|---|---|---|
| `typescript` client — byte-identical to official | no | **yes** |
| `typescript` client — official segments reproduced | (not in population) | 0 / 52 — 40 missing, 12 wrong |
| `typescript` client — out-of-range | 0 | **4** |
| `sourcemap-offsets` client — out-of-range | 0 | **1** |

Both directions in one change: generated-code parity *improved* (54 → 55
byte-identical, which is precisely why `typescript` newly qualifies for
`map-parity` at all), while map quality *regressed* — 5 new segments pointing
outside the source.

Totals moved accordingly:

| | before | after |
|---|---|---|
| official segments reproduced | 164 / 712 (23.0%) | 164 / 764 (21.5%) |
| out-of-range | 32 / 559 (5.7%) | 37 / 545 (6.8%) |
| anchors passing | 10 / 23 | 10 / 23 (unchanged) |

Two things worth noting about *how* it moved:

- **Server totals are byte-for-byte unchanged** (164/284 exact, 113 missing, 7
  wrong). #1772 is client-only, and the gate says so independently rather than
  taking the PR's word for it.
- **No other sample's counts moved at all** — the ratchet diff is three pure
  additions, no modifications. So nothing else that landed on main in this window
  touched source maps, and the two samples that did move are fully explained.

This is #1781's degradation, reproduced from the outside with numbers, on a
change that had already gone green through every other suite. It is the first
thing the gate has caught, and it caught it on day one.

Baseline updated to the new base: ratchet 71 → 74 entries (13 `anchor`, 45
`map-parity`, 16 `out-of-range`), `EXPECTED_IDENTICAL_OUTPUTS` 54 → 55, all
regenerated with `UPDATE_SOURCEMAP_RATCHET=1`. The `.md` gains a "First catch:
#1772" section with the table above, so the next person to touch
`take_chunk_region` sees which samples are the canaries.

Re-verified on the new base: `cargo fmt --check` ok, `sourcemaps_gate` 15 passed
+ 1 ignored, `sourcemaps` 16 passed.

Issue #1766, opened 2026-07-25
Wave 4 M5: replace svelte-vscode — grammar, settings compatibility, commands, other editors


baseballyama:
Part of the Wave 4 epic. Depends on M0; the rest can land in parallel with M1–M3.

## Goal

Make rsvelte a **replacement** for `svelte-vscode`, not something you run alongside it. Today
`apps/npm/vscode` contributes three settings and nothing else — no language definition, no grammar —
so it only works when the official extension is installed.

## VS Code extension

- [x] Vendor the TextMate grammars from `svelte-vscode` (MIT): `source.svelte` plus the pug / markdown-codeblock / postcss variants
- [x] `contributes.languages` for `svelte` (+ the `svelte-start-tag` helper language), indentation rules, `breakpoints`, snippets
- [x] Detect the official extension being enabled at the same time and warn — otherwise diagnostics appear twice
- [x] **Accept upstream `svelte.*` settings** in addition to `rsvelte.*`, so switching needs no reconfiguration (~45 keys: `svelte.plugin.{typescript,css,html,svelte}.*`)
- [x] `rsvelte.restartLanguageServer` command + output channel
- [x] Restart on save of `svelte.config.*`, `vite.config.*`, `.prettierrc*`, `rsvelte-lint.json`
- [x] `$/getCompiledCode` + a `svelte-compiled://` content provider (show compiled JS / CSS to the side)
- [x] Extract Component command (input box → `executeCommand`)
- [x] `findAllFileReferences` / `findComponentReferences` commands and the CodeLens middleware that turns server results into `editor.action.showReferences`
- [x] Import updates on file rename via `$/getEditsForFileRename`
- [x] Tag auto-close client wiring (`html/tag`, honouring `html.autoClosingTags`)
- [x] SvelteKit file-generation context menu (`+page` / `+layout` / `+server` / `+error`, TS and runes detection)
- [x] Publish to the VS Code Marketplace **and** Open VSX

## Other editors (the distribution advantage)

A single static binary with no Node requirement is something upstream cannot offer.

- [x] Neovim: `nvim-lspconfig` entry + docs
- [x] Zed extension (#1203)
- [x] Sublime `LSP-rsvelte`, Helix, Emacs `lsp-mode` configuration docs
- [x] Binary distribution: per-platform npm packages (mirroring `@rsvelte/fmt-*`) plus plain GitHub release archives for non-npm users

Tracked by #1768.

PR #1797, opened 2026-07-25
feat(vscode): restart command, config-file restarts and server trace settings

baseballyama:
## Summary

Adds command/lifecycle UX on top of #1776's bundled grammar/language extension. Server-independent — only `apps/npm/vscode` changes.

**Commands added**
- `rsvelte.restartLanguageServer` — command-palette entry, title `rsvelte: Restart Language Server`. Clears the output channel and restarts (or starts) the client.

**Settings added** (`contributes.configuration`)
- `rsvelte.languageServer.path` — path to a language server entry script/binary, for pointing at a dev build. Relative paths resolve against the first workspace folder; empty (default) keeps the bundled `dist/server.mjs`. Marked `restrictedConfigurations` under `capabilities.untrustedWorkspaces` since it's an arbitrary executable path (mirrors the official extension's `svelte.language-server.ls-path`).
- `rsvelte.trace.server` — `off` / `messages` / `verbose` (default `off`), wired into the `LanguageClient` via `setTrace()` after start and re-applied on config change.

**Auto-restart on save**

Saving any of the following inside an open workspace folder silently restarts the server (no notification, matching the official extension's behavior for its equivalent list):
- `svelte.config.{js,mjs,cjs,ts,mts}`
- `vite.config.{js,mjs,cjs,ts,mts,cts}`
- `rsvelte-lint.json` / `.rsvelte-lintrc.json`
- `.oxfmtrc.{json,jsonc}` / `oxfmt.config.{ts,mts}`

The extension list matches what rsvelte's own config discovery actually recognizes (`crates/rsvelte_core/src/svelte_check/config.rs`, `crates/rsvelte_fmt/src/config.rs`, `crates/rsvelte_lint/src/main.rs` — read for reference only, not modified). To avoid misfires: the save handler requires `scheme === "file"`, requires the document to resolve to an open workspace folder (`workspace.getWorkspaceFolder`), matches the exact basename via anchored regexes (not a substring match), and skips anything under a `node_modules` segment.

**Stderr visibility**

The client now creates its own `LogOutputChannel` ("rsvelte Language Server") and passes it via `clientOptions.outputChannel`; `vscode-languageclient` already pipes the spawned server's stderr into that channel for the stdio transport, so it's now a stable, restart-cleared channel instead of an implicitly-created one.

**Concurrency / restart safety**

- A module-level `restarting` boolean collapses concurrent restart triggers (the command plus several config saves in a row) into a single restart — repeat calls while one is in flight are no-ops.
- `BaseLanguageClient.stop()`/`restart()` throws unless the client is `Running` (it rejects while `Starting`, `Stopped`, or `StartFailed`). `restartLanguageServer()` first awaits `waitWhileStarting()` (resolves once `onDidChangeState` reports a state other than `Starting`), then calls `restart()` only if `Running`, otherwise falls back to `start()` — so a restart triggered mid-launch or after a failed launch can't throw or hang.
- `deactivate()` applies the same `waitWhileStarting` guard before calling `stop()`, so tearing down mid-startup doesn't throw either. The output channel and command/listener disposables are all registered on `context.subscriptions`, so they're cleaned up automatically.

## Out of scope (per issue)

`svelte.*` config acceptance, `$/getCompiledCode`, Extract Component, SvelteKit file-generation menu, `findFileReferences`/other server-dependent commands, Marketplace publishing, and anything under `crates/` or `apps/npm/language-server`.

## Test plan

- [x] `package.json` is valid JSON (`node -e "JSON.parse(...)"`)
- [x] `pnpm --filter rsvelte-vscode run typecheck` passes (`tsc --noEmit`, no errors)
- [x] `node apps/npm/vscode/build.mjs` bundles `src/extension.ts` successfully via esbuild (fails only at the expected step — copying `language-server/dist`, which isn't built in this worktree and is out of scope)
- [ ] Manual smoke test in a real VS Code window (not done in this environment)

Refs #1766.

https://claude.ai/code/session_019tsoBsfzrKu4kfbfrdFYJS

### DIFF
--- a/apps/npm/vscode/package.json
+++ b/apps/npm/vscode/package.json
@@ -42,7 +42,22 @@
     "onLanguage:scss",
     "onLanguage:less"
   ],
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "restrictedConfigurations": [
+        "rsvelte.languageServer.path"
+      ],
+      "description": "The extension can execute a language server binary/script specified by the workspace."
+    }
+  },
   "contributes": {
+    "commands": [
+      {
+        "command": "rsvelte.restartLanguageServer",
+        "title": "rsvelte: Restart Language Server"
+      }
+    ],
     "languages": [
       {
         "id": "svelte",
@@ -199,6 +214,21 @@
           "type": "string",
           "default": "",
           "description": "Absolute path to a rsvelte-fmt binary. When empty, the extension resolves node_modules/.bin/rsvelte-fmt from the workspace."
+        },
+        "rsvelte.languageServer.path": {
+          "type": "string",
+          "default": "",
+          "description": "Path to a language server entry script/binary, for development. Relative paths resolve against the first workspace folder. When empty, the extension uses its own bundled server."
+        },
+        "rsvelte.trace.server": {
+          "type": "string",
+          "enum": [
+            "off",
+            "messages",
+            "verbose"
+          ],
+          "default": "off",
+          "description": "Traces the communication between VS Code and the rsvelte language server, shown in the \"rsvelte Language Server\" output channel."
         }
       }
     }
--- a/apps/npm/vscode/src/extension.ts
+++ b/apps/npm/vscode/src/extension.ts
@@ -7,19 +7,27 @@
 import * as path from "node:path";
 import {
   IndentAction,
+  commands,
   extensions,
   languages,
   window,
+  workspace,
   type ExtensionContext,
+  type LogOutputChannel,
+  type TextDocument,
 } from "vscode";
 import {
   LanguageClient,
+  State,
+  Trace,
   TransportKind,
   type LanguageClientOptions,
   type ServerOptions,
 } from "vscode-languageclient/node";
 
 let client: LanguageClient | undefined;
+let outputChannel: LogOutputChannel | undefined;
+let restarting = false;
 
 /** Languages the server attaches to (formatting + diagnostics). */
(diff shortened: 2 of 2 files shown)


baseballyama:
Addressed the internal review feedback in b1357592:

1. **StartFailed permanently disabled restart** — `restartLanguageServer()` now discards the previous `LanguageClient` instance and constructs a fresh one (`createClient`/`startClient`) instead of calling `restart()`/`start()` on the same instance. `start()` on an already-`StartFailed` instance just returns the old cached rejection in `vscode-languageclient@10.1.0` (`_onStart` is never cleared on that path), so reuse could never recover.
2. **`rsvelte.languageServer.path` changes ignored on restart** — fixed by the same change: `buildServerOptions()` is re-evaluated on every restart (inside the freshly constructed client), so a changed path takes effect without a window reload.
3. **Race with `deactivate()`** — `restartLanguageServer()` and `deactivate()` now capture `client`/`outputChannel`/`extensionContext` into local consts before their first `await`, so a concurrent `deactivate()` clearing the module-level globals can't produce a `TypeError` on resume. A `deactivated` flag also prevents `startClient` from leaving an orphaned client running past a `deactivate()` that occurred mid-restart.
4. **Unbounded wait on a hung server** — `waitWhileStarting()` now takes a 10s timeout instead of waiting forever on `onDidChangeState`; on timeout it shows a warning and proceeds to discard/recreate the client anyway, so the `restarting` guard can't wedge permanently.

`pnpm --filter rsvelte-vscode run typecheck` and JSON validity both still pass.

PR #1805, opened 2026-07-25
feat(lint): expose fix data alongside the validator wrap

### LINKED ISSUE #1789: rsvelte_lint: no API returns both the validator wrap and fix/suggestion data
`crates/rsvelte_lint/src/runner.rs` exposes two entry points and neither is sufficient for editor code actions:

- `lint_source(source, file, options, config) -> Vec<Diagnostic>` runs the **validator wrap** — compiler warnings/errors and a11y — but returns `svelte_check::Diagnostic`, which carries no `fix` or `suggestions`.
- `lint_source_raw(source, file, config) -> Vec<LintDiagnostic>` carries `Fix` / `Suggestion` / `help`, but skips the validator wrap entirely, so compiler warnings and a11y diagnostics disappear.

The language server (#1761) currently calls `lint_source`, which is correct for diagnostics parity with the published Node server. But M1 (#1762) needs to turn lint fixes into `CodeAction`s, and doing that today would mean either dropping compiler/a11y diagnostics or running the file through both paths and reconciling two diagnostic sets.

Wanted: one API that returns `LintDiagnostic` (fix + suggestions + help intact) *including* the validator wrap, so a single lint pass drives both `publishDiagnostics` and `textDocument/codeAction`.

Blocks the code-action tasks in #1762.

Refs #1768.

baseballyama:
Neither lint entry point could drive an editor: `lint_source` runs the validator
wrap (compiler warnings / errors / a11y) but returns `svelte_check::Diagnostic`,
which has no fix data; `lint_source_raw` keeps `Fix` / `Suggestion` / `help` but
skips the wrap. Turning lint findings into code actions meant either dropping the
compiler and a11y diagnostics or running both paths and reconciling two sets.

## New API

```rust
pub fn lint_source_messages(
    source: &str,
    file: &Path,
    options: &CompileOptions,
    config: &LintConfig,
) -> Vec<LintMessage>

pub struct LintMessage {
    /// Identical to the element `lint_source` yields.
    pub diagnostic: Diagnostic,
    /// `(start, end)` UTF-8 byte offsets; `None` for validator-wrap findings.
    pub span: Option<(u32, u32)>,
    pub help: Option<String>,
    pub fix: Option<Fix>,
    pub suggestions: Vec<Suggestion>,
}
```

One pass now feeds both `publishDiagnostics` (`m.diagnostic`) and `codeAction`
(`m.fix` / `m.suggestions`, in byte offsets the editor maps with its own line
index).

### Why not `Vec<LintDiagnostic>`

The issue asks for `LintDiagnostic` including the wrap, but the two sources do
not share a position model: rule findings carry UTF-8 byte spans, while the
validator wrap carries compiler line/column positions — and a hard compile error
carries none at all (`range: None`). Materialising `LintDiagnostic` for wrap
findings would mean inventing byte offsets by reversing the UTF-16 column
mapping, and collapsing "no range" to `0..0`, which then reappears as line 1
column 0 and changes the sort key. Keeping the output `Diagnostic` verbatim and
attaching the byte-span payload only where it genuinely exists loses nothing on
either side, and gives the editor the LSP-shaped range it needs anyway.

## Relationship to the existing two functions

- `lint_source` — unchanged signature and output; now
  `lint_source_messages(..).map(|m| m.diagnostic)`. It is the same list, in the
  same order.
- `lint_source_raw` — unchanged signature and output; still the wrap-free,
  rule-only view used by the eslint-plugin-svelte oracle.

## Shared internals

`lint_source_messages` is the single implementation of the full pipeline
(inline config → wrap → rule passes → meta rules → comment-directive →
suppression → sort). The AST-driven passes that `lint_source_raw` shares with it
moved into one helper, `svelte_rule_findings`, which does the single lenient
parse, builds the scope resolver once, runs the native template walk, the
`<script>` walk and the scope rules, and reports whether the parse succeeded (the
flag that gates the block-lang source-scan fallback). Nothing runs twice, and
neither the wrap nor any rule pass is implemented in two places.

## Order, suppression and inline config

The pipeline was moved, not rewritten: inline config is still applied first, the
wrap still contributes before the rule passes and the source-scan meta rules,
the comment-directive meta rule still reads the pre-suppression finding set,
suppression still filters on `(code, range.start.line)` of the output
diagnostic, unused-directive reports are still appended after it, and the final
sort is still the same stable sort by `(line, column)` over the same input
order.

Verified empirically as well as by construction: the `rsvelte-lint` CLI built
from `main` and from this branch produces byte-identical output over 32 files
(the repo's `.svelte` sources plus fixtures covering a parse error, an unknown
`<style lang>`, `svelte-ignore`, `eslint-disable-next-line`, an inline
`/* eslint … */` config and a `.svelte.ts` module) in all six formats — human,
human-verbose, machine, machine-verbose, sarif, github-actions — and `--fix`
leaves identical files and identical output.

## Tests

`crates/rsvelte_lint/src/runner.rs`:

- `messages_pair_the_validator_wrap_with_rule_fixes_in_one_pass` — one call over
  a source with both an `a11y_*` warning and a fixable rule finding: the wrap
  entry is positioned with no fix payload, the rule entry carries a byte span
  that slices back to the reported text and a fix that applies correctly.
- `messages_carry_suggestions` — suggestion-only rules keep their suggestions.
- `messages_reproduce_lint_source_output_verbatim` — the projected diagnostics
  equal `lint_source`'s over a source exercising the wrap, both walks, a meta
  rule and a disable directive.

CLI output is additionally covered by the existing suites in
`crates/rsvelte_lint/tests/` (`comment_directive`, `valid_compile`,
`valid_style_parse`, `experimental_and_dispatcher`,
`browser_globals_ts_assertion`, `eslint_plugin_oracle`), all of which go through
`lint_source` / `lint_source_raw`.

Commands: `cargo fmt`, `cargo clippy -p rsvelte_lint --all-targets
--all-features -- -D warnings`, `cargo test -p rsvelte_lint` (all green),
`cargo check -p rsvelte_lint --no-default-features` (wasm-shaped build),
`cargo check -p rsvelte_lint_bindings --all-features`,
`cargo check -p rsvelte_language_server --all-targets`.

No changeset: this is an additive Rust API with no user-visible change to the
published `@rsvelte/lint` CLI, whose output is verified unchanged above.

Closes #1789.

https://claude.ai/code/session_019tsoBsfzrKu4kfbfrdFYJS

### DIFF
--- a/crates/rsvelte_lint/src/diagnostic.rs
+++ b/crates/rsvelte_lint/src/diagnostic.rs
@@ -78,6 +78,56 @@ pub struct LintDiagnostic {
     pub suggestions: Vec<Suggestion>,
 }
 
+/// One finding of a full lint pass: the output-ready [`Diagnostic`] plus the
+/// fix payload the editor needs for code actions.
+///
+/// Findings reach the output list from two sources with different position
+/// models — native rules carry UTF-8 byte spans, the validator wrap carries
+/// compiler line/column positions (and none at all for a hard compile error).
+/// Keeping the output `Diagnostic` verbatim and attaching the byte-span payload
+/// only where it exists means neither side is round-tripped through the other's
+/// coordinates.
+#[cfg(feature = "native")]
+#[derive(Debug, Clone)]
+pub struct LintMessage {
+    /// Identical to the element [`lint_source`](crate::lint_source) yields.
+    pub diagnostic: Diagnostic,
+    /// `(start, end)` UTF-8 byte offsets; `None` for validator-wrap findings.
+    pub span: Option<(u32, u32)>,
+    pub help: Option<String>,
+    /// The `--fix` autofix, when the rule offers one.
+    pub fix: Option<Fix>,
+    /// Editor-offered suggestions (never auto-applied).
+    pub suggestions: Vec<Suggestion>,
+}
+
+#[cfg(feature = "native")]
+impl LintMessage {
+    /// Wrap a rule finding, keeping its fix payload alongside the converted
+    /// line/column diagnostic.
+    pub(crate) fn from_lint(d: LintDiagnostic, file: &Path, line_index: &LineIndex) -> Self {
+        Self {
+            diagnostic: d.to_output(file, line_index),
+            span: Some((d.start, d.end)),
+            help: d.help,
+            fix: d.fix,
+            suggestions: d.suggestions,
+        }
+    }
+
+    /// Wrap a diagnostic that has no fix payload (validator wrap, source-scan
+    /// meta rules).
+    pub(crate) fn from_diagnostic(diagnostic: Diagnostic) -> Self {
+        Self {
+            diagnostic,
+            span: None,
+            help: None,
+            fix: None,
+            suggestions: Vec::new(),
+        }
+    }
+}
+
 #[cfg(feature = "native")]
 impl LintDiagnostic {
     /// Convert to the shared output diagnostic. `Off`-severity findings should
--- a/crates/rsvelte_lint/src/lib.rs
+++ b/crates/rsvelte_lint/src/lib.rs
@@ -62,10 +62,14 @@ pub mod json_api;
 pub const CRATE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 pub use config::LintConfig;
+#[cfg(feature = "native")]
+pub use diagnostic::LintMessage;
 pub use diagnostic::{Fix, LintDiagnostic, Suggestion, TextEdit};
 pub use rule::{Fixable, Rule, RuleCategory, RuleConditions, RuleMeta, Severity};
 
 #[cfg(feature = "native")]
 pub use output::{LintFormat, render};
 #[cfg(feature = "native")]
-pub use runner::{FixResult, fix_source, lint_file, lint_source, lint_source_raw};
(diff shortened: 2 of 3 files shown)


baseballyama:
Applied `skip-changeset`: this adds a Rust API and changes nothing a consumer of the published `@rsvelte/lint` package can observe, so there is no version to bump.

For the record, the precedent cited in the PR description doesn't apply — #1803 needed no changeset because the gate is keyed on the PR title (`perf(...)` is exempt, `feat`/`fix` are not), not because lint-crate changes are exempt.

Independently reproduced the byte-identical claim: built the `rsvelte-lint` CLI from `origin/main` and from this branch, ran both over 21 inputs (repo `.svelte` sources plus fixtures for a parse error, an unknown `<style lang>`, `svelte-ignore`, and `eslint-disable-next-line`) in the human / machine / sarif / github-actions formats — all four byte-identical, so ordering and suppression are unchanged.

One note on the test rather than the code: `messages_reproduce_lint_source_output_verbatim` compares `lint_source_messages(..).diagnostic` against `lint_source`, but `lint_source` is now implemented as exactly that mapping, so the assertion is tautological and cannot catch a regression against the pre-refactor behaviour. The 258 existing tests and the CLI diff above are what actually pin that down.

PR #1807, opened 2026-07-25
perf(lint): one template serialization and one script walk per file (-20%)

baseballyama:
## Why

Follow-up to #1803, working down the same profile. Two more instances of the same pattern — whole-file work redone per rule.

## What changed

**1. One script walk per program instead of one per rule.** Every script rule descended the ESTree JSON itself: 40 whole-program walks per script block across the enabled set, each paying a `type` probe and a child scan on every object. The engine now walks the program once into a DFS index (nodes in pre-order + each node's ancestor count) and hands rules a `ProgramView`; `view.walk(f)` replays that index against a truncate-then-push stack, reproducing `walk_js`'s callback contract exactly. `ProgramView` derefs to the `Value`, so field access in rules is unchanged. Rules that build their own sub-programs keep the recursive walker.

**2. One template serialization per file instead of five.** `no-dom-manipulating`, `no-immutable-reactive-statements` and `no-reactive-reassign` each **re-parsed the source and serialized the whole template**, and `prefer-destructured-store-props` serialized the fragment of the `Root` it was already handed. They now go through the `LintContext` caches introduced in #1803 — the three source-scanners share `template_fragment_json()` (the same default parse options they used themselves), and `prefer-destructured-store-props` reads the `fragment` of the shared `root_json()` (the same lenient `Root` it was serializing).

## Measured

Interleaved A/B, three rounds each, on an idle machine over the 3,412-file benchmark corpus:

| Change | Before | After | Delta |
|---|---:|---:|---:|
| Shared script walk | 1,146 ms | 1,116 ms | −2.7% (−2.7 / +0.3 / −3.2) |
| One template serialization | 1,108 ms | 943 ms | **−14.9%** (−14.9 / −15.1 / −13.3) |

Cumulative for the lint work so far: **1,665 ms → 943 ms (−43%)**, i.e. roughly 24.9× → ~39× against the unchanged ESLint baseline.

## Correctness

- Lint parity corpus: **80 current / 80 known, 0 new divergences**.
- `cargo test -p rsvelte_lint --release`: 226 + 29 tests pass.
- `cargo clippy -p rsvelte_lint --all-targets`: clean.

## Two things tried and reverted (measured, not kept)

- **Cross-pass shared cache.** A file is linted in three passes, each with its own context; sharing one cache across them measured **±0%** and is provably a no-op for any current config — every `analyze_scope` consumer is a script-pass rule and `scope_rules()` is empty. Not worth threading a cache through three entry points.
- **Reusing the validator's analysis.** The validator wrap compiles every component, so handing its `ComponentAnalysis` to the rules that call `analyze_scope` removes a whole parse + analyze per file — worth **−8.3%**. But it is not behaviour-preserving: `compile()` resolves lazy expressions and strips TS before analyzing, so `reassigned` differs, and `prefer-const` lost three findings the oracle reports (caught by the parity gate). Making the two analyses equivalent is a real follow-up, not a free win.

## What's left

Per-file, the profile now shows: the lint parse (~20%), the validator wrap's own compile (~11%, a second full parse + analyze), `analyze_scope` for `prefer-const` et al (~another full analysis), the script-program JSON, and `serde_json::Value`'s IndexMap + SipHash (~12%). The two analyses are the biggest single remaining item, and unifying them needs the parse-option/TS-strip difference resolved first.

https://claude.ai/code/session_01GrqdRdS2YdA12XDA24KUHp

### DIFF
--- a/crates/rsvelte_lint/src/context.rs
+++ b/crates/rsvelte_lint/src/context.rs
@@ -6,9 +6,13 @@
 //! also borrows the resolved [`LintConfig`] so a rule can read its own parsed
 //! options via [`LintContext::option_bool`] / [`LintContext::option_str_list`].
 
+use std::cell::OnceCell;
 use std::path::Path;
+use std::rc::Rc;
 
 use serde_json::Value;
+
+use rsvelte_core::compiler::phases::phase2_analyze::ComponentAnalysis;
 
 use crate::config::LintConfig;
 use crate::diagnostic::{Fix, LintDiagnostic, Suggestion};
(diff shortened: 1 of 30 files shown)


baseballyama:
Superseded by a rebased branch — #1803 was squash-merged, so this branch's history conflicted. Reopening with the same commits (plus one more) on top of current main.

PR #1829, opened 2026-07-25
chore(fmt): collapse Text match guard to satisfy clippy collapsible_match

baseballyama:
Newer clippy (1.95) flags `collapsible_match` on the nested `if` inside the `Text` arm of `fragment_has_collapse_candidate`, which breaks `cargo clippy --all-targets --all-features -- -D warnings` (and thus the pre-commit hook) on current main. Fold the blank-text check into a match guard; behavior is unchanged — blank `Text` still falls through to the `element_container` probe.

Verified: `cargo clippy -p rsvelte_formatter --all-targets --all-features -- -D warnings` clean; `cargo test -p rsvelte_formatter --lib` 81/81.

Found while reviewing #1828, whose agent hit this as a pre-existing workspace clippy failure.

https://claude.ai/code/session_014ER8ncJhRQKMdxB1e7a7wH

### DIFF
--- a/crates/rsvelte_formatter/src/collapse/mod.rs
+++ b/crates/rsvelte_formatter/src/collapse/mod.rs
@@ -63,11 +63,7 @@ pub(crate) fn fragment_has_collapse_candidate(fragment: &Fragment) -> bool {
         // top-level text and `{@html}` runs are collapse targets too, not just
         // element bodies.
         match n {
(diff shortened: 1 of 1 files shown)


baseballyama:
Superseded: the identical clippy fix landed on main via #1830's first commit.

Issue #1735, opened 2026-07-23
css-prune: multi-relative sibling Chain resolution doesn't guard snippet/dynamic-element ancestor blind spot


baseballyama:
## Summary

The multi-relative ancestor `Chain` resolution added in #1728 (for `:global(X)`'s
inner descendant/child chain and for a bare `&` resolving against a
multi-relative parent prelude) walks `parent_idx` directly via
`structural_ancestors_satisfy_links` / `structural_element_matches_compound`
without the two safety guards that the *other* caller of those same
primitives, `is_structural_descendant_chain_unused`
(`crates/rsvelte_core/src/compiler/phases/3_transform/css.rs`, ~L3169), already
applies before trusting a raw ancestor walk:

- `ctx.has_dynamic_elements` bail
- bail when any element in the whole `dom_structure` is `in_snippet` or is a
  `<selectedcontent>` (because a `{#snippet}`-declared element's real ancestor
  is determined by its `{@render}` call site, not its lexical position, and a
  lexical `parent_idx` walk cannot model that)

The new code paths this affects: `chain_is_structurally_evaluable`,
`level_is_structurally_evaluable`, `build_parent_chain`,
`resolve_bare_nesting_chain`, `resolve_global_inner_matcher`,
`resolve_sibling_matcher` — none of these take `ctx`, so they have no way to
even check these conditions.

## Repro

```svelte
<div class="foo"><span>irrelevant</span></div>
{#snippet pair()}
  <div class="a"></div><div class="a"></div>
{/snippet}
<section>
  {@render pair()}
</section>
<style>
  .foo > .a { & + & { color: red; } }
</style>
```

- Official compiler (`svelte@5.56.4`): prunes the whole nested rule —
  `/* (empty) .foo > .a { & + & { color: red; } }*/` — because the only real
  render site of the `.a` pair is under `<section>`, not `.foo`.
- rsvelte (current `main`, with #1728 applied): keeps it —
  `.foo.svelte-xxx > .a:where(.svelte-xxx) { & + & { color: red; } }`.

## Not a regression from #1728

The exact same wrong output reproduces on `main` **before** #1728 was applied
(verified by running the identical test against the pre-#1728 tree). Pre-#1728,
`& + &` fell back to the single-compound `Info` path, which ignored the `.foo`
ancestor constraint entirely but still matched the two literally-adjacent `.a`
siblings inside the snippet body — arriving at the same "kept" answer by a
[...]

PR #1851, opened 2026-07-26
fix(css): guard multi-relative chain resolution against non-lexical ancestry

baseballyama:
## Summary

The multi-relative ancestor `Chain` resolution added in #1728 (`resolve_global_inner_matcher`
for `:global(.a .z) + .b`, `resolve_bare_nesting_chain` for `.foo > .a { & + & }`) walks
`parent_idx` directly, without the ancestry guard the other caller of the same primitives,
`is_structural_descendant_chain_unused`, already applies. `parent_idx` is *lexical*, and two
constructs make it a lie:

- a `{#snippet}`-declared element's real ancestors come from its `{@render}` call sites, not
  its lexical parent;
- `<selectedcontent>` mirrors the selected `<option>`'s subtree, so elements inside an
  `<option>` are also descendants of the `<selectedcontent>`.

This PR extracts that condition into `structural_ancestry_is_lexical(ctx)` and gates the two
`Chain` producers on it.

## Why a new `SiblingMatcher::Unresolvable` instead of "bail to the `Info` fallback"

The issue suggested bailing to the compound-only `Info` path. Measured against the official
compiler, that is actively wrong: `Info` *discards* the ancestor constraint, the operand
resolves to an empty (matches-nothing) selector, and with no opaque-boundary fallback in play
the rule gets pruned. It regressed two cases that `main` gets right, e.g.

```svelte
<script>let tag = 'div';</script>
<svelte:element this={tag} class="foo"><div class="a"></div><div class="a"></div></svelte:element>
<style>.foo > .a { & + & { color: red; } }</style>
```

So the guard produces `SiblingMatcher::Unresolvable`, and the two call sites in
`is_sibling_combinator_unused` bail conservatively (`return false` — "not unused", keep the
rule) rather than falling through with a weaker matcher.

## Why the `has_dynamic_elements` half of the guard is *not* applied to the `Chain` paths

`is_structural_descendant_chain_unused` also bails on `ctx.has_dynamic_elements`; that stays
there unchanged, but is deliberately not carried over to the `Chain` paths. `svelte:element`
does not remap ancestry — it is a real node at a known position, and
`structural_element_matches_compound` already relaxes tag/attribute matching via
`is_dynamic_tag`. Including it regressed two cases against official output:

```svelte
<script>let tag = 'div';</script>
<div class="foo"><svelte:element this={tag}><div class="a"></div><div class="a"></div></svelte:element></div>
<style>.foo > .a { & + & { color: red; } }</style>
```

official: `/* (empty) … */`; with `has_dynamic_elements` in the guard rsvelte kept it. Both
cases are covered by regression tests.

## Effect on the issue's repro — NOT fixed, documented as a limitation

```svelte
<div class="foo"><span>irrelevant</span></div>
{#snippet pair()}
  <div class="a"></div><div class="a"></div>
{/snippet}
<section>
  {@render pair()}
</section>
<style>
  .foo > .a { & + & { color: red; } }
</style>
```

Official `svelte@5.56.8` prunes: `/* (empty) .foo > .a { & + & { color: red; } }*/`.
rsvelte still keeps it, before and after this PR.

Verified root cause: this is not the sibling-combinator code at all. rsvelte also keeps the
plain `.foo > .a { color: red; }`, `.foo .a { … }` and `.foo > .a + .a { … }` variants of the
same template, because `is_structural_descendant_chain_unused` bails on *any* snippet element —
`dom_structure` has no model of `{@render}` call sites, so it cannot tell the `.a` pair's real
parent is `<section>`. Fixing the repro requires teaching phase 2's `CssDomElement` a
render-site ancestry link (`parent_idx` is snippet-boundary-unaware, so this also touches
`children_idx` / `is_root_child` and the recursive ancestor walk, which would have to branch
over multiple ancestor paths). That is a separate architectural change, so #1735 stays open.

The repro is landed as an explicit known-limitation test
(`snippet_rendered_under_mismatched_ancestor_over_kept`), so the assertion flips the day
render-site modelling lands.

## What this PR does buy

- Fixes one real over-prune on `main`: `selectedcontent > .a { & + & }` was emitted as
  `/* (empty) */` where official keeps it.
- Stops the #1728 `Chain` paths from extending the lexical-ancestry blind spot, and makes both
  callers of the structural primitives share one named predicate.

## Verification

13 shapes compiled with both `submodules/svelte` (`svelte@5.56.8`) and rsvelte, diffing
`css.code`: matching/mismatched snippet render sites, snippet declared inside vs. outside the
required ancestor, `:global()` inner chains with snippets, `selectedcontent` as ancestor and as
bystander, `svelte:element` as ancestor and as an intervening node, plus the snippet-free
positive/negative controls. After this PR all 13 match official except the documented repro; on
`main`, `selectedcontent > .a { & + & }` also diverged.

Refs #1735

---

https://claude.ai/code/session_01GrqdRdS2YdA12XDA24KUHp

### DIFF
--- a/.changeset/fix-css-chain-non-lexical-ancestry.md
+++ b/.changeset/fix-css-chain-non-lexical-ancestry.md
@@ -1,0 +1,15 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(css): guard multi-relative chain resolution against non-lexical ancestry (#1735)
+
+The `+`/`~` prune check resolves a multi-relative operand (`:global(.a .z) + .b`,
+or a bare `&` against a `.foo > .a` parent prelude) into an ancestor chain
+verified by walking `parent_idx`. That walk is lexical, so it silently
+mis-answers for `{#snippet}` bodies (whose real ancestors come from their
+`{@render}` call sites) and for `<selectedcontent>` (which mirrors the selected
+`<option>`'s subtree). Both `Chain` producers now share the predicate the
+descendant-chain check already used and bail conservatively when the ancestry is
+not lexical, fixing `selectedcontent > .a { & + & }` being emitted as
+`/* (empty) */` where the official compiler keeps it.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/css.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/css.rs
@@ -2371,7 +2371,10 @@ fn is_sibling_combinator_unused(rel_selectors: &[Value], ctx: &CssContext) -> bo
             // like `:global(.a .z)` becomes a `Chain` so the `.a` ancestor of a
             // candidate `.z` sibling is verified (rather than left unresolved,
             // which over-prunes even when the ancestor really exists).
-            let inner_matcher = resolve_global_inner_matcher(&rel_selectors[0]);
+            let inner_matcher = resolve_global_inner_matcher(&rel_selectors[0], ctx);
+            if matches!(inner_matcher, SiblingMatcher::Unresolvable) {
+                return false;
+            }
 
             // `:global(X) + Y` is used when some Y is preceded by a node X could
             // be: a real previous sibling matching X, an opaque boundary, or a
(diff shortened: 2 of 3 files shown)


baseballyama:
Note: the `Test unit` check failure (`diagnostics_test::constructor_counts_are_pinned` / `diagnostic_codes_and_messages_are_pinned`) is **pre-existing on `main`**, not from this PR — `main` @ 6a487171 (#1849) is red with the same two assertions, and this branch touches no diagnostic definitions. It needs `ERROR_MACRO_COUNT` / `DIAGNOSTICS_DIGEST` (→ `0xc8f80792df41fb82`) updated on `main`.

Issue #782, opened 2026-06-08
`--tsgo`: cross-package `.svelte` import resolves to ambient `*.svelte` (default-only) → module-context named/`default` re-exports reported missing


baseballyama:
## Summary

When a `.svelte` component in **another workspace package** is imported through that package's `exports` (a TS barrel that re-exports the component's **module-context** members), `--tsgo` resolves the `.svelte` to the ambient `declare module '*.svelte'` (default-only) instead of the component's real module, so its named `<script module> export const …` members — and even `export { default }` re-exports — are reported missing:

```
Module '"*.svelte"' has no exported member 'SIDEBAR_DEFAULT_WIDTH'. Did you mean to use 'import SIDEBAR_DEFAULT_WIDTH from "*.svelte"' instead?
Module '"./highlight/Highlight.svelte"' has no exported member 'default'.
Circular definition of import alias 'HIGHLIGHT_WRAPPER_ATTRIBUTE'.
```

Official `svelte-check`: 0 errors. The same pattern **within a single package** resolves correctly — only the cross-package (imported-via-`exports`) case fails.

## Shape (real)

Package B (`@flyle-lib/design-system`):
```svelte
<!-- sidebar/Sidebar.svelte -->
<script module lang="ts">export const SIDEBAR_DEFAULT_WIDTH = 256 as const;</script>
<script lang="ts">/* … */</script>
```
```ts
// components/index.ts  (barrel, the package's exports entry)
export { SIDEBAR_DEFAULT_WIDTH } from './sidebar/Sidebar.svelte';
export { default as Highlight, HIGHLIGHT_WRAPPER_ATTRIBUTE } from './highlight/Highlight.svelte';
```
Package A (`@flyle-lib/shared-ui`):
```svelte
<script lang="ts">
  import { SIDEBAR_DEFAULT_WIDTH } from '@flyle-lib/design-system/components';
</script>
```

Running `--tsgo` on package A pulls B's `index.ts` into the program; every re-export from a B `.svelte` then errors as above. (A clean single-package repro of `<script module> export const` + a `.ts` barrel + a consumer does **not** error, which is what points at the cross-package boundary.)

## Likely cause

[...]

PR #1884, opened 2026-07-29
fix(svelte-check): resolve cross-package .svelte imports reached via a paths alias (#1883)

### LINKED ISSUE #1883: `--tsgo`/`svelte-check`: paths-aliased (non-node_modules) sibling-package .svelte import resolves to ambient `*.svelte` — module-context exports reported missing
## Summary

When a `.svelte` file lives in a **sibling pnpm workspace package** and is reached through a plain `tsconfig.json` **`paths` alias** (e.g. SvelteKit's `kit.alias` / Vite `resolve.alias`, resolved as a raw relative path — not a package `exports` barrel, and not a `node_modules` symlink), the overlay resolves the import to the ambient `declare module '*.svelte'` (default-only) instead of the component's real module. Its `<script module> export type/const …` members are then reported missing.

Official `svelte-check`: **0 errors**.

This looks related to #782 / #805, but is a distinct trigger: #805's fix discovers sibling packages by walking `node_modules` symlinks. A `paths`-aliased sibling reached without any `node_modules` entry for that package (common in SvelteKit monorepos — `$lib`-style aliases pointing straight at another workspace package's source tree, with no `exports` field or declared dependency involved) isn't found by that discovery, so the bug still reproduces on the latest release.

## Minimal repro

```
repo/
├── pnpm-workspace.yaml       # packages: [pkg-a, pkg-libs]
├── pkg-libs/
│   ├── package.json          # { "name": "libs" }
│   └── components/survey-options/survey-options.svelte
└── pkg-a/
    ├── package.json          # no dependency on "libs" at all
    ├── tsconfig.json
    └── src/consumer.svelte
```

`pkg-libs/components/survey-options/survey-options.svelte`:
```svelte
<script module lang="ts">
  export type WithOther<T extends string> = T | `OTHER: ${string}`;
</script>

<script lang="ts">
  let { id }: { id: string } = $props();
</script>

<div>{id}</div>
```

`pkg-a/tsconfig.json` (relevant bit — no `baseUrl`, just a relative `paths` mapping, the SvelteKit-alias equivalent):
```json
{
  "compilerOptions": {
    "moduleResolution": "bundler",
    "paths": { "$libs/*": ["../pkg-libs/*"] },
    "types": ["svelte"]
  },
  "include": ["src/**/*.ts", "src/**/*.svelte", "../pkg-libs/**/*.svelte"]
}
```

`pkg-a/src/consumer.svelte`:
```svelte
<script lang="ts">
  import SurveyOptions, { type WithOther } from '$libs/components/survey-options/survey-options.svelte';
  type X = WithOther<'a' | 'b'>;
</script>

<SurveyOptions id="a" />
```

Note: `pkg-a/package.json` does **not** list `libs` as a dependency at all — there is no `node_modules/libs` entry, symlink or otherwise. Resolution goes entirely through the `paths` alias.

Official `svelte-check --tsconfig ./tsconfig.json`: `0 errors and 0 warnings`.

```
ERROR .svelte-check/svelte/src/consumer.svelte.tsx:3:30 (ts): Module '"*.svelte"' has no exported member 'WithOther'. Did you mean to use 'import WithOther from "*.svelte"' instead?
```
Same error with the `tsc` backend (no `--tsgo`), so it's not tsgo-specific — it's the overlay/module resolution.

## Likely cause

Per #805's writeup, sibling-package discovery walks `node_modules` symlinks whose realpath resolves outside the checked workspace. A package reached purely through a `tsconfig.json`/bundler `paths` alias (no corresponding `node_modules` entry) never enters that discovery, so no shadow `.tsx`/`.d.ts` is emitted for it and no `rootDirs` bridge is set up — same end symptom as #782, different root trigger (alias-based resolution vs. `node_modules`-symlink-based resolution).

## Environment

- `@rsvelte/svelte-check` 0.5.2 (also reproduces on the `tsc` backend, not just `--tsgo`)
- `@typescript/native-preview` latest (`tsgo`)
- Found while trying `rsvelte-check --tsgo` as a `svelte-check` replacement on a large SvelteKit monorepo (pnpm workspaces, `kit.alias` mapping `$libs` → a sibling package's source dir). Real workspace triggers 475 diagnostics of this shape across the app, all false positives — official `svelte-check` reports 0 errors/warnings on the same tree.

Mokto:
Fixes #1883.

## Problem

`discover_external_svelte_packages` only found sibling workspace packages reachable through a `node_modules` symlink (#805's fix for #782). A `.svelte` in a sibling package reached purely through a `tsconfig.json`/bundler `paths` alias (SvelteKit's `kit.alias`, `resolve.alias`, etc.) with **no `node_modules` entry for that package at all** still fell back to the ambient `declare module '*.svelte'` (default-only), so the component's `<script module>` named exports (and, in a barrel, its `default` re-export) were reported missing — even though official `svelte-check` resolves it with 0 errors.

## Fix

- `discover_external_svelte_packages` now also resolves `compilerOptions.paths` alias targets (via a new `resolve_paths_alias_dirs_abs`, following `extends` the same way `resolve_root_dirs_abs` already does) and treats any target landing outside the workspace as an external package to mirror, same as the `node_modules`-symlink case.
- `rewrite_aliased_svelte_imports` — which re-points an aliased `.svelte` specifier at its shadow `.tsx` — previously only did this for targets *under the workspace*; a target under one of the newly-discovered external packages is now rewritten to that package's mirror shadow too. (Picks the longest-matching external root, so a more specific alias prefix wins over a broader one.)
- External-package discovery + shadow emission now runs **before** the per-file loop (it used to run after), since the per-file alias rewrite needs the mirrors to already exist.

### A second bug this surfaced

`build_svelte_import_resolver` handed `oxc_resolver` the `--tsconfig` path verbatim. With a **relative** `config_file`, `oxc_resolver`'s tsconfig discovery returns `NotFound` for any `paths` target that escapes the CWD via `..` — which is exactly what every cross-package alias does (a sibling package is reached by climbing up and over). `--tsconfig ./tsconfig.json` is the CLI's own documented usage, so this silently defeated alias rewriting for precisely the case this fix targets. Fixed by absolutising before constructing the resolver.

## Verification

- Two new unit tests: `cross_package_paths_alias_named_export_resolves_to_its_shadow` (the #1883 repro shape: no `node_modules` entry, resolution entirely through `paths`) and `relative_tsconfig_path_still_resolves_paths_aliases_that_escape_cwd` (regression test for the second bug).
- Full `cargo test -p rsvelte_core --lib`: **1527 passed, 0 failed** (up from 1525 — the 2 new tests — no regressions).
- `cargo fmt` clean; `cargo clippy -p rsvelte_core --lib -- -D warnings` clean on the touched file (there's one pre-existing, unrelated `collapsible_match` failure elsewhere in the crate on `main` today, confirmed via `git stash` — not touched by this PR).
- Real-world: ran the release binary against a ~900-file SvelteKit monorepo workspace with a `$libs` alias into a sibling pnpm package (no `node_modules` entry for it). `rsvelte-check --tsgo` false-positive count: **475 → 15**. The residual 15 were triaged into 4 further distinct bugs, each filed and fixed separately: #1886/#1892 (SvelteKit hooks written as `const` arrow exports), #1887/#1893 (`ComponentProps` through a self-referential alias inside an external package), #1888/#1895 (a plain `.ts` file's aliased `.svelte` import), #1889/#1896 (`<svelte:boundary onerror>` untyped). Combining all five branches together and re-running against the same real app: **0 errors**, matching official `svelte-check` exactly.


baseballyama (approved):
Thank you so much!

PR #1892, opened 2026-07-29
fix(svelte-check): type SvelteKit hooks written as const arrow/function-expression exports (#1886)

### LINKED ISSUE #1886: svelte-check: SvelteKit hook typed as `export const handleFetch = (...) => {...}` (arrow form) doesn't get its params typed — only `export function handleFetch(...)` does
## Summary

A SvelteKit hook exported as a `const` arrow function (`export const handleFetch = async ({ request, fetch, event }) => {...}`, the form shown in [SvelteKit's own docs](https://svelte.dev/docs/kit/hooks#Server-hooks-handleFetch) and probably the more common style in real code) doesn't get its destructured parameter typed by the hook-file augmentation, so every binding element is reported as implicit `any`.

The **function-declaration** form (`export function handleFetch({ request, fetch, event }) {...}`) works correctly — this is specifically about the `const` + arrow-function form.

Official `svelte-check`: **0 errors** on both forms.

## Minimal repro

`src/hooks.server.ts`:
```ts
export const handleFetch = async ({ request, fetch, event }) => {
  return fetch(request);
};
```
(needs a `src/routes/+page.svelte` present so the workspace has at least one `.svelte` file, and `svelte.config.js` present so the project is detected as SvelteKit)

`tsconfig.json`:
```json
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler",
    "strict": true,
    "types": ["@sveltejs/kit"]
  },
  "include": ["src/**/*.ts", "src/**/*.svelte"]
}
```

```
ERROR src/hooks.server.ts:1:36 (ts): Binding element 'request' implicitly has an 'any' type.
ERROR src/hooks.server.ts:1:45 (ts): Binding element 'fetch' implicitly has an 'any' type.
ERROR src/hooks.server.ts:1:52 (ts): Binding element 'event' implicitly has an 'any' type.

svelte-check found 3 errors and 0 warnings in 1 file
```

Changing only the declaration form fixes it:
```ts
export function handleFetch({ request, fetch, event }) {
  return fetch(request);
}
```
→ `svelte-check found 0 errors and 0 warnings in 1 file`.

## Likely cause

`kit_file.rs`'s `add_hooks_type` (used for `visit_server_hooks_statement` / `visit_client_hooks_statement` / `visit_universal_hooks_statement`, covering `handle`, `handleError`, `handleFetch`, `reroute`) only matches:

```rust
let oxc::Statement::ExportNamedDeclaration(ex) = stmt else { return; };
let Some(oxc::Declaration::FunctionDeclaration(f)) = &ex.declaration else { return; };
```

An `export const handleFetch = (...) => {...}` is an `ExportNamedDeclaration` wrapping a `Declaration::VariableDeclaration` whose initializer is an arrow function expression, not a `Declaration::FunctionDeclaration` — so the matcher silently no-ops and no parameter-type augmentation is inserted. Since this one function backs all four hook names, the gap likely affects `handle` / `handleError` / `reroute` too whenever they're written in the arrow-const style.

## Environment

- Found while auditing residual false positives after fixing #1883/#1884 on a real SvelteKit monorepo workspace.

Mokto:
## Problem

`--tsgo`/`svelte-check` only augments a SvelteKit hook (`handle`, `handleError`, `handleFetch`, `reroute`) with `Parameters<T>[0]`/`ReturnType<T>` when it's written as a `FunctionDeclaration`:

```ts
export function handleFetch({ request, fetch, event }) { ... } // ✅ typed
```

The equally common `const` + arrow/function-expression form fell through untouched, leaving every destructured param `implicit any`:

```ts
export const handleFetch = async ({ request, fetch, event }) => { ... } // ❌ implicit any × 3
```

Official `svelte-check` types both forms identically (`findExports` in the JS reference folds `FunctionDeclaration`/`ArrowFunction`/`FunctionExpression` into one `'function'` variant).

## Fix

- `add_hooks_type` now also matches `Declaration::VariableDeclaration` whose single declarator initializes to an `ArrowFunctionExpression` or `FunctionExpression`, applying the same param/return-type insertion (factored into a shared `add_hooks_type_to_function_like` helper). For an arrow function, the return-type insertion point is the params list's own span end (right before ` => `) rather than the function body's start.
- Fixing that surfaced a second, adjacent bug: naively inserting `: ReturnType<HandleFetch>` on an `async` arrow trips `TS1064` ("the return type of an async function ... must be the global Promise<T> type"), since `HandleFetch` returns `MaybePromise<Response>`, not a literal `Promise<T>`. Official svelte-check never hits this because every kit-file insertion is wrapped in `/*Ωignore_startΩ*/…/*Ωignore_endΩ*/` (`surroundWithIgnoreComments`) and diagnostics landing inside are dropped unless their code is in a small allowlist (`TS2307`/`TS2694`/`TS2355`) — this repo already implements that exact mechanism (`mapper.rs::is_in_generated_code`) for svelte2tsx `.tsx` shadows, but `kit_file.rs`'s insertions were never wrapped and `mapper.rs`'s kit-file branch never consulted it. Fixed both: `build_added_code` now wraps every insertion in the ignore markers, and `map_tsgo_diagnostics`'s kit-file branch now drops diagnostics landing inside them (same allowlist as official).

## Verification

- `cargo test -p rsvelte_core --lib` — 1529/1529 pass (8 new: 3 hooks-augmentation tests, 1 mapper ignore-region test, 4 existing assertions updated for the new markers).
- `cargo fmt --check`, `cargo clippy -p rsvelte_core --lib -- -D warnings` clean (pre-existing unrelated `collapsible_match` failure on `main` aside).
- Isolated repro (`export const handleFetch = async ({ request, fetch, event }) => { return fetch(request); }`): official `svelte-check` 0 errors; this branch's `--tsgo` and `tsc` backends both now 0 errors (previously 3 `implicit any`).
- Sanity check that suppression isn't over-broad: a genuine type error inside the same hook body (`const bad: string = 123`) still surfaces identically to official svelte-check.

Fixes #1886.


baseballyama (approved):
Thank you so much!

PR #1893, opened 2026-07-29
fix(svelte-check): resolve a self-referential aliased .svelte import inside an external package's own shadow (#1887)

### LINKED ISSUE #1887: svelte-check: ComponentProps<typeof X>['prop'] loses its signature when X is reached via a self-referential alias inside its own external package
## Summary

`ComponentProps<typeof Foo>['someProp']`, used to type another component's prop, resolves to a signature that loses its parameter types — turning any inline callback assigned to that prop into an implicit-`any` false positive — specifically when `Foo` is itself reached through a `paths`/bundler alias **from within its own external (sibling) package** (i.e. the package aliases its own absolute import path back to itself, a common monorepo pattern: `$libs/components/input.svelte` used both by external consumers and internally by other components inside `libs` itself).

Official `svelte-check`: **0 errors**. This reproduces on both the `--tsgo` and `tsc` backends.

This is different from #1149 (which fixed a *bare npm package* import inside an external shadow resolving to `any`) — here the import that goes unresolved is another **`.svelte` file inside the same external package**, reached via the alias rather than a relative specifier.

## Minimal repro

Two-package pnpm workspace (`pkg-a` consumes `libs` purely through a `paths` alias, no `node_modules` entry for `libs` at all):

`pkg-libs/components/input.svelte`:
```svelte
<script lang="ts">
  interface Props {
    onChange?: (value: string) => void;
  }
  let { onChange }: Props = $props();
</script>
<input oninput={(e) => onChange?.((e.target as HTMLInputElement).value)} />
```

`pkg-libs/components/selection-menu.svelte` — imports its sibling `input.svelte` through the **same alias** other packages use, not a relative path:
```svelte
<script lang="ts">
  import Input from '$libs/components/input.svelte';
  import type { ComponentProps } from 'svelte';

  interface Props {
    search?: {
      onChange?: ComponentProps<typeof Input>['onChange'];
      placeholder?: string;
    };
  }
  let { search }: Props = $props();
</script>
{#if search}
  <Input onChange={search.onChange} />
{/if}
```

`pkg-a/tsconfig.json`:
```json
{
  "compilerOptions": {
    "paths": { "$libs/*": ["../pkg-libs/*"] }
  },
  "include": ["src/**/*.svelte", "../pkg-libs/**/*.svelte"]
}
```

`pkg-a/src/consumer.svelte`:
```svelte
<script lang="ts">
  import SelectionMenu from '$libs/components/selection-menu.svelte';
</script>
<SelectionMenu
  search={{
    onChange: async value => {
      console.log(value.toUpperCase());
    },
  }}
/>
```

```
ERROR src/consumer.svelte:7:30 (ts): Parameter 'value' implicitly has an 'any' type.
```
Official `svelte-check`: 0 errors.

## Likely cause

`emit_external_shadows` (which materialises shadows for a sibling package discovered by `discover_external_svelte_packages`, e.g. via #1884's `paths`-alias discovery or the pre-existing `node_modules`-symlink discovery) never calls `rewrite_aliased_svelte_imports` on the `.tsx` it emits — only the primary in-workspace per-file loop does. So `selection-menu.svelte`'s own `import Input from '$libs/components/input.svelte'` specifier is left untouched inside its mirror shadow. `$libs/*` still resolves (via the *consumer*'s inherited `paths`) to the real `input.svelte` source, but with no shadow reachable at that specifier (`rootDirs` doesn't apply to non-relative specifiers), so `Input`'s type falls back to the ambient default-only `declare module '*.svelte'`. `ComponentProps<typeof Input>['onChange']` therefore can't recover a concrete function signature, and any inline callback assigned to it loses parameter-type inference.

A fix likely needs `emit_external_shadows` to run its own alias-rewrite pass, resolved against the external package's own tsconfig if it has one (falling back to the consumer's, since in practice the alias mapping is usually monorepo-wide/consistent).

## Environment

- Found while auditing residual false positives after fixing #1883/#1884 on a real SvelteKit monorepo workspace — this pattern (a design-system package that imports its own components via the same public alias apps use) accounted for 4 of the remaining false positives there.

Mokto:
## Problem

`ComponentProps<typeof Foo>['someProp']`, used to type another component's prop, loses its parameter types — turning any inline callback assigned to that prop into an implicit-`any` false positive — when `Foo` is itself reached through a `paths`/bundler alias **from within its own external (workspace-sibling) package**. This is a common monorepo pattern: a design-system package imports its own components through the same public alias its consumers use (`$lib/Input.svelte`), not a relative path.

Different from #1149 (a *bare npm package* import inside an external shadow resolving to `any`) — here the unresolved import is another **`.svelte` file inside the same external package**, reached via the alias rather than a relative specifier.

Official `svelte-check`: 0 errors on this pattern.

## Root cause

`emit_external_shadows` (which materialises shadows for a sibling package discovered via a `node_modules` symlink, #782) never called `rewrite_aliased_svelte_imports` on the `.tsx` it emits — only the primary in-workspace per-file loop did. So a component's own aliased import of a sibling in the *same* package was left untouched inside its shadow. The alias itself still resolves (via the *consumer*'s inherited `paths`) to the real source `.svelte`, but with no shadow reachable at that specifier (`rootDirs` doesn't bridge non-relative specifiers), so the imported component's type falls back to the ambient default-only `declare module '*.svelte'`.

## Fix

- `rewrite_aliased_svelte_imports` now also matches specifiers that resolve under one of the discovered external packages' own real dirs (not just the workspace), with longest-real-dir-prefix-wins when packages nest.
- External-package discovery now runs *before* the primary per-file loop (rather than after), so its `rootDirs` pairs are available both to that loop's own rewrite call and to `emit_external_shadows`, which now invokes the same rewrite on its own emitted shadows.
- Bycatch: a relative `--tsconfig` path made oxc_resolver's tsconfig discovery silently return `NotFound` for any `paths` target escaping the CWD via `..` — exactly every cross-package alias, and exactly the CLI's own documented `--tsconfig ./tsconfig.json` usage. Anchored it on the CWD before building the resolver, otherwise this fix (and the pre-existing in-workspace alias rewrite) is a no-op under normal CLI usage.

## Verification

- `cargo test -p rsvelte_core --lib` — 1526/1526 pass (3 new: an integration test asserting the external shadow's self-referential import is rewritten, plus a relative-`--tsconfig` regression test for the node_modules-sibling case).
- Isolated 2-package repro (consumer imports `SelectionMenu` via a `node_modules` symlink; `SelectionMenu` imports its own sibling `Input` via the same alias; consumer assigns an inline async callback to `search.onChange`, typed as `ComponentProps<typeof Input>['onChange']`): official `svelte-check` 0 errors; this branch's `--tsgo` and `tsc` backends both now 0 errors (previously 1 false `implicit any`).
- Sanity check that the fix isn't over-broad: a genuine typo in the same callback (`value.toUppercase()`) still surfaces the real error identically to official svelte-check.

Fixes #1887.


baseballyama (approved):
Thank you so much!

PR #1895, opened 2026-07-29
fix(svelte-check): resolve a plain .ts/.js file's paths-aliased .svelte import via an exact tsconfig paths override (#1888)

### LINKED ISSUE #1888: svelte-check: a plain .ts/.svelte.ts file's aliased .svelte import is never rewritten to its shadow (only .svelte-file importers are)
## Summary

A **plain `.ts` / `.svelte.ts`** source file (not itself a `.svelte` file, so never run through svelte2tsx) that imports a `.svelte` component through a `paths`/bundler alias gets no alias rewriting at all — `rewrite_aliased_svelte_imports` is only invoked on the `.tsx` shadow generated for a `.svelte` file. The import therefore resolves to the ambient `declare module '*.svelte'` fallback, which surfaces as one of two symptoms depending on how the import is used:

- a **named** import (`import type { Foo } from '$lib/x.svelte'`) → `Module '"*.svelte"' has no exported member 'Foo'`
- a **generic default** import used as a type (`import type Comp from '$lib/x.svelte'; ... Comp<T>`) → `Type 'Comp' is not generic` (the ambient fallback's default export isn't generic)

Official `svelte-check`: **0 errors** in both cases. This reproduces on both the `--tsgo` and `tsc` backends, and with or without a `node_modules` entry for the target package (i.e. it's not specific to #1883/#1884's cross-package case — it also happens for a plain in-workspace alias).

Downstream, this also poisons any `.svelte` file that consumes a class/type declared this way: a field typed `Comp<Row> | null` in a `.ts` file gets the *wrong* (non-generic, or member-less) shape, so a correctly-typed `.svelte` usage of the same component gets reported as not assignable to it.

## Minimal repro ("not generic" flavor)

Two-package pnpm workspace, `pkg-a` → `libs` via `paths` alias:

`pkg-libs/components/data-table.svelte`:
```svelte
<script module lang="ts">
  export interface RowShape { id: string; }
</script>
<script lang="ts" generics="T extends RowShape">
  interface Props { rows: T[]; }
  let { rows }: Props = $props();
</script>
<ul>{#each rows as row}<li>{row.id}</li>{/each}</ul>
```

`pkg-a/tsconfig.json`: `"paths": { "$libs/*": ["../pkg-libs/*"] }`, `include` covers both `src/**/*.ts` and the sibling package's `.svelte` files.

`pkg-a/src/store.ts` — a **plain `.ts` file**:
```ts
import type DataTable from '$libs/components/data-table.svelte';

interface Row { id: string; }

export class Store {
  public table: DataTable<Row> | null = null;
}
```

```
ERROR src/store.ts:8:17 (ts): Type 'Comp' is not generic.
```

Swapping the import for a named type export reproduces the other flavor:
```ts
import type { RowShape } from '$libs/components/data-table.svelte';
```
→
```
ERROR src/store.ts:1:15 (ts): Module '"*.svelte"' has no exported member 'RowShape'. Did you mean to use 'import RowShape from "*.svelte"' instead?
```

Official `svelte-check`: 0 errors in both cases.

## Likely cause

`materialize_overlay_with`'s per-file loop only runs svelte2tsx (and the subsequent `rewrite_aliased_svelte_imports` pass) for `.svelte` sources. A plain `.ts`/`.js`/`.svelte.ts` file that also imports a `.svelte` component via an alias enters the TS program as-is (verbatim, no shadow, no rewrite), so its aliased `.svelte` specifiers are never redirected to the target's shadow the way a `.svelte` file's would be.

A fix likely needs the same specifier-rewrite treatment applied to non-`.svelte` sources that are part of the checked file set (or a `.d.ts` module-augmentation approach targeting the raw alias specifier itself, similar to the existing same-name-companion augmentation for #800).

## Environment

- Found while auditing residual false positives after fixing #1883/#1884 on a real SvelteKit monorepo workspace — this pattern accounted for 3 direct false positives (`.ts`/`.svelte.ts` files) plus 2 cascading type-mismatch false positives in the `.svelte` files that consumed the resulting wrong type.

Mokto:
## Problem

A **plain `.ts` / `.js` / `.svelte.ts`** source file (not itself a `.svelte` file, so never run through svelte2tsx) that imports a `.svelte` component through a `paths`/bundler alias gets no alias rewriting at all — `rewrite_aliased_svelte_imports` is only invoked on the `.tsx` shadow generated for a `.svelte` file. The import therefore falls back to the ambient `declare module '*.svelte'`, surfacing as one of two symptoms:

- a **named** `<script module>` export → `Module '"*.svelte"' has no exported member 'X'`
- a **generic default** import used as a type (`Comp<T>`) → `Type 'Comp' is not generic`

This also cascades: any `.svelte` file that consumes a type declared this way gets reported as not assignable to the (correctly-typed) component.

Official `svelte-check`: 0 errors in both cases. Reproduces on both `--tsgo` and `tsc`, with or without a `node_modules` entry for the target package.

## Fix

For every discovered `.svelte` file reachable through a `paths` alias (both in-workspace and external packages, discovered via `node_modules` symlinks per #782), the overlay tsconfig now adds an **exact (non-wildcard)** `paths` entry redirecting that one specifier straight at the component's shadow `.tsx`:

```json
"paths": {
  "$lib/*": ["../lib/*"],
  "$lib/components/data-table.svelte": ["/abs/.svelte-check/svelte/components/data-table.svelte.tsx"]
}
```

TypeScript prefers an exact `paths` match over a wildcard pattern, and since the resolved target no longer ends in `.svelte`, the ambient `*.svelte` wildcard is never consulted for it — regardless of which kind of file does the importing. This is simpler and more robust than trying to detect-and-rewrite every plain source file's import text (the approach used for `.svelte`-generated shadows): no augmentation-syntax restrictions to work around, and no need to enumerate/scan arbitrary `.ts`/`.js` files in the project.

The original `paths` object (including unrelated entries) is read through the same `extends`-chain resolution `rootDirs` already needs (a child `compilerOptions.paths` replaces the parent's wholesale, so it can't just be left to `extends`) and preserved — only the affected component gets a more specific entry alongside it.

(An earlier version of this fix tried a `declare module "<exact alias>"` ambient augmentation instead, reusing the existing same-name-companion mechanism's shape — but a specifier already reachable through the `*.svelte` wildcard makes any such declaration an *augmentation*, and TS forbids `export`/`export default` re-export syntax there. The `paths`-override approach sidesteps that restriction entirely.)

**Update:** a second commit fixes a bug this uncovered when verified against a real SvelteKit project rather than an isolated repro. A real project's `tsconfig.json` extends a *generated* `.svelte-kit/tsconfig.json` that owns `paths` — a two-hop chain — and checked with the CLI's own documented relative `--tsconfig ./tsconfig.json`. Two path-anchoring bugs only manifest through that combination (a single-hop tsconfig masks both):
- `resolve_paths_object_abs` walked the chain with `file.parent()` staying relative at each hop, compounding into a garbled, unresolvable path (`././.svelte-kit/../src/utils`) instead of an absolute one.
- an override's shadow-path target can also stay CWD-relative when `workspace` itself is relative, but `paths` values are resolved relative to the *overlay tsconfig's own directory*, not the CWD.

Both are now anchored with the same `absolutize()` helper `build_svelte_import_resolver` already uses for the identical reason, with a two-hop-chain regression test added. Verified end-to-end against a real ~900-file SvelteKit monorepo app (0 errors, matching official `svelte-check`, after combining with the other 4 PRs from this same investigation).

## Verification

- `cargo test -p rsvelte_core --lib` — 1527/1527 pass (2 new: an overlay-tsconfig test asserting the exact override is added while the original wildcard + unrelated exact entries survive the merge; a two-hop-`extends`-chain regression test for the path-anchoring fix).
- Isolated 2-package repro, both symptom flavors (named export / generic default), plus the cascading `.svelte`-file-consumer scenario: official `svelte-check` 0 errors; this branch's `--tsgo` and `tsc` backends both now 0 errors (previously 1 false positive each).
- Sanity check that the fix isn't over-broad: a genuine type mismatch (wrong generic argument) still surfaces on both the `.ts` file and the `.svelte` consumer, identically to official svelte-check.
- End-to-end: combined with the other 4 PRs from the same investigation (#1884, #1892, #1893, #1896) against a real ~900-file SvelteKit app — 0 errors, matching official `svelte-check` exactly.

Fixes #1888.


baseballyama (approved):
Thank you so much!

PR #1896, opened 2026-07-29
fix(svelte-check): add missing 'svelte:boundary' IntrinsicElements entry to the svelte-jsx-v4 shim (#1889)

### LINKED ISSUE #1889: svelte-check: <svelte:boundary onerror={e => ...}> callback parameter is implicit any
## Summary

`<svelte:boundary onerror={e => ...}>`'s callback parameter isn't typed — `e` is reported as implicit `any`, with no cross-package or alias involvement at all (fully self-contained, single-file repro).

Official `svelte-check`: **0 errors**.

## Minimal repro

```svelte
<script lang="ts">
  function reportSentry(e: unknown) {
    console.error(e);
  }
</script>

<svelte:boundary onerror={e => reportSentry(e)}>
  content
  {#snippet failed()}
    <p>failed</p>
  {/snippet}
</svelte:boundary>
```

```
ERROR src/app.svelte:9:21 (ts): Parameter 'e' implicitly has an 'any' type.
```
Official `svelte-check`: `svelte-check found 0 errors and 0 warnings`.

## Environment

- Found while auditing residual false positives after fixing #1883/#1884 on a real SvelteKit monorepo workspace — accounted for 3 of the remaining false positives there (all plain `<svelte:boundary onerror>` usages, no other pattern involved).

Mokto:
## Problem

```svelte
<svelte:boundary onerror={e => reportSentry(e)}>
  content
  {#snippet failed()}
    <p>failed</p>
  {/snippet}
</svelte:boundary>
```

reports `Parameter 'e' implicitly has an 'any' type` on `--tsgo` and the `tsc` backend. Official `svelte-check`: 0 errors. No cross-package or alias involvement — reproduces in a single, self-contained file.

## Root cause

`svelte:boundary` is handled generically as a special element, generating `svelteHTML.createElement("svelte:boundary", { onerror: e => ..., failed: () => {...} })` — same shape as `svelte:window`/`svelte:body`/etc. The embedded `svelte-jsx-v4.d.ts` shim's `IntrinsicElements` interface had no `'svelte:boundary'` entry (unlike those three, which *are* declared), so the call fell through to the interface's own catch-all:

```ts
interface IntrinsicElements {
  // ...
  [name: string]: { [name: string]: any };
}
```

Every property of the attrs object — including `onerror` — is then contextually typed as bare `any`. Contextual typing against `any` doesn't propagate a *parameter* type into an inline arrow function the way an actual function-typed prop would (e.g. `onerror?: (error: unknown, reset: () => void) => void`), so `e` gets no type at all, tripping `noImplicitAny`.

(Digging further: the *currently-published* `svelte-check@4.7.4`'s own bundled legacy `svelte-jsx-v4.d.ts` has this exact same gap — official svelte-check doesn't hit it because, for a Svelte 5 project, it loads a *different*, non-`-v4` shim whose `IntrinsicElements` derives from Svelte's own `svelte.JSX.IntrinsicElements` instead of hand-listing every element. rsvelte only ever embeds the `-v4` variant. That's a bigger, separate question (worth its own investigation/issue) about shim selection by Svelte version — out of scope here; this PR just fixes the concrete reported false positive by adding the missing entry to the shim rsvelte actually ships.)

## Fix

Added a `'svelte:boundary'` entry to `IntrinsicElements`, using the same `HTMLProps<'svelte:boundary', HTMLAttributes>` pattern the other `svelte:` elements already use — which (since the shim's own `HTMLAttributes` override is an empty interface) resolves to exactly `svelte/elements`' own `SvelteHTMLElements['svelte:boundary']` shape: `{ onerror?, failed?: Snippet<[error, reset]>, pending?: Snippet }`.

## Verification

- `cargo test -p rsvelte_core --lib` — 1526/1526 pass (1 new: asserts the shim embeds a `'svelte:boundary'` entry rather than falling through to the catch-all).
- Isolated repro: official `svelte-check` 0 errors; this branch's `--tsgo` and `tsc` backends both now 0 errors (previously 1 false `implicit any`).
- Extended repro exercising `onerror`'s 2nd param (`reset`) and both `{#snippet failed(error, reset)}` / `{#snippet pending()}`: 0 errors, matching official.
- Sanity check that the fix isn't over-broad: a genuine type error inside the `onerror` callback body still surfaces identically to official svelte-check.

Fixes #1889.


baseballyama (approved):
Thank you so much!

PR #1908, opened 2026-07-29
feat(svelte-check): prefer TypeScript 7 native tsc over legacy tsgo

Mokto:
## Summary

TypeScript 7.0 GA ships the native compiler as `tsc` — `@typescript/native-preview`'s separate `tsgo` binary is only needed on TypeScript 6 projects.

When `rsvelte-check --tsgo` runs against a workspace that resolves `typescript@>=7`, prefer `node_modules/.bin/tsc` over legacy `tsgo`. TypeScript 6 behaviour is unchanged (`tsgo` first, `tsc` fallback).

**On `typescript@^7`, `--tsgo` is optional.** Default `rsvelte-check` (no flag) already prefers native `tsc`. Keep `--tsgo` in scripts during the TS 6→7 transition; drop it once the catalog is on TS 7.

## Why

Consumers migrating to `typescript@^7` should not need to keep `@typescript/native-preview` installed just to get fast type-checking. The overlay + diagnostic mapping already works with TS 7's `tsc` (verified on a ~900-file SvelteKit app); only compiler discovery needed updating.

## Changes

- `find_compiler`: detect `typescript` major version by walking up from the workspace; flip `--tsgo` preference to `["tsc", "tsgo"]` on TS 7+
- Unit test: both binaries present + `typescript@7.0.2` → picks `tsc`
- Docs / CLI help / warning messages updated to recommend `typescript@^7` over native-preview and note `--tsgo` is optional on TS 7

## Test plan

- [x] `cargo test -p rsvelte_check --lib find_compiler` (3 tests pass)
- [x] Manual: `svelte_check --tsgo` on oceanio/platform webapp-v2 with `typescript@7.0.2` → 0 errors
- [x] Manual: same with `typescript@6` + both binaries → still picks `tsgo` (existing test)

## Follow-up (out of scope)

- `rsvelte_lint_types` corsa resolver still pins `@typescript/native-preview` for the `--api` protocol — separate from batch `svelte-check`
- Platform can drop `@typescript/native-preview` and `--tsgo` from check scripts once this ships and catalog bumps to `typescript@^7`


baseballyama (approved):
Thank you so much!

PR #1995, opened 2026-07-30
perf(server): drop the whole-output TypeScript strip from SSR

baseballyama:
## What

Phase 3 re-parsed the **entire generated SSR program** as TypeScript and stripped it, once per TS component. That was **7.96% of server-prod** in the wave-3 profile. This removes it, and moves the erasure it was backstopping to the one place that actually needed it.

Rebased onto `main` after `perf/compile-wave2` was squash-merged as #1912; only the two commits below are this PR's.

## Why the whole-output pass can go

The pass's stated premise — that template-expression interpolations are "spliced straight from source positions" — is *mostly* no longer true. Almost every expression now reaches the output through `visit_expr_raw` (`server/ast/mod.rs`), which is TS-safe on both branches:

- **Structural path**: `jsnode_to_oxc_expr` builds from the typed `JsNode`, and `remove_typescript_from_ast` (`compiler/mod.rs`, run immediately after parse) has already cleared every TS node from the fragment — ExpressionTag, HtmlTag, ConstTag, RenderTag, the `{#if}`/`{#each}`/`{#await}`/`{#key}` heads, snippet parameters, and all attribute kinds.
- **Fallback path**: the source-slice fallback re-parses via `reparse_expression`.

## …except for the source-slice reparse (second commit)

`{@const}` is the exception, and it is not a fallback: `visit_const_tag_sync` *always* decomposes by source slice, reading the declarator's `id` / `init` spans straight out of `state.source` and handing them to `reparse_expression`. That helper parsed TypeScript-aware and then collapsed only the four TS **expression wrappers** (`as`, `satisfies`, `!`, `<T>x`). Type-only syntax that is not a wrapper survived:

- parameter annotations on an inline arrow — `{@const startX = (d: { startDate: Date }) => …}`
- call type arguments — `xs.map<number>(…)`
- a `type` alias or annotated `const` inside an arrow's block body

The whole-output strip was silently rescuing all of it. Removing it regressed two real corpus files — layerchart's `Dodge/duration-bars-dense-lanes.svelte` and `Waffle/Waffle.svelte` — whose SSR output stopped being parseable at all. **Both were outside the six corpora the "0 changes" measurement covered** (the corpus has 34 sources / 13,987 entries), which is why the measurement did not see them.

The fix runs the analyzer's positional stripper over the already-parsed slice and re-parses the result as plain ES. `strip_typescript_from_program` returns its input unchanged when there is nothing to remove, so a non-TS slice pays no second parse — and the walk covers one small expression instead of every byte of generated output, which is the cost this PR is removing.

## Measurement (first commit)

Over 6 corpora (8657 `.svelte` × server dev **and** prod = 17314 compiles):

| metric | value |
|---|---|
| pass invocations | 7398 |
| invocations that changed the output | 0 |
| invocations where the strip's own TS parse failed | 0 |

Caveat, now known: that sample did not include layerchart, and the pass **was** load-bearing there. The per-mode figure (7398 / 2 = 3699) independently reproduces the prod-only count of 3695 measured separately during wave 3.

## Verification

- `pnpm run corpus:collect && corpus:compile && corpus:verify` over the full 13,987-entry corpus (client + server + client-dev): **no regressions**; client / server known failures unchanged at 1 each.
- `cargo test --release` green, including `tests/ssr` and `tests/runtime`.
- `cargo fmt` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean for the touched crates.

## Test

`tests/ssr_typescript_erasure.rs` pins the invariant the pass used to backstop: the SSR output of a TypeScript component must parse as **plain ES** (`SourceType::mjs()`) with zero diagnostics — precisely the condition rolldown enforces when it rejects `Type assertion expressions can only be used in TypeScript files`.

The second commit also **fixes** that suite: `type_annotations_in_const_and_snippet_are_erased` placed its `{@const}` at the fragment root, which is a `const_tag_invalid_placement` validation error, so the test panicked on `.expect("component should compile")` rather than asserting anything. Two cases are added for the paths that decompose by source slice — the `{@const}` initializer, and `{#each}` / `{#await}` bindings.

## Also

Two comments that justified themselves by this pass are updated: `server/ast/script.rs` referred to it as the "OLD oracle", and the SIMD `declare ` probe in `2_analyze/types.rs` was motivated by "generated output, which this runs over in full".

### DIFF
--- a/.changeset/ssr-drop-whole-output-ts-strip.md
+++ b/.changeset/ssr-drop-whole-output-ts-strip.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Drop the whole-output TypeScript strip from SSR codegen and erase type-only syntax in the template source-slice reparse instead. Fixes `{@const}` initializers with TypeScript-annotated arrow parameters (e.g. `{@const f = (d: T) => …}`) leaking TypeScript into the generated server output.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
@@ -613,8 +613,7 @@ fn strip_typescript_from_program_impl(
     // to ensure they're removed.
     //
     // Every keyword below starts with `declare `, so one SIMD scan for that prefix
-    // decides all three at once — the common case (generated output, which this runs
-    // over in full) has none and skips three whole-source `str::find` passes.
+    // decides all three at once and skips three whole-source `str::find` passes.
     if memchr::memmem::find(source.as_bytes(), b"declare ").is_some() {
         for keyword in &["declare global", "declare module", "declare namespace"] {
             let bytes = source.as_bytes();
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/mod.rs
@@ -166,24 +166,6 @@ pub(crate) fn transform_component_with_scripts(
         }
         GenerateMode::Server => {
             let code = server::transform_server(analysis, ast, source, options)?;
-            // Template-expression interpolations in the SSR output are spliced
-            // straight from source positions, so for a TypeScript component
-            // (`<script lang="ts">`) bits of TS-only syntax — `as T` casts,
-            // `<T>` generics, `! ` non-null assertions, `: T` annotations on
-            // inline destructures — can leak into the emitted JS inside
-            // `$.escape(...)`, `$.stringify(...)`, `$.attr_style(...)`, etc.
-            // rolldown then rejects the result with
-            // `Type assertion expressions can only be used in TypeScript files`.
-            // Run the same TypeScript strip the analyzer uses over the final
-            // output to catch every leak in one pass, rather than threading
-            // a per-expression strip through every visitor source-slice site.
-            // Real-world surface: shadcn-svelte's `base-color-picker.svelte`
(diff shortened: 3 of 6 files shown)


baseballyama:
## A/B measurement

Isolated measurement of this PR's single commit (`fee9dc72`) against its merge
base `2443dcec`, on a quiet machine.

**Protocol.** Both variants built as separate binaries and kept side by side —
never compared across a rebuild. One round runs `base change change base`
(palindromic, so a monotonic load drift inside the round cancels); per round each
variant's time is the mean of its two slots and the statistic is the paired
per-round delta. Harness `corpus_profile --reps 9` (min of 9 interleaved passes)
over bits-ui + shadcn-svelte = 2300 files / 3.97 MB.

| mode | base ms | this PR ms | median delta | faster in |
|---|---|---|---|---|
| **server-prod** | 308.63 | 273.19 | **−9.06%** | **9/10 rounds** |
| client-prod (control) | 380.75 | 376.19 | −0.48% | 6/10 rounds |
| client-dev (control) | 479.61 | 470.97 | +1.02% | 4/10 rounds |

The two client modes are controls: this change only touches the
`GenerateMode::Server` arm, so it cannot affect them. Their coin-flip win rates
(6/10, 4/10) and sub-1% medians are the noise floor of this box — which is what
makes server-prod's −9.06% at 9/10 rounds a real effect rather than drift.

−9.06% end-to-end is larger than the 7.96% the pass held in the profile, because
removing it also drops the allocation and the `String` round-trip that fed it.

<sub>Measured by an agent on `perf/compile-wave4`. Note that
`origin/perf/compile-wave2` has moved to `2443dcec` (wave3 merged in), so the
base above is wave3, not `21cfc981`.</sub>

PR #1997, opened 2026-07-30
Reuse borrowed svelte2tsx generics attributes

baseballyama:
## Summary
- borrow the instance script generics attribute instead of allocating its value
- extract it once and reuse it for generic-name collection, render generation, and component export
- preserve the existing malformed-instance-script gate
- add exact coverage for TypeScript, JSDoc, empty, absent, and both quote styles

## Validation
- rustfmt
- git diff --check


baseballyama:
CodSpeed reports all 126 active benchmarks as untouched, including the pinned TypeScript-generics svelte2tsx corpus. Closing this performance-only change because CI found no measurable instruction-count improvement.

PR #1998, opened 2026-07-30
Speed up case-insensitive svelte2tsx tag scans

baseballyama:
## Summary
- return directly from one-byte case-insensitive searches such as the style-tag closing delimiter
- accept the common exact lowercase tag match before ASCII case folding
- preserve mixed-case matching and byte offsets

## Validation
- rustfmt
- git diff --check
- existing focused tests cover lowercase, mixed-case, partial candidates, UTF-8 offsets, and start offsets


baseballyama:
CodSpeed reports all 126 active benchmarks as untouched, including the new pinned svelte2tsx corpus. Closing this performance-only change because CI found no measurable instruction-count improvement.

PR #2009, opened 2026-07-30
Skip unused svelte2tsx hoist dependency analysis

baseballyama:
## Summary
- return before type dependency analysis when an untyped props rune has no movable props interface
- return after candidate indexing when a named props type is imported or global rather than local
- match the official HoistableInterfaces early-return order
- verify both paths avoid building dependency graph edges while preserving declaration placement

## Validation
- rustfmt
- git diff --check


baseballyama:
CodSpeed measured all 127 active benchmarks, including project[untyped-props], as unchanged. The shortcut is semantically valid, but without a CI-measurable performance improvement it is not worth carrying. Closing rather than merging.

PR #2037, opened 2026-07-30
Skip unused svelte2tsx export metadata

baseballyama:
## Summary

- detect local export specifiers during the existing top-level declaration prepass
- skip `possible_exports` name allocation, type-text copying, JSDoc scans, and HashMap insertion when no local specifier can observe that metadata
- preserve the existing path for `export { local }` and add focused gate coverage

This follows the official `ExportedNames` flow: possible declarations are only observed while handling a named local export declaration.

## Performance

The fixed CodSpeed svelte2tsx workload has no local export specifiers across its ten projects, so this removes metadata collection for roughly 57 variable bindings and more than 50 leading-JSDoc probes per full workload.

## Validation

Local validation was intentionally limited to `rustfmt` and `git diff --check`; build, test, compatibility, and CodSpeed validation are delegated to CI.

### DIFF
--- a/.changeset/skip-unused-export-inventory.md
+++ b/.changeset/skip-unused-export-inventory.md
@@ -1,0 +1,7 @@
+---
+"@rsvelte/compiler": patch
+"@rsvelte/svelte2tsx": patch
+"@rsvelte/svelte-check": patch
+---
+
+Skip unused local export metadata collection during Svelte-to-TypeScript projection.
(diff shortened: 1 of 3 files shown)


baseballyama:
CodSpeed measured all 127 active benchmarks as unchanged against the exact 5956bc2 main baseline. The gate is semantically valid, but without a CI-measurable improvement it is not worth carrying. Closing rather than merging.

PR #2038, opened 2026-07-30
Reuse adjacent svelte2tsx source locations

baseballyama:
## Summary

- cache the exact original location at the end of each unedited MagicString chunk
- reuse it only for a forward lookup on the same original line
- retain the existing binary-search and UTF-16 prefix fallback for moved, backward, and cross-line chunks
- cover edited-chunk reuse and moved-chunk fallback explicitly

The restricted fast path preserves official MagicString semantics: generated-order chunks are not assumed to be globally monotonic.

## Performance

The latest local profile attributed up to 2.45% of the svelte2tsx workload to repeated original-line lookup and UTF-16 line-prefix scans. Every fixed CodSpeed svelte2tsx project generates a source map and exercises this path.

## Validation

Local validation was intentionally limited to `rustfmt` and `git diff --check`; build, unit, compatibility, source-map, and CodSpeed validation are delegated to CI.

### DIFF
--- a/.changeset/reuse-svelte2tsx-original-locations.md
+++ b/.changeset/reuse-svelte2tsx-original-locations.md
@@ -1,0 +1,7 @@
+---
+"@rsvelte/compiler": patch
+"@rsvelte/svelte2tsx": patch
+"@rsvelte/svelte-check": patch
+---
+
+Reuse adjacent source locations while generating Svelte-to-TypeScript source maps.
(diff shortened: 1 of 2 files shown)


baseballyama:
CodSpeed measured all 127 active benchmarks as unchanged against the exact 5956bc2 main baseline. The cursor is correctness-preserving, but the removed lookup work is below the CI-measurable threshold, so this is not worth carrying. Closing rather than merging.

Issue #2177, opened 2026-08-01
compiler (client): destructure `should_cache` keyed off the unvisited RHS — each-item RHS stays a sequence where upstream emits the `$$value` IIFE


baseballyama:
Found while fixing #2163 (PR #2176), pre-existing on main (reproduces identically with `$state` targets).

`should_cache` for a legacy destructuring assignment is decided from the *unvisited* RHS. When the RHS is an each-block item — visited form `$.get(item)` — upstream caches into a `$$value` IIFE (`(($$value) => …)($.get(item))`), but rsvelte keeps a plain sequence.

Reference: upstream `3-transform/shared/assignments.js` (`inserts.length > 0 || should_cache` on the **visited** expression), rsvelte `crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs`.

Issue #2163, opened 2026-08-01
compiler (client): legacy destructure assignment whose targets are only props is never recognised as reactive


baseballyama:
Found while fixing #2138 (PR #2158), pre-existing on main.

A legacy assignment destructure whose targets are *only* props (no `$state`/store targets) is not recognised as reactive at all — no prop-assignment calls are emitted, diverging from upstream's `visit_assignment_expression` (`3-transform/shared/assignments.js`) which routes each extracted path through the normal assignment lowering.

Issue #2162, opened 2026-08-01
compiler (client): single-expression parenthesized statement loses its parens — `({ a } = obj)` emits `$.set(a, obj.a);` without upstream's wrapping parens


baseballyama:
Found while fixing #2138 (PR #2158), pre-existing on main.

Upstream emits `($.set(a, obj.a));` for `({ a } = obj)` (the statement keeps its parentheses); rsvelte's redundant-paren swallow in `ast_state_transform.rs` drops them for a single-element sequence.

Issue #2139, opened 2026-08-01
compiler (client): legacy declaration with nested destructure pattern is not expanded — nested state leaf loses `$.mutable_source`


baseballyama:
Found while fixing #2078 (PR #2137).

`let { a: { b } } = obj` in legacy mode is not expanded at all, so a nested state leaf never gets `$.mutable_source` (and `$.tag` in dev). Upstream `_extract_paths` recurses into nested patterns; rsvelte's string-based expansion in `client/state_transforms.rs::transform_legacy_destructure_declarations` is single-level.

Making the expansion recursive is a materially bigger change than #2078 — likely wants the AST-based `extract_paths` port rather than extending the string pipeline (cf. epic #2089 direction).

PR #2192, opened 2026-08-02
fix(compiler): flatten nested destructuring assignments (#2186)

### LINKED ISSUE #2186: compiler (client): nested destructuring *assignment* emits nested `$$value` IIFE — upstream flattens to `$.set(b, $$value.a.b)`
Found while fixing #2139 (PR #2185), pre-existing on main.

`$: ({ a: { b } } = src)` emits a nested `(($$value) => …)($$value.a)` IIFE instead of upstream's flat `$.set(b, $$value.a.b)`. Lives in `generate_destructure_iife` (`crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs`) — the #2138/#2163 assignment family; the declaration path was made recursive in #2185 and can serve as the model.

baseballyama:
Fixes #2186

## What was wrong

`generate_destructure_iife` expanded a destructuring **assignment** one level deep. `({ a: { b } } = src)` produced `{ b } = $$value.a` as a statement inside the IIFE, which the very same transform then picked up and rewrote into a *second* IIFE:

```js
(($$value) => {
  (($$value) => { $.set(b, $$value.b); })($$value.a);
})($.get(src));
```

Upstream builds the lowering from `extract_paths`, whose `_extract_paths` (`utils/ast.js`) *recurses* through every nested `ObjectPattern` / `ArrayPattern` / `AssignmentPattern`, so every leaf carries its whole path and the result is flat:

```js
(($$value) => { $.set(b, $$value.a.b); })($.get(src));
```

Two more parts of `visit_assignment_expression` (`3-transform/shared/assignments.js`) were off in the same function, and are only decidable once the pattern is walked recursively:

- **The shape.** Upstream emits the IIFE only when `inserts.length > 0 || should_cache`, otherwise a plain `SequenceExpression`. rsvelte decided this per bracket type — an object pattern with any nested pattern *always* became an IIFE (with an always-cached `$$value`), an array pattern *always* became one.
- **`should_cache = value.type !== 'Identifier'`,** evaluated on the **visited** right-hand side. rsvelte's `force_cache_rhs` covered a prop / store RHS (which visit into calls) but not a state one, so `$: [a, b] = src` emitted `((src) => { var $$array = $.to_array(src, 2); … })($.get(src))` where upstream caches into `$$value`. That one is not nesting-specific — it reproduces with a flat pattern.

## The fix

`extract_destructure_paths` is a structural port of `_extract_paths`: same switch, same traversal order, same two output lists (`paths` + `inserts`). Each level passes the member access it built down as the next level's base expression. The emitter then mirrors upstream: **every** `$$array` helper first (so a nested helper is declared before the paths that read it), then **every** assignment, then the `return` when the destructure is part of a larger expression — or, with no helper and an uncached identifier RHS, a sequence expression instead of an IIFE.

It is the **same function** #2185 wrote for the declaration side, shared: the only difference between the two lowerings is how a path reads the `$$array` helper an array pattern contributes (`$$array[i]` for the assignment lowering's plain `var`, `$.get($$array)[i]` for the declaration lowering's `$.derived`), which is now an `ArrayHelperRead` parameter. `state_transforms.rs`'s copy is deleted; the two lowerings stay byte-identical by construction.

Two things fell out of the rewrite:

- A candidate `X = Y` that sits inside another candidate's **pattern** is that pattern's `AssignmentPattern` default, not an assignment — `({ a: { b } = { b: 3 } } = src)` used to lower the inner `{ b } = { b: 3 }` into an IIFE and plant a call in an LValue slot. The declaration form of this was already guarded (`is_inside_enclosing_pattern`); the assignment form now falls out of comparing candidate spans.
- A trailing comma is no longer counted as an element, so `[{ a }, ] = src` passes `1` to `$.to_array`, not `2` (upstream counts `param.elements`, and a trailing comma creates none) — the same alignment #2185 made on the declaration side.

## Verification

**Nested-shape matrix (46 hand-written components, official vs rsvelte, `client` / `client-dev` / `server`, byte-identical after normalization):** object-in-object (to 3 levels), array-in-object, object-in-array, array-in-array, `[...[a, b]]`, rest-of-rest, nested rest, leaf defaults, defaults *on* a nested pattern, doubly-nested fallbacks (`[[a = 3] = []]`), nested computed / quoted / numeric keys, elisions, trailing commas, `await` right-hand sides, non-standalone (`out = ({ x: { a } } = src)`), member-expression leaves, state / store / prop / non-reactive right-hand sides, legacy `$:` bodies, event handlers, runes and legacy mode, plus a nested destructure written **inline in the template** (a separate lowering, `expression_converter.rs`, already correct — checked for regressions) and nine **flat** control shapes covering the `should_cache` change.

- `cargo test --release -p rsvelte_core`: **2288 passed, 0 failed** (8 ignored)
- Destructure regression suites all still pass: #2078, #2138, #2139, #2162, #2163, #2177.
- New regression suite `crates/rsvelte_core/tests/nested_assignment_destructure_2186.rs` (12 tests).
- New pattern-corpus repro `pattern/issues/2186-nested-assignment-destructure.svelte` (README row added) — matches on all three targets.
- Corpus (`svelte` + `pattern`, 5162 entries × 3 targets): **no regressions**, no ratchet entry becomes fixed. (The "133 known failures now PASS" line is exactly the 133 baseline ids whose source submodules are not checked out in this environment — verified by intersecting the baseline against the collected sources — so no baseline shrink is warranted.)
- `cargo fmt` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean.

### Pre-existing divergences found while building the matrix (not touched here)

Both reproduce on `origin/main` (with this branch's output, not the old nested-IIFE one), and both live outside `generate_destructure_iife`:

1. **A member-expression leaf on a non-signal target** — `({ b: o.p } = src)` where `o` is a runes `$.proxy` emits `o.p = src.b`, while upstream keeps the original assignment verbatim because `build_assignment` returns `null` for every path (`changed === false`). That is the `changed` check, which the instance-script lowering does not have at all (the template one does). It reproduces with a **flat** pattern, so it is not nesting-specific.
2. **A runes prop-only script in non-dev** — `({ a: { b } } = src)` with `b` a prop and *no* state variable anywhere is left untransformed; the same file in dev, in legacy mode, or with any state variable present is now correct, and so is the flat `({ b } = src)` form in all of them. The instance script there is lowered by the retained-program AST path (`ast_state_transform`), which handles a flat destructuring assignment but not a nested one — a separate lowering from the one this PR fixes.

https://claude.ai/code/session_01RQSYoh4Wc353KgUnWHjazu


baseballyama:
Rebased onto `main` (past #2188 / #2189 / #2190). The only conflict was the pattern-corpus README table row, resolved by keeping both rows in numeric order (`2186` then `2187`). #2190 is entirely in the **server** oxc-AST lowering (`server/ast/{mod,script}.rs`, its own `extract_paths` + `state.array_counter`); this PR is the **client** text lowering and its `SCRIPT_ARRAY_COUNTER`, so the two `$$array` counters do not interact.

Re-verified after the rebase: destructure regression suites (#2078 / #2138 / #2139 / #2162 / #2163 / #2177 / #2186) + `--test server_destructure_divergences_2033_2034` + `--test ssr` all green, `--lib` 1358 passed, `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

⚠️ Heads-up, **not from this PR**: the corpus gate is currently red on `main` itself. `pattern/issues/2187-server-array-counter.svelte` (added by #2190 for the server fix) diverges on the **client** and **client-dev** targets — a destructured runes `$state` array pattern's update expression stays `a++` where upstream emits `$.update(a)`:

```
client/js line 14
  expected:     $.update(a);
  actual:       a++;
```

I reproduced it with a build of `origin/main` alone (this branch's four source files reverted), and the diff is byte-identical with and without this PR, so this branch neither causes nor worsens it. It needs its own fix (or baseline rows) on the #2187 side.

Issue #2061, opened 2026-07-31
svelte-check: two overlay-resolution divergences from official — missing TS5090 for non-relative `paths` without `baseUrl`, and companion-module named imports resolved leniently


baseballyama:
Two pre-existing divergences from official svelte-check observed while building #1942's fixture (PR pending; neither is caused by that fix):

1. **Missing TS5090.** When a tsconfig `paths` value is non-relative (e.g. `"src/lib/*"` without a leading `./`) and `baseUrl` is unset, official svelte-check reports TS5090 on the tsconfig; rsvelte-check reports nothing. Likely the synthesized overlay tsconfig normalizes the value before tsgo sees it.

2. **Companion-module named imports are too lenient.** For a component `Widget.svelte` with a companion `Widget.svelte.ts`, importing a named export of the companion through an alias (`import { helper } from '$lib/Widget.svelte'`) errors with TS2614 in official (the specifier resolves to the component), but resolves successfully in rsvelte-check. rsvelte is more permissive than official here, which can hide real errors.

Both reproduce on a nodenext + `type: module` + `paths` alias project; see `compatibility/check-fixtures/ts-aliased-rune-module-nodenext/` (the named-import case was deliberately left out of that fixture to keep it green).

baseballyama:
Third observation from #1942's review (same overlay area, pre-existing): a `.svelte.js`-only rune module (no `.ts` companion) in a project without `allowJs` yields TS2614 on named imports through the #1941 bridge — reproduces identically with a plain relative import, so it is a constraint of the bridge mechanism itself, not the alias override.

baseballyama:
PR #2114 (merged) fixed the bulk of this: TS5090 is now replayed with official positions/severity (incl. `${configDir}` substitution, also applied on the resolution side), companion named imports now fail with TS2614 like official, and `.svelte.js`-only modules surface official's TS2307/TS7016 instead of fake TS2614s (ambient `*.svelte` wildcard blanked + redirected, matching official byte-for-byte).

Two sub-cases remain open — both need official's `resolveModuleNames` hook, which has no equivalent for a stock compiler over an on-disk overlay:
1. relative-bundler companion named import (`import { helper } from './widget.svelte'` with a sibling `widget.svelte.ts`): TS finds the real companion in the importer's directory before `rootDirs`/`paths` can intervene → false negative vs official's TS2614.
2. nodenext `.js` rune module: no extension completion in ESM mode → TS2307 instead of official's TS7016.

[...]

PR #2212, opened 2026-08-04
fix(check): resolve a dependency's `svelte` type reference to a stub

### LINKED ISSUE #2211: svelte-check: a dependency's `/// <reference types="svelte" />` reloads svelte's declarations beside the blanked copy — false TS2322 on every snippet prop
Regression in 0.5.5 (`svelte-types.d.ts`, the `*.svelte` wildcard fix for #2061), still present in 0.5.9. `--tsgo` and `tsc` alike.

The blanked copy of `svelte/types/index.d.ts` is reached two ways: `paths` entries for every module it declares, and `blank_svelte_type_reference` for the `/// <reference types="svelte" />` svelte2tsx writes into our own shadows. Neither covers a **dependency** whose shipped `.d.ts` opens with that directive — `@sveltejs/kit/types/index.d.ts` and `@tanstack/svelte-table/dist/flex-render.svelte.d.ts` both do — because a type reference resolves through `typeRoots` / node resolution, not `paths`:

```
node_modules/svelte/types/index.d.ts
   Type library referenced via 'svelte' from file 'node_modules/@sveltejs/kit/types/index.d.ts'
   Type library referenced via 'svelte' from file 'node_modules/@tanstack/svelte-table/dist/flex-render.svelte.d.ts'
```

So the original loads beside the copy and every ambient svelte module is declared twice. `Snippet`'s brand is a `unique symbol` per declaration, so a snippet built against one is not assignable to the other:

```
ERROR src/routes/design/chip/+page.svelte:179:43 (ts): Type '() => ReturnType<import('svelte').Snippet>'
  is not assignable to type 'string | Snippet<[]> | undefined'.
```

130 of these in one real SvelteKit app (895 files), all false — official `svelte-check` reports 0. Nothing in the project can be changed to avoid it; the only workaround is pinning 0.5.4.

## Repro

```
src/box.svelte      let { title }: { title?: string | Snippet } = $props();
src/consumer.svelte {#snippet heading()}…{/snippet}  <Box title={heading} />
src/ambient.d.ts    /// <reference types="svelte" />
```

`ambient.d.ts` stands in for the dependency: any file in the program carrying that directive is enough. 0.5.4 → 0 errors, 0.5.5 through 0.5.9 → TS2322 on the prop.

## Notes

- Generated `.tsx` is byte-identical between 0.5.4 and 0.5.5 apart from the blanked directive, so this is purely the overlay's type environment.
- Worth knowing for a fix: setting `typeRoots` at all replaces TypeScript's default, and a name in `types` resolves through that option alone with no node-resolution fallback — so `"types": ["@sveltejs/kit"]`, which SvelteKit projects have, turns into TS2688 unless the `node_modules` dirs are among the roots.

Mokto:
Fixes #2211.

## The hole

Since the `*.svelte` wildcard fix (#2061), the blanked copy of `svelte/types/index.d.ts` is meant to be the program's only source of svelte's ambient modules. Two channels lead to it: `paths` entries for every module it declares, and `blank_svelte_type_reference` for the `/// <reference types="svelte" />` svelte2tsx writes into our own shadows.

A type reference inside a file we *don't* generate goes through neither — it resolves via `typeRoots` / node resolution, which `paths` cannot touch. Plenty of packages ship one:

```
node_modules/svelte/types/index.d.ts
   Type library referenced via 'svelte' from file 'node_modules/@sveltejs/kit/types/index.d.ts'
   Type library referenced via 'svelte' from file 'node_modules/@tanstack/svelte-table/dist/flex-render.svelte.d.ts'
```

So the original loaded beside the copy and every ambient svelte module was declared twice. `Snippet`'s brand is a `unique symbol` per declaration, so a snippet built against one stopped being assignable to the other:

```
ERROR src/routes/design/chip/+page.svelte:179:43 (ts): Type '() => ReturnType<import('svelte').Snippet>'
  is not assignable to type 'string | Snippet<[]> | undefined'.
```

130 of those in one real SvelteKit app (895 files) — official `svelte-check` reports 0 there. Regressed in 0.5.5, still in 0.5.9; `--tsgo` and `tsc` alike. Generated `.tsx` is byte-identical to 0.5.4 apart from the blanked directive, so it is purely the overlay's type environment.

## The fix

The reference resolves to an empty types package materialised at `<cache>/svelte-type-ref/svelte/`, placed first in `typeRoots`. Empty is enough: the blanked copy is in the overlay's `files` unconditionally, so svelte's declarations are in the program either way — now from exactly one file.

`typeRoots` replaces TypeScript's default wholesale, so the value absorbs two consequences:

- the ancestor `node_modules/@types` dirs are restated, or `@types/*` packages stop being auto-included;
- when the project pins `compilerOptions.types`, the `node_modules` dirs are appended as well. A name in `types` resolves through `typeRoots` with **no** node-resolution fallback (unlike a `reference types`), so `"types": ["@sveltejs/kit"]` — what SvelteKit projects have — otherwise becomes TS2688 and takes every real diagnostic in the project with it. I hit exactly that with a first cut of this patch: six `kit-*` scenarios silently lost their diagnostics. The widening is inert because a pinned `types` is what turns automatic inclusion off in the first place.

## Verification

- New fixture `compatibility/check-fixtures/svelte-type-reference` — a snippet passed to a `Snippet`-typed prop in a project that imports from `@sveltejs/kit`, so the reference arrives through the real channel. Confirmed it fails without the code change (`+ERROR src/consumer.svelte:12 2322`) and passes with it.
- `check-verify` full corpus: 0 divergences on both backends (`--rsvelte-backend tsgo` and the default `tsc`).
- `cargo test -p rsvelte_check`: 228 tests green, including two new unit tests (stub is first in `typeRoots` and declares nothing; a pinned `types` keeps the widened roots, an unpinned one does not).
- The real app above: 130 → 0 errors, matching official svelte-check.


baseballyama (approved):
Thank you so much!

Issue #1732, opened 2026-07-23
lint: catch undefined and unused variables in component scripts


mustafa0x:
I want to use `@rsvelte/lint` for Svelte files and Oxlint for everything else, but `@rsvelte/lint` currently does not catch undefined or unused variables in component scripts.

This component currently passes with no warnings:

```svelte
<script>
  import { writable } from 'svelte/store';

  const count = writable(0);
  const current = $count;
  const unused = 1;

  void missing;
</script>

<p>{current}</p>
```

I would expect `unused` and `missing` to be reported, while `count`, `$count`, and `current` remain valid because of the store subscription and markup reference.

Would it make sense for rsvelte-lint to own these checks for Svelte files? That would let projects use Oxlint for JavaScript and TypeScript without keeping ESLint or a custom Oxlint build just for component scripts.

baseballyama:
Current-tree check: the unused-variable half is now implemented as `svelte/no-unused-vars` (registry-enabled; its source explicitly cites #1732 and uses Phase-2 scope/template/store information). The issue remains open because no corresponding `no-undef` implementation exists yet; the original example still needs a diagnostic for `missing`. Future work should therefore scope to undefined-name parity rather than reimplementing unused binding analysis.

baseballyama:
Audit on current main: the remaining undefined-variable gap is ESLint core `no-undef`, not an eslint-plugin-svelte rule. The lint oracle deliberately enables only `svelte/*`, and the compatibility gate excludes the core no-undef comment-directive case.

OXC has the needed semantic primitive, but rsvelte_lint has no globals/environment configuration contract. A small identifier-list rule would falsely flag normal browser/project globals and needs Svelte-specific handling for runes, `$store`, TS type references, `typeof`, `arguments`, and suppressions. Keeping this open as a scoped `svelte/no-undef` feature rather than landing an unsafe partial rule. The unused-variable half is already implemented.

baseballyama:
Resolved by #2878 (`3b2e3b465`). `svelte/no-unused-vars` was already registered and covered; the new opt-in `svelte/no-undef` uses OXC semantic unresolved references and honors configured globals/env, runes, `$store` subscriptions, TypeScript type references, `typeof`, `arguments`, and inline suppressions. Full PR CI passed.

PR #2263, opened 2026-08-05
feat(lint): add svelte/no-unused-vars

baseballyama:
Refs #1732

## Scope

The issue asks for **two** checks — undefined variables and unused variables.
This PR ships **only `svelte/no-unused-vars`**, fully done.

`no-undef` is **left out** deliberately: it additionally needs a globals /
environment list (browser, node, es20xx, `svelte/*` ambient names, project-level
`globals` config), and without one it is extremely noisy on real components. A
rule with false positives is worse than no rule, so it stays for a follow-up.

## What the rule does

`svelte/no-unused-vars` flags top-level `<script>` bindings that are never read
anywhere in the component. ESLint core's `no-unused-vars` and oxlint both stop
at the `.svelte` boundary, so these went unchecked unless a project kept a
Svelte-aware ESLint around — exactly the gap the issue describes.

It is implemented as a `ScriptRule` on top of the compiler's **existing** Phase-2
scope tree (`ctx.scope_analysis()` → `ComponentAnalysis::root.bindings`), so
template reads, `$store` auto-subscriptions and `bind:` targets already resolve
back to their binding. No scope analysis is re-implemented, and the rule is pure
analysis (no `svelte_check` / validator dependency), so it works in the wasm
build too.

The issue's example now reports exactly one finding (`unused`), leaving `count`,
`$count` and `current` clean.

## Conservative by construction

A false positive on a real component is worse than a miss, so:

- only **top-level** module/instance-script bindings are judged — each items,
  snippet params, `let:` directives and function locals live in child scopes and
  are never reported;
- props (`export let`, `$props()` destructuring, `$$props`/`$$restProps`/
  `$$slots`), stores read as `$name`, reactive `$:` declarations, exported
  declarations (`export const`, `export function`, `export { x }`) and
  reassigned/mutated bindings all count as used;
- a name that occurs anywhere else in the source vetoes the report — this covers
  reads Phase 2 does not record as references (TypeScript type positions, JSDoc
  `@type`, generics).

## Tests

19 inline tests in `crates/rsvelte_lint/src/rules/no_unused_vars.rs`, written
false-positive-matrix first: every Svelte-specific bullet above has a case that
must NOT report (legacy props, `$props()` destructuring, template-only reads,
`$$props`/`$$restProps`/`$$slots`, snippet params, `{#each … as x}`, store `$`
auto-subscription, `bind:` targets, exported module declarations, `export { x }`,
`$:` declarations, TS type-only usage, function locals, `$state` read from the
template, `style:` directives, template-only component imports).

The rule is rsvelte-only (eslint-plugin-svelte has no such rule — upstream defers
to ESLint core, which cannot see template reads), so it is added to
`NO_FIXTURE_RULES` in the fixture-coverage gate with that reason, alongside the
existing `svelte/no-companion-module-shadow` precedent. Because the lint parity
corpus intersects rsvelte's rule set with the plugin's, the new rule is outside
that universe and cannot add entries to the shrink-only
`compatibility/lint-known-failures.json`.

### DIFF
--- a/.changeset/lint-no-unused-vars.md
+++ b/.changeset/lint-no-unused-vars.md
@@ -1,0 +1,15 @@
+---
+"@rsvelte/lint": patch
+---
+
+Add `svelte/no-unused-vars`, a Svelte-aware unused-variable rule for component
+scripts. ESLint core's `no-unused-vars` and oxlint both stop at the `.svelte`
+boundary, so top-level `<script>` bindings went unchecked unless a project kept
+a Svelte-aware ESLint around. The rule reads the compiler's Phase-2 scope tree,
+so template reads, `$store` auto-subscriptions and `bind:` targets all count as
+uses. It is deliberately conservative: only top-level module/instance-script
+declarations are judged, and props (`export let`, `$props()` destructuring,
+`$$props`/`$$restProps`/`$$slots`), exported declarations, reactive `$:`
+declarations, reassigned/mutated bindings, and names that occur anywhere else
+in the source (covering TypeScript type positions the scope tree does not
+record) are never reported.
--- a/apps/npm/oxlint-plugin/recommended.json
+++ b/apps/npm/oxlint-plugin/recommended.json
@@ -104,6 +104,7 @@
     "svelte/no-unknown-style-directive-property": "warn",
     "svelte/no-unnecessary-state-wrap": "warn",
     "svelte/no-unused-class-name": "warn",
+    "svelte/no-unused-vars": "warn",
     "svelte/no-useless-children-snippet": "warn",
     "svelte/no-useless-mustaches": "warn",
     "svelte/node_invalid_placement_ssr": "warn",
--- a/crates/rsvelte_lint/src/registry.rs
+++ b/crates/rsvelte_lint/src/registry.rs
@@ -219,5 +219,6 @@ pub fn all_script_rules() -> Vec<Box<dyn crate::script::ScriptRule>> {
             crate::rules::no_export_load_in_svelte_module_in_kit_pages::NoExportLoadInSvelteModuleInKitPages,
         ),
         Box::new(crate::rules::infinite_reactive_loop::InfiniteReactiveLoop),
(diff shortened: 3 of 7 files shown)


baseballyama:
### Real-world verification (step 3)

The lint parity corpus (`node scripts/compat-corpus/lint-verify.mjs`) cannot run in this
worktree — `compatibility/lint-manifest.json` is produced by `lint-collect.mjs`, which
found 0 entries because the corpus submodules are not initialised here (disk-constrained
environment). It is structurally impossible for this rule to add entries to the
shrink-only `compatibility/lint-known-failures.json` anyway: `ruleUniverse()` in
`scripts/compat-corpus/lint-universe.mjs` **intersects** rsvelte's `--list-rules` with
`Object.keys(eslint-plugin-svelte.rules)`, and both sides of the diff are filtered by
`universeSet.has(ruleId)`. `svelte/no-unused-vars` does not exist upstream, so it is
never in the universe and never in the diff. CI will run the gate for real.

Instead I ran the built CLI over real component trees:

- **shadcn-svelte** (1683 `.svelte` files): **0 findings** — no false positives.
- **skeleton + attractions + layercake**: 10 findings, all manually verified true
  positives (each reported name occurs exactly once in its file — its own declaration):
  unused `const seriesNames` in three layercake examples, two unused `@lucide/svelte`
  imports plus two unused `const`s in a skeleton `+page.svelte`, two unused class-string
  `let`s in a skeleton example, one unused `globals` import.
- The issue's exact reproduction reports exactly one finding:
  `'unused' is defined but never used.` — `count`, `$count` and `current` stay clean.

Local checks: `cargo clippy -p rsvelte_lint --all-targets --all-features -- -D warnings`
clean; `cargo test -p rsvelte_lint --lib` 273 passed (19 of them the new rule's);
`--test eslint_plugin_oracle`, `--test comment_directive`,
`--test experimental_and_dispatcher`, `--test browser_globals_ts_assertion`,
`--test valid_style_parse` all pass.

PR #2264, opened 2026-08-05
fix(compiler): anchor the client instance-script map chunk at its first token

baseballyama:
Refs #1781

The client instance-script chunk was anchored at `ScriptContent::start` — the byte
immediately after `<script>`, i.e. the newline ending that line. Every map segment derived
from that chunk therefore resolved to a column past the end of the `<script>` line, which
is exactly the class of segment that breaks a downstream consumer resolving a frame. The
chunk is now anchored at the script's first non-whitespace byte.

Measured on `d2e30d90` with `UPDATE_SOURCEMAP_RATCHET=1 cargo test -p rsvelte_core --test sourcemaps_gate -- --ignored sourcemap_gate_measure`:

| | before | after |
|---|---|---|
| client `out-of-range` segments | 37 | **19** (−49%) |
| client official segments reproduced | 0 / 488 | **9 / 488** |
| client `wrong` segments | 81 | **72** |
| `compatibility/sourcemap-known-failures.json` | 75 | **73** |

Ratchet movement is deletions and shrinks only — no entry grew, none added. Two entries
disappear (`out-of-range two-scripts/client`, `out-of-range typescript/client`); nine
`out-of-range` budgets shrink; four `map-parity` budgets shrink (`attached-sourcemap`
24→22, `script-after-comment` 18→16, `sourcemap-empty-source` 24→21, `two-scripts` 24→22).
The 9 reproduced client segments are the first non-zero client `exact` count this gate has
recorded.

Generated JS is unchanged — the offset only ever reaches the source map, never codegen.
`compatibility/known-failures.client-dev.json` is untouched.

While measuring I also established what does *not* fix the rest, and recorded it in the
paired `.md` plus a comment on #1781: the official map's segments are overwhelmingly
identifier/literal start-end pairs from esrap's `Context.write(content, node)`, which
`rsvelte_esrap` never ports — but adding those anchors makes the gate worse, because
`to_oxc.rs::parse_chunk` leaves comment-free chunks with **chunk-local 0-based spans** that
`print_with_map` then reads as absolute `.svelte` offsets. Chunk-local and real source
offsets share one number space with nothing to tell them apart, so per-node client
resolution needs the Wave-4 script AST-visitor migration, not a printer change.


baseballyama:
Heads up: this PR's changeset names `@rsvelte/check`, which is not a workspace package — the correct name is `@rsvelte/svelte-check` (see `apps/npm/svelte-check/package.json`).

The same typo landed on `main` via #2248 and broke the `Release` workflow at "Create or refresh Release Pull Request":

```
🦋  error Error: Found changeset css-render-site-ancestry for package @rsvelte/check which is not in the workspace
    at mapGetOrThrow (@changesets/assemble-release-plan)
```

No Release PR can be produced until it is fixed, so please correct the name before merging.

#2267 fixes `main` and adds a Changeset-gate check that rejects changesets naming packages outside the workspace. Once that lands, this PR's gate will fail on the bad name until it is corrected — so fixing it now avoids a red run later.

Issue #1229, opened 2026-06-26
compiler(CSR): leading JSDoc `@typedef` dropped + store-getter declaration order differs


baseballyama:
## Track
Compile (CSR) — awesome-svelte compat corpus (PR #1219)

## Summary
Two divergences in client codegen for store-heavy legacy components:

1. **Leading JSDoc `@typedef` / comment blocks are dropped** from the client output.
2. The generated store-getter consts (`const $x = () => $.store_get(...)`) are emitted in a **different order** than the official compiler.

## Repro
`layercake/src/_components/Column.svelte` (also `GroupLabels.html.svelte`, `QuadTree.html.svelte`, `AxisRadial.svelte`).

## Official vs rsvelte (client)
```diff
-  const $y = () => $.store_get(y, "$y", $$stores);
-  const $x = () => $.store_get(x, "$x", $$stores);
+  const $x = () => $.store_get(x, "$x", $$stores);
+  const $y = () => $.store_get(y, "$y", $$stores);
```
```diff
-    /**
-     * @typedef {Object} Props
-     * @property {string} [fill='#00e047'] - The shape's fill color.
-     * ...
-     */
   (dropped by rsvelte)
```

## Expected
Store-getter declaration order matches official iteration order, and leading JSDoc/typedef comments are preserved (these compile through the AST path that currently strips them).

PR #2269, opened 2026-08-05
fix(compiler): unwrap each-item reads in switch discriminants (#2254)

### LINKED ISSUE #2254: compiler (client, dev + prod): each-item used as a `switch` discriminant is not `$.get`-unwrapped → no case ever matches, silently

Inside an `{#each}` block, a reference to the each-item used as a **`switch` discriminant** is not
unwrapped with `$.get(...)`. It is emitted against the raw signal object, so the discriminant is
`undefined` and no `case` ever matches.

Every other reference to the same binding in the same function is unwrapped correctly — including
`if` and `while` conditions, and `$.get(item).value` inside the very same handler.

This reproduces with `dev: false` as well, so `vite build` succeeds and the app silently takes the
wrong branch at runtime. There is no error and no warning.

### Repro (minimal)

```svelte
<!-- B.svelte -->
<script>
	const items = $state([{ value: 'a' }]);
	const { onPick } = $props();
</script>

{#each items as item (item.value)}
	<button
		onclick={() => {
			if (item.value === 'a') onPick('if');
			while (item.value === 'never') break;
			switch (item.value) {
				case 'a':
					onPick('switch');
			}
		}}>x</button
	>
{/each}
```

```js
compile(src, { filename: 'B.svelte', generate: 'client', dev: false })
```

### rsvelte

```js
if ($.get(item).value === 'a') $$props.onPick('if');   // ok
while ($.get(item).value === 'never') break;           // ok
switch (item.value) {                                  // <-- raw signal
```

```js
if ($.get(item).value === 'a') $$props.onPick('if');
while ($.get(item).value === 'never') break;
switch ($.get(item).value) {
```

With `dev: true` the same thing happens (`if`/`while` become `$.strict_equals($.get(item).value, …)`,
the `switch` stays `switch (item.value)`), and the rest of the emitted module is byte-identical to
upstream — the `switch` line is the only difference.

### Impact

Found in a real app: a checkbox list rendered from `{#each}` where the change handler dispatches on
`switch (option.value)`. Checking a box did nothing at all, while unchecking worked — the uncheck
path is in the `else` branch two lines below and uses `$.get(option).value`, which compiles
correctly.

Of the issues I have hit so far this is the one I would most want fixed before trusting a build:
the other codegen bugs fail loudly at parse or import time, whereas this one produces a green
`vite build` whose output is wrong.

baseballyama:
Closes #2254

## Symptom

Inside an `{#each}` block, an each-item read used as a `switch` discriminant was emitted against the raw signal object instead of `$.get(item)`, so the discriminant was `undefined` and no `case` ever matched. Every sibling reference in the same function (`if`, `while`, …) was unwrapped correctly. This reproduced with `dev: false` too, so `vite build` succeeded and the app silently took the wrong branch — no error, no warning.

A generated differential shape matrix (7 binding kinds × 47 syntactic positions × 3 targets) run by the coordinating session showed the same hole in three more positions, plus two further symptoms of it:

| position | official | rsvelte (before) |
|---|---|---|
| `SwitchStatement.discriminant` | `switch ($.get(item).value) {` | `switch (item.value) {` |
| `SwitchCase.test` | `case $.get(item).value:` | `case item.value:` |
| class-expression field initializer | `f = $.get(item).value;` | `f = item.value;` |
| class-expression computed method key | `[$.get(item).value]() {}` | `[item.value]() {}` |

When the each-**index** binding was read *only* from one of those four positions, rsvelte judged it unreferenced and dropped it from the `$.each` callback signature (`($$anchor, item)` instead of `($$anchor, item, i)`). And when a `$store` was read only from a `case` test, no auto-subscription was created at all (`$s` undefined → `ReferenceError`, and on the server no `$$store_subs`).

## Root cause

Two independent causes, both in walks that enumerate child fields per node type.

**1 (four positions + index survival + store getter in the discriminant).** The client transform's recursive rewrite in `3_transform/client/visitors/shared/utils.rs`:

- `apply_transforms_to_statement_with_shadowed` had **no `JsStatement::Switch` arm**; the trailing `_ => stmt.clone()` cloned the whole switch verbatim, so neither the discriminant nor any case test nor any consequent statement was ever visited.
- `apply_transforms_to_expression_with_shadowed` listed **`JsExpr::Class(_)` among the terminal "expressions that don't need transformation"**, so a class expression's field initializers, computed keys and method bodies were cloned verbatim.

That single walk does three jobs at once — it rewrites reads to `$.get(...)`, it sets `state.each_index_used` from its `JsExpr::Identifier` arm, and it resolves store references — which is why the omission produced all three symptoms rather than only the missing unwrap. Upstream has no `SwitchStatement` visitor at all; the discriminant is transformed by the ordinary generic zimmerframe descent, so the fix is to make the walk descend, not to special-case `switch`.

**2 (store subscription in a `case` test) — genuinely separate.** The store-subscription pre-scan (`2_analyze/store_subscriptions.rs`) is a source-text scanner, and `is_dollar_ident_object_property_key` classified any `$store` followed by `:` as an object property key (`{ $s: 1 }`). That misfires on `case $s:`. It already carved out the ternary consequent (`cond ? $x : y`, #1229); a `case` test is now carved out the same way. Confirmed separate by bisection: after fix 1, `switch ($s)` matched but `case $s:` still did not.

## Fields audited

Descent was checked for every field of every node type reachable from an event-handler body in this walk. **Broken and fixed:** `SwitchStatement.discriminant`, `SwitchCase.test`, `SwitchCase.consequent`, `ClassExpression.superClass`, `PropertyDefinition.value`, `PropertyDefinition.key` (computed), `MethodDefinition.key` (computed), `MethodDefinition.value` (method body), `StaticBlock.body`.

**Verified already correct** (matrix found no divergence, and the arms exist): `if`/`while`/`do-while` tests, `for` init/test/update/body, `for-of`/`for-in` left/right/body, `return.argument`, `throw.argument`, array elements, object values, object computed keys, call arguments, member object/computed property, optional member, conditional test/consequent/alternate, logical/binary operands, unary/typeof argument, assignment LHS/RHS, sequence expressions, template-literal expressions, tagged-template tag + expressions, `await` argument, arrow/function bodies, default parameters, spread arguments, `try`/`catch`/`finally` bodies, destructuring defaults, labeled-statement body.

`JsExpr::Chain` and `JsExpr::Void` remain in the terminal list **deliberately**: they are only ever synthesized by `js_ast::builders` around already-transformed subtrees (never produced from user source by the expression converter), so recursing into them would transform twice. Probed empirically — `void item.value`, `item?.value` and `item?.value?.[item.value]` all `MATCH` because the converter lowers them to `Unary` / `Member`.

Scoping is mirrored too: all `case` clauses share one lexical scope (so a `let` in any consequent shadows the each item across the whole switch body), the class binding name is in scope inside its own body, and method parameters shadow.

## Verification

One-file differ against the official compiler (`scripts/compat-corpus/one.mjs`), all `MATCH` on **`client`, `client-dev` and `server`** for: the issue repro; `case item.value:`; `class { f = item.value }`; `class { [item.value]() {} }`; the each-index variants of all four; and the `case $s:` store case. Every one of these produced a diff before the change.

New regression test `crates/rsvelte_core/tests/switch_and_class_expr_transforms_2254.rs` — 15 tests covering the four positions × {each-item read, each-index survival, store subscription} in both dev and prod, plus scope anchors (case-local `let` shadowing, method-parameter shadowing, `{ $s: 1 }` still not a store ref) and "still correct" anchors (`if`/`while` tests, `throw`/`return` arguments).

`cargo test -p rsvelte_core --release`, `cargo fmt --all` and `cargo clippy --all-targets --all-features -- -D warnings` run locally.

### DIFF
--- a/.changeset/switch-and-class-expression-transforms.md
+++ b/.changeset/switch-and-class-expression-transforms.md
@@ -1,0 +1,20 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Client output now applies the read/store transforms inside `switch` statements
+and class expressions. A `{#each}` item read used as a `switch` discriminant
+(`switch (item.value)`), as a `case` test (`case item.value:`), as a class
+expression field initializer (`class { f = item.value }`) or as a class
+expression computed method key (`class { [item.value]() {} }`) was emitted
+against the raw signal instead of `$.get(item)`, so the value was `undefined`
+and no `case` ever matched — silently, in production builds as well as dev. The
+recursive transform walk had no `switch` arm (the catch-all cloned the statement
+verbatim) and listed class expressions among the terminal "nothing to transform"
+nodes. Because that same walk marks the each-index binding as used and registers
+store getters, the omission also dropped the `i` parameter from the `$.each`
+callback and skipped the `$.store_get` getter whenever the binding was read only
+from one of those positions, turning an undefined `$store` read into a
+`ReferenceError`. Separately, the store-subscription pre-scan classified any
+`$store` followed by `:` as an object property key, which misfired on
+`case $store:`; a `case` test is now recognised as a value expression.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/store_subscriptions.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/store_subscriptions.rs
@@ -616,7 +616,8 @@ fn dollar_param_body_range(
 /// A ternary consequent (`cond ? $x : y`) is also `$x` followed by `:`, but `$x`
 /// there is a real reference, not a property key. Such a `$x` is preceded
 /// (ignoring whitespace) by `?`, which never precedes a property key in a runtime
-/// object literal, so we exclude it (issue #1229).
+/// object literal, so we exclude it (issue #1229). A `switch` case test
+/// (`case $x:`) is excluded for the same reason.
 fn is_dollar_ident_object_property_key(
     chars: &[char],
     ident_start: usize,
@@ -648,6 +649,17 @@ fn is_dollar_ident_object_property_key(
         }
         if k >= 0 && chars[k as usize] == '?' {
             return false;
+        }
+        // Exclude a `switch` case test: `case $x:` is a value expression, not a
+        // property key.
+        if k >= 3 {
+            let pos = k as usize;
+            if chars[pos - 3..=pos].iter().collect::<String>() == "case" {
+                let before = if pos >= 4 { chars[pos - 4] } else { ' ' };
+                if !before.is_alphanumeric() && before != '_' && before != '$' {
+                    return false;
+                }
+            }
         }
         return true;
     }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
@@ -1415,7 +1415,39 @@ pub fn apply_transforms_to_expression_with_shadowed(
             })
         }
 
-        // Expressions that don't need transformation
+        JsExpr::Class(class) => {
+            // The class binding name is in scope inside its own body.
+            let mut class_scope = local_scope.clone();
+            if let Some(id) = &class.id {
+                class_scope.add_shadowed(id.to_string());
+            }
+            JsExpr::Class(JsClassExpression {
+                id: class.id.clone(),
+                super_class: class.super_class.map(|sc| {
+                    context
+                        .arena
+                        .alloc_expr(apply_transforms_to_expression_with_shadowed(
+                            context.arena.get_expr(sc),
+                            context,
+                            &class_scope,
+                        ))
+                }),
+                body: JsClassBody {
+                    body: class
+                        .body
+                        .body
+                        .iter()
+                        .map(|member| {
+                            apply_transforms_to_class_member(member, context, &class_scope)
+                        })
+                        .collect(),
+                },
+            })
+        }
+
+        // Expressions that don't need transformation. `Chain` and `Void` are only
+        // ever synthesized by the builders around already-transformed subtrees,
+        // never produced from user source, so recursing would transform twice.
         JsExpr::Literal(_)
         | JsExpr::This
         | JsExpr::Super
@@ -1423,7 +1455,6 @@ pub fn apply_transforms_to_expression_with_shadowed(
         | JsExpr::ImportExpression { .. }
         | JsExpr::Raw(_)
         | JsExpr::OpaqueIdentifier(_)
-        | JsExpr::Class(_)
         | JsExpr::Chain(_)
         | JsExpr::Void(_) => expr.clone(),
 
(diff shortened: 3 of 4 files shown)


baseballyama:
Independent verification from the coordinating session — I rebuilt this branch and re-ran the same generated shape matrix that produced the widened scope (7 binding kinds × 47 syntactic positions × 3 targets = 987 comparisons against `svelte/compiler` 5.56.8).

**Before this branch: 21 divergences. After: 2.**

```
generated cases:      329 (7 bindings x 47 positions)
compiled comparisons: 987
both compilers reject: 0
DIVERGENCES:          2

### label.body  (2 divergences)
    derived-local__label.body.svelte [server] js-mismatch
      official: if (doubled) break lbl;
      rsvelte : if (doubled()) break lbl;
    bindings affected: derived-local, store-auto-sub
```

Every `switch` / class-expression / each-index / store-in-`case` divergence is gone across `client`, `client-dev` **and** `server`.

The 2 remaining are **not** regressions and are **not** in scope here — they are an upstream Svelte bug where rsvelte's output is the correct one. `submodules/svelte/.../3-transform/server/visitors/LabeledStatement.js` returns early without calling `context.next()` for non-`$` labels, so zimmerframe never descends into the labeled subtree; the client visitor at the same guard *does* call `context.next()`. Since `$.derived()` returns a function in `svelte/internal/server`, upstream's `if (doubled)` is always truthy. Tracking separately.

The split into two root causes (the shared-walk descent vs. the `case $s:` store pre-scan), and the bisection evidence for it, matches what I measured independently. Nice catch on `is_dollar_ident_object_property_key` — that one would not have been visible from the matrix alone without the first fix landing.

MathiasWP:
> Independent verification from the coordinating session — I rebuilt this branch and re-ran the same generated shape matrix that produced the widened scope (7 binding kinds × 47 syntactic positions × 3 targets = 987 comparisons against `svelte/compiler` 5.56.8).
> 
> **Before this branch: 21 divergences. After: 2.**
> 
> ```
> generated cases:      329 (7 bindings x 47 positions)
> compiled comparisons: 987
> both compilers reject: 0
> DIVERGENCES:          2
> 
> ### label.body  (2 divergences)
>     derived-local__label.body.svelte [server] js-mismatch
>       official: if (doubled) break lbl;
>       rsvelte : if (doubled()) break lbl;
>     bindings affected: derived-local, store-auto-sub
> ```
> 
> Every `switch` / class-expression / each-index / store-in-`case` divergence is gone across `client`, `client-dev` **and** `server`.
> 
> The 2 remaining are **not** regressions and are **not** in scope here — they are an upstream Svelte bug where rsvelte's output is the correct one. `submodules/svelte/.../3-transform/server/visitors/LabeledStatement.js` returns early without calling `context.next()` for non-`$` labels, so zimmerframe never descends into the labeled subtree; the client visitor at the same guard _does_ call `context.next()`. Since `$.derived()` returns a function in `svelte/internal/server`, upstream's `if (doubled)` is always truthy. Tracking separately.
> 
> The split into two root causes (the shared-walk descent vs. the `case $s:` store pre-scan), and the bisection evidence for it, matches what I measured independently. Nice catch on `is_dollar_ident_object_property_key` — that one would not have been visible from the matrix alone without the first fix landing.

Should the upstream Svelte bug be filed so they can fix it as well?

baseballyama:
@MathiasWP Submitted!

https://github.com/sveltejs/svelte/issues/18615

PR #2271, opened 2026-08-05
fix(compiler): honour svelte-ignore comments inside object and array literals (#2256)

### LINKED ISSUE #2256: compiler: `// svelte-ignore` is not honored when it precedes an object-literal property

A `// svelte-ignore <code>` comment is honored when it precedes a statement, but not when it
precedes a **property in an object literal**. Upstream honors both positions.

### Repro (minimal)

```svelte
<!-- D.svelte -->
<script>
	const { dims } = $props();

	// svelte-ignore state_referenced_locally
	const statementLevel = dims.length > 0;

	const opts = $state([
		{
			// svelte-ignore state_referenced_locally
			propertyLevel: dims.length > 0
		}
	]);
</script>

{statementLevel}{opts.length}
```

```js
compile(src, { filename: 'D.svelte', generate: 'client', dev: true })
```

### rsvelte

```
state_referenced_locally at 10:18   // the `propertyLevel` line
```

The statement-level ignore on line 5 is respected; the property-level one on line 9 is not.

No warnings.

### Impact

Warning noise only — the emitted code is fine. But it is noise you cannot suppress: the whole point
of the comment is that the initial-value capture is deliberate, and in this shape (a `$state` array
of option descriptors, each seeded from a prop) there is no statement to attach the ignore to. One
component in my app produces six of these on every compile.

baseballyama:
Closes #2256

## Root cause

Phase 1 attached JS comments to AST nodes through `distribute_comments_to_node`, a
hand-maintained allowlist of *statement-body* fields (`BlockStatement`, `SwitchStatement`,
`TryStatement`, `VariableDeclaration`, `ClassBody`, …). Anything that is not a statement
body — object-literal properties, array elements, call arguments — had no entry, so a
`// svelte-ignore` in those positions never became a leading comment of the node that
followed it, never reached `Program.ignore_comment_map`, and suppressed nothing.

Upstream's rule (`add_comments` in `phases/1-parse/acorn.js`) has no node-type table at
all: the walk is generic and positional. A comment goes to the first node in pre-order
that starts after it, and Phase 2's ignore stack then covers that node's whole subtree —
unless an earlier node claims the comment as a *trailing* comment first.

## Fix

`distribute_comments_to_node` / `harvest_ignore_comments` are replaced by
`CommentAttacher`, a direct port of upstream `add_comments`, reduced to the single
consumer rsvelte has for JS comment attachment (`svelte-ignore` suppression): it walks
every node generically and records the ignore texts keyed by the owning node's absolute
start, which is exactly what Phase 2 already looks up.

This is the general rule, not a widened allowlist. Both of upstream's trailing-comment
rules are ported with it, because without them the generic rule *over*-suppresses:

* `is_last_in_body` — a comment after the last child of a `Program`/`BlockStatement`/
  `ArrayExpression`/`ObjectExpression` belongs to that child, not to whatever follows the
  block.
* the `/^[,) \t]*$/` separator rule — a comment separated from the previous node by only
  `,`/`)`/spaces/tabs is that node's trailing comment, so `f(1, // svelte-ignore …` does
  **not** cover the next argument.

One rsvelte-specific gap is closed alongside: `convert_statement_for_program` returns
`None` for type-only statements (`type X = …`, `interface`, `import type`), which have no
typed node, so comments inside them used to escape into the next statement. Upstream binds
them to TS nodes whose subtree Phase 2 never visits, so they are now discarded with the
statement.

## Verification

Every shape below was compiled with both the official compiler and rsvelte
(`generate: 'client', dev: true`) and the emitted warning sets compared. **29/29 match.**

| # | Shape | official | rsvelte |
|---|---|---|---|
| 01 | issue repro: statement-level + object-property ignore | none | none |
| 02 | ignore before an array element | none | none |
| 03 | ignore before the first call argument | none | none |
| 04 | ignore before a later call argument (newline before it) | none | none |
| 05 | `/* svelte-ignore … */` block comment in property position | none | none |
| 06 | `// svelte-ignore a, b` comma-separated codes | none | none |
| 07 | ignore before a class field and a class method | none | none |
| 08 | ignore on the enclosing `$state([...])` declaration | none | none |
| 09 | **negative** — must not leak to the next statement | warns | warns |
| 10 | **negative** — must not leak to the sibling property | warns | warns |
| 11 | **negative** — comment after the last statement in a block | warns | warns |
| 12 | ignore before a property of a plain object literal | none | none |
| 13 | **negative** — an unrelated code must not suppress | warns | warns |
| 14 | ignore inside a nested function body | none | none |
| 15 | ignore in an arrow function's returned object literal | none | none |
| 16 | ignore on an array element nested in an object | none | none |
| 17 | **negative** — same-line comment after `,` in a call | warns | warns |
| 18 | **negative** — same-line comment after `,` in an array | warns | warns |
| 19 | **negative** — comment after the last array element | warns | warns |
| 20 | ignore before an object method, covering its body | none | none |
| 21 | dangling comment at the end of a call's arguments | none | none |
| 22 | **negative** — TS: comment inside a `type` alias body | warns | warns |
| 23 | TS: object-property ignore (the repro, `lang="ts"`) | none | none |
| 24 | `<script module>` + instance script, both levels | none | none |
| 25 | **negative** — TS: comment inside an inline type literal | warns | warns |
| 26 | **negative** — TS: comment inside an `interface` body | warns | warns |
| 27 | **negative** — TS: leading comment on a `type` alias | warns | warns |
| 28 | **negative** — TS: leading comment on `import type` | warns | warns |
| 29 | TS: comment inside a generic call's type arguments | none | none |

Additionally, every `.svelte` file under `submodules/svelte/packages/svelte/tests` that
contains a `svelte-ignore` comment (33 files) was compiled with both compilers: **0**
warning-code differences.

New regression suite: `crates/rsvelte_core/tests/svelte_ignore_comment_attachment_2256.rs`
— the positive shapes plus six must-still-warn negatives (next statement, sibling
property, unrelated code, both same-line-after-comma forms, last-in-block and
last-array-element).

### DIFF
--- a/.changeset/svelte-ignore-comment-attachment.md
+++ b/.changeset/svelte-ignore-comment-attachment.md
@@ -1,0 +1,16 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): honour `svelte-ignore` comments inside object and array literals.
+Phase 1 distributed script comments through a hand-maintained allowlist of
+statement-body fields (`BlockStatement`, `SwitchStatement`, `VariableDeclaration`,
+`ClassBody`, …), so a `// svelte-ignore` in front of an object-literal property,
+an array element or a call argument bound to nothing and suppressed nothing —
+producing warning noise the author had no way to silence. Comment attachment now
+mirrors upstream `add_comments` (`phases/1-parse/acorn.js`): the walk is generic
+and positional, a comment binds to the first node in pre-order that starts after
+it, and that node's whole subtree inherits the ignore. Upstream's trailing-comment
+rules are ported with it, so a comment after the last element of a block or literal,
+or one separated from the previous node by only `,`/`)`/spaces, still belongs to
+the node before it and does not over-suppress the node after it.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -6860,63 +6860,47 @@ fn convert_parsed_program<'ast>(
         // output.
         let mut ignore_comment_map: Vec<(u32, Vec<CompactString>)> = Vec::new();
         let body: Vec<JsNode> = if has_comments && has_ignore {
-            let mut comment_idx = 0;
             let mut body_nodes: Vec<JsNode> = Vec::with_capacity(program.body.len());
 
-            // Pre-compute comment entries (absolute positions + Value) once, used for
-            // distributing comments onto nested statement bodies.
             let comment_entries: Vec<CommentEntry> = all_comments
                 .iter()
                 .map(|comment| {
-                    let comment_start = offset + comment.span.start as usize;
-                    let comment_end = offset + comment.span.end as usize;
+                    let raw_text = content
+                        .get(comment.span.start as usize..comment.span.end as usize)
+                        .unwrap_or("");
                     CommentEntry {
-                        start: comment_start as u32,
-                        end: comment_end as u32,
-                        value: build_comment_value(comment, content, offset),
+                        start: offset as u32 + comment.span.start,
+                        text: CompactString::from(extract_comment_value(raw_text, comment.kind)),
                     }
                 })
                 .collect();
 
-            for stmt in program.body.iter() {
+            let mut attacher = CommentAttacher {
+                comments: &comment_entries,
+                next: 0,
+                content,
+                offset: offset as u32,
+                map: &mut ignore_comment_map,
+            };
+            let last_index = program.body.len().saturating_sub(1);
+
+            for (index, stmt) in program.body.iter().enumerate() {
                 if let Some(stmt_node) =
                     convert_statement_for_program(arena, stmt, offset, line_offsets)
                 {
-                    let stmt_start = stmt.span().start;
-
-                    // Collect comments that appear before this statement (its own leading
-                    // comments).
-                    let mut stmt_leading = Vec::new();
-                    while comment_idx < all_comments.len()
-                        && all_comments[comment_idx].span.end <= stmt_start
-                    {
-                        let comment = all_comments[comment_idx];
-                        stmt_leading.push(build_comment_value(comment, content, offset));
-                        comment_idx += 1;
-                    }
-
-                    // Skip comments that are inside the statement
-                    while comment_idx < all_comments.len()
-                        && all_comments[comment_idx].span.start < stmt.span().end
-                    {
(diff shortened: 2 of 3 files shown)


baseballyama:
Independent verification from the coordinating session. Worth stating up front: **no CI gate on this repo can verify this PR** — `scripts/compat-corpus/compile.mjs` keeps only `result.js` / `result.css` and discards `result.warnings`, so the corpus scores warning divergences as `MATCH`. A green board here says the fix broke nothing; it says nothing about whether the fix works. So I measured it directly.

**1. The reported repro** (`D.svelte` from #2256), rebuilt from this branch:

```
--- official warnings ---   (none)
--- rsvelte warnings ---    (none)
```
(on `main` rsvelte emitted `state_referenced_locally @ 10:18`.)

**2. Corpus-wide warning parity, all 14,026 entries**, both compilers, `generate: 'client', dev: true`, 0 worker aborts:

| metric | main | this branch |
|---|---|---|
| entries with any warning-payload divergence | 699 | **696** |
| entries with a true warning **code-set** divergence | 73 | **70** |
| rsvelte over-warns (code level) | 66 | **63** |
| rsvelte under-warns (code level) | 6 | **6** |

Three entries fixed, none introduced, and the under-warn count did not move — which is the number that matters, since **over-suppression** was the dangerous failure mode for this change.

**3. Over-suppression audit.** I checked all 6 remaining under-warn entries individually. Five contain zero `svelte-ignore` comments, so this PR cannot be their cause. The sixth, `melt-ui/docs/src/components/motion.svelte`, contains two `svelte-ignore` comments — but they are **HTML template comments** (`<!-- svelte-ignore a11y_mouse_events_have_key_events -->`, line 170) for a11y codes, while the missing warning is `state_referenced_locally@107:57` in the script. Unrelated, and pre-existing.

**4. Positive control.** `layerchart/docs/src/routes/+layout.svelte` is a real corpus entry carrying exactly the #2256 shape (`// svelte-ignore state_referenced_locally` before `themes: data.themes` inside `$state({...})`). On `main` rsvelte over-warned there while the corpus reported `MATCH`; on this branch it no longer diverges. That file is the reason I am confident this fixes the class and not just the repro.

Replacing the allowlist with a port of upstream `add_comments` — including both trailing-comment rules, which are what stop the generic rule from over-suppressing — is the right call. Merging.

The remaining 70 code-set divergences and the 592 entries where rsvelte omits `start`/`end` on a warning are a separate, pre-existing gap; I'm writing that up as a proposal for a warning-parity corpus gate so this class stops being unmeasurable.

PR #2279, opened 2026-08-05
fix(compiler): unwrap each-item signals in switch discriminants

### LINKED ISSUE #2254: compiler (client, dev + prod): each-item used as a `switch` discriminant is not `$.get`-unwrapped → no case ever matches, silently

baseballyama:
Closes #2254

### DIFF
--- a/.changeset/switch-discriminant-each-item.md
+++ b/.changeset/switch-discriminant-each-item.md
@@ -1,0 +1,8 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Unwrap each-item signals read inside a `switch` statement. The client identifier
+transform had no `SwitchStatement` arm, so the discriminant, every `case` test
+and every consequent statement were emitted against the raw signal object — no
+`case` ever matched, in both dev and prod builds, with no error or warning.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
@@ -1897,6 +1897,43 @@ fn apply_transforms_to_statement_with_shadowed(
             })
         }
 
+        JsStatement::Switch(switch_stmt) => {
+            // Upstream scopes the whole `SwitchStatement` as one block, so a
+            // declaration in any case shadows the discriminant and every case.
+            let mut switch_scope = local_scope.clone();
+            for case in &switch_stmt.cases {
+                register_block_local_vars(&case.consequent, &context.arena, &mut switch_scope);
+            }
+            let transform_in_switch = |e: &JsExpr| {
+                apply_transforms_to_expression_with_shadowed(e, context, &switch_scope)
+            };
+            let discriminant = context.arena.alloc_expr(transform_in_switch(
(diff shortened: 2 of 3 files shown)


baseballyama:
Superseded by #2269, which landed the same fix as commit `6162f60d` while this was in flight — hence the `unreachable pattern` clippy failure here (two `JsStatement::Switch` arms).

#2269 covers the same ground: one shared block scope registered across all case consequents, then the discriminant transformed in the enclosing scope and every case test plus consequent transformed inside it. So the case-test and consequent gaps this PR also found were already closed, not just the discriminant.

The independent diagnosis matched exactly — `apply_transforms_to_statement_with_shadowed` hand-rolls the statement walk and had no `Switch` arm, so switches hit the catch-all `_ => stmt.clone()`; upstream cannot have the bug because it uses a generic `walk` plus a scope-consulting `Identifier` visitor with `SwitchStatement: create_block_scope`.

The sibling-position audit from this work is worth keeping on record: `throw`, `for...of` RHS, tagged-template tags and computed member keys are all handled correctly, and `with` is unreachable (no `JsStatement::With` variant and no `convert_statement` case, so it is dropped at conversion). After #2269, `Switch` was the only real gap in that statement match.

Issue #907, opened 2026-06-08
NAPI compiler not thread-safe under Vite 8 + Rolldown — non-deterministic spurious [PARSE_ERROR] on valid rune modules


MathiasWP:
Following up on #815 (thanks — `0.4.0` loads fine on `vite@8.0.15`!). With the override in place, `vite build` fails in the SSR environment with `[PARSE_ERROR]` on valid third-party `.svelte.js` rune modules (from `runed`). The failing **file set changes between runs** and error messages vary ("Cannot assign to this expression", "Expected \`,\` or \`)\` but found \`;\`/Identifier", "Unexpected token").

Each affected file compiles cleanly in **isolation**, in **200× sequential** `compileModule` loops, and across **8 Node `worker_threads`** — only the real Vite 8 + Rolldown build reproduces it. This looks like shared-state / thread-safety in `@rsvelte/vite-plugin-svelte-native@0.2.3`'s `compileModule` (cf. the previously-closed #128), now triggered because Rolldown executes transform hooks across parallel threads.

### Evidence of non-determinism (two consecutive builds, same commit)
- Run 1 failing modules: `debounced`, `element-rect`, `persisted-state`
- Run 2 failing modules: `animation-frames`, `debounced`, `persisted-state`

### Stack
```
[plugin vite-plugin-svelte:compile-module] .../runed/dist/utilities/persisted-state/persisted-state.svelte.js
    at Module.compileModule (.../@rsvelte/vite-plugin-svelte-native@0.2.3/index.cjs:69:32)
    at TransformPluginContextImpl.handler (.../@rsvelte/vite-plugin-svelte@0.4.0/src/plugins/compile-module.js:85:43)
```

### Environment
- SvelteKit 2.61.1, svelte 5.56.0
- vite 8.0.15 (Rolldown-powered)
- @rsvelte/vite-plugin-svelte 0.4.0, @rsvelte/vite-plugin-svelte-native 0.2.3
- Node 25 (large real-world SvelteKit app, heavy on `.svelte.ts` rune modules)

Happy to test a thread-safe build against our codebase — it's a good stress case for concurrent `compileModule`.

baseballyama:
@MathiasWP Hmm, I already tested concurrency, but it seems there are still some issues. Maybe I missed something. I’ll investigate it.

Anyway, thank you! I’m really happy that you opened these issues. Issues like yours are very helpful!

MathiasWP:
[...]

PR #2283, opened 2026-08-05
fix(compiler): read class-transform terminators lexically, not by byte hunt (#2253)

### LINKED ISSUE #2253: compiler: `)` closing an injected `$.set(...)` is emitted at a comment offset when a `#private` `$state` field is assigned in a constructor

When a `#private` `$state` field is assigned **inside a constructor** and the right-hand side is a
multi-line object/array literal containing a `//` line comment, the `)` that closes the injected
`$.set(...)` call is emitted at the comment's start offset instead of at the end of the literal.
The output is not parseable.

The same assignment in an ordinary method compiles correctly, so the repro below contains both.

### Repro (minimal)

```js
// A.svelte.js
export class R {
	#x = $state.raw({});

	constructor(s) {
		this.#x = {
			a: s,
			// c
			b: s
		};
	}

	m(s) {
		this.#x = {
			a: s,
			// c
			b: s
		};
	}
}
```

```js
compileModule(src, { filename: 'A.svelte.js', generate: 'client', dev: true })
```

### rsvelte

```js
export class R {
		#x = $.tag($.state({}), 'R.#x');

		constructor(s) {
			$.set(this.#x, {
a: s,)// c
b: s
};
		}
	m(s) {
		$.set(this.#x, {          // <-- method: correct
			a: s,
			// c
			b: s
		});
	}
	}
```

Note the constructor body is also re-indented to column 0, while the method body preserves its
original formatting — the two seem to go through different printing paths, and only the
constructor one misplaces the paren.

```js
export class R {
	#x = $.tag($.state({}), 'R.#x');

	constructor(s) {
		$.set(this.#x, {
			a: s,
			// c
			b: s
		});
	}
	...
}
```

### Narrowing

Only the combination breaks it — each of these compiles correctly:

- the same assignment in a method rather than the constructor (see above)
- the same constructor assignment with the comment removed
- a `/* block */` comment instead of a `//` line comment
- an assignment inside an `if` block in a method

An array literal breaks the same way: `$.set(this.#x, [\ns,)// c\ns\n];`

### Impact

Hit in a real app on a `#private` `$state.raw` baseline field initialised from a constructor
argument, where the object literal carried an explanatory comment. Vite/Rolldown fails on the
emitted module:

```
RolldownError: Parse failure: Parse failed with 1 error: Unexpected token
```

so the module cannot be imported at all.

baseballyama:
Closes #2253

## Root cause

The class transforms rewrite private-`$state` assignments as **text**, and
several of their scanners locate brackets and statement terminators by hunting
for a byte. A `}`, `)` or `;` that appears inside a comment or a literal is
text, not code — reading it as code is the whole bug family. Four distinct
scanners had it:

**1. `split_rhs_at_top_level_semi` — the reported bug.** Its `//` arm had no
bracket-depth guard, unlike the neighbouring `;` arm:

```rust
b'/' if bytes.get(i + 1) == Some(&b'/') => break, // line comment → tail
```

A line comment nested inside a multi-line literal was therefore mistaken for
the statement's trailing comment: the value was truncated at the comment's
start offset and the `)` closing the injected `$.set(` was emitted there.

```js
// input                    // before
this.#x = {                 $.set(this.#x, {
	a: s,                   a: s,)// c
	// c                    b: s
	b: s                    };
};
```

The result does not parse — Vite/Rolldown aborts with `Parse failure:
Unexpected token`, so the module cannot be imported at all.

**2. Four `rest.find(';')` scans** in the method / non-`this` paths
(`class_transforms.rs`, the compound-operator and direct-assignment loops,
twice each) took the first `;` *anywhere* — inside a comment, inside a string,
inside a nested function body — and ignored bracket depth entirely. This is
why `// } ) ; c` in a **method** produced `// } )); c` with the real `)` gone,
a shape the constructor fix alone did not touch.

**3. `transform_class_fields_server` — silent content loss, not a parse
error.** It found the class body's closing brace with a bare char loop
(`'{' => depth += 1, '}' => depth -= 1`), no string or comment awareness at
all. A `}` written inside a comment — `// returns { ok, err }`, `// move this
into the if () {} block` — closed the class body early and **every member
after it was deleted from the output**, with no error and no warning. This is
the #2254 failure mode (green build, wrong program) rather than the #2253 one.

**5. `static { … }` blocks (server).** A class member was recognised as a
block only when its line contained a `(`, so a static initialization block fell
through to the *field* emitter: its body was written out line by line with a
`;` appended to each — including the comment line, which then swallowed the
property after it. Found by the coordinator's matrix, which has a
`static-block` host mine did not.

**4. Member-scan brace counters.** The client `parse_section_members` counted
`{`/`}` with `for ch in trimmed.chars()`, and the server's
`update_member_brace_depth` did the same. A `// … } …` line decremented the
depth mid-method, so the scan **split a method in two**; the first half then
had no trailing `;` at all, which is why the `)` landed at the *end* of the
comment line rather than at its `;`.

## Fix

All four now share `phases/3_transform/shared/js_scan.rs::skip_opaque`, which
steps over strings, template literals, **regex literals** and both comment
forms in one place. Regex mattered on its own: `{ a: /[});]+/g }` was
unparseable on every target with no comment involved, so it would have survived
any comment-specific patch.

Two hardenings beyond skipping: `rhs_value_end` also ends the value at a
closing bracket it never opened (so a `;`-less RHS such as
`m() { this.#x = 1 }` no longer swallows the method's `}`), and bracket depth
is now respected at all four assignment sites.

Also fixed, in the same re-printer: synthesized class members were emitted at a
hard-coded one-level indentation, which assumed the class sat one tab deep
inside a component `<script>`. A top-level `class` in a `.svelte.js` module
came out with its fields, accessors, constructor and closing brace one tab too
deep. They now follow the class's own source indentation, and a grouped
multi-line constructor statement keeps the relative indentation of its
continuation lines instead of being flattened to column 0.

## Verification

**Generated matrix** — 264 class shapes (6 hosts × 8 RHS shapes × 5 comment
slots × 3 operators, keeping only shapes whose *input* parses) × 3 targets =
792 comparisons. Invariant: does rsvelte's emitted module parse (acorn,
`sourceType: 'module'`)?

| revision | unparseable / 792 |
|---|---|
| `origin/main` | **354** |
| + comment-depth guard (cause 1) | 354 |
| + four `find(';')` sites (cause 2) | **192** |
| + server class-body brace match (cause 3) | **66** |
| + member-scan brace counters (cause 4) | **0** |

(Cause 1 alone does not move the total because the shapes it fixes are also
broken by causes 2–4; each row is cumulative.)

That matrix had no `static { … }` host. Adding it — both the multi-line form
and the one-line `static { const s = 1;` form — plus the nested-function-body
RHS (`{ f: () => { g(); }, b: s }`) gives **416 cases / 1248 comparisons, still
0 unparseable**. The static-block cell needed cause 5; it was failing on all
three targets before this PR, and causes 1–4 fixed only its two client targets.

**Normalized output parity** against the official compiler over the same 264
cases (oxfmt + blank-line stripping, the corpus gate's normalization):

| revision | differing / 792 |
|---|---|
| `origin/main` | 394 |
| this PR | **100** |

(Re-run after the static-block fix: unchanged at 100, and still no new
differences relative to either the baseline or the preceding commit.)

The remaining 100 are a **strict subset** of the baseline's 394 — `comm -13`
between the two complete lists is empty, so the PR introduces no new
divergence. They are pre-existing shapes unrelated to this bug (constructor
`+=` / `??=` lowering, and block-comment placement inside single-line
literals), and they differ on `origin/main` with or without a comment present.

**Issue repro**, `scripts/compat-corpus/one.mjs` vs the official compiler:

```
A.svelte.js  client-dev  MATCH
A.svelte.js  client      MATCH
A.svelte.js  server      MATCH
```

The issue's second, indentation-only snippet (`class S` with `for await`)
matches too, and the equivalent **component** (`<script>` with the same class)
is a full **raw** (un-normalized) MATCH on `client` and `client-dev`, so
defect (b) is fixed rather than deferred.

`node --check` on the emitted JS for 3 sources × {client, client-dev, server}:
9/9 parse (0/9 for the constructor cases before).

**Regression test**:
`crates/rsvelte_core/tests/ctor_private_state_set_paren_2253.rs`, 13 tests —
pins the exact emitted constructor body (not just "contains `$.set`") and
covers the object literal, an array literal, a comment in the last property
position, a comment whose text contains `}` `)` `;` in both a constructor and a
method, a block comment with the same characters, a string literal, a regex
literal, a nested function body, the compound/logical operators, dev mode, the
server target, class-body truncation (a member declared *after* the offending
comment must survive), a `static { … }` block in both its multi-line and
one-line forms, the module-level indentation, and a negative test that a
depth-0 trailing comment is still treated as the statement tail (#907).

## Known, out of scope

`origin/main` prints a verbatim (non-rune) class member at column 0 in module
output and drops the blank line between class members. Both are pre-existing —
measured on a build of `origin/main` with this PR's changes reverted — and are
untouched here; the corpus gate normalizes both away.

### DIFF
--- a/.changeset/ctor-private-state-set-paren.md
+++ b/.changeset/ctor-private-state-set-paren.md
@@ -1,0 +1,40 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Class-field lowering no longer reads a `}`, `)` or `;` that appears inside a
+comment or a literal as code. Two failure modes are fixed.
+
+**Unparseable output.** A `#private` `$state` field assigned from an
+object/array literal that contains a `//` line comment closed the injected
+`$.set(` at the comment's offset — `$.set(this.#x, {\na: s,)// c` — and
+Vite/Rolldown rejected the module with `Parse failure: Unexpected token`. The
+scanner that locates the end of the assigned value treated a `//` at *any*
+bracket depth as the statement's trailing comment, and four sibling scans in the
+method paths took the first `;` anywhere at all, including inside a comment, a
+string or a nested function body.
+
+**Silent content loss.** On the server target the class body's closing brace was
+found with a bare character loop, so a `}` written inside a comment (`// returns
+{ ok, err }`) closed the class early and *every member after it was dropped from
+the output* — no error, no warning, just a class missing its methods. The client
+member scan had the same defect one level down and split a method in two at such
+a comment.
+
+A fifth site: the server treated a class member as a block only when its line
+had a `(`, so a `static { … }` initialization block was never recognised and its
+body was emitted line by line as class fields — each with a `;` appended,
+comment lines included.
+
+All of these scanners now share `shared::js_scan::skip_opaque`, which steps over
+strings, template literals, regex literals and both comment forms in one place;
+a regex such as `{ a: /[});]+/g }` was mis-parsed on every target even with no
+comment involved.
+
+Class lowering also printed its synthesized members at a hard-coded one-level
+indentation, which assumed the class sat one tab deep inside a component
+`<script>`. A top-level `class` in a `.svelte.js` module came out with its
+fields, accessors, constructor and closing brace one tab too deep. Synthesized
+members now follow the class's own source indentation, and a grouped multi-line
+constructor statement keeps the relative indentation of its continuation lines
+instead of being flattened to column 0.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/class_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/class_transforms.rs
@@ -7,6 +7,7 @@ use super::REGEX_INVALID_IDENTIFIER_CHARS;
 use super::expression_needs_proxy;
 use crate::compiler::phases::phase1_parse::utils::find_matching_bracket;
 use crate::compiler::phases::phase3_transform::shared::class_body::split_class_members_onto_lines;
+use crate::compiler::phases::phase3_transform::shared::js_scan::skip_opaque;
 
 /// JS-lexical-aware replacement for `find_matching_paren`: given `s` positioned
 /// just after an opening `(`, return the byte offset of the matching `)`,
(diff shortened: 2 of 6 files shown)


baseballyama:
Re-verified `0aa4b72c` independently: rebuilt the branch and re-ran my matrix — 516 generated cases (6 hosts: constructor, method, nested-if, ctor-loop, getter, static-block × 10 RHS shapes × comment slots × 4 assignment operators) × 3 targets = **1548 comparisons**, acorn `sourceType: 'module'`.

```
UNPARSEABLE rsvelte out: 0
```

Progression on my harness across the three revisions of this branch: **48 → 4 → 0**.

Cause 5 is the interesting one, and worth keeping in the record: it was a *classification* defect (`static {` has no `(`, so it never matched `is_method_start` and fell through to the field emitter), not a *scanning* one. `skip_opaque` — a genuinely good consolidation — was therefore not a complete theory of this bug family, and the only reason we know that is that the two matrices had different host sets. Neither of us would have found it alone.

I also pushed one commit of my own here, `0aa4b72c`: the module doc for `js_scan` used an intra-doc link to `skip_opaque`, which is `pub(crate)`, and `-D rustdoc::private-intra-doc-links` failed the `Documentation` and `MSRV, package, and docs` jobs. Clippy does not look at rustdoc, which is why it passed locally. Link turned into plain code formatting; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p rsvelte_core` is clean.

All 30 checks pass, including `Compiler parity` (the corpus gate). Merging.

PR #2287, opened 2026-08-05
ci: fail when a ratchet lists entries that already pass

baseballyama:
## Root cause

`known-failures.client-dev.json` drifts because **every ratchet gate is one-sided**.

`scripts/compat-corpus/verify.mjs` computed `fixedKnown` (baseline entries that now pass) and
only `console.log`-ed a nudge. The single `process.exit(1)` lived *inside* the
`if (regressions.length)` block, so "N known failures now PASS" left the exit code at 0.
That is why the same line was seen accompanying a red run: those runs also had a NEW failure,
which is what actually failed them.

Evidence — merged PR #2261 (`fix(compiler): report a null prop alias for legacy ownership
mutations`), its own `Compiler parity` job (run `30994362026`, job `92267769732`,
conclusion **success**):

```
[verify] 🎉 39 known failures now PASS (client 0, server 0, client-dev 39) — shrink the baselines:
[verify] ✅ no regressions (client 0, server 0, client-dev 148 known failures remain …)
```

Same picture on #2278's run `31004700195` (job `92301666072`, **success**): 39 now PASS,
141 remaining — plus `[fmt-verify] 🎉 1 known failures now PASS` in the green Formatter-parity
job. So the backlog is carried forward untouched by every green PR until one re-baselines and
absorbs the lot.

**The motivating sequence.** Three consecutive re-baselines each found staleness the previous
one had missed, and all of it was invisible because the gate exited 0:

- #2285 re-baselined on a base predating #2282, so three `$.strict_equals` entries survived it
  and had to be cleaned up by the *next* PR.
- #2250 re-baselined again (ratchet 85).
- #2291 re-baselined again, 91 → 46 on base `94764029`.

A PR whose entire purpose is re-baselining can still leave staleness behind if it measures on
the wrong base. That is exactly what this gate turns from invisible into a hard failure.

Hypotheses ruled out:

- **Not** `continue-on-error` / a swallowed exit code — the steps are bare `run:` lines.
- **Not** a path filter — the `pull_request` filter includes `crates/**`, and the jobs did run.
- **Not** a subset of targets / different normalization — CI runs all three targets and reports
  `client-dev 148`, matching a local full run.
- **Not** an admin override — the runs were genuinely green.

Separately worth noting: the `main` ruleset has **no required status checks** (only `deletion`
+ `non_fast_forward`), so nothing outside auto-merge etiquette forces `Corpus Compat` green
before a merge. That is a repo-settings decision for the owner, not a file change, and this
PR-level gate stands on its own regardless.

## What the fix guarantees

Every ratchet becomes two-sided — a listed entry that already passes fails the run, prints the
offending ids, and names the exact command to fix it:

| Gate | file | shared the flaw |
|---|---|---|
| compiler CSR/SSR/CSR-dev | `verify.mjs` | yes — fixed (all three targets, target-scoped runs stay scoped) |
| formatter parity | `fmt-verify.mjs` | yes — fixed |
| svelte2tsx TSX **and** source map | `svelte2tsx-verify.mjs` | yes — fixed (both ratchets, via the shared `ratchet()`) |
| lint parity | `lint-verify.mjs` | yes — fixed |
| svelte-check parity | `check-verify.mjs` | yes — fixed |
| svelte-check e2e parity | `check-e2e-verify.mjs` | yes — fixed |
| source maps | `crates/rsvelte_core/tests/sourcemaps_gate.rs` | yes — fixed (`fixed` is now asserted empty) |

Notes:

- `fmt-verify` deliberately carries over baseline ids whose oracle is absent from the local
  parity set; those are excluded from the stale set so the documented macOS/Linux carry-over
  does not turn into a false failure.
- A PR cannot dodge the gate by not touching `compatibility/`: the `Corpus Compat` path filter
  triggers on `crates/**`, `Cargo.lock`, `scripts/compat-corpus/**` and the submodule pins, and
  the sourcemap gate is a plain `cargo test`.
- Also fixes `MAX_PRINT` in `verify.mjs`: `Number(args[i + 1] || 20)` evaluated to `NaN`
  whenever `--max-print` was absent but another flag was present (e.g. `--no-fmt`), and
  `slice(0, NaN)` is empty — so a real regression list could print zero entries.
- Consequence: the PR that fixes entries must re-baseline **in that PR**, so deltas stay small
  and a real regression can no longer hide inside a large, normal-looking "now PASS" delta.

## The gate proving itself

The first CI run of this branch failed `Formatter parity` (run `31011731212`, job
`92325369377`) with exactly the pre-existing staleness the old gate had been printing green:

```
[fmt-verify] ❌ 1 baseline entries already PASS — the ratchet is stale.
  - skeleton/sites/skeleton.dev/src/components/landing-page/design-system.svelte
  fix: node scripts/compat-corpus/fmt-verify.mjs --update-baseline
```

## Re-baseline commit

Kept as a **separate commit** from the gate change so the diff shows plainly which entries the
new gate forced out. The fmt entry above is removed surgically from
`fmt-known-failures.json` + its justification in the paired `.md`, **not** via
`--update-baseline`: `scripts/compat-corpus/README.md` (lines 239-248) forbids rewriting that
baseline from macOS, because Linux CI includes ~13 loose-declaration-tag entries macOS `oxfmt`
skips and a macOS rewrite would silently drop them.

Attribution: none of the ratchet entries removed here are this PR's own work — they are
pre-existing staleness that the new gate is the first thing to report. If a compiler PR merges
between the re-baseline this branch sits on (#2291, `71398d28`) and this run, its residue lands
in this commit too; that is the gate working, and the alternative — another wrong-base
measurement — is strictly worse.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -58,7 +58,9 @@ rsvelte for CSR, SSR **and** dev-mode CSR (the three targets declared in
 `scripts/compat-corpus/targets.mjs`). Outputs must be byte-identical after comparison-side normalization
 (oxfmt + blank-line stripping — never compiler post-passes). To grow the corpus, add a submodule
 plus a line to `corpus-sources.json`. CI ratchet: `compatibility/known-failures.{client,server,client-dev}.json`
-may only shrink, and each remaining failure is justified in `compatibility/known-failures.md`. The
+may only shrink, and each remaining failure is justified in `compatibility/known-failures.md`. Every
+ratchet is two-sided: a new failure **and** a listed entry that already passes both fail CI, so the PR
+that fixes entries must re-baseline in the same PR instead of leaving a backlog for a later one. The
 same directory holds three sibling shrink-only ratchets, each with per-entry justification in a paired
 `.md`: the formatter-parity gate (`fmt-known-failures.json` / `fmt-oracle-excluded.json`), the
 svelte2tsx output-parity gate (`svelte2tsx-known-failures.json`), and the lint output-parity gate
--- a/compatibility/fmt-known-failures.json
+++ b/compatibility/fmt-known-failures.json
@@ -7,7 +7,6 @@
 	"powertable/app/src/lib/components/PowerTable.svelte",
 	"powertable/app/src/routes/examples/+layout.svelte",
 	"skeleton/sites/plus.skeleton.dev/src/routes/(app)/content/blocks/+page.svelte",
-	"skeleton/sites/skeleton.dev/src/components/landing-page/design-system.svelte",
 	"skeleton/sites/skeleton.dev/src/components/ui/header/theme.svelte",
 	"skeleton/sites/skeleton.dev/src/components/ui/preview.svelte",
 	"skeleton/sites/themes.skeleton.dev/src/lib/components/generator/Controls/ControlsColors.svelte",
--- a/compatibility/fmt-known-failures.md
+++ b/compatibility/fmt-known-failures.md
@@ -5,7 +5,7 @@ The formatter-parity corpus formats every `.svelte` component with both
 Svelte structure + oxc for embedded JS/CSS — rsvelte-fmt's exact layering) and
 requires **byte-identical** output. The ratchet may only shrink.
 
-**Current baseline: 20 entries**, concentrated in real-world corpus repos
+**Current baseline: 19 entries**, concentrated in real-world corpus repos
 (skeleton, layerchart, svelte-ux, layercake, cmsaasstarter, and a long tail).
 Oracle-bug / invalid-input / migrate cases are NOT here — those are permanently
 excluded in `fmt-oracle-excluded.json` (see `fmt-oracle-excluded.md`). Every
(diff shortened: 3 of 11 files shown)


baseballyama:
この赤は staleness ゲートの誤検出ではなく、**測定に使った main が古い**ためのようです（別ブランチから、検証できる事実のみ）:

```
失敗した run の作成時刻   2026-08-05T14:04:35Z  (job 92325369615)
#2285 が main に入った時刻 2026-08-05T14:08:24Z  「shrink the client-dev ratchet to 91」= 42 件削減
このブランチの merge-base   origin/main より 5 コミット後ろ（#2285 #2288 #2268 #2286 #2289 が未取込）
```

つまり `client-dev 45` は **133 件版の baseline** に対する数字です。`origin/main` に rebase してから再測定すると、45 のうち 42 は #2285 で既に削られているはずで、残りが本 PR で縮めるべき実数になります（#2286 / #2289 の compiler 修正でさらに減る可能性あり）。

なお `--max-print` のインデックス修正（フラグ無しのとき `args[0]` を拾う）と、「大きな now-PASS デルタの中に本物の regression が隠れる」という WHY コメントは、ゲート本体とは独立に効く良い修正だと思います。

PR #2298, opened 2026-08-05
chore(compat): shrink the client-dev ratchet to 49

baseballyama:
The client-dev corpus baseline had not been re-measured since #2248. The five
dev fixes that landed since (#2286, #2289, #2290, #2293 and the ones already
merged before the last update) clear 42 of the 91 entries.

Re-derived from a full `compat-corpus` run (oxfmt-normalized, AST-structural
comparison): `client` 0, `server` 0, `client-dev` 49. `known-failures.md` is
updated to match — header count, the cluster table (recounted by helper-family
under/over-emit, not by first differing line), the cluster/residue
reconciliation, and the per-row prose for the rows whose remaining entries are
now a different bug than the one described.

The residue section also records the CSS-sourcemap model derived while
characterising that row: upstream's guard is `dev && analysis.inject_styles &&
css.code`, and MagicString emits a segment at every chunk start, every line
start inside an unedited chunk, and every `addSourcemapLocation` — which
`css/index.js`'s `_` visitor calls on every CSS AST node's `start` and `end`.

https://claude.ai/code/session_014EJ8yjsgEh6CxqJhQoGZHP


baseballyama:
Closing as a subset of #2291, which makes the identical 91→49 ratchet shrink **and** adds a regression test.

Both PRs rewrite `compatibility/known-failures.client-dev.json` to the same 49 entries, so keeping both open guarantees one wasted rebase-and-remeasure cycle — those cost a full NAPI rebuild plus a 14027×3 corpus recompile each, and this file has already gone through six of them today.

#2291 additionally carries `regrouped_reactive_mutations_keep_their_own_location` in `crates/rsvelte_core/tests/dev_ownership_validator_mutation.rs`, which pins the ownership-mutation location pairing for legacy `$:` statements regrouped in dependency order. #2289 landed that fix with no test covering that angle, so the gap is worth closing.

Note both branches are now behind `origin/main` anyway: #2250 merged as `88daef15` carrying a ratchet of **85**, so 49 has to be re-measured on top of it. #2291 is doing that now.

Reopen this if #2291 stalls.

Issue #2299, opened 2026-08-05
vite-plugin: dependency .svelte.js modules reach the browser uncompiled (breaks bits-ui's entire test suite)


MathiasWP:
## Summary

`@rsvelte/vite-plugin-svelte` does not compile `.svelte.js` / `.svelte.ts` modules that live inside **dependencies**. They are served to the browser with their runes intact, so the first one evaluated throws.

Swapping only the plugin — same frozen dependency tree, same tests — takes bits-ui from a fully green suite to zero:

| plugin | result |
|---|---|
| `@sveltejs/vite-plugin-svelte@6.2.0` | **1239 passing** / 43 files |
| `@rsvelte/vite-plugin-svelte@0.5.0` | **0 passing** / 43 files |

Every one of the 43 browser test files fails to load, in both browsers:

```
Caused by: ReferenceError: $derived is not defined          (chromium)
Caused by: ReferenceError: Can't find variable: $derived    (webkit)
  ❯ boxWith  svelte-toolbelt/dist/box/box-extras.svelte.js:5:20
  ❯ packages/bits-ui/dist/internal/body-scroll-lock.svelte.js:19:19
```

## This is a plugin defect, not a compiler defect

Both halves were isolated:

- **The compiler is correct.** Calling `compileModule` directly on that exact file through the NAPI binding emits the expected `$.derived(...)` — no unreplaced runes.
- **`setup-optimizer.js` is byte-identical** to official 6.2.0's, so the rolldown-vs-esbuild optimizer branch is not the difference.

So the defect is in plugin-layer file selection / dependency prebundling: which files the plugin hands to `compileModule` at all. Not root-caused further.

## Why no existing gate catches this

Every corpus track compares compiler **output**. bits-ui's own 677 sources are byte-identical across all three corpus targets, and `svelte-toolbelt`'s modules compile correctly in isolation — so the corpus is green while every real app breaks. Only running the plugin against a real dependency graph exposes it.

## Reproducing

```bash
pnpm run plugin-swap
```

Currently baselined as 43 entries in `compatibility/plugin-swap-known-failures.json` (justified in the paired `.md`) so the new gate can land; this issue is that ratchet's burn-down target. Fixing the prebundling bug should drop the entire block at once.

Pinned versions: `vite@7.1.5`, `vitest@3.2.4`, `svelte@5.46.4`. **Not** version-specific drift — it reproduces at the lockfile pins.

baseballyama:
[...]

Issue #2309, opened 2026-08-05
codegen: constant expression not folded inside a style-attribute template


MathiasWP:
## Summary

A compile-time-constant expression inside a `style` attribute template is folded by official but emitted verbatim by rsvelte.

**Target:** `client`, `client-dev`

```js
// official
background-size: 192px 256px;
// rsvelte
background-size: 192px ${64 * sheetRows}px;
```

`sheetRows` is a module-level constant, so official folds `64 * sheetRows` to `256` at compile time.

## Affected corpus entry

- `runed/sites/docs/src/lib/components/demos/animation-frames.svelte`

## Severity

Low — the runtime result is identical; rsvelte's form just evaluates on every update instead of once at build time. Output-parity blocker with a minor perf edge.

Note the neighbouring interpolations (`${-$.get(currentCol) * 64}px`) are genuinely reactive and correctly left unfolded by both, so the gap is specifically constant propagation of a non-reactive binding into the template.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

Issue #2308, opened 2026-08-05
codegen (server): `x = x + 1` is contracted to `x += 1` in a .svelte.(js|ts) module


MathiasWP:
## Summary

The server target preserves a compound assignment where official expands it.

**Target:** `server`

```js
// official
runs = runs + 1;
// rsvelte
runs += 1;
```

## Affected corpus entry

- `runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts`

## Severity

Low — output-parity only; the two forms are equivalent for a numeric binding. Worth confirming the expansion is unconditional upstream rather than tied to a reactive-binding path, since for a **reactive** target the expanded form is what lets the assignment be intercepted; if so, the same gap could be a correctness bug on a different input.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

baseballyama:
Reproduced, but the direction in the title is inverted — worth recording before anyone fixes it.

At the pinned source, `watch.test.svelte.ts` line 14 already reads `runs = runs + 1` (a plain `=`, not a compound assignment). Official keeps it verbatim; **rsvelte contracts it to `runs += 1`**. So the bug is a spurious contraction, not a missing expansion.

Minimal repro (`.svelte.js`, target `server`):

```js
export function t() {
	let runs = $state(0);
	let plain = 0;
	function f() {
		runs = runs + 1;   // official: unchanged     rsvelte: runs += 1
		plain = plain + 1; // both unchanged
		runs = runs - 2;   // official: unchanged     rsvelte: runs -= 2
		runs = 1 + runs;   // both unchanged
	}
	return f;
}
```

Only `$state` bindings are affected — `plain` is untouched, and the contraction only fires when the state variable is the *left* operand.

Root cause: the `.svelte.js` module path is the one server path that is still string-based (per AGENTS.md, SSR components are pure-AST but `compileModule` is not). `transform_module_server` runs the **client** module transform and then undoes it in `post_process_for_server`. The client transform lowers both `runs += 1` and `runs = runs + 1` to the identical text `$.set(runs, $.get(runs) + 1)` (`client/state_pipeline_ast.rs`, the `Assign` and compound arms converge after read-wrapping). By the time `post_process_for_server` rewrites `$.set` back to an assignment, the original operator is gone, so `recompact_compound_set` (`server/mod.rs:370`) *guesses* — and it always guesses "compound".

[...]

Issue #2306, opened 2026-08-05
codegen (dev): missing $.log_if_contains_state console instrumentation


MathiasWP:
## Summary

Dev mode does not wrap `console.*` calls in the `$.log_if_contains_state` helper official emits.

**Target:** `client-dev` only

```js
// official
console.log(...$.log_if_contains_state("log", m));
// rsvelte
console.log(m);
```

Note the polarity: **rsvelte emits the plain call**, official emits the instrumented one.

## Affected corpus entry

- `runed/packages/runed/src/lib/utilities/persisted-state/persisted-state.svelte.ts` (test module)

## Severity

Dev-only fidelity. No production impact — the helper exists so that logging a rune-bearing value in dev prints the unwrapped state rather than the internal signal object. Users logging `$state` values see raw internals instead of the friendly form.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

baseballyama:
Reproduced, but **the polarity in the report is inverted** — rsvelte emits the *instrumented* call and official emits the plain one.

Minimal repro is the sibling test module `packages/runed/src/lib/utilities/persisted-state/persisted-state.test.svelte.ts` (not `persisted-state.svelte.ts`, which matches on all three targets):

```js
vi.spyOn(console, "error").mockImplementation((...args) => {
  console.log(...$.log_if_contains_state("log", "args:", args)); // both agree — `args` is a rest param
  const m = `console.error was called with args ${args.concat(", ")}`;
- console.log(m);                                   // official
+ console.log(...$.log_if_contains_state("log", m)); // rsvelte
  throw new Error(m);
});
```

Root cause: upstream gates the wrap on `scope.evaluate(arg).has_unknown`, and its `TemplateLiteral` case (`phases/scope.js`) adds `STRING` — never `UNKNOWN` — so a `const` bound to a template literal folds to a string and is *not* wrapped.

rsvelte has that rule (`console_wrap.rs::shape_can_be_unknown` returns `false` for `E::TemplateLiteral`), but the `.svelte.(js|ts)` module path decides the wrap from *generated text*, where `m` is just a bare `Identifier`. `identifier_can_be_unknown` then only resolves names declared **exactly once at `analysis.root`**; `m` is declared inside a nested arrow function, so the lookup misses and the predicate errs toward "unknown" → wrap.

[...]

PR #2310, opened 2026-08-05
test(ci): plugin-swap parity gate

MathiasWP:
Rebased onto main now that #2321 has landed. The corpus/submodule commits are gone; this is the plugin-swap gate alone — 1 commit, 10 files.

**The original description of this PR was wrong and has been replaced.** It claimed the gate "found a real bug within an hour" (#2299). It did not: that measurement was the harness. What follows is the corrected version.

## What the gate does

Every corpus track compares compiler **output**. None observes *which files the Vite plugin hands to `compileModule`*, so a plugin that silently skips a dependency's `.svelte.js` module is green across all of them and still breaks every real app. This runs a pinned project's own suite twice against one frozen tree — official plugin, then rsvelte's — and requires the same tests to pass.

## Two resolution invariants it asserts

This is the "make the harness prove it did the work" point from your review, applied. Each of these produced a confident and entirely fictional result *before* it was asserted.

**1. The shim and the target must resolve the same `vite`.** A symlinked package resolves its imports from its realpath, so the earlier swap had the shim importing vite 8 (rolldown) while driving bits-ui's vite 7 (esbuild) server. `setup-optimizer.js` branches on `rolldownVersion`, so it registered its optimizer plugins where nothing read them, and every prebundled dependency `.svelte.js` reached the browser uncompiled. That was the 1239 → 0 measurement.

The shim is now staged as a real directory with peers linked from the **target**, and the versions are compared before the run. **All 86 baselined entries disappear** — the ratchet ships empty.

**2. The target's `svelte` must equal the version rsvelte mirrors.** This one is not a bug and does not go away:

| | compiler | runtime | matched? |
|---|---|---|---|
| official plugin | resolved **from the target** | target | ✅ by construction |
| rsvelte plugin | pinned to the mirrored svelte | target | only if the versions coincide |

bits-ui runs `svelte@5.46.4`; the mirror is `5.56.8`. Between them `rest_props`'s `exclude` changed from an Array (`.includes`) to a Set (`.has`), so rsvelte emits Set-shaped code into a runtime calling `.includes` — 2436 "regressions" measuring the version gap. Verified not to be the compiler: the failing component compiles **byte-identically** under both at identical options.

## Why this is still draft

Invariant 2 means the gate **currently refuses to run against bits-ui**. It hard-fails on skew rather than reporting a diff it cannot attribute, so there is nothing baselined and nothing to review as a finding.

Living with it is an open decision, and I do not think it is mine to make:

- **Pin the target's svelte** to the mirrored version during setup — no longer the project as shipped, and it breaks `--frozen-lockfile`.
- **Only enrol matching targets** — fragile; an upstream bump becomes a gate outage.

If neither is acceptable, that is a legitimate reason to close this. The premise still holds (nothing else in CI observes plugin-layer file selection), but a gate that cannot run is not worth carrying.

## Other review items, applied

- **Two-sided ratchet from the start**, matching what #2287 established — a listed entry that starts passing fails as loudly as a regression. `--update` also refuses to *grow* the baseline without `--allow-growth`.
- **`baseline-failure` and a crashed swapped run are hard failures.** Both previously exited 0; the crashed-swap case produced exactly the 86 baselined entries and reported green.
- **`process.exit` inside `try` replaced with `throw`** so the restore runs, plus signal handlers and a stale-`*.official` guard.
- **One vitest project per invocation**, keyed `project|file|test` — the reporter carries no discriminator at 3.2.4, but instances are selectable as `--project="browser (chromium)"`, which removes the merge that let one engine mask another.
- **Placement** as you suggested: own workflow, per-PR only on plugin paths, scheduled + push-to-main otherwise. Playwright installs `--with-deps` on Linux; the declared `timeoutMinutes` is now applied.
- **`auto-shrink-ratchets.yml` dropped** — I had drafted it as a safety net for one-sided ratchets, and #2287 makes it redundant.
- **`test:vps-fixture` wired into CI** (shipped in `package.json`, never run), and `collect.mjs` skips `dist/` so a local build cannot leak generated output into the corpus.

Still not run on `ubuntu-latest` — no checks have executed on this branch, so wall clock and the `--with-deps` fix remain unverified.


baseballyama:
Niceeee!!!!!!🚀🚀🚀🚀🚀

baseballyama (commented):
Thanks for this — it is a genuinely well-argued contribution, and the write-up did most of my homework for me. I read `plugin-swap-verify.mjs` line by line, the new ratchet + `.md`, the workflow wiring, the corpus enrolment, and #1087 for the prior art. I did **not** run the gate (installing bits-ui + Playwright browsers was out of budget here), so everything below is reasoned from the scripts; where that matters I say so.

## What is clearly good

- **The premise is correct and the repo has a real hole here.** Every existing track compares compiler *output*. Nothing observes which files the plugin hands to `compileModule`. Byte-identical output over 677 bits-ui sources plus a broken app is exactly the situation you describe, and #2299 is the proof. That argument lands.
- **Symlink-swap over `overrides` + reinstall is the right mechanism**, and the measured drift numbers (vite 7.1.5→7.3.6, vitest 3.2.4→3.2.7) are the concrete thing #1087 was missing. Replacing the realpath'd package directory rather than a consumer-side symlink is also the correct level — under pnpm every consumer funnels through that one store directory.
- **`diff()` collapsing a dead suite into one `suite-load-failure` instead of N test regressions** (L119–128) is a good call: it keeps the ratchet proportional to defects rather than to test counts, and it is why the 43-entry baseline reads as one bug.
- **The `finally` restore + `artifacts` cleanup** (L213–229) shows you thought about leaving the submodule pristine, and the `collect.mjs` `dist` interaction you found is a real cross-gate hazard that nobody would have predicted.
- **`plugin-swap-known-failures.md` is the best-written of the ratchet `.md`s in this repo.** Isolating the defect to the plugin layer (compileModule correct, `setup-optimizer.js` byte-identical) before baselining it is exactly the standard the other `.md`s aim for.

## Findings, most severe first

### 1. `baseline-failure` makes the whole gate exit 0 — a green run that verified nothing (`plugin-swap-verify.mjs:188–192`)

```js
if (basePassing === 0) {
  console.error(`[plugin-swap] ${t.id}: BASELINE FAILURE — official plugin produced no passing tests`);
  return { id: t.id, result: 'baseline-failure', regressions: [], fixed: [] };
}
```

`result: 'baseline-failure'` is written into the report and then **never read again**. The aggregation at L251–257 only walks `r.regressions`, which is empty, so `fresh` is empty and the script falls through to `log('✅ no new divergences')` and exits 0. The swap is never even performed.

So any failure that zeroes the baseline — Playwright browsers not installed, a vitest config error, an OOM, a `prepare` step that half-succeeded — produces a **green gate**. And several of those are reachable without any other failure signal, because the Playwright install ignores its own result:

```js
run(`pnpm exec playwright install ${t.playwright.join(' ')}`, runDir, { quiet: true });   // L162
```

`run()` returns a boolean that is discarded here. Install fails quietly → both browser projects fail to launch → `basePassing === 0` → `baseline-failure` → ✅.

This is the one finding I would call blocking, because it inverts the gate's failure mode: the situation in which the gate is least able to tell you anything is the situation in which it is loudest that everything is fine. Suggested shape: `baseline-failure` should be a hard `process.exit(1)` (the target is a committed pin — if the official plugin cannot pass its own suite at that pin, that *is* something a human needs to look at), or at minimum feed a distinct non-zero exit. Also `if (!run(playwright install)) process.exit(1)`.

Related: at L191 the early return omits `basePassing` / `swapPassing`, so those come out `undefined` in `plugin-swap-report.json`.

### 2. `process.exit(1)` inside the `try` skips the `finally` — the no-op path leaves the target broken (`plugin-swap-verify.mjs:196–205`)

```js
try {
  fs.renameSync(official.dir, stash);
  fs.symlinkSync(SHIM, official.dir);
  const now = pluginIdentity(runDir);
  if (!now.name.startsWith('@rsvelte/')) {
    ...
    process.exit(1);          // <- finally at L213 does not run
  }
```

`process.exit()` does not unwind `finally`. The comment at L214 ("Always restore, even on crash") is not true for this path. The consequence is nastier than a dirty tree: the store directory is left as a symlink to the shim and the real plugin is parked at `*.official`, so the **next** run resolves `@rsvelte/vite-plugin-svelte` as its "official" baseline, `fs.renameSync(official.dir, stash)` hits an existing `.official`, and the run is garbage. Since swap-noop is precisely the state where you most want a clean retry, this should `throw` (and let the `finally` restore) rather than `exit`.

Worth adding a `process.on('SIGINT'|'SIGTERM')` restore too — a cancelled CI job or a local Ctrl-C at the wrong moment leaves the same wreckage.

### 3. Stale `--outputFile` reports can produce a false green (`plugin-swap-verify.mjs:177–182, 211–212`)

`baseOut` / `swapOut` are never unlinked before the runs. If vitest dies before writing its report (config error, crash, OOM), `readVitestReport` reads the file **left over from the previous invocation** and the run is compared against last time's results. Locally that is a plausible false green, and locally is where `--update` gets run.

The fix is one line each (`fs.rmSync(baseOut, { force: true })`), and the signal you need is already computed and thrown away: `readVitestReport` sets `crashed: true` at L92/L97 and **nothing ever reads `crashed`**. `if (swapped.crashed) → hard fail` closes both this and part of #1.

### 4. Swap-noop detection is a static resolve, not evidence the plugin ran — and it assumes exactly one copy (`plugin-swap-verify.mjs:65–76, 200–205`)

Two gaps in the check that the gate's whole value rests on:

- `resolvePluginDir` resolves **one** copy, from `runDir`. In a pnpm monorepo the store can legitimately hold two `@sveltejs+vite-plugin-svelte@*` entries (different versions, or the same version with different peer hashes). Swapping the one that `tests/` resolves would then pass the noop check while some other consumer — notably `vite.config.ts` loading, or a workspace package's own config — still gets official. Cheap hardening: glob `node_modules/.pnpm/@sveltejs+vite-plugin-svelte@*` under the target and assert exactly one match before swapping.
- The check runs `createRequire` **in the gate's process**, which proves resolution, not execution. The strongest cheap positive control is in-band: have the shim emit something identifiable (a banner, a marker in the vite cache dir name, an env-gated log line) and assert it appears in the swapped run's output. Today, if the swapped vite silently fell back for any reason, the gate's evidence that "the swap really happened" is a `package.json` read.

I want to be explicit that I could not test either of these — I did not install bits-ui. They may both be fine in practice for this one target; they are load-bearing assumptions that are currently unasserted.

### 5. A whole browser project can fail under swap and be invisible (`plugin-swap-verify.mjs:99–110`)

```js
// A file run under N browser projects appears N times under the same `name`
// with no project discriminator, so statuses are merged pessimistically
files.set(rel, prev + (suite.assertionResults?.length ?? 0));
```

Summing across projects means `files.get(file) > 0` as long as **one** project produced assertions. So if webkit dies entirely under the swap and chromium is fine, there is no `suite-load-failure`; and the test ids that exist only from chromium all read as passing. A plugin defect that only manifests in one engine is silently ungated.

The same merge bites at baseline in the other direction and interacts with #1: if webkit fails to launch at baseline (very plausible on `ubuntu-latest` — see #7 below, `playwright install` is called without `--with-deps`), every id merges to `ok === false`, `diff()` skips them all at L130, and `basePassing` can reach 0 → `baseline-failure` → green.

Running one project per invocation (`--project=chromium`, then `--project=webkit`, keyed `project|file|test`) removes the merge hack, the blind spot and the pessimistic-baseline amplification in one move, at the cost of doubling the number of vitest invocations (not the work).

### 6. The new ratchet is one-sided, exactly what #2287 is fixing (`plugin-swap-verify.mjs:270–276`)

```js
if (stale.length) {
  log(`🎉 ${stale.length} known failure(s) now PASS — shrink the baseline:`);
  ...
}
if (fresh.length) { ...; process.exit(1); }
```

Stale entries warn and exit 0 — the identical shape #2287 is currently converting to a hard failure across the seven existing gates. Please make this two-sided from the start (`if (stale.length || fresh.length) process.exit(1)`), so #2287 doesn't have to come back for an eighth.

**Ordering caveat, and it matters:** with #1 unfixed, two-sidedness would be actively harmful — a `baseline-failure` empties `seen`, which marks *all 43* known entries stale, which under a two-sided gate is a spurious red on an unrelated PR. Fix #1 first, then #6.

Also `--update` (L264–267) writes `seen` unconditionally, so `pnpm run plugin-swap:update` after a bad run happily **grows** the baseline. Given the shrink-only house rule I would add a guard: refuse to write a strictly-larger baseline without an explicit `--allow-growth`.

### 7. CI cost and placement (`corpus-compat.yml:535–574`)

The job has no `paths` of its own, so it inherits the workflow filter — including `crates/**`. That is essentially every compiler PR, and each one pays: root install + NAPI build + bits-ui `--frozen-lockfile` install + `pnpm -F bits-ui build` + `svelte-kit sync` + chromium & webkit + **two** full browser suites.

Two things make me doubt the ~5 min figure:

- `gh pr checks 2310` reports **no checks have run on this branch at all**, so this job has never executed on `ubuntu-latest`. The 40-minute `timeout-minutes` suggests you weren't confident either.
- The number is measured in the currently-broken state, where the rsvelte side dies immediately at 0 passing. **When #2299 is fixed the swapped side starts actually running 1238 tests × 2 browsers**, so the job's steady-state cost is roughly double today's. The gate gets more expensive precisely as its burn-down succeeds.
- `pnpm exec playwright install` without `--with-deps` (L162) usually is not enough for webkit on `ubuntu-latest`.

`timeoutMinutes: 25` in `plugin-swap-targets.json:11` is declared but never read by the script.

On placement: I'd argue **per-PR only on paths that can plausibly move plugin-layer file selection** (`apps/npm/vite-plugin-svelte/**`, the NAPI crate, `.gitmodules`/target pin) and **scheduled + push-to-main** otherwise. The reasoning: for `crates/**` changes the corpus already proves byte-identical output on 14k entries, so the plugin-swap gate's marginal information is near zero except on entries that are themselves baselined. That keeps the gate on the PRs where it actually discriminates and off the ones where it is a 5–10 minute tax. If you'd rather keep it per-PR, I'd want a measured post-#2299 wall clock first.

### 8. Hermeticity: the claim is *nearly* right, and the residual is worth stating

"Nothing else in the dependency graph moves" is true for the official side and for the target's tree. It is not quite true for the swapped side: Node resolves through the symlink's realpath, so the shim's own runtime deps (`@rsvelte/vite-plugin-svelte-native`, `@sveltejs/vite-plugin-svelte-inspector`, `deepmerge`, `magic-string`, `obug`, `vitefu`) come from the **rsvelte root** `pnpm-lock.yaml`, not from bits-ui's. That is defensible — you *want* to test the shipped shim — and it is still hermetic, because both lockfiles are committed. But the two sides are pinned by two different lockfiles, and a root-lockfile bump can therefore move the swapped side alone. One sentence in the header comment would make that explicit rather than something a reader has to derive.

On the "weekly submodule bump moves drift into its own PR" argument: I checked, and yes — `auto-update-submodules.yml` gains `runed` / `svelte-toolbelt` entries, and bits-ui was already there. That part of the #1087 rebuttal holds up.

### 9. Corpus enrolment (separate assessment)

Mechanically this follows the CLAUDE.md recipe exactly: submodule + one line in `corpus-sources.json` (`markdown: false`, matching the other project entries), `auto-update-submodules.yml` matrix entries, both push and PR `paths` lists updated, `.gitmodules` with `ignore = dirty` / `shallow = true` like its siblings. `known-failures.md` header counts are updated to `10` / `4`, which is what `known-failures-md-check.mjs` parses, so that check should stay green. The gap-filling argument (module share 3.4% → 5.1%) is a good reason to pick these two specifically.

The one thing I want to flag for @baseballyama rather than for you: this moves `known-failures.client.json` and `.server.json` **off zero for the first time**, adding 14 baselined entries for what your own `.md` calls genuine correctness bugs (#2300/#2301/#2304 produce stale values or a runtime `ReferenceError`). That is a maintainer policy call, not a code-review one. You already offer the alternative in the PR body. I'd note that the enrolment and the plugin-swap gate are independent and could land separately if that call goes the other way.

Smaller: the `dist` skip in `collect.mjs:52–61` is a **global** change to what "the corpus" means, made to defend against an artifact that the gate already cleans up itself via `artifacts`. It is belt-and-braces, which is fine, but it means any future corpus source that legitimately commits a `dist/` silently loses those entries. I could not verify "no corpus repo commits a `dist/`" locally (the submodules aren't populated in my worktree) — how did you check that?

### 10. Flakiness mitigation — I don't think it's adequate yet

You document it honestly, and the current 43 entries being deterministic `suite-load-failure`s is a fair reason it isn't biting today. But the mitigation is "it hasn't bitten yet", and your own numbers already show the jitter (PR body says 1239 baseline passing, the `.md` says 1238). Once #2299 is fixed and the swapped side runs 1238 real browser tests, a single flaky test becomes a red on someone else's PR with no obvious cause — historically the exact failure mode that got ecosystem-ci deleted, arriving by a different route.

The cheap mitigation that fits the existing structure: on a `test-regression` (not `suite-load-failure`), re-run just that test id on both sides once, and only report it if it reproduces. `suite-load-failure` — the class that found the actual bug — stays single-run and fast.

## Open questions

1. Has the plugin-swap job been run on `ubuntu-latest` at all? `gh pr checks 2310` says no checks have run. Specifically: does webkit launch without `playwright install --with-deps`, and what is the real wall clock?
2. What is the projected job duration **after** #2299 is fixed, when the swapped side runs the full suite instead of dying at load?
3. Does bits-ui's tree contain exactly one `@sveltejs/vite-plugin-svelte` in the pnpm store at the pinned lockfile? (Finding #4.)
4. Does the vitest JSON reporter genuinely carry no project discriminator at vitest 3.2.4, or would a newer reporter/`--project` split give you per-project ids cheaply? (Finding #5.)
5. Do you have a preference on landing order — the gate is independently valuable and could go in before the corpus enrolment, which is the part that carries the ratchet-policy question.

None of the above touches the core idea, which I think is right and which has already paid for itself once. Findings #1, #2 and #6 are the ones I'd want addressed before this becomes a required check; #3–#5 are hardening; #7 is a judgement call for the maintainer.

MathiasWP:
**Moving this to draft — the headline claim in the description is wrong, and I want to correct it before anyone spends more review time.**

I root-caused #2299 (now closed as not-a-bug). @baseballyama's read was right: it was my harness.

The swap symlinks the shim into the target's pnpm store, and Node resolves a symlinked package's imports from its **realpath**, so `setup-optimizer.js`'s `import * as vite from 'vite'` resolves out of the rsvelte root rather than the target:

```
shim resolves vite:    8.2.0   rolldownVersion: 1.2.1   → rolldown branch runs
bits-ui actually runs: 7.1.5   (esbuild-based)          → reads esbuildOptions
```

The optimizer plugins get registered on a path the running vite never reads, so every prebundled dependency `.svelte.js` reaches the browser uncompiled. Confirmed by forcing `rolldownVersion = undefined` — the 43 suites that produced zero tests then load and run.

`vite` is a `peerDependency`, so a real install resolves it to the user's own vite. **No user hits this.**

### What that invalidates

- **"Found a real bug within an hour" — it did not.** The 1239 → 0 measurement was real; what it measured was the harness. That was the main argument for this PR and it doesn't hold.
- **The 86 baselined entries are artifacts**, not a burn-down target.
- **The hermeticity residual I called "defensible" is load-bearing.** You flagged it as worth stating explicitly. It's fine for `deepmerge`/`magic-string` and wrong for a peer dep the plugin branches on — which is precisely the thing that broke.

### What survives

The **premise** is untouched: nothing in CI observes which files the plugin hands to `compileModule`, and that remains a real hole. Also everything from your review is fixed and pushed (including a `swapped.crashed` false-green I'd left half-done — a crashed swapped run produced exactly the 86 baselined entries and reported green).

But a gate that fabricates its own failure is worse than no gate, so this shouldn't merge as-is.

### Proposed path

1. **Redesign the swap** so the shim resolves the *target's* vite — copy into the store path and link only the non-peer deps, letting `vite`/`vitefu` resolve upward. Then re-measure. The honest possibility is the 86 entries go to zero and the gate finds nothing, which is a fine outcome and worth knowing.
2. **Split off the corpus enrolment** (`runed` / `svelte-toolbelt`) — it is independent of all of this, unaffected by the bug, and carries the real findings (#2300–#2309, incl. `$state` never lowered and a missing rune unwrap in production output). That half is ready; this half is not.

That also answers your Q5 on landing order, from the other direction than either of us expected. Happy to open the enrolment as its own PR if you'd prefer to look at it separately.

baseballyama:
Status update, since this PR has been open while the ground moved under it.

**The enrolment half has landed** — #2321 merged this morning. So this PR is now only the plugin-swap parity gate, and you should drop the corpus/submodule commits on your next rebase.

**What the enrolment bought, in the end**: eleven real divergences, of which ten are fixed and merged — #2300, #2301, #2302, #2303, #2304, #2305, #2306, #2309, #2330 and #2343. `known-failures.client-dev` went 46 → 20 over the same window and `client`/`server` are still at 0, so enrolling runed + svelte-toolbelt cost exactly one entry on `server` (#2308, architectural) and one on `client-dev` (#2064's `$.assign` cluster). Picking two rune-in-module libraries was the right call — that shape had zero coverage in 14,028 existing entries.

**On the swap-mechanism redesign you identified**: your diagnosis was right and it is worth restating, because we hit an adjacent instance of the same class today. Your symlink caused the shim to resolve `vite` from its realpath, so a vite-8 plugin drove a vite-7 server and the rolldown branch ran against an esbuild-based optimizer. The general shape is *a harness whose own resolution differs from the target's*, and the failure mode is that it manufactures results that look like findings.

We found the same shape in the corpus gate itself, from the other direction: on the oxfmt 0.62 branch, the normalization pass matched **zero files and exited 0** (0.62 applies the repo's VCS ignore rules to its walk, and `compatibility/{expected,actual}` are gitignored), so every one of ~42,000 outputs was compared **raw** while the gate reported normally. That manufactured 22 "failures" that normalization absorbs entirely. Fixed by walking from a symlink in `os.tmpdir()` — no `.git` above the walk root — plus a canary file that hard-exits if the pass leaves the tree untouched.

The lesson we took, which applies to your gate too: **make the harness prove it did the work**, rather than infer it from the absence of an error. For the plugin swap that would mean asserting, at runtime, that the shim and the target resolve the *same* `vite` — one check that would have caught your peer-dep misresolution on the first run instead of after a 1239 → 0 measurement.

No rush on this from our side. When you do pick it back up, the scheduled + push-to-main placement with per-PR only on plugin paths still seems right, and the two-sided ratchet from #2287 means a listed entry that starts passing now fails CI as loudly as a regression — so the gate's own ratchet needs to be exact from the start rather than approximately right.

Thanks again for the reports. Two of them (#2306, #2308) turned out to point the opposite way from the description — rsvelte was wrong in the other direction — and #2318 was a comparison against a stale upstream, but the divergences were real in every case that mattered, and #2299 you root-caused yourself. That is a good ratio.

MathiasWP:
Rebased — corpus/submodule commits dropped, now 1 commit / 10 files. Description replaced (the old one led with the #2299 claim, which was the harness).

**On "make the harness prove it did the work"** — that is exactly the check that was missing, and it is now in. Two assertions, both of which would have caught their respective fiction on the first run:

- shim and target must resolve the same `vite` — this is your suggestion verbatim, and it collapses the 86 entries to zero
- the target's `svelte` must equal the version rsvelte mirrors

The second is the one I would flag for you, because it is not fixable from inside the gate. Official's plugin compiles with `svelte/compiler` resolved *from the target*, so its output always matches that target's runtime. rsvelte's compiler is pinned to its mirror. bits-ui at 5.46.4 against a 5.56.8 mirror gave 2436 "regressions", all from `rest_props`'s `exclude` changing Array (`.includes`) → Set (`.has`).

So the gate currently **refuses to run against bits-ui** rather than reporting a diff it cannot attribute. Options are pin-the-target's-svelte (breaks `--frozen-lockfile`, and it is no longer the project as shipped) or enrol-only-matching-targets (an upstream bump becomes an outage). If neither appeals, closing this is a reasonable call — the premise holds, but a gate that cannot run is not worth carrying.

Your oxfmt 0.62 case is the same shape from the other side and a better example than mine, because the failure was *silent success*: a normalization pass that matched zero files, exited 0, and manufactured 22 failures. The canary that hard-exits when the pass leaves the tree untouched is the generalisation — assert the work happened, do not infer it from the absence of an error. I have tried to apply that consistently here: a zero baseline, a crashed swapped run, a no-op swap, and a peer mismatch are now all hard failures rather than paths that quietly produce a comparison.

Thanks for the #2318 correction too — `rolldownOptions` is right and 6.2.0 was stale. I diffed against whatever bits-ui's lockfile happened to pin rather than against current upstream, which is the same mistake in a different costume: trusting a local artifact as the reference. Along with #2306 and #2308 pointing the opposite way, that is three of eleven where my description was wrong even though the divergence was real — worth me being more careful about direction and reference version before writing the prose.

Good to hear the enrolment paid off. Ten fixed and `client-dev` 46 → 20 is a much better return than I expected from two small libraries.

baseballyama:
Closing because the gate is currently not runnable against its only target: bits-ui resolves Svelte 5.46.4 while rsvelte mirrors 5.56.8. The resulting runtime mismatch is non-attributable, as documented in this PR. Reopen only with an explicit version-skew policy and a matching target.

Issue #2304, opened 2026-08-05
codegen: $.template_effect emitted without its deps array while the body references $0/$1 (ReferenceError)


MathiasWP:
## Summary

`$.template_effect` is emitted with a zero-argument callback, but the callback body still references the `$0` / `$1` parameters that official binds via the deps array. Those identifiers are unbound at runtime.

**Target:** `client`, `client-dev`

```js
// official
$.template_effect(
  ($0, $1) => {
    $.set_text(text_6, `${$0 ?? ""}%`);
    $.set_text(text_7, `${$1 ?? ""}%`);
  },
  [() => scroll.progress.x.toFixed(0), () => scroll.progress.y.toFixed(0)],
);

// rsvelte
$.template_effect(() => {
  $.set_text(text_6, `${$0 ?? ""}%`);
  $.set_text(text_7, `${$1 ?? ""}%`);
});
```

## Affected corpus entry

- `runed/sites/docs/src/lib/components/demos/scroll-state.svelte`

## Severity

Correctness, in **production** output, and it throws rather than degrading: `$0` is not defined in scope, so the effect raises a `ReferenceError` the first time it runs. The deps array is also dropped, so even with the parameters bound the effect would not track `scroll.progress`.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

PR #2317, opened 2026-08-05
fix(compiler): lower $state declarations in .svelte.(js|ts) modules

### LINKED ISSUE #2300: codegen: $state declaration in a .svelte.(js|ts) module is not lowered at all
## Summary

A top-level `$state(...)` declaration inside a `.svelte.ts` module is emitted as a plain assignment — the state call disappears entirely, so the value is never reactive.

**Target:** `client`, `client-dev`

```js
// official
let userId = $.state(1);
// rsvelte
let userId = 1;
```

```js
// official (client-dev)
let count = $.tag($.state(0), "count");
// rsvelte
let count = 0;
```

## Affected corpus entries

- `runed/packages/runed/src/lib/utilities/resource/resource.test.svelte.ts`
- `runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts`

## Severity

Correctness, in **production** output. Any read of these bindings returns a non-reactive value, so nothing that depends on them ever updates. This is one of the divergences that moved `known-failures.client.json` off zero for the first time.

Found by adding `runed` / `svelte-toolbelt` to the corpus (rune-heavy module libraries — the corpus's module share was 3.4% before them).

Baselined in `compatibility/known-failures.client.json` / `.client-dev.json`; see the "runed / svelte-toolbelt enrolment" section of `compatibility/known-failures.md`.

baseballyama:
Closes #2300

Reported by @MathiasWP while enrolling `runed` / `svelte-toolbelt` into the compat corpus (#2310).

## Cause

`transform_module_script_runes` classifies a module's `$state` bindings as reactive or not
through **name-keyed** lists derived from `analysis.root.bindings`. That vector is flat across
every scope, so two same-named `$state` locals in sibling function scopes collapse into a
single entry. As soon as one of them is never reassigned, the name lands in
`module_non_reactive_vars` and **every** binding of that name loses its `$.state(...)` wrapper,
its `$.get(...)` reads and its `$.set(...)` writes:

```js
// runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts
testWithEffect("...", async () => {
	let count = $state(0);   // reassigned    → official: $.state(0)
	count++;                 //               → rsvelte:  0   (plain, non-reactive)
});
testWithEffect("...", () => {
	const count = $state(0); // never written → official & rsvelte: 0
});
```

Upstream's `is_state_source` gates on `binding.reassigned` for the *specific* binding it is
visiting, never on a name.

## Fix

Only names whose `$state` bindings **disagree** on reassignment are ambiguous, and only those
are resolved per binding:

* `state_call_ast` builds a `Semantic` for the module program and, for an ambiguous name,
  decides the declaration rewrite from the resolved symbol's own write references — mirroring
  `binding.reassigned` per binding. Unambiguous names keep trusting the analysis, which is
  required: a `<script module>` `$state` can be reassigned from the instance script, a write
  the module program alone cannot see.
* Ambiguous names are additionally handed to the read/write pipeline, whose existing
  symbol-identity matching (`find_state_var_symbols`) already separates the reactive
  declarations from the ones lowered away.

## Verification

* Reported corpus entry (`runed/.../watch.test.svelte.ts`): 8 wrong lowerings → 0. The single
  remaining divergence in that file is unrelated (a spurious third `$.set(..., true)` proxy flag).
* Corpus ratchets: `client` 0 → 0, `server` 0 → 0, `client-dev` **42 → 39** (`derived-in-expression`,
  `mutation-both`, `wrapped-$state` now match — the whole "signal read/write" cluster).
* `cargo test --release -p rsvelte_core --test runtime --test ssr` green, plus a new
  `same_named_siblings_resolve_per_binding` unit test.

## On #2304

Not the same defect, so it stays open. #2304 is `$.template_effect` emitted with a
zero-argument callback whose body still references the `$0`/`$1` deps parameters — a component
**template** codegen path (deps-array construction) in a `.svelte` file, with no module `$state`
classification involved and no same-name binding collision. This fix does not touch it.


MathiasWP:
your speed is 🏎️ !!

Issue #2307, opened 2026-08-05
codegen: spurious `/* @__PURE__ */` annotation on a default parameter initializer


MathiasWP:
## Summary

rsvelte adds a `/* @__PURE__ */` annotation to a default parameter initializer that official leaves bare.

**Target:** `client`, `client-dev`

```js
// official
function serializeValue(value, key, dateFormats = {}, codecEncoders = new Map()) {
// rsvelte
function serializeValue(value, key, dateFormats = {}, codecEncoders = /* @__PURE__ */ new Map()) {
```

## Affected corpus entry

- `runed/packages/runed/src/lib/utilities/use-search-params/use-search-params.svelte.ts`

## Severity

Low — output-parity only, no behavioural difference. `/* @__PURE__ */` is a hint to downstream minifiers; on a default-parameter initializer it is harmless (the expression is only evaluated when the argument is omitted). Blocks byte-identical parity.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

baseballyama:
Reproduced, and it is **not** client-only — it diverges on `server` too. Minimal case:

```ts
export function serializeValue(
	value: unknown,
	key: string,
	dateFormats: Record<string, string> = {},
	codecEncoders: Map<string, (value: unknown) => unknown> = new Map()
): string { ... }
```

```js
-export function serializeValue(value, key, dateFormats = {}, codecEncoders = new Map()) {          // official
+export function serializeValue(value, key, dateFormats = {}, codecEncoders = /* @__PURE__ */ new Map()) { // rsvelte
```

The annotation is not invented by rsvelte: the corpus pipeline strips TS with esbuild before `compileModule` (both compilers see the same text), and esbuild emits `/* @__PURE__ */ new Map()` for the erased-generics `new` expression. Both compilers therefore *receive* the comment; official drops it in the default-parameter-initializer position and rsvelte keeps it.

Upstream's `add_comments` (`phases/1-parse/acorn.js`) does attach it as `leadingComments` on the `NewExpression` — the `AssignmentPattern` and its `left` both start before the comment, so the walk lands on `right`. The divergence is therefore on the **printing** side: esrap does not emit `leadingComments` in that position, while `rsvelte_esrap` does. Fix belongs in the printer's leading-comment emission rules, mirroring which esrap handlers actually call the comment-emitting path — not in comment attachment.

Not fixed in #2323 (which only covers #2305). Thanks @MathiasWP.

[...]

Issue #2305, opened 2026-08-05
codegen: reactive parameter thunk not called in an assignment target (`key` vs `key()`)


MathiasWP:
## Summary

Inside a generated `set value($$value)` accessor, a reactive parameter used as a computed member key is emitted without its call, so the key is the thunk function rather than its value.

**Target:** `client`, `client-dev`

```js
set value($$value) {
-   params.fields[key()] = $$value;   // official
+   params.fields[key] = $$value;     // rsvelte
},
```

## Affected corpus entry

- `runed/sites/docs/src/lib/components/demos/use-search-params.svelte`

## Severity

Correctness, in **production** output. The assignment lands on a property keyed by the stringified function instead of the intended key, so the write goes to the wrong slot and the intended field never updates.

Likely related to #2301 (reactive value not unwrapped at a use site), but this one is an assignment **target** in a generated accessor rather than a call argument, so filing separately.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

Issue #2303, opened 2026-08-05
codegen: private class-field state read emits `.v` instead of `$.get(...)`


MathiasWP:
## Summary

Reading a `$state` held in a **private class field** goes through the raw `.v` property instead of `$.get(...)`.

**Target:** `client`, `client-dev`

```js
// official
if (!$.get(this.#pressedKeys).includes(key)) {
// rsvelte
if (!this.#pressedKeys.v.includes(key)) {
```

## Affected corpus entry

- `runed/packages/runed/src/lib/utilities/pressed-keys/pressed-keys.svelte.ts`

## Severity

Correctness, in **production** output. `.v` bypasses `$.get`, so the read is never registered as a dependency — effects and deriveds that read this field will not re-run when it changes. It also reaches into a runtime-internal field shape rather than the public accessor.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

Issue #2302, opened 2026-08-05
codegen: $state initializer missing the $.proxy wrapper (deep reactivity lost)


MathiasWP:
## Summary

`$.state(...)` is emitted without the `$.proxy(...)` wrapper official applies to the initializer.

**Target:** `client`, `client-dev`

```js
// official (client)
let loading = $.state($.proxy(initialValue === void 0 && !lazy));
// rsvelte
let loading = $.state(initialValue === void 0 && !lazy);
```

```js
// official (client-dev)
let loading = $.tag($.state($.proxy($.strict_equals(initialValue, void 0) && !lazy)), "loading");
// rsvelte
let loading = $.tag($.state($.strict_equals(initialValue, void 0) && !lazy), "loading");
```

## Affected corpus entry

- `runed/packages/runed/src/lib/utilities/resource/resource.svelte.ts`

## Severity

Correctness, in **production** output. `$.proxy` is what gives `$state` its deep reactivity — without it, mutations to nested properties of the held value do not trigger updates.

Found by adding `runed` / `svelte-toolbelt` to the corpus.

PR #2321, opened 2026-08-05
test(corpus): enrol runed + svelte-toolbelt (rune-in-module libraries)

MathiasWP:
Split out of #2310 (now draft) so the half that stands on its own can be reviewed independently. **This PR contains no plugin-swap machinery** — it is corpus enrolment only, and is unaffected by the harness bug that sank the other half.

## Why these two

The corpus was component-heavy. Modules were **3.4%** of entries, so runes written *outside* components were the thinnest surface it covered. `runed` (24 `.svelte.js`, 0 components) and `svelte-toolbelt` are almost pure module libraries, and two of the most-depended-on utility packages in the Svelte 5 ecosystem.

| | before | after |
|---|---|---|
| Corpus entries | 14,027 | **14,130** |
| Module share | 3.4% | **5.1%** |

## What they found

They moved `known-failures.client.json` and `.server.json` **off zero for the first time** — 13 entries across the two production targets, every one a real bug, filed individually:

| symptom | official | rsvelte | issue |
|---|---|---|---|
| `$state` declaration not lowered | `let userId = $.state(1)` | `let userId = 1` | #2300 |
| reactive value not unwrapped | `setTimeout(…, $.get(wait$))` | `setTimeout(…, wait$)` | #2301 |
| missing deep-reactivity proxy | `$.state($.proxy(x))` | `$.state(x)` | #2302 |
| private class-field state read | `$.get(this.#pressedKeys)` | `this.#pressedKeys.v` | #2303 |
| `$.template_effect` loses its deps array | `($0, $1) => …, [ … ]` | body references unbound `$0`/`$1` | #2304 |

Plus #2305, #2307, #2308, #2309, and the dev-only #2306.

**#2300 / #2301 / #2304 are correctness bugs in production output**, not formatting: a non-unwrapped reactive read returns a stale value, a `$state` that is never lowered is never reactive at all, and #2304 throws a `ReferenceError` the first time the effect runs.

## The call this asks you to make

This takes `client` and `server` from **0 → 9 and 0 → 4**. Those lists have been empty, and that is a meaningful property to give up.

The alternative is to leave the PR red until the nine issues are fixed. I baselined instead, shrink-only with per-entry justification in `known-failures.md`, so the enrolment can land and hold the line while the burn-down happens — but it is your policy call, and reversing it is deleting the entries.

Worth noting the bugs exist either way; the corpus simply had no source that exercised them.

## Verification

Measured against `main` at 02a315c0, on macOS:

- compiler parity ✅ 14,130 entries — `client 9, server 4, client-dev 56`
- svelte2tsx ✅ 0 `ts-mismatch`, 0 `map-invalid` — **clean against the new sources**
- fmt ✅ clean against the new sources (measured before the split; unchanged here)
- `known-failures-md-check` ✅

`client-dev` is `main`'s 46 plus 10 enrolment entries. I deliberately did **not** re-baseline `main`'s existing entries: I can only measure on macOS, and shrinking a ratchet on unverified ground risks turning this PR red for reasons unrelated to it. Keeping extra entries is always safe; removing live ones is not.

Both submodules join the weekly `auto-update-submodules` rotation, so upstream drift arrives as its own PR.


baseballyama (commented):
## Re-measured on current `main` — the answer changed, and it changed in your favour

First, the headline: **the cost of this enrolment is now 1 client + 1 server + 2 client-dev entries, not 9 + 4 + 10.** Nine of the ten bugs you found were fixed today (#2300, #2301, #2302, #2303, #2304, #2305, #2306, #2309, #2330), and the ratchets you were asked to move off zero mostly stay at their new floor of one.

You wrote "it is your policy call, and reversing it is deleting the entries." The measurement says the policy call is much cheaper than you priced it — which is the best possible outcome for a PR framed this honestly.

### Method

`gh pr checkout 2321`, rebased onto `origin/main` @ `7f834ade`. On the conflicts in the three ratchet files I took **main's** versions (`client []`, `server []`, `client-dev` 20 entries) and kept only your submodules + `corpus-sources.json` line, so `verify` reports exactly what enrolment adds and nothing else. Release `rsvelte_napi` + `ast_equiv_batch` staged, `pnpm -C submodules/svelte install`, then `collect` → `compile` → `verify` across all three targets **with** oxfmt normalization (not `--no-fmt`).

```
[collect] runed: 91 files + 0 markdown snippets
[collect] svelte-toolbelt: 12 files + 0 markdown snippets
[collect] total: 14131 corpus entries
```

+103 entries over the 14028 baseline, so the submodules really did clone and the measurement is live.

### Result

```
match 13161 · error-parity 948 · js-mismatch 22 · css-mismatch 0 · error-mismatch 0

[verify] ❌ 4 NEW failures (not in baseline)
```

| target | your baseline | measured on current `main` | now passing |
|---|---|---|---|
| `client` | 9 | **1** | 8 |
| `server` | 4 | **1** | 3 |
| `client-dev` | 10 (on top of 46) | **2** (on top of main's 20) | 8 |

No stale-entry complaint against main's existing 20 `client-dev` entries, so that side of the ratchet is clean.

### The four survivors, enumerated and attributed

**1–2. `runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts` — `client` + `server`**, both from one source line, `runs = runs + 1` where `runs` is `$state(0)`:

```js
// server — expected: runs = runs + 1;   actual: runs += 1;
// client — expected: $.set(runs, $.get(runs) + 1);   actual: $.set(runs, $.get(runs) + 1, true);
```

The server half is **#2308** exactly, and it is the residue you predicted: the `.svelte.(js|ts)` server path round-tripping through the client transform.

The client half is **not filed anywhere** — I checked #2300–#2309, #2330 and the open list. It is the third argument to `$.set`, the `should_proxy` flag. Upstream decides it with `should_proxy()` in `3-transform/client/utils.js:133`, which returns `false` outright for a `BinaryExpression` (line 141). `$.get(runs) + 1` is a `BinaryExpression`, so official omits the flag and rsvelte passes `true`. Severity is genuinely low — `proxy()` on a number or string is a no-op, so there is no behavioural difference on this input, and I could not construct one, since `BinaryExpression` cannot yield a proxyable object. It is a wrong-by-construction flag and a parity blocker, not a correctness bug. **Worth a new issue**; it is a small, well-localised port gap (add the `BinaryExpression`/`UnaryExpression` early-out to whichever `should_proxy` equivalent feeds this call site).

My guess is that this line was always diverging but was invisible to you, because `verify` scores one entry per file and this file was already baselined for #2301. It is residue exposed by the fixes, not a regression from them.

**3. same file, `client-dev`** — the dev-mode face of #1, same `$.set(..., true)`.

**4. `runed/sites/docs/src/lib/components/demos/scroll-state.svelte` — `client-dev` only:**

```js
// expected: $.assign(scroll, "x", "=", $.get(x), "…scroll-state.svelte:41:69")
// actual:   scroll.x = $.get(x)
```

Missing dev-mode `$.assign` ownership instrumentation. **#2064** names `$.assign` explicitly, so this is #2064 residue and lands in an already-justified cluster.

**#2307 costs nothing.** `use-search-params.svelte.ts` is in your `client` baseline but **passes today**: the spurious `/* @__PURE__ */` is comment-only, and the comparator's AST-equivalence fallback does not see comments. So the esrap epic (#2336) is not blocking this PR at all.

### Two factual corrections, offered as corrections and not as criticism

Both divergences were real and both reports were worth filing — only the direction was described backwards:

- **#2308** — the source is `runs = runs + 1` and rsvelte *contracts* it to `runs += 1`. The issue title has since been fixed to say "contracted", but `known-failures.md` in this PR still reads "`+=` not expanded to `x = x + 1`", which is the inverted framing. Worth a one-line fix when you re-baseline.
- **#2306** — same inversion, per the maintainer's triage.

For the record, the rest of your day was unusually careful work: root-causing **#2299** as a harness artifact rather than a compiler bug is exactly the diligence that makes a bug report trustworthy, and retracting a claim you had reached by grepping instead of diffing is the part almost nobody does.

### On the merits

- **Enrolment procedure** — followed exactly: gitmodules entry + `corpus-sources.json` line + README table row. You also updated `auto-update-submodules.yml` and both `corpus-compat.yml` path filters and the shallow-checkout list, which is more than the documented minimum and the right call.
- **Pins** — `runed@a12cd68f`, `svelte-toolbelt@7b0e2a6d`, both gitlinked with `branch = main`, `shallow = true`, `ignore = dirty`, matching every other corpus source. Reproducible; both cloned cleanly here.
- **CI cost** — +103 entries on 14028 is **+0.73%**. Full 3-target compile of 14131 entries ran in 19.7s locally, so enrolment adds ~0.15s of compile; verify scales the same way. Negligible.
- **Scope** — nothing in the diff outside enrolment, ratchet baselining and the paired justification. Nothing to split out.

**One gap:** `corpus:sync` in `package.json` (line 60) was not updated, though the two CI checkout lists were. Anyone running `pnpm run corpus` locally will silently get a 14028-entry corpus and conclude the enrolment is clean, because the submodules never clone. That is the exact failure mode where a measurement looks green for the wrong reason. Two names to add.

### Recommendation (verdict left to the maintainer)

The `client` and `server` ratchets do not go to 9 and 4. They go to **1 and 1**, and both entries are one source line in one test file. That is a very different trade from the one your description asks the maintainer to weigh, and I think it makes this land-able.

To make CI green, the ratchets need re-baselining on a fresh rebase — the gate is **two-sided** (`verify.mjs:392-410`: a listed entry that already passes is fatal, "the ratchet is stale"), so as submitted it would fail with 8 client + 3 server + 8 client-dev stale entries. Post-rebase the correct contents are:

- `known-failures.client.json` → `["runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts"]`
- `known-failures.server.json` → `["runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts"]`
- `known-failures.client-dev.json` → main's current 20 + `watch.test.svelte.ts` + `runed/sites/docs/src/lib/components/demos/scroll-state.svelte` = 22

Note `main`'s `client-dev` is now **20**, not the 46 you branched from (#2291 and others shrank it today), so that file will conflict on rebase — take main's side and add the two.

Your instinct not to re-baseline main's existing entries from macOS was right, and it is why I measured with main's ratchets rather than yours. The rewrite above is a shrink of your own entries only.

`known-failures.md` will need the counts and the cluster prose updated to match (9→1, 4→1, and the client-dev "10 arrived with the enrolment" → 2), plus the #2308 polarity fix, plus a line for the unfiled `should_proxy` flag once it has an issue number.

Thank you for this. Ten real codegen bugs out of one enrolment, nine of them already fixed within a day, is the single best argument anyone has made for why the corpus should keep growing — and the module surface you targeted is exactly where it was thinnest.

baseballyama:
(Rewriting this comment — the original was posted from a shell that mangled several code spans and spliced unrelated console output into it. Apologies for the noise; content below is what was intended.)

Maintainer follow-up to the measurement review above — the cost of this PR has dropped again since that review was taken.

**#2343 is now fixed** (#2344, in CI): the spurious third argument in `$.set(runs, $.get(runs) + 1, true)`. That was 1 of the 3 entries this enrolment would have added. Upstream's `should_proxy` returns false outright for a `BinaryExpression`; rsvelte was missing exactly that one arm of the whitelist, and probing it also turned up a second latent divergence (`mk() instanceof Map`).

So once #2344 lands, enrolling runed + svelte-toolbelt costs:

| target | added entries | attribution |
|---|---|---|
| client | **0** | — |
| server | 1 | #2308 — the `.svelte.(js\|ts)` server path round-trips through the client transform |
| client-dev | 1 | #2064's `$.assign` ownership cluster |

`client` back at **0** is the part that matters. That ratchet sitting at zero is what makes any future entry there read as a correctness alarm rather than noise, and it was the main reason to hesitate on enrolment when this was first proposed.

**Please rebase onto current `main` and re-baseline.** The two-sided gate from #2287 means a listed entry that now *passes* fails CI just as loudly as a regression, and `main`'s `client-dev` baseline has moved 46 → 20 since you branched, so that file will conflict — take main's side and add the one remaining entry.

Also worth fixing in the same push: **`corpus:sync` in `package.json:60` was not updated** alongside the two CI checkout lists. Anyone running `pnpm run corpus` locally silently gets the old 14028-entry corpus and concludes the enrolment is clean because the new submodules never clone — green for the wrong reason, which is the exact failure mode this pipeline exists to prevent.

---

For the record on what this enrolment bought: it surfaced **eleven** real divergences, of which #2300, #2301, #2302, #2303, #2304, #2305, #2306, #2309, #2330 and #2343 are now fixed and merged.

Two of your reports (#2306, #2308) turned out to have **inverted polarity** — rsvelte was wrong in the opposite direction from the description — but the divergences were real either way, and #2308's title has been corrected. #2299 you root-caused yourself as a harness artifact. #2318 turned out to be a stale-baseline comparison where our code already matched upstream 6.2.2+.

That is a good ratio on eleven reports, and the two libraries were well chosen: rune-in-module is exactly the shape the existing corpus did not cover.

baseballyama (approved):
Approved. All checks green on the rebased head (`7ec019f3`), including `Compiler parity` — the corpus gate this PR moves. Details in the comment below.

baseballyama:
Rebased onto current `main` (`04a50404`), re-baselined, and merged. Thank you — this was a well-aimed enrolment.

## What changed on top of your commit

**Rebase.** Three ratchet files conflicted; I took `main`'s side on all three and kept only your submodules + the `corpus-sources.json` line, so what the run reports is exactly what enrolment adds. Note `main`'s `client-dev` had reached **0** by the time this rebased — not the 46 you branched from, and not the 20 quoted in the review above.

**Re-baseline, measured not assumed.** Full local run: `collect` → `compile` → `verify` with oxfmt normalization over all **14,131** entries × 3 targets (`+103` over the 14,028 baseline, so the submodules really did clone). The ratchets are far smaller than the ones you submitted:

| target | as submitted | measured on `main` | remaining entry |
|---|---:|---:|---|
| `client` | 9 | **0** | — |
| `server` | 4 | **1** | #2308 |
| `client-dev` | 10 | **1** | #2064's `$.assign` row |

Nine of the ten codegen bugs were fixed while this was open (#2300, #2301, #2302, #2303, #2304, #2305, #2306, #2309, #2330, #2343). #2307 stays invisible to this gate — it is comment-only and the comparator is AST-structural — so it costs nothing here and burns down with #2336.

`client` back at **0** is the part that matters: that ratchet sitting empty is what makes any future entry there read as a correctness alarm rather than noise.

**Warning gate.** #2314 landed after you branched, so the enrolment meets a gate your measurement could not have seen. It finds two more divergences, both baselined with a fresh issue and justification:

- **#2348** — `perf_avoid_nested_class` over-warns at function depth 1 in a standalone `.svelte.(js|ts)`. `analyze_module` passes `ast_type: null`, not `'module'`, so upstream's `allowed_depth` is 1 there and only depth ≥ 2 warns; we use 0. Code ratchet 70 → 71.
- **#2349** — the same warning carries no span (`?:?` vs `12:4`). Position ratchet 626 → 627.

**`corpus:sync`.** Fixed, as the review asked. Both submodules were in the two CI checkout lists but not in `package.json:60`, so a local `pnpm run corpus` would have compiled the old 14,028-entry corpus and pronounced the enrolment clean — green for the wrong reason.

**`known-failures.md`.** Counts and prose brought in line, including the #2308 polarity: the source is `runs = runs + 1` and rsvelte **contracts** it to `runs += 1`, not the other way round.

## One thing found while re-baselining

`verify.mjs --update-baseline --update-warning-baseline` writes **nothing** and exits 0 — `UPDATE_BASELINE` skips the warning loop, and the warning path's `finish(0)` then returns before the output baselines are written. Each flag disables the other's write. Filed as **#2350**; worked around here by running verify twice.

That is the failure mode this pipeline exists to prevent, so it is a fitting thing for an enrolment PR to have shaken loose.

## For the record

Eleven real divergences from one enrolment, ten already fixed, plus two warning bugs and a hole in the re-baseline tooling. `runed` and `svelte-toolbelt` were the right two libraries: rune-in-module was the corpus's thinnest surface, and it was thin enough that ten production-codegen bugs were sitting in it undisturbed.

PR #2325, opened 2026-08-05
fix(compiler): fold constants reaching a template through a non-literal initializer

### LINKED ISSUE #2309: codegen: constant expression not folded inside a style-attribute template

baseballyama:
Closes #2309

Reported by @MathiasWP while enrolling `runed` / `svelte-toolbelt` into the compat corpus.

## What was wrong

`runed/sites/docs/src/lib/components/demos/animation-frames.svelte`:

```svelte
const sprites = 10;
const sheetCols = 3;
const sheetRows = Math.ceil(sprites / sheetCols);
...
style="background-size: {64 * sheetCols}px {64 * sheetRows}px;"
```

Official folds both chunks (`background-size: 192px 256px`); rsvelte folded only
the first and emitted `${64 * sheetRows}px` for the second.

Root cause: Phase 2 stores `binding.initial` only for *literal* initializers.
A non-literal one is kept separately as AST JSON in `binding.init_expr_json`,
gated by `init_needs_expr_json`, which did not include `CallExpression` — and
neither the client evaluator (`get_literal_value_json`, Identifier arm) nor the
server evaluator (`evaluate_binding_initial`) ever consulted that JSON. So any
identifier with a non-literal initializer resolved to "unknown" and stayed
reactive. Upstream's `scope.evaluate` recurses into `binding.initial` whatever
the node shape (`phases/scope.js`), so this mirrors that.

Three changes, all in the evaluation path:

- `init_needs_expr_json`: accept `CallExpression` (`Math.ceil(...)` is in
  upstream's known-function table).
- client `get_literal_value_json`, Identifier arm: when `initial` is absent, fall
  back to `init_expr_json_parsed()` and recurse (depth-capped for cyclic
  initializers).
- server `evaluate_binding_initial`: same fallback before returning unknown.

Both fallbacks skip `$derived` bindings, whose runtime value is the signal read,
not the compile-time initializer.

## Verification

`scripts/compat-corpus/one.mjs` on the reported file: `client`, `server` and
`client-dev` all MATCH (previously `client` / `client-dev` diverged as reported,
and `server` diverged the same way).

## Not in this PR — #2308

#2308 (`server`: `+=` not expanded) reproduces, but the direction in the issue
title is inverted: the source is `runs = runs + 1` and it is **rsvelte** that
contracts it to `runs += 1`. Details in a comment on the issue; the fix needs the
`.svelte.js` module path to stop guessing, which is a separate change.

### DIFF
--- a/.changeset/fold-non-literal-binding-initials.md
+++ b/.changeset/fold-non-literal-binding-initials.md
@@ -1,0 +1,10 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Fold constants that reach a template through a non-literal initializer. A `const`
+whose initializer is a call, binary or conditional expression (`const rows =
+Math.ceil(sprites / cols)`) is now evaluated at compile time when it is read from
+a template chunk, so `style="background-size: {64 * rows}px"` emits the folded
+literal instead of a reactive interpolation — matching the official compiler on
+client, dev-client and server output.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/variable_declarator.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/variable_declarator.rs
@@ -965,7 +965,8 @@ fn init_needs_expr_json(init: &JsNode) -> bool {
         JsNode::TemplateLiteral { expressions, .. } => !expressions.is_empty(),
         JsNode::BinaryExpression { .. }
         | JsNode::UnaryExpression { .. }
-        | JsNode::ConditionalExpression { .. } => true,
+        | JsNode::ConditionalExpression { .. }
+        | JsNode::CallExpression { .. } => true,
         _ => false,
     }
 }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
@@ -3816,6 +3816,13 @@ fn get_literal_value_json(
     }
 }
 
+/// Mirrors upstream `get_global_keypath`, which yields a rune keypath only when
+/// the name is unbound: in legacy mode `$state` is the store subscription of an
+/// imported `state`, not a rune.
+fn is_rune_callee(name: &str, context: &ComponentContext) -> bool {
+    context.state.analysis.runes && context.state.get_binding(name).is_none()
+}
+
 const MAX_INITIAL_EVAL_DEPTH: u8 = 8;
 
 thread_local! {
@@ -3930,7 +3937,11 @@ fn get_literal_value_complex(
                 }
 
                 // Fix C: $state.raw(arg) — MemberExpression callee with object=$state, property=raw
-                if obj_type == "Identifier" && obj_name == "$state" && prop_name == "raw" {
+                if obj_type == "Identifier"
+                    && obj_name == "$state"
+                    && prop_name == "raw"
+                    && is_rune_callee(obj_name, context)
+                {
                     let args = obj.get("arguments").and_then(|a| a.as_array());
                     if let Some(args) = args
                         && let Some(first_arg) = args.first()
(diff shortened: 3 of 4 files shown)


baseballyama:
**Ratchet not re-measured locally — needs a follow-up before merge.**

This change widens constant folding globally (any identifier with a non-literal
initializer can now fold in a template chunk), so it is exactly the shape of
change that can move `known-failures.{client,server,client-dev}.json` in both
directions. I could not run
`node scripts/compat-corpus/verify.mjs --update-baseline <target>` in this
session: the 33 corpus submodules were not checked out here and cloning them
plus three full passes over ~14027 entries did not fit the time box.

What *was* verified:

- `scripts/compat-corpus/one.mjs` on `runed/sites/docs/src/lib/components/demos/animation-frames.svelte`
  — `client`, `server`, `client-dev` all MATCH (all three diverged before).
- `cargo test --release -p rsvelte_core --test runtime --test ssr` — 5/5 and 2/2 pass
  (runtime legacy, runtime runes, hydration, browser, ssr).
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — only the
  known pre-existing `collapsible_match` at `1_parse/read/expression.rs:7062`, which is
  clean on CI's clippy.

So the three ratchets must be re-baselined (rebased onto `origin/main` first) and
`compatibility/known-failures.md` updated before this merges.

baseballyama:
**Ratchet measured — the caveat above is resolved.**

Rebased onto `origin/main` (`f3baaec6`), all 34 corpus submodules initialised, NAPI binding rebuilt, collect reports **14027 entries**. All three targets measured with oxfmt normalization:

| target | before (main) | after |
|---|---|---|
| `client` | 0 | **0** |
| `server` | 0 | **0** |
| `client-dev` | 39 | **39** |

`git diff compatibility/` is empty — the three ratchet JSONs are byte-identical to main's, so nothing to add and nothing to delete. `known-failures-md-check.mjs` passes unchanged.

**One regression was found and fixed in code, not baselined.** The first `client` run came back at 1: `svelte/packages/svelte/tests/validator/samples/no-runes-mode/input.svelte`

```svelte
<script>
    import { state } from './store';
    const x = $state();
</script>
{x}
```

folded to `text.nodeValue = x` instead of staying reactive. In legacy mode `$state` is not a rune — it is the store subscription of the imported `state` — but two evaluator sites (`get_literal_value_json` and `is_expression_known_json`) matched the callee by name alone and treated `$state()` as the rune. Widening `init_needs_expr_json` to cover call initializers made those sites reachable from an identifier's initializer for the first time, which is what surfaced it. Upstream expresses the same condition through `get_global_keypath`, which returns null for a bound name, so both sites now go through `is_rune_callee` (runes mode **and** the name unbound). Second commit on the branch.

After that fix both the reported file and the regressed fixture are correct:

- `runed/sites/docs/.../animation-frames.svelte` — `client`, `server`, `client-dev` all MATCH
- `no-runes-mode/input.svelte` — MATCH

Re-ran after the narrowing: `cargo test --release -p rsvelte_core --test runtime --test ssr` 5/5 + 2/2; clippy clean apart from the known pre-existing `collapsible_match` at `1_parse/read/expression.rs:7062`.

Separately filed #2330 for the opposite-direction divergence found with a synthetic probe while working on this (inline `{Math.ceil(a / b)}` folded by rsvelte, memoized by official). Not in the corpus, so no ratchet entry covers it.

PR #2334, opened 2026-08-05
fix(compiler): resolve dev console arguments against the generated program's own consts

### LINKED ISSUE #2306: codegen (dev): missing $.log_if_contains_state console instrumentation

baseballyama:
Closes #2306

Reported by @MathiasWP.

**The report's polarity is inverted.** rsvelte does not *miss* the `$.log_if_contains_state` wrap — it emits one where official emits the plain call, so the fix is the opposite of what the issue asks for:

```js
// official
console.log(m);
// rsvelte (before)
console.log(...$.log_if_contains_state("log", m));
```

## Cause

Upstream wraps a `console.*` call only when some argument's `scope.evaluate(arg).has_unknown` holds. `scope.evaluate` resolves an `Identifier` through `scope.get(name)` and, for an un-updated binding with an initializer, evaluates that initializer — a `TemplateLiteral` contributes `STRING` and never `UNKNOWN`.

rsvelte's script paths rewrite *generated* JS text, so they use `console_wrap::shape_can_be_unknown`, the scope-free half of the same lattice. That half already returns `false` for a `TemplateLiteral`, but the reported shape is

```ts
const m = `console.error was called with args ${args.concat(", ")}`;
console.log(m);
```

where the argument is a bare `Identifier`. `identifier_can_be_unknown` resolved only against `ComponentAnalysis`, which carries root / instance / template bindings — `m` is declared inside a nested arrow function, so the lookup missed and the predicate erred toward `UNKNOWN`.

## Fix

`collect_local_consts` indexes the parsed generated program's own `const` declarators (name → the initializer's shape verdict) and `identifier_can_be_unknown` consults it first, reconstructing the reach of upstream's `scope.get(name)`. Restricted to `const` so `binding.updated` is false by construction, and to names declared exactly once in the program so a shadowing declaration still errs toward the wrap — the same direction the predicate already erred in.

## Not fixed here

#2307 (`/* @__PURE__ */` on a default parameter initializer) is left open; it is an independent printer-side divergence with a much larger blast radius. Details in a comment on that issue.

### DIFF
--- a/.changeset/console-wrap-nested-const-scope.md
+++ b/.changeset/console-wrap-nested-const-scope.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Resolve a dev console argument against the generated program's own `const` declarations. The script text passes only had the component analysis, which carries no binding for a name declared inside a nested function, so `const m = \`…\`; console.log(m)` in a `.svelte.(js|ts)` module was wrapped in `$.log_if_contains_state` even though upstream's `scope.evaluate` resolves `m` to a string and emits the plain call.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/console_dev_ast.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/console_dev_ast.rs
@@ -33,7 +33,9 @@ use oxc_span::{GetSpan, SourceType};
 use crate::compiler::phases::phase2_analyze::ComponentAnalysis;
 
 use super::ast_rewrite::{self, Edit};
-use super::console_wrap::{CONSOLE_METHODS, shape_can_be_unknown};
+use super::console_wrap::{
+    CONSOLE_METHODS, LocalConsts, collect_local_consts, shape_can_be_unknown,
+};
 
 thread_local! {
     static MODULE_CONSOLE_ALLOC: RefCell<Allocator> = RefCell::new(Allocator::default());
@@ -115,6 +117,7 @@ pub(super) fn collect_console_edits(
     let mut collector = ConsoleCollector {
         source,
         analysis,
+        locals: collect_local_consts(program, analysis),
         replacements: Vec::new(),
     };
     collector.visit_program(program);
@@ -124,6 +127,7 @@ pub(super) fn collect_console_edits(
 struct ConsoleCollector<'src> {
     source: &'src str,
     analysis: Option<&'src ComponentAnalysis>,
+    locals: LocalConsts,
     replacements: Vec<Edit>,
 }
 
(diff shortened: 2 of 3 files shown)


baseballyama:
## Rebased onto `origin/main` + gates now actually run

Earlier I had to report the gate as **unrun** (the run aborted with empty output — not a test failure, but equally not a pass). It has now been run for real, on the tree rebased onto `origin/main` at `00cbfba7` (picking up #2320 / #2331 on top of #2323 / #2325 / #2328 / #2329). Rebase was conflict-free; force-pushed as `d706a6d5`.

### Gates — all green

| Gate | Result |
|---|---|
| `cargo test --release -p rsvelte_core --test runtime` | ✅ 5 passed, 0 failed (136.6s) — covers `test_runtime_legacy`, `test_runtime_runes`, `test_hydration`, `test_runtime_browser` |
| `cargo test --release -p rsvelte_core --test ssr` | ✅ 2 passed, 0 failed |
| `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings -A clippy::collapsible_match` | ✅ clean |
| `cargo test --release -p rsvelte_core --lib console` | ✅ 21 passed, 0 failed — includes this PR's 3 new cases |

Fixtures were regenerated from scratch first (`pnpm run generate-fixtures`), so nothing here is reading a stale tree.

### Corpus ratchets — re-measured on the rebased tree

Full `corpus:collect` (14,028 entries) → `corpus:compile` (all 3 targets) → `corpus:verify`:

```
match            13060
error-parity       948
js-mismatch         20
js-unparseable       0
css-mismatch         0
error-mismatch       0

✅ no regressions (client 0, server 0, client-dev 20 known failures remain)
```

All three baselines are **unchanged**: `known-failures.client.json` 0, `known-failures.server.json` 0, `known-failures.client-dev.json` 20. Because the ratchet is two-sided, this is also a positive result in the other direction — verify reports no listed entry that already passes, so the baselines are not stale either. Working tree is clean; no re-baseline is needed in this PR.

Worth stating explicitly so it isn't read as a miss: **#2306 was never a corpus ratchet entry** (it does not appear in `known-failures.md`, and none of the 20 remaining `client-dev` entries is it). So this fix is not expected to shrink any baseline — its coverage is the three new unit tests, not a corpus delta.

Artifact trees (`compatibility/{sources,actual,expected}`, ~570 MB) were dropped immediately after measuring.

PR #2335, opened 2026-08-06
chore(corpus): clean up corpus artifacts by default and guard against a full disk

baseballyama:
Corpus runs left `compatibility/{sources,actual,expected}` on disk with nothing to delete them — every operator was expected to remember `rm -rf`. That failed twice in one day, the second time hard enough that the tool harness could not write its own output file and an agent lost an uncommitted fix. This makes the producer clean up after itself, fails early when disk is short, and adds a one-command reclaim.

## Dependency graph (established by reading `.github/workflows/corpus-compat.yml`)

Job **`corpus` (Compiler parity)** — one job, sequential steps:

```
collect.mjs        -> compatibility/sources/ , manifest.json
compile.mjs        <- sources/, manifest.json      -> expected/, actual/   (all 3 targets in ONE pass)
verify.mjs         <- expected/, actual/, manifest.json  -> report.json
svelte2tsx-compile <- sources/, manifest.json      -> expected-s2t/, actual-s2t/
svelte2tsx-verify  <- expected-s2t/, actual-s2t/, sources/, manifest.json -> report-s2t.json
css-prune-sweep    <- (synthetic; own ratchet)
cluster.mjs        <- expected/, actual/, report.json     [if: failure()]
upload-artifact    <- report.json, report-s2t.json, manifest.json, cluster.txt  [if: failure()]
```

Job **`fmt-parity`**: `collect.mjs` -> `fmt.mjs` (reads `sources/`) -> `fmt-verify.mjs`.
Jobs **`lint-parity`**, **`svelte-check`**, **`check-e2e`**: independent, touch none of these trees.

Two consequences that shaped the design:

1. **`sources/` is shared across three consumers** (`compile.mjs`, `svelte2tsx-compile.mjs`, `fmt.mjs`) and is read *after* `verify.mjs` in the same CI job. It is therefore **never** deleted by a verify step — only by the explicit `corpus:clean`.
2. **`cluster.mjs` reads `expected/` + `actual/` under `if: failure()`**, which fires on *any* earlier step's failure — including `svelte2tsx-verify` failing after `verify` passed. So a "delete on green" rule alone would still break clustering in CI. Cleanup is skipped unconditionally when `CI` is set.

Note the targets are **not** a per-target sequence: `compile.mjs` writes `client`, `server` and `client-dev` in one pass into the same `expected/<id>/` directory, and `verify.mjs` reads all three. There is no per-target cleanup point.

## What changed

New `scripts/compat-corpus/artifacts.mjs` owns the policy; `clean.mjs` is the reclaim command.

- **`verify.mjs` deletes `expected/` + `actual/` after a passing run.** `svelte2tsx-verify.mjs` does the same for the `-s2t` trees. A **failing** run keeps them — that is exactly when someone diffs `expected/<id>` against `actual/<id>` to attribute a cluster — and prints the reclaim command. `--keep-artifacts` / `CORPUS_KEEP_ARTIFACTS=1` keep unconditionally; `--clean-artifacts` deletes even after a failure; `CI` keeps unconditionally.
- **`compile.mjs` preflights free disk** and exits 3 below `180 MiB x targets + 512 MiB`. It runs after the existing unfiltered wipe of `expected/`+`actual/`, so the figure reflects the space the run actually needs.
- **`pnpm run corpus:clean`** removes every regenerable artifact in this checkout and in every `.claude/worktrees/*` sibling (`--here`, `--dry-run`, `--all`). It works from an explicit allowlist in `artifacts.mjs`, plus a startup assertion that no allowlist entry contains `known-failures` or ends in `.md`, so the ratchets can never be deleted.

## The FALSE-SHRINK guard (the important part)

`verify.mjs` reads a missing output file as `""` on **both** sides, so a verify against an absent tree scores every entry `match` — and `--update-baseline` *deletes* every baseline id it did not observe failing. That path exists on `main` today; deleting trees by default would make it much easier to walk into. Two assertions close it:

- **Coverage assertion**: before any normalization or comparison, `verify.mjs` requires >=99% of manifest entries to have compiled output (`<target>.js` or `error.json`) on at least one side, and exits 2 otherwise with `run: node scripts/compat-corpus/compile.mjs`. The union rather than per-tree, because a crashed worker legitimately leaves only the rsvelte-side `error.json`. This is checked *before* the oxfmt pass so a cleaned tree fails legibly instead of throwing ENOENT out of `flattenTreeTemplateHoles`.
- **Full-corpus floor**: `--update-baseline` and `--from-report` refuse to rewrite a ratchet from fewer than 12000 corpus entries (the corpus is 14025 with every submodule present). `svelte2tsx-verify.mjs --update-baseline` enforces the same floor, which also catches `--filter` + `--update-baseline`.

## What I measured

`du -sk` on a completed 3-target run in the main checkout (14025 manifest entries):

| tree | allocated | apparent | files | dirs |
|---|---|---|---|---|
| `sources/` | 60.5 MiB | 16.0 MiB | | |
| `expected/` | 254.3 MiB | 140.6 MiB | 42460 | 21201 |
| `actual/` | 253.8 MiB | 140.3 MiB | | |
| **total** | **568.6 MiB** | 296.9 MiB | | |

Allocated is ~1.9x apparent: ~42k tiny files plus 21k directories round up to 4 KiB APFS blocks. The disk budget uses the **allocated** figure — 508 MiB of `expected`+`actual` across 3 targets is ~170 MiB/target, rounded to 180 MiB, plus 512 MiB headroom for the in-place oxfmt normalization pass. With 8 worktrees on the machine, `corpus:clean --dry-run` reported 0.38 GiB reclaimable right now (apparent bytes; ~0.7 GiB allocated).

## Verified

End-to-end `collect` -> `compile --targets server` -> `verify --targets server` on the real corpus, then re-ran `verify` against the cleaned tree to confirm the coverage assertion fires instead of a vacuous pass. Details in a comment below.

## Deliberately not changed

- **`sources/` is never auto-deleted.** Three scripts read it, one of them after `verify.mjs` in the same CI job. Only `corpus:clean` removes it.
- **`collect.mjs`'s `MIN_MANIFEST_ENTRIES = 1000` floor stays.** Agents routinely collect with a partial submodule checkout; raising that floor would break legitimate local runs. The 12000 floor is applied only where it matters — the ratchet *rewrite* path.
- **`fmt/`, `lint-sources/` and the check stages are not auto-cleaned** and are behind `corpus:clean --all`: the fmt oracle is expensive to rebuild (prettier over 14k files) and is CI-cached by svelte.dev SHA.
- **No CI workflow edits.** The `CI` env var is set by GitHub Actions for free, so the corpus job's behaviour is byte-for-byte what it was.
- **`compile.mjs`'s existing wipe of `expected/`+`actual/` is untouched**, as is every ratchet file.

No changeset: `.github/workflows/changeset.yml` (line 25) exempts every type other than `fix`/`feat`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,19 @@ maps are segmented too differently to diff (byte, decoded-set and lookup-equalit
 than equal to official's — using official only to calibrate the invariants. See
 [scripts/compat-corpus/README.md](scripts/compat-corpus/README.md).
 
+**Corpus artifacts clean themselves up.** A full run writes ~0.57 GiB of regenerable trees per
+checkout (`sources/` 60 MiB, `expected/` 254 MiB, `actual/` 254 MiB), and N parallel agent
+worktrees each hold a set — this filled the dev machine's disk twice. `verify.mjs` therefore
+deletes `expected/` + `actual/` after a **passing** run (`svelte2tsx-verify.mjs` likewise for the
+`-s2t` trees); a **failing** run keeps them so a divergence can still be diffed, as does CI and as
+does `--keep-artifacts`. `compile.mjs` aborts up front when free disk is below
+`180 MiB × targets + 512 MiB`. `pnpm run corpus:clean` reclaims everything regenerable across
+this checkout and every `.claude/worktrees/*` sibling — never the checked-in `*known-failures*`
+ratchets. Because a verify against an absent tree would score every entry `match`, `verify.mjs`
+asserts ≥99% of manifest entries have compiled output before comparing, and refuses
+`--update-baseline` below 12000 corpus entries (the FALSE-SHRINK trap: `--update-baseline` deletes
+every baseline id it did not measure).
+
 The svelte-check diagnostic-parity gate is the odd one out: its unit is a **type-checked project**,
 not per-file text, so module resolution / workspace layout / the `.d.ts` environment are observable
 there and nowhere else. Layer 1 (`check-verify.mjs`, ratchet `check-known-failures.json`) runs
--- a/package.json
+++ b/package.json
@@ -61,6 +61,8 @@
 		"corpus:compile": "node scripts/compat-corpus/compile.mjs",
 		"corpus:verify": "node scripts/compat-corpus/verify.mjs",
 		"corpus:known-failures-check": "node scripts/compat-corpus/known-failures-md-check.mjs",
+		"corpus:clean": "node scripts/compat-corpus/clean.mjs",
+		"corpus:clean:all": "node scripts/compat-corpus/clean.mjs --all",
 		"corpus:compile:dev": "node scripts/compat-corpus/compile.mjs --targets client-dev",
 		"corpus:verify:dev": "node scripts/compat-corpus/verify.mjs --targets client-dev",
 		"corpus:fmt": "node scripts/compat-corpus/fmt.mjs",
--- a/scripts/compat-corpus/README.md
+++ b/scripts/compat-corpus/README.md
@@ -196,6 +196,52 @@ node scripts/compat-corpus/cluster.mjs              # group failures by diff sig
 node scripts/compat-corpus/cluster.mjs --show 'JS client: E:…'   # list ids in a cluster
 ```
 
+### Disk: who deletes the artifacts
+
+A full run writes ~0.57 GiB of regenerable trees per checkout — measured with
+`du -sk` on a 3-target run: `sources/` 60 MiB, `expected/` 254 MiB, `actual/`
+254 MiB (the apparent bytes are ~40% lower; ~42k tiny files plus 21k directories
+round up to 4 KiB blocks). N parallel agent worktrees each hold their own set,
+which is how the machine runs out of disk. The rules, all implemented in
+`artifacts.mjs`:
+
+- **`verify.mjs` deletes `expected/` + `actual/` after a passing run.** Nothing
+  downstream reads them (`svelte2tsx-compile.mjs` and `fmt.mjs` read `sources/`,
+  which is kept), so the producer cleans up instead of the operator remembering.
+  `svelte2tsx-verify.mjs` does the same for `expected-s2t/` + `actual-s2t/`.
+- **A failing run keeps them** — that is when someone diffs `expected/<id>`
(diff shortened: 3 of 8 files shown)


baseballyama:
## End-to-end verification

Real corpus in an isolated worktree (`submodules/svelte` + `compatibility/pattern-corpus` only, so 5166 of the 14025 entries — the other submodules were not initialised), using a `rsvelte.node` / `ast_equiv_batch` staged from a release build.

| # | command | result |
|---|---|---|
| 1 | `collect.mjs` | `5166 corpus entries -> compatibility/sources` |
| 2 | `compile.mjs --targets server` | `[compile] disk: 40.20 GiB free, ~0.68 GiB needed` then `done in 4.3s` — preflight prints and passes |
| 3 | `verify.mjs --targets server` | `match 4388`, `error-parity 778`, 0 failures; ratchet read correctly (`known-failures.server.json` is empty, so no stale-ratchet noise); then `[verify] removed expected, actual (keep them with --keep-artifacts)` |
| 4 | `verify.mjs --targets server` again, on the cleaned tree | `[verify] only 0/5166 manifest entries have compiled output for server` + `run: node scripts/compat-corpus/compile.mjs`, **exit 2** — the coverage assertion refuses the vacuous pass instead of scoring 5166 × `match` |
| 5 | recompile, then `verify.mjs --targets server --no-fmt --update-baseline` | `refusing to rewrite baselines from 5166 corpus entries (expected >= 12000)` — the FALSE-SHRINK path is closed; `git status` clean afterwards, no ratchet touched |
| 6 | `verify.mjs --targets server --no-fmt --keep-artifacts` (fails, since `--no-fmt` skips normalization) | `artifacts kept for inspection — reclaim with: pnpm run corpus:clean` |
| 7 | `CI=1 verify.mjs --targets server` (green) | no removal message, `compatibility/{expected,actual}` still on disk — this is what the `Cluster failures` step needs |
| 8 | `clean.mjs --here` | removed the 6 artifacts, `git status` still clean |

Step 4 is the one that matters: without the coverage assertion, that run would have printed `match 5166` and a `--update-baseline` on top of it would have emptied every ratchet.

Sizes above are apparent bytes (`clean.mjs` sums `lstat.size`); the disk budget in `artifacts.mjs` uses the `du -sk` allocated figure from a full 3-target run, which is ~1.9× larger.

Issue #2351, opened 2026-08-06
SIGSEGV: a comment containing `}` or `)` inside a `$:` block body crashes the client compiler


baseballyama:
Found by the corpus-seeded mutation fuzz (#2281 Gate 3), seeded from `svelte/packages/svelte/tests/migrate/samples/reactive-statements-reorder-not-deleting-additions/input.svelte`.

The compiler **aborts the process with SIGSEGV**. It does not throw a `CompileError` — the host process dies, so a build tool, editor or language server embedding rsvelte goes down with it.

## Repro

```svelte
<script>
	let bar
	$: {
		/* } c */
		bar = []
	}
</script>
```

```
$ node -e "require('./rsvelte.node').compile(src, { generate: 'client' })"
$ echo $?
139        # 128 + SIGSEGV
```

The official compiler compiles the same input without complaint (337 bytes of output).

## Conditions

| variation | result |
|---|---|
| `/* } c */` in a `$:` block body | **SIGSEGV** |
| `// } c` in a `$:` block body | **SIGSEGV** |
| `/* ) c */` in a `$:` block body | **SIGSEGV** |
| `/* c */` (no `}` / `)`) in a `$:` block body | compiles |
| `/* } c */` in a plain (non-reactive) block | compiles |
| `generate: 'server'` | compiles |

So it needs all three: `generate: 'client'`, a **reactive** (`$:`) block, and a closing delimiter **inside a comment** in its body.

## Class

Same root class as #2253 and #2347: a text-level rewrite locates a terminator by scanning bytes instead of lexing, so a delimiter inside a comment is read as code. #2283 consolidated five such scans behind `shared/js_scan.rs::skip_opaque`; the reactive-statement path is evidently not among them. Here the mis-located terminator does not merely produce wrong output — it crashes.

## Severity

Worse than #2347. That one emits parseable-but-wrong code; this one takes the process down. `compile.mjs` already treats an rsvelte panic as a recordable verdict (`rust_panic`) precisely because this can happen, but the shape is not in the collected corpus — it only appears once a comment is inserted.

## How it was found

The collected corpus is at 0 known failures on all three targets. The fuzz inserts one semantics-preserving comment at a line boundary inside a `<script>` region of a real corpus file and requires parity on the mutant. This one killed the sweep outright, which is why Gate 3 compiles in child processes (mirroring `compile.mjs`): the parent names the crashing seed, records `compiler-crash` as a finding, and resumes.

Issue #2347, opened 2026-08-06
Line comment before the closing brace of a `$props()` pattern swallows the `rest_props` initializer


baseballyama:
Found by the corpus-seeded mutation fuzz (#2281 Gate 3), seeded from `shadcn-svelte/docs/src/lib/registry/ui/input-group/input-group.svelte`.

## Repro

```svelte
<script>
	let {
		class: className,
		...props
	// c
	} = $props();
</script>

<div class={className} {...props}></div>
```

## rsvelte output (`generate: 'client'`)

```js
let props;
var // c = $.rest_props($$props, rest_excludes)
	div = root();
```

The `= $.rest_props($$props, rest_excludes)` initializer lands **inside the line comment**. `props` is declared and never assigned.

## Official output

```js
let props = $.rest_props($$props, rest_excludes);
```

## Why this is worse than it looks

The output **parses**. There is no compiler error, no warning, and no syntax error at load time — so nothing catches it.

At runtime `props` is `undefined`, and `{ class: $$props.class, ...props }` spreads it. `{...undefined}` is legal JavaScript and yields nothing, so **every forwarded attribute silently disappears** with no exception and no console output.

## Conditions

- Only `//` line comments. A `/* */` block comment in the same slot is fine.
- Only when the pattern has a rest element (`...props`). Without one the bug does not appear.
- All three targets (`client`, `client-dev`, `server`) diverge; the swallowed-initializer form is the client one.

## Class

Same family as #2253: a text-level rewrite that decides where to split without lexing, so a comment absorbs what follows it. #2283 consolidated five such scans behind `shared/js_scan.rs::skip_opaque`; this call site was not among them.

## How it was found

The collected corpus is at 0 known failures on all three targets — it has nothing left to teach. This shape is not in it: the fuzz inserts one semantics-preserving comment at a line boundary inside a `<script>` region of a real corpus file, and requires parity on the mutant. 300 sampled seeds produced this, the only mutant in the sample whose divergence survives normalizing away comments, whitespace and trailing commas.

PR #2353, opened 2026-08-06
feat(corpus): fuzz the corpus as a seed set, not a test set (#2281)

### LINKED ISSUE #2281: epic: close the two corpus blind spots — warning parity and generated input shapes
## Why a 14,026-entry byte-parity corpus did not catch any of these

Four compiler bugs were reported from one real application on the same day (#2253, #2254,
#2255, #2256). The compat corpus is at **0 known failures on `client` and `server`** — it is
saturated — and it caught none of them. This issue records the measurements that explain why,
because the explanation splits cleanly in two and the two halves need different fixes.

All numbers below were measured on `main` at the time of writing, against
`svelte/compiler` 5.56.8, with 0 worker aborts.

### Finding 1 — for two of the four bugs, the SHAPE IS NOT IN THE CORPUS

A census over all 14,026 corpus source files:

| bug shape | occurrences in the corpus |
|---|---|
| #2254 — an `{#each … as X}` item used as a `switch` discriminant | **0 / 14,026** |
| #2253 — a `#private` `$state` field assigned in a constructor from a multi-line literal containing a `//` comment | **0 / 14,026** |
| #2256 — `svelte-ignore` immediately before an object-literal property | 6 / 14,026 |

(The census initially flagged one candidate for #2254; on inspection it was a false positive —
a `switch (color)` on a *function parameter* in a file that separately contains
`{#each colors as color}`. The true count is zero.)

This is not a corpus that is too small. It is a corpus of the wrong shape. A **found** corpus
samples the marginal distribution of code people actually published; every one of these bugs is
an **interaction** — a binding kind × a syntactic position, or a construct × a comment slot —
and interactions are under-sampled exponentially. Growing the corpus from 14k to 140k real
files would move these counts from 0 to approximately 0.

### Finding 2 — for the third bug, the shape IS in the corpus and the ORACLE DOES NOT LOOK

`scripts/compat-corpus/compile.mjs::compileOne` returns `{ js, css }` or `{ error }`.
**`result.warnings` is discarded.** Warning divergences are therefore invisible to the gate by
construction, at any corpus size.

Positive control — `layerchart/docs/src/routes/+layout.svelte`, an entry that is in the corpus
right now and carries exactly the #2256 shape:

```
official warnings: (none)
rsvelte  warnings: state_referenced_locally @ 20:10
corpus verdict   : MATCH
```

Sizing the blind spot across all 14,026 entries (`generate: 'client', dev: true`):

| | entries |
|---|---|
| any warning-payload divergence | **699 (5.0%)** |
| — true warning **code-set** divergence (rsvelte warns where official does not, or vice versa) | **73** |
| — rsvelte omits `start`/`end` position metadata on ≥1 warning | **592** |

The 592 are a second, separate parity gap: rsvelte's warning objects carry no position for many
warning kinds (`event_directive_deprecated`, `element_invalid_self_closing_tag`,
`a11y_role_supports_aria_props`, `export_let_unused`, `non_reactive_update`, …), so any editor or
CLI that places a squiggle from `warning.start` gets nothing. Also invisible today.

(#2271 has since taken the code-set number from 73 to 70 and the total from 699 to 696.)

### Finding 3 — generated inputs find in seconds what the found corpus cannot find at all

Two throwaway prototypes, written in an afternoon, both using the official compiler as the
oracle so no notion of "correct output" is needed:

**(a) Shape matrix** — 7 reactive-binding kinds × 47 syntactic positions × 3 targets = 987
comparisons. Runs in seconds. Found **21 divergences in 5 positions**: #2254 as reported, plus
`SwitchCase.test`, class-expression field initializers, class-expression computed method keys —
and two much worse symptoms of the same missing traversal, an each-**index** binding being
dropped from the `$.each` callback signature entirely, and a store auto-subscription never being
created (`$s` undefined → `ReferenceError`). After #2269 the same matrix reports **2**, and both
survivors are an upstream Svelte bug where rsvelte's output is the correct one (see below).

**(b) Comment-position fuzz** — 6 base sources × one comment inserted at each line boundary ×
3 comment kinds = 219 mutants, 657 comparisons. After applying the corpus's own normalization
(oxfmt + blank-line stripping, so formatting-only differences are tolerated exactly as
`verify.mjs` tolerates them; 303 were formatting-only and correctly dropped), **188 comparisons
(29%) still diverge**. This axis contains #2253's family and overlaps the known comment-
preservation cluster.

For contrast: the found corpus needed ~35 submodules, a multi-minute sync, and 14,026 files to
reach 0 remaining `client`/`server` findings. 329 generated cases found 21.

### Upstream bug found along the way

`submodules/svelte/.../3-transform/server/visitors/LabeledStatement.js` returns early for
non-`$` labels **without calling `context.next()`**, so zimmerframe never descends into the
labeled subtree. The client visitor calls `context.next()` at the same guard. Since
`$.derived()` returns a function in `svelte/internal/server`, upstream emits `if (doubled)` —
always truthy — where every other position emits `doubled()`. Store auto-subscriptions inside a
labeled body are mis-emitted the same way. rsvelte is correct here; to be reported upstream.

---

## Proposal

Three gates, ordered by measured value per unit of work.

### Gate 1 — warning parity over the existing corpus  *(closes Finding 2; near-zero marginal cost)*

`compile.mjs` already compiles every entry with both compilers; it just throws the warnings
away. Persist them and diff them.

- `compile.mjs`: write `warnings.json` per (entry, target) on both sides — normalized
  `(code, line, column)`, sorted.
- `verify.mjs`: add a `warning-mismatch` verdict.
- Ratchet `compatibility/warning-known-failures.<target>.json`, shrink-only, per-entry
  justification in a paired `.md`, exactly like the existing three.
- Seed baseline: 70 code-set entries + 592 position entries. Consider two ratchets so the
  position gap (a single systemic cause — warning emission sites not attaching a span) can be
  burned down independently of the semantic gap.

This makes the whole existing corpus a warning-parity gate for free, and #2256 becomes a
regression the gate would have caught.

### Gate 2 — generated shape matrix  *(closes Finding 1; best measured bug-per-line ratio)*

Promote the prototype into `scripts/compat-corpus/matrix/` as declarative axes:

- binding kind × syntactic position (the 987-comparison prototype)
- comment kind × insertion slot (the 657-comparison prototype)
- later: rune × script context (instance / `<script module>` / `.svelte.js` / legacy),
  attribute and directive kinds, control-flow block nesting, `{#snippet}` / `{@render}` sites

Fully generated — no submodules, no corpus sync, seconds to run, so it can be a required check
on every PR rather than a nightly. Ratcheted the same way.

### Gate 3 — mutation fuzzing seeded from the corpus  *(the general form)*

Stop treating the 14,026 corpus files as the test set and start treating them as a **seed set**:
apply small semantics-preserving mutations (insert a comment at each token boundary; wrap an
expression in parens; move a statement into a block / `switch` / labeled block; rename a
binding) and require parity on every mutant. Budgeted sampling on PRs, full sweep nightly.

Gate 2 is Gate 3 with hand-picked seeds; build 2 first, then generalize.

### What this changes about how we think about the corpus

`client` and `server` sitting at 0 known failures does not mean the compiler is correct — it
means this input distribution has nothing left to teach us. Corpus **size** is no longer the
axis worth growing. The two axes that are: **what we compare** (Gate 1) and **how inputs are
constructed** (Gates 2 and 3).

baseballyama:
Closes #2281 (Gate 3).

The collected corpus is at 0 known failures on all three targets — saturated. That does not
mean the compiler is correct; it means this input distribution has nothing left to teach. So
the 14,131 entries stop being the test set and become a **seed set**: insert one
semantics-preserving comment at a line boundary inside a `<script>` region and require parity
on the mutant.

## What is gated

A divergent mutant is classified by whether the difference survives normalizing comments,
whitespace and trailing commas away:

| verdict | ratcheted | meaning |
|---|---|---|
| `code-mismatch` | yes | the generated **code** changed because a comment moved |
| `compiler-crash` | yes | rsvelte aborted the process on the mutant |
| `error-mismatch` | yes | exactly one compiler rejected the mutant |
| `comment-mismatch` | no | comment dropped/duplicated/relocated, or a line broke differently |

Comment fidelity diverges on ~34% of mutants. Ratcheting it per id here would mean a
13,000-entry file that churns on every submodule bump and buries the class that matters; it is
already ratcheted per id by Gate 2 on **generated** seeds that do not move when a submodule
bumps.

---

## Provenance of the numbers in this PR

The branch was 13 commits behind `main`, including #2249 (oxfmt 0.61 → 0.62). The code/comment
split is *defined* by what the normalizer absorbs, so the original baseline had been computed
against a normalizer that no longer exists in the tree. Everything below is split by who
measured it.

### Re-measured on this branch, after rebasing onto `main`

Full sweep, oxfmt **0.62.0**: 14,129 seeds → 12,116 mutants → 36,348 comparisons.

| verdict | before (oxfmt 0.61) | now (oxfmt 0.62) |
|---|---|---|
| `code-mismatch` (gated) | 213 | **525** |
| `comment-mismatch` | 13,242 | 13,758 |
| `compiler-crash` | 1 | **0** |
| `error-mismatch` | 0 | 0 |

Entry-level: 137 kept, 77 dropped, 388 added.

**The gated bucket grew.** The expectation going in was the opposite — 0.62 normalizes strictly
more, so entries should have moved *out* of the code bucket. That expectation was wrong, and it
is worth stating plainly rather than restating.

I attributed the growth instead of assuming it. Three things changed at once (oxfmt version, 13
commits of `main`, corpus 14,027 → 14,131), so I held the compiler and the corpus fixed and
varied only the normalizer, over the 193 seeds behind the added entries:

| | oxfmt 0.61.0 | oxfmt 0.62.0 |
|---|---|---|
| `match` | 331 | 45 |
| `comment-mismatch` | 244 | 146 |
| `code-mismatch` | 4 | 388 |

Same seeds, same 193 mutants, same 579 comparisons. **384 of the 388 additions are the
normalizer** — not `main`, not corpus growth.

The mechanism: 0.61 collapsed redundant parentheses on *both* sides, which made a real rsvelte
divergence compare equal. 0.62 preserves them. Sampling the first difference in the string each
verdict is computed from, **353 of 525 (91%)** are a parenthesis — official emits
`() => (items())` where rsvelte emits `() => items()`. On the *unmutated* seed the two agree,
so inserting a comment is what changes rsvelte's paren emission. These are real findings that
0.61 was hiding; the larger ratchet is the more honest one.

Worth carrying forward: a normalizer that absorbs more can *expose* more, because what it stops
rewriting on the expected side is what the actual side is compared against.

### Why a normalizer bump is not contained by the verdict's definition

It is tempting to reason that because the verdict is "does the difference *survive*
normalization", a strictly-more-normalizing version can only move entries **out of**
`code-mismatch` into `comment-mismatch` — so no new gated entry can appear from a version bump
alone, and the gate can only weaken loudly rather than silently.

That reasoning is sound but its premise does not hold here, and the table above is the
counterexample: `code-mismatch` went 4 → 388 on identical seeds from the version change alone.
0.62 is not strictly-more-normalizing. It normalizes *differently* — it **preserves**
parentheses that 0.61 collapsed. In that dimension it does less, and the differences it stops
erasing land in the gated bucket.

So a normalizer bump has to be treated as re-defining the verdicts in both directions and
re-measured, not argued about from the classifier's shape.

What *is* true is that the movement is never silent. The two-sided `--full` check reports
entries that have started passing as **stale** and fails the run: re-deriving this baseline
reported `❌ 77 baseline entries already PASS`. Anyone who bumps the normalizer without
re-baselining gets a loud failure, not a quietly weakened gate — and that failure is expected,
not a broken ratchet. Re-baseline with `pnpm run corpus:mutate:update` and check the
`mutation-known-failures.md` header counts, which `known-failures-md-check.mjs` gates.

Two consequences for the doc:

- `compiler-crash` and `error-mismatch` are now **0**. #2351 and #2347 are both fixed and
  closed, and the sweep reproduces neither. The child-process isolation and the
  `error-mismatch` verdict stay — they are what made those findings attributable.
- The delimiter-vs-plain finding rate falls from **2.81× to 1.30×**. The plain-comment rate
  rose 8.0 → 35.4 per 1,000 while the delimiter rate only doubled, so the doc no longer claims
  the delimiter *is* the mechanism — it is one mechanism, and no longer the majority.

### Inherited from the original sweep, not re-measured

- "Trailing-comma normalization took the code class from 45 apparent findings to 2 real ones"
  — a 300-seed sample under 0.61.
- "Quote-style normalization reclassifies 0 of 213 entries" — measured under 0.61; the doc now
  says so rather than presenting it as current.
- The filename-suffix finding (9 spurious `error-mismatch` entries when the mutant tag was
  appended after the extension).

---

## Instrument fixes

The gate has to prove it did the work, not infer it from the absence of a failure. A sweep that
generates nothing prints "no divergences found", which reads as a pass.

- **Mutant and comparison counts are printed** (`mutants 12116, comparisons 36348`), and a run
  is rejected if `comparisons === 0` or fewer than 50% of seeds yielded a mutant.
- **Shard-count equality**: a worker that dies without writing its shard would silently remove
  its whole range from every count. `shardFiles.length !== ranges.length` now exits 2.
- **Pair-count reconciliation**: the number of written artifact pairs must equal the divergent
  count, and a missing `actual` exits 2.
- **`compileOne` returns an explicit error** instead of `{ js: '' }`, so "compiler returned no
  code" cannot compare equal to another empty output.
- **The oxfmt version is printed** next to the counts, and a non-runnable oxfmt exits 2 rather
  than silently producing incomparable verdicts. This PR is the reason: the version *is* part
  of the measurement.
- **`OXFMT_BIN` is honoured**, like the sibling fmt gates. Without it the A/B above could not
  have been run at all.
- **The per-kind finding-rate table is emitted by the gate**, so its headline claim comes out of
  the run instead of being recomputed by hand into the paired `.md` — where it had gone stale.

## Ratchet key

Mutant ids now carry the mutant **index** (`__m0__line-with-paren`) rather than the slot's line
(`__L41__…`). `n` and the comment kind derive from the seed id alone; the line does not, so an
edit anywhere in a seed file rewrote every entry for that file — surfacing as a regression and
a staleness simultaneously, for a divergence that had not changed.


baseballyama:
Reviewed for landability, ratchet soundness, and the instrument rule. **One correction to the premise, one ratchet finding, and four instrument gaps — the last with a fix pushed to `test/2353-fuzz-instrument` (`b7061746`), branched from this PR's head.**

## 1. The Changeset check is not failing

`gh pr checks 2353` reports `Changeset pass 7s`, and the PR is `MERGEABLE` / `mergeStateStatus: CLEAN` with every check green. The `CANCELLED` entry in the rollup is a superseded run: `31073449497` cancelled at 05:12:39, `31073465480` succeeded 19 s later at 05:12:58 — ordinary concurrency-group cancellation. **Nothing to fix here, and no commit-type change is needed.** `feat(corpus)` is fine as-is.

## 2. Ratchet split: sound, but the ids are more fragile than the design intends

The split itself is right, and for the reason claimed. The 13,242 comment-fidelity divergences are **not gated here at all** — only printed — so they cannot produce false failures in this gate regardless of what they do on a submodule bump. And Gate 2, which does ratchet that class per id, runs on **generated** inputs from `matrix/axes.mjs`, which by construction do not move when a submodule bumps. So the "does the Gate 2 half churn?" risk does not materialise.

**But the 213 entries that *are* ratcheted here churn harder than a normal corpus ratchet.** The baseline id embeds the insertion line:

```
attractions/attractions/file-input/paperclip__L8__svelte-ignore.svelte [code-mismatch] (server)
```

`slot = slots[h % slots.length]` where `h = fnv1a("<id>#<n>")`. The hash is content-independent — that is the stated design property #2 — but **`slots.length` is not**, and neither is `slot.line`. So editing a seed file *anywhere* can change both which slot is chosen and the line it lands on, and **every mutant id for that file changes**. On the ratchet's two-sided check that surfaces as a simultaneous regression (new id) and staleness (old id) for the same underlying divergence.

This is strictly worse than `known-failures.{client,server}.json`, which key on the file id alone and are unaffected by edits within a file. Design property #2 protects against *adding or removing corpus entries*; it does not protect against *editing* one.

Not blocking — the repo's convention is to re-baseline in the same PR, and #2281's other ratchets have the same manual step. But it is worth either recording in `mutation-known-failures.md` as a known maintenance cost, or removing `__L<line>__` from the id (the kind name plus the seed id is already unique per `--per-file 1`, and the line number is recoverable from the artifact tree when a failure is being investigated).

## 3. Instrument rule: four paths reached "success" without work

All four end at the same line — `[mutate] ✅ no mutant changes generated code`, then `finish(0)`:

| # | path | why it reported success |
|---|---|---|
| A | `insertionSlots()` returns `[]` for every seed | each seed hits `mutants.length === 0` → `seed-skipped` → `continue`; all counts 0, `ids.size` 0 |
| B | anything throws during generation | `catch { mutants = [] }` swallows it into the same skip, with no diagnostic |
| C | a worker dies without writing its shard | `readdirSync(SHARDS)` just yields fewer files; that range vanishes from every count |
| D | a compiler returns no `js.code` | `result.js?.code ?? ''` makes both sides `''`, and `expected === actual` scores `match` |

D is the shape of #2335 exactly — a missing output read as `""` on both sides so everything scores as agreement. There was also a fifth, milder one: the pair walk did `if (!fs.existsSync(actualPath)) continue`, silently dropping a comparison whose rsvelte half went missing.

### Fix (`b7061746`)

- tally `mutants` and `comparisons` per shard, merge them, and print `[mutate] mutants N, comparisons M, raw-divergent K`
- `exit 2` unless at least half the seeds produced a mutant **and** `comparisons > 0`
- `exit 2` unless the number of shard result files equals the number of workers
- count generation exceptions as `seed-error` separately from `seed-skipped`, and report them
- treat a missing/empty `js.code` as an oracle error, not an empty program
- `exit 2` when the pair count on disk disagrees with `divergent`, or an `actual` half is missing

### Verified three ways

Ran against a two-seed local fixture (the real corpus is not staged in this checkout):

1. **unbroken** — `mutants 2, comparisons 6, raw-divergent 0`; guards silent; the run proceeds to the ratchet exactly as before (it then correctly reports 214 stale entries, since a 2-seed manifest measures none of them — which incidentally confirms the two-sided check works)
2. **`slots.length = 0` injected** — `mutants 0, comparisons 0`, then `only 0/2 seeds produced a mutant (0 comparisons). The insertion-slot scanner is finding nothing — this run measured nothing.`, non-zero exit
3. **break reverted** — back to `mutants 2, comparisons 6`

The `mutants × targets == comparisons` identity (2 × 3 = 6) is printed every run, so the instrument is checkable from the log rather than trusted.

## Note on what I could not verify

The two-seed fixture exercises the plumbing, not the measurement. I did **not** re-run the full sweep, so I am not confirming or disputing the 213 / 13,242 / 22.4-vs-8.0 numbers — those remain as measured by the PR author. The borrowed NAPI binding I used is from another checkout and is load-bearing only for "does the script run", never for a divergence count.

baseballyama:
Follow-up on the oxfmt 0.61 → 0.62 correction. **The exposure here is stronger than a stale local install, and the ratchet's own design contains it.** Version reporting added in `ea98a164`.

## This branch pins 0.61 by declaration, not by accident

```
branch base:            3ecdfdae   (predates #2249 / 8c7851ca, the 0.62 bump)
package.json on branch: "oxfmt": "^0.61.0"
main today:             "oxfmt": "^0.62.0"   (installed binary: 0.62.0)
```

So the 213 / 13,242 split was not measured on 0.61 because an install had gone stale — **the branch asks for 0.61**. Anyone re-running the sweep on this branch as it stands reproduces the 0.61 numbers, which is worse than a one-off environment slip because it is reproducible and therefore looks trustworthy.

## The direction is favourable, and the two-sided ratchet catches it

0.62 normalizes strictly more (optional-chain parens, `;;`, paren multiplication). Since the verdict is "does the difference **survive** normalization", stronger normalization can only move entries **out of** `code-mismatch` and into `comment-mismatch`. So on rebase:

- no new gated entries can appear from the version change alone — the gate cannot silently weaken in the dangerous direction;
- some of the 213 will start passing, which the `--full` two-sided check reports as **stale baseline entries** and fails on.

That is the design working: the split cannot rot quietly, it breaks loudly on the next full run. **No blocking change is needed for correctness — but the re-baseline must happen in the rebase commit**, per the repo's ratchet protocol, or the first nightly after the rebase fails on staleness rather than on a regression.

## What I did not do

I did **not** re-derive the split. It needs the full corpus staged (~0.57 GiB, which I was asked not to stage) plus a complete sweep of 14,027 seeds. The exact procedure, for whoever rebases:

```
git rebase origin/main            # picks up ^0.62.0
pnpm install                      # resolves oxfmt 0.62.x
node scripts/compat-corpus/collect.mjs
pnpm run corpus:mutate:update     # --full --update-baseline
```

and the header counts in `mutation-known-failures.md` need updating with it, since `known-failures-md-check.mjs` gates them.

## Instrument change (`ea98a164`)

Per the generalised rule — an instrument must report *what it did*, and for a versioned oracle that includes which version:

```
[mutate] oxfmt 0.62.0…
```

printed next to `[mutate] mutants N, comparisons M, raw-divergent K`. And oxfmt failing to run is now `exit 2` rather than a silent proceed: normalization *defines* the verdicts, so a run without it is not comparable, and `--no-fmt` remains the explicit opt-out.

Verified the extraction against the real binary (`Version: 0.62.0` → `"0.62.0"`). I could **not** exercise the line end-to-end in this checkout: the block is behind `fs.existsSync(TREE)`, the tree is only created by a divergent comparison, and my two-seed fixture produces none — so that one line is verified as an expression and by syntax check, not by a full run.

## My own earlier measurements are unaffected

Checked rather than assumed, because "probably fine" is how the 0.61 run happened:

| measurement | normalizer involved? |
|---|---|
| CRLF differential (#2377) | **no** — raw `diff` of compiler output, oxfmt never invoked |
| axis-B 12-guard mutation sweep | **no** — `cargo test -p rsvelte_core --lib` |
| this PR's plumbing verification | **no** — run with `--no-fmt`, which short-circuits at line 366 |

Nothing of mine needs re-running. The `MATCH under weaker normalization implies MATCH under stronger` asymmetry does not even need invoking, since none of the three passed through a normalizer at all.

baseballyama:
## Review — the gate is real, and the classifier is measuring the wrong relation

Recommendation: **merge after one specific fix.** The mechanism, the CI wiring and the failure
behaviour all check out — I verified each rather than reading them. What does not check out is the
`code-mismatch` verdict itself: it is not the equivalence relation the corpus gate uses, and the
consequence is that **76% of the 525 ratcheted entries are non-divergences, while a severe class
that the corpus gate would report loudly is filed inside them as ordinary noise.**

Everything below was measured. Environment: `.corpus-cache/rsvelte.node` and
`target/release/ast_equiv_batch` both built 2026-08-06 11:34; mini corpus = the **278 seed files
behind this PR's own ratchet**, reconstructed from the submodules, so the cross-tabs below cover
the ratchet population exactly. That binding reproduced **511 of your 525 entries with 0
regressions**, which is what cross-validates it against your baseline.

---

### 1. Is the gate wired, or only written? — **wired, and it did work**

Traced workflow → step → assertion, and then read the job log rather than the YAML:

```
.github/workflows/corpus-compat.yml  job `corpus` ("Compiler parity")
  → step "Mutation fuzz parity"  → node scripts/compat-corpus/mutate-corpus.mjs --seeds 1500
```

Job `92615642501` on head `a3261aa2`:

```
[mutate] manifest 14130, already-diverging 1, eligible 14129
[mutate] mutants 1291, comparisons 3873, raw-divergent 1588
[mutate] oxfmt 0.62.0…
[mutate] ✅ no code regressions (58 known remain …)
```

Non-zero mutants and comparisons, the version printed, the ratchet consulted. The
`comparisons === 0` / `<50% seeds mutated` / shard-count / pair-count guards are the right guards
and they are the reason that log is evidence instead of decoration. This is the part of the PR I
have no criticism of.

### 2. Can it fail? — **yes, and it names the right thing**

N4 control, one input and two opposite results, no compiler patching:

| arm | command | result |
|---|---|---|
| baseline intact | `mutate-corpus.mjs --seeds 278` | `✅ no code regressions (511 known remain)`, exit 0 |
| two entries deleted from the baseline | same command | `❌ 2 NEW divergences`, exit **1** |

and the red arm printed the actual cause, not a line number:

```
- attractions/.../checkbox-group__m0__line-with-paren.svelte [code-mismatch] (client)
    official: …$.legacy_pre_effect(()=>($.deep_read_state(items())),()=>{…
    rsvelte : …$.legacy_pre_effect(()=>$.deep_read_state(items()),()=>{…
```

The `codeDiffWindow`-over-`codeIdentity` choice earns its keep here: a line diff on that pair leads
with a quote style.

### 3. The blocking finding — `codeIdentity` is not the relation the corpus gate uses

`verify.mjs` has not been a byte gate since #2184. Bytes differ ⇒ it asks
`crates/rsvelte_ast_equiv` (`verify.mjs:279-316`, verdicts consumed at `:350-366`), and
`compatibility/ast-equivalence.md` lists under **"Formatting — collapses"**: whitespace, quote
style, optional semicolons, **optional parentheses**, trailing commas, prose comments.

`codeIdentity` (`mutate-corpus.mjs:413`) collapses comments, whitespace and trailing commas —
**not parentheses.** So the two gates disagree about what "the code changed" means, and the module
header's "must match verify.mjs exactly" is no longer true.

I ran the project's own comparator over all 729 divergent pairs from the 278-seed run:

| Gate 3 verdict | `ast_equiv` (what `verify.mjs` uses) | count | share of gated |
|---|---|---|---|
| `code-mismatch` | **`equivalent`** | **387** | **75.7 %** |
| `code-mismatch` | `unparseable` (rsvelte side) | 78 | 15.3 % |
| `code-mismatch` | `code-differs` | 46 | 9.0 % |
| `comment-mismatch` | `equivalent` | 216 | — |

387 + 78 + 46 = 511, the full reproduced ratchet.

**387 of 511 gated entries are `equivalent` under the comparator this repo uses to decide whether
output diverged.** That is your own "91% are a parenthesis" measurement, cross-checked from the
other side: parentheses are collapsed by `ast_equiv` *by contract*. Which also re-reads the
oxfmt 0.61 → 0.62 story: 0.61 was collapsing redundant parens, i.e. doing by accident what
`ast_equiv` does by design. "These are real findings that 0.61 was hiding" is not supported — by
the repo's own definition they are not findings, and the normalizer-version sensitivity that made
the bucket jump 213 → 525 disappears entirely once the comparator, not oxfmt, decides.

Why this is blocking rather than cosmetic: **a ratchet entry suppresses everything about its key.**
The key is `<id> [code-mismatch] (<target>)`. `checkbox-group__m0__line-with-paren [code-mismatch]
(client)` is listed today for a redundant paren; if rsvelte later drops a `$.get()` on that same
mutant, the key still matches and the gate stays green. 387 noise entries are 387 blind spots on
exactly the mutants that have already proven able to break codegen.

**Fix:** classify with `ast_equiv_batch` instead of `codeIdentity`. The binary is already built in
this same job (step "Build rsvelte NAPI binding + AST comparator"), it already emits the three
verdicts you need, and it is `--comments`-configurable. Expected effect: 525 → ~120 entries, all of
which mean something, and immunity to the next oxfmt bump.

### 4. What is hiding in that bucket — **rsvelte emits code that does not parse, 78 times**

The 78 `unparseable` rows are all `side: right` — **rsvelte's** output. Confirmed independently
with Node's own parser, not just the project comparator: **78/78 rejected by `node --check`**.
48 distinct components, all three targets, `.svelte` and `.svelte.ts`, six of the eight comment
kinds — including three from the bare `// c` with no delimiter at all.

```js
// bits-ui/…/accordion__m0__line-with-paren.svelte.ts, rsvelte, server
#props = $.derived(() => ({
  id: this.opts.id.current,
  // ) c
  "data-orientation": this.opts.orientation.current,
  …
});          // ← the arrow's closing `)` is gone. SyntaxError: missing ) after argument list
```

```
pattern/issues/2139-legacy-nested-destructure__m0__line.svelte  (client, client-dev)
  SyntaxError: Unexpected token 'function'      ← from a plain `// c`
```

That last one is this repo's own pattern fixture for #2139, which 5a18d1af closed.

`verify.mjs` gives this its own verdict (`js-unparseable`, `:354`) precisely because "the output is
not JavaScript" is a different and worse statement than "the output differs". This PR files all 78
under `code-mismatch` with the blanket justification, and the PR body — which reports the bucket as
"the generated **code** changed because a comment moved" and attributes 91% of it to parentheses —
never mentions that a sixth of it is invalid output. The 91% figure is computed by
`codeDiffWindow`, which cannot see this: it reports the first differing character, and for a
truncated program that is still a paren. Worth adding to the doc as its own lesson — it is the
same shape as the warning-parity gap: *ask what the oracle does not look at.*

Please split `output-unparseable` out as its own verdict with its own ratchet (or no ratchet), and
say in the `.md` how many there are.

### 5. Is `comment-mismatch` genuinely covered by Gate 2? — **yes, with one caveat worth writing down**

Checked against the code, not the sentence. `matrix/run.mjs:166-172` compares normalized text with
`expected === actual` and files any difference as `js-mismatch`, so comment-only differences **do**
fail there and **are** ratcheted per id: `matrix-known-failures.json` holds **348** `comment-slot`
entries out of 350. The claim is true.

The caveat is the axis. Gate 2's comment population is **6 hand-picked seeds → 336 cases**
(`COMMENT_SEEDS`: class-private-state, class-static-block, legacy-reactive, snippet-render,
await-block, module-script). So "the class is watched" is true; "this population is watched" is
not. 13,758 corpus comment divergences sit behind 6 seeds.

That said, the hole is narrower than it first looks and you should say so in the doc, because it is
the strongest available defence of the decision: `verify.mjs` invokes `ast_equiv_batch` **with no
`--comments` flag** (`verify.mjs:310`), i.e. `CommentPolicy::Ignore`. Comment fidelity is not gated
on the *unmutated* corpus either. Not ratcheting it on mutants is consistent with that, and the
right framing is "Gate 1 does not gate comments at all yet (ast-equivalence.md, 'Known gap')"
rather than "Gate 2 covers it".

One cheap improvement while it stays ungated: assert the aggregate `comment-mismatch` count against
a ceiling on `--full`. Right now nothing at all constrains 34% of every mutant, not even its size.

### 6. Does the mutation preserve semantics? — **yes on 278/278 real seeds; no inside multi-line template literals**

Oracle = the official compiler: compile seed and mutant with official only, compare the two
official outputs with `ast_equiv`. If the comment were not inert, official's own output would move.

```
seeds 278, no slot 0, official rejects both 0, official rejects ONLY the mutant 0
official(seed) vs official(mutant): { equivalent: 278 }
```

ASI is safe structurally too, and I checked it rather than assuming: `insertionSlots`
(`matrix/mutate.mjs:41-54`) only yields **line-start** offsets and the insertion keeps the original
newline, so the line-terminator structure is unchanged — 0/48 mutants moved meaning on a
`return`-newline probe.

The exception, from targeted probes over all slots × all 8 kinds:

| probe | mutants | official rejects | **meaning changed** |
|---|---|---|---|
| multi-line template literal | 40 | 0 | **16** |
| string line-continuation | 32 | 8 | 0 |
| nested block comment | 40 | 6 | 0 |
| ASI after `return` | 48 | 0 | 0 |
| regex literal | 32 | 0 | 0 |

Inside a multi-line template literal the "comment" becomes **string content**, and the compiler
accepts it silently — 16 of 40. This does not manufacture a false `code-mismatch` (both compilers
receive the same bytes, so a divergence is still a real divergence), but it contradicts the word
*semantics-preserving*, and it contradicts `mutate.mjs`'s own stated exclusion: markup is excluded
because "`// x` is literal text there … it would only dilute the signal" — a template literal
inside `<script>` is the same situation and is not excluded. The line-continuation and nested-block
rows are a narrow noise channel into `error-mismatch` (currently 0, so invisible today).
`normalize.mjs` already has the exact lexical state machine needed to skip slots whose state is
`template` / `squote` / `dquote` / `block-comment`.

Incidence on real seeds was 0/278, so this is a correctness-of-claim issue, not an urgent one.

### 7. A worker that fails before its first seed makes the gate spin forever

`runRange` (`:285-313`) resumes at `last + 1`, and `last` starts at `start - 1`. If a worker dies
**before** printing its first `IDX`, `last + 1 === start` and it respawns the identical range.
Reproduced by removing `submodules/svelte`:

```
[mutate] worker aborted (1) on undefined
… 145 respawns in 20 seconds, still going
```

In CI that is the six-hour job timeout instead of a failure. Two small fixes: exit 2 when
`last < start` (the worker never started, so there is no seed to blame — note it currently blames
`undefined`), and add the `OFFICIAL` existence check that `matrix/run.mjs:70-75` already has. The
parent checks the NAPI binding but not the compiler it imports.

### 8. The headline rate is inflated up to 3×

`tally.byKind` counts **mutants** (incremented once per mutant, `:226`); `hits` counts **failures**,
which are per (mutant × target) — so `code findings per 1,000 mutants` (`:502`) divides a
3-targets numerator by a 1-target denominator. My run shows it plainly:

```
block   61 / 29   2103.4      ← 2103 findings per 1000 mutants
```

The delimiter-vs-plain **ratio** survives (both sides are inflated alike), but the per-1,000 figures
in the log and in the `.md` do not. `mutantsByKind[k] * TARGETS.length`.

### 9. Documentation

- **`AGENTS.md` was not re-derived.** The new section still says **213** / **13,242** / **2.81×** /
  **14,027** — the pre-rebase numbers this PR's own body retracts (525 / 13,758 / 1.30× / 14,131).
  It also still asserts the delimiter *is* the mechanism, which the body explicitly withdraws.
  `known-failures-md-check.mjs` gates `mutation-known-failures.md` and cannot see `AGENTS.md`,
  which is exactly why it drifted.
- **The premise sentence is off by two.** "0 known failures on all three targets" — `client` is 0,
  but `server` and `client-dev` each hold 1 (`runed/.../watch.test.svelte.ts`,
  `runed/.../scroll-state.svelte`), which is also why the gate prints `already-diverging 1`. The
  saturation argument is unaffected; the number should still be right.

### 10. Prior review

The only existing review is `copilot-pull-request-reviewer[bot]`, which says it could not review the
PR (quota). There are no substantive points outstanding from it.

---

### Summary

Keep the gate — it earns its place. In one 278-seed run it produced **46 genuine `code-differs`**
including a dropped `$.get()` around a state read (`layercake/.../Beeswarm`) and
`(a?.b?.[k])?.(e)` → `a?.b?.[k]?.(e)`, which changes optional-chaining short-circuit semantics
(`ast-equivalence.md` names that exact pair as meaning-changing) — **plus 78 outputs that are not
valid JavaScript.** That is a real bug harvest from a class the collected corpus provably cannot
reach.

**Blocking:** swap `codeIdentity` for `ast_equiv_batch` and re-derive the baseline. It is the same
binary the job already builds, it removes 76% of the ratchet as non-divergences, it removes the
387 suppression keys those entries create, it makes the "same relation as `verify.mjs`" claim true,
and it makes the ratchet immune to the next normalizer bump instead of tripling on it.

**Fix alongside:** split `output-unparseable` out (§4) and the `last < start` respawn loop (§7).
**Non-blocking:** the rate denominator (§8), the `AGENTS.md` numbers and the premise count (§9),
the template-literal slots (§6), an aggregate ceiling on `comment-mismatch` (§5).

baseballyama:
## A second classification defect, distinct from the 387/511 one already raised

Filing this because it shares a fix with the parenthesis issue already on this review but is **not the same defect** — and if the two are read as one item, only half will get fixed.

### The two defects

| | already raised | this one |
|---|---|---|
| symptom | **387 of 511** gated entries (75.7%) are `equivalent` under `crates/rsvelte_ast_equiv` | **78** `node --check` rejections are filed as ordinary `code-mismatch` |
| mechanism | `codeIdentity` does not collapse optional parentheses | `unparseable` has no class of its own, so it is folded into `code-mismatch` |
| what goes wrong | **false entries admitted** — non-divergences reported as divergences | **true entries flattened** — a strictly stronger fact demoted to a weaker one |
| effect on the reader | noise inflates the population | signal becomes indistinguishable from the noise above |

They are opposite failures. The first adds entries that should not be there; the second destroys the information that would have made 78 real ones stand out. **Collapsing parentheses alone does not fix the second**: after that change the 78 are still binned as `code-mismatch`, just in a smaller pile.

### Why the second one matters more than its count suggests

"Does not parse" is a **strictly stronger** assertion than "differs", and it is available here for free because `node --check` is an external oracle. The harness currently discards that strength at the moment of classification. The consequence is visible in this PR: the fuzzer **found** 78 mutants across 48 components that make rsvelte emit non-parsing JavaScript, and the PR body does not mention it — not because the evidence was missing, but because the label put it beside formatting noise. That class is now tracked independently as #2434.

The general form: a harness that can prove a strong property and reports a weak one has thrown away the reason it was worth running.

### Shared fix, one pass

Both resolve by classifying with **`ast_equiv_batch`** — the comparator the corpus gate itself has used since #2184 — rather than the hand-rolled `codeIdentity`, plus giving `unparseable` its own verdict class that is **never** folded into `code-mismatch`. Same file, same function; worth doing as one change rather than two passes.

Note `ast_equiv_batch` already models exactly this distinction: unparseable input is its own `Comparison` variant and is deliberately never downgraded to text matching.

### Acceptance should be a discriminating control, not a green run

A harness that merely runs is not evidence it classifies correctly. Suggested:

- take one of the 78 known-unparseable mutants,
- confirm the **current** harness labels it `code-mismatch`,
- confirm the **fixed** harness labels it `unparseable` **and names the component**.

One input, two opposite results. Without the first half the test cannot distinguish "the fix works" from "the input never reached the branch".

### Not touching the branch

I have not pushed anything here and will not — this is a finding, not a patch. Flagging it for whoever picks the implementation up.

baseballyama:
## Fourth commit added after review began — `bf3ae08b`

Flagging explicitly because it changes what is under review.

`mutate-corpus.mjs` is an `--update-baseline` target with no defence against a truncated input tree. `:128`'s `eligible` filter dropped a vanished source **silently**, and `MIN_MUTATED_SEED_RATIO` cannot see that — `seeds` derives from the same filter, so a truncated `sources/` moves the ratio's own denominator and the check passes on whatever survived. That is the #2455 hazard reaching a **ratchet** rather than only a measurement.

The commit reports the missing entries and refuses `--update-baseline` while any exist. The guards beside it all key on **flags** (sampled run, `--no-fmt`, `--targets` subset, crash-orphaned seeds) — flags describe how the run was invoked. This one keys on the **input tree**, which no flag describes and which another process can change mid-run.

Threshold-free on purpose: an absolute floor would need a number calibrated against today's corpus that rots as it grows, and this reports the hazard itself rather than a proxy for it.

**Observed firing** against the real corpus, one source moved aside:

```
[mutate] 1/14131 manifest sources are missing from compatibility/sources:
  - attractions/attractions/accordion/accordion-section.svelte
  the input tree changed under this run — re-run collect.mjs before trusting any number from it.
[mutate] manifest 14131, already-diverging 2, eligible 14128
```

`eligible` falling 14129 -> 14128 is the silent filter this fixes. The detection arm is observed; the **refusal** arm shares the identical predicate and array but is gated behind `--full`, so I have not paid for a full sweep to watch it fire — stated rather than implied.

Also fixed in the same commit: the warning emitted once per worker process, because the script re-executes itself as its own shards. Now parent-only.

Credit to `s2tparity`, who measured this and corrected my earlier claim that the 12000 floor blocked the catastrophic path — true of `verify.mjs`, false here.

PR #2354, opened 2026-08-06
fix(compiler): make the reactive-statement assignment scan lexical-aware

### LINKED ISSUE #2351: SIGSEGV: a comment containing `}` or `)` inside a `$:` block body crashes the client compiler

baseballyama:
Closes #2351

## Diagnosis

The SIGSEGV is **stack exhaustion from unbounded recursion**, not an FFI or memory-safety problem. The crash reproduces in a plain `cargo test` (`thread … has overflowed its stack`), and the macOS crash report names the cycle:

```
extract_destructure_targets   (x N, until the guard page)
transform_reactive_statement
transform_instance_script_for_visitors
transform_client
```

Two defects compose:

1. **What overran.** `client/state_transforms.rs::find_assignment_position` counted `(`/`[`/`{` character by character with no lexical awareness. For the repro body `{ /* } c */ bar = [] }` the `}` **inside the comment** decremented the depth back to 0, so the `=` of `bar = []` was reported as a top-level assignment. `transform_reactive_statement` then took `lhs = "{ /* } c */ bar"` — an unbalanced fragment.

2. **What recursed.** `client/destructure_transforms.rs::extract_destructure_targets` strips the outer brackets only when the string starts *and* ends with a matching pair. The fragment starts with `{` and ends with `r`, so nothing was stripped; `split_on_commas` returned the identical single part; and the `part.starts_with('{')` branch recursed on that same string. No termination condition, no progress — infinite recursion.

That is why all three conditions from the issue are required: `client` (only the client path runs these text transforms), a **reactive block** (only `$:` feeds a `{…}` body into `find_assignment_position`), and a closing delimiter **inside a comment** (without it the depth never returns to 0 and no `=` is reported, so the else-branch handles the block correctly).

## Fix

`find_assignment_position` now steps over comments, strings, template literals and regex literals with the shared `shared::js_scan::skip_opaque` — the same consolidation #2283 applied to five sibling scans; this call site was not among them. It also now returns **byte** offsets: it previously collected `chars()` and returned a char index that every caller sliced with as a byte offset, which was a latent panic on any non-ASCII reactive statement.

`extract_destructure_targets` recurses only into a *closed* bracket pair, so the callee is always strictly shorter than the caller's input. This is the safety net, not the fix — with the scan corrected the malformed fragment is never constructed — but it makes the function structurally unable to diverge on any future caller.

No post-pass string hack; both changes are in the scanners themselves.

## #2347

Not fixed here — different scanner. #2347 is in the `$props()` destructuring / `rest_props` emission path, which does not go through `find_assignment_position`. Left open.

## Test

`crates/rsvelte_core/tests/reactive_block_comment_delimiter_2351.rs` covers all three crashing variants from the issue (`/* } c */`, `// } c`, `/* ) c */`). Every one of them aborts the test binary on `main`.

### DIFF
--- a/.changeset/reactive-comment-brace-segv.md
+++ b/.changeset/reactive-comment-brace-segv.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Fix a stack-overflow crash when a comment containing `}` or `)` appears inside a `$:` reactive block body
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
@@ -566,8 +566,12 @@ pub(super) fn extract_destructure_targets(pattern: &str) -> Vec<String> {
             targets.push(root);
         }
 
-        // Also recurse into nested patterns
-        if part.starts_with('[') || part.starts_with('{') {
+        // Also recurse into nested patterns. Only a *closed* bracket pair is
+        // recursed into: an unbalanced fragment strips nothing, so the callee
+        // would re-derive the identical part and never terminate.
+        if (part.starts_with('[') && part.ends_with(']'))
+            || (part.starts_with('{') && part.ends_with('}'))
+        {
             let nested = extract_destructure_targets(part);
             targets.extend(nested);
         }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/state_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/state_transforms.rs
@@ -13,6 +13,7 @@ use super::rune_transforms::{
 };
 use super::{STATE_TMP_COUNTER, get_or_compile_regex};
 use crate::compiler::phases::phase2_analyze::scope::DeclarationKind;
+use crate::compiler::phases::phase3_transform::shared::js_scan::skip_opaque;
 
 // ---------------------------------------------------------------------------
 // Identifier reference detection (lines 7653-8602 of mod.rs)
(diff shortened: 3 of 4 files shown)


baseballyama:
## Verification

- `cargo test --release -p rsvelte_core --test runtime --test ssr` — green (runtime legacy/runes/browser, hydration, ssr).
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings -A clippy::collapsible_match` — clean.
- `pnpm run corpus:matrix` — 1639 match / 356 js-mismatch, **no regressions** (unchanged).
- Corpus ratchets, all three targets, with oxfmt normalization — **unchanged, no regressions, no stale entries**, so no re-baseline is needed:

| target | js-mismatch before | after |
|---|---|---|
| client | 0 | 0 |
| server | 1 | 1 |
| client-dev | 1 | 1 |

  Warning parity likewise unchanged (71 code / 627 position).

## Note on a *separate* divergence this repro exposes

rsvelte keeps the comment inside the generated `legacy_pre_effect` body where official drops it:

```js
	// } c            <-- rsvelte keeps this, official does not
	$.set(bar, []);
});
```

That is **pre-existing and independent** of this fix — a delimiter-free `/* c */`, which compiles fine on `main`, behaves identically. `strip_reactive_statement_comments` drops comments that trail a `$:` statement but not comments inside a `$:` **block body**. Left out of scope here rather than folded into a crash fix.

Issue #2350, opened 2026-08-06
corpus: verify.mjs silently no-ops when --update-baseline and --update-warning-baseline are passed together


baseballyama:
`scripts/compat-corpus/verify.mjs` accepts both `--update-baseline` and
`--update-warning-baseline`. Passing them together writes **nothing** and exits 0.

```
node scripts/compat-corpus/verify.mjs --update-baseline --update-warning-baseline
...
[verify] results:
  js-mismatch      2
[verify] warning parity:
  warning-code-mismatch      71
  warning-position-mismatch  627
$ echo $?
0
```

All nine ratchet files are untouched, and nothing in the output says so — the run
looks like a successful re-baseline.

## Mechanism

Two interacting guards, both individually deliberate:

- `for (const ratchet of UPDATE_BASELINE ? [] : WARNING_RATCHETS)` (verify.mjs:486)
  skips the whole warning loop when `--update-baseline` is set, so the
  `UPDATE_WARNING_BASELINE` write inside it never runs.
- `if (UPDATE_WARNING_BASELINE) finish(0);` (verify.mjs:510) then exits **before**
  `if (UPDATE_BASELINE) { … writeBaselines(…) }` at line 542.

So each flag disables the other's write, in the one order that produces silence
rather than an error.

## Why it matters

This is the documented re-baseline path for an enrolment PR, which moves output
*and* warning ratchets in the same push. The failure mode is the one the corpus
pipeline exists to prevent: an operator re-baselines, sees exit 0, commits, and
the ratchets still hold the pre-run contents. CI catches it on the next push, but
only after the fact, and the local signal was green.

## Fix

Either reject the combination up front, or make it do both writes. Cheapest
correct version is to let the two writes run independently and only `finish(0)`
after both — the two ratchet families do not overlap, which is the reason the
guard at line 486 exists in the first place.

Found while re-baselining #2321; worked around there by running verify twice.

PR #2361, opened 2026-08-06
fix(compiler): match upstream nested-class depth and span in standalone modules

### LINKED ISSUE #2348: compiler (warnings): perf_avoid_nested_class over-warns at function depth 1 in a standalone .svelte.(js|ts) module
Surfaced by the `runed` / `svelte-toolbelt` corpus enrolment (#2321), on
`runed/packages/runed/src/lib/utilities/on-click-outside/on-click-outside.test.svelte.ts`.

rsvelte emits a `perf_avoid_nested_class` warning that the official compiler does not.

## Repro

```js
// x.svelte.js — compileModule
describe("x", () => {
  class A extends Event {}
});
```

| | official | rsvelte |
|---|---|---|
| `describe(() => { class A {} })` (depth 1) | *no warning* | `perf_avoid_nested_class` |
| `describe(() => { it(() => { class A {} }) })` (depth 2) | `perf_avoid_nested_class` | `perf_avoid_nested_class` |

So the divergence is exactly at **function depth 1 in a standalone module**.

## Root cause

`2-analyze/visitors/ClassDeclaration.js` picks the threshold from `state.ast_type`:

```js
const allowed_depth = context.state.ast_type === 'module' ? 0 : 1;
if (context.state.scope.function_depth > allowed_depth) w.perf_avoid_nested_class(node);
```

For a **standalone** `.svelte.(js|ts)` module, `analyze_module` does not pass
`'module'` — it passes `null` (`2-analyze/index.js:309`, with a `TODO` saying the
field is not needed for modules). So `allowed_depth` is **1** there, not 0, and only
depth ≥ 2 warns. `'module'` is only ever set for a component's `<script module>`
(`index.js:721,794`), where the threshold really is 0.

`crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/class_declaration.rs:28-35`
uses `AstType::Module` for both cases, so the standalone-module path warns one level
too early.

Whether upstream's `null` is intentional or an oversight, parity means matching it —
and the two cases must stay distinguishable, since a component's `<script module>`
does use `allowed_depth = 0`.

## Gate

Baselined in `compatibility/warning-known-failures.{client,server,client-dev}.json`
with justification in `compatibility/warning-known-failures.md`.

### LINKED ISSUE #2349: compiler (warnings): perf_avoid_nested_class is emitted with no span
Surfaced by the `runed` / `svelte-toolbelt` corpus enrolment (#2321), on
`runed/packages/runed/src/lib/utilities/is-document-visible/is-document-visible.test.svelte.ts`.

Both compilers agree on the warning **code**; rsvelte attaches no position.

```
official: perf_avoid_nested_class @ 12:4
rsvelte : perf_avoid_nested_class @ ?:?
```

## Root cause

Upstream passes the node: `w.perf_avoid_nested_class(node)`
(`2-analyze/visitors/ClassDeclaration.js:21`).

rsvelte calls
`context.emit_warning(warnings::perf_avoid_nested_class())`
(`crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/class_declaration.rs:34`)
with no span, so `warning.start` is `undefined` and an editor or CLI has nowhere
to put the squiggle. The `ClassDeclaration` span is available right there.

This is one instance of the systemic cause already documented for the
`warning-position-known-failures.<target>.json` ratchet, and it is the mechanical
kind: attach the span already in hand at the emission site.

## Gate

Baselined in `compatibility/warning-position-known-failures.{client,server,client-dev}.json`
with justification in `compatibility/warning-known-failures.md`.

baseballyama:
Two bugs in the same warning, surfaced by the `runed` / `svelte-toolbelt` corpus enrolment (#2321).

## #2348 — over-warns at function depth 1 in a standalone `.svelte.(js|ts)` module

Upstream picks the threshold from `state.ast_type`:

```js
const allowed_depth = context.state.ast_type === 'module' ? 0 : 1;
```

and `analyze_module` passes **`null`**, not `'module'` (`2-analyze/index.js:309`, with a TODO saying the field is not needed for modules). `'module'` is set only for a component's `<script module>` (:721, :794). rsvelte used `AstType::Module` for both, so `describe(() => { class A {} })` in a `.svelte.js` warned one function level early.

The distinction already exists on the analysis (`ComponentAnalysis::is_module_file`, whose doc comment already records that `ast_type` is null for module files), so the threshold now keys off it and the two cases stay distinguishable.

| | before | after / official |
|---|---|---|
| `.svelte.js`, depth 1 | warns | silent |
| `.svelte.js`, depth 2 | warns | warns |
| `<script module>`, depth 1 | warns | warns |

## #2349 — emitted with no span

Upstream passes the node (`w.perf_avoid_nested_class(node)`); rsvelte called `emit_warning(warnings::perf_avoid_nested_class())` with nothing, so `warning.start` was undefined and an editor had nowhere to put the squiggle. The `ClassDeclaration` span was already in hand at the emission site.

## Tests

Five lib unit tests in `class_declaration.rs` (no new test binary): the three rows above, the instance-script component-scope carve-out, and the span assertion. Verified against upstream's source rather than re-derived — the `ast_type: null` line at index.js:309 is verbatim.

Closes #2348
Closes #2349


baseballyama:
## Warning-ratchet movement (needs a corpus re-baseline before merge)

This changes which warnings are emitted, so two entries flip from FAIL to PASS. Both ratchets are two-sided, so they go **stale** (loud red, not silent) until re-baselined:

| ratchet | entry | why it flips |
|---|---|---|
| `warning-known-failures.<target>.json` (71 each) | `runed/…/on-click-outside/on-click-outside.test.svelte.ts` | #2348 — the over-warning is gone |
| `warning-position-known-failures.<target>.json` (627 each) | `runed/…/is-document-visible/is-document-visible.test.svelte.ts` | #2349 — the warning now carries a span |

Both are named explicitly in `compatibility/warning-known-failures.md` as the single entry each issue contributed, so the expected movement is 71 → 70 and 627 → 626 per target — but the fix could also flip other `.svelte.(js|ts)` entries the enrolment did not attribute, and I could not measure that locally (a full corpus run needs ~35 submodules and ~570 MB of artifacts). I have deliberately **not** hand-edited the ratchets.

Re-baseline with a full run:

```
pnpm run corpus:sync && pnpm run corpus:collect && pnpm run corpus:compile
node scripts/compat-corpus/verify.mjs --update-warning-baseline
```

then update the two justification bullets in `compatibility/warning-known-failures.md`. Note that #2360 (PR #2360) makes `--update-warning-baseline` compose with `--update-baseline`; before it, passing both wrote nothing and exited 0.

baseballyama:
## Warning ratchets re-baselined (from CI's own measurement, not my prediction)

Rebased onto `origin/main` (#2359 landed since; #2354 and #2356 were already in the base).

The `Compiler parity` failure was the predicted staleness, and my earlier caveat was right — CI found **one more entry than I predicted**. Pulled the `corpus-report` artifact from the failing run instead of staging the corpus locally, and derived the ratchets from it:

| ratchet | was | now | ids that now pass |
|---|---|---|---|
| `warning-known-failures.<target>.json` | 71 | **70** | `runed/…/on-click-outside.test.svelte.ts` |
| `warning-position-known-failures.<target>.json` | 627 | **625** | `runed/…/is-document-visible.test.svelte.ts` **+ `svelte/…/validator/samples/inline-new-class-2/input.svelte`** |

× 3 targets = the 9 stale entries the job reported.

The unattributed one is the nice confirmation: `inline-new-class-2` is the upstream validator fixture that asserts `perf_avoid_nested_class` at `3:2`–`5:3`, which is precisely the span #2349 was missing. No issue named it because the corpus reports counts, not per-code attribution — so it could only be found by measuring. I have added that lesson to `warning-known-failures.md` for the next person burning down this list.

Cross-checks, since a derive is only as good as its agreement with the tool:

- derived set sizes (70 / 625) reproduce the job's own `warning-code-mismatch 70` / `warning-position-mismatch 625` counters exactly;
- the diff is 9 pure line deletions — no reordering or reformatting — so the serialization matches what `--update-warning-baseline` writes;
- nothing moved the other way: zero new entries in either warning family;
- output ratchets unchanged and non-stale (`client` 0→0, `server` 1→1, `client-dev` 1→1), so no regression got absorbed.

`known-failures-md-check.mjs` is green.

Note on method: the warning derive from `report.json` is exact rather than lossy. The artifact only records the first differing line per target for *output* diffs, but warning partitioning needs just (id, target, kind), all of which the report carries in full.

One irony worth recording: the correct way to do this in one run is `--update-baseline --update-warning-baseline` together, which on this branch's base writes nothing and exits 0 — that is #2350, fixed by #2360, which has not landed yet. I used the single-flag path.

PR #2364, opened 2026-08-06
fix(compiler): stop contracting plain `$state` assignments in server modules (#2308)

### LINKED ISSUE #2308: codegen (server): `x = x + 1` is contracted to `x += 1` in a .svelte.(js|ts) module

baseballyama:
Fixes #2308.

## What was wrong

On the `server` target, a `.svelte.(js|ts)` module rewrote `runs = runs + 1` into
`runs += 1` whenever `runs` was a `$state` binding. Official leaves it verbatim.

`transform_server_module` runs the **client** module transform and then undoes it in
`post_process_for_server`. The client transform lowers *both* `runs += 1` and
`runs = runs + 1` to the identical text `$.set(runs, $.get(runs) + 1)` — the operator is
destroyed before the server undo pass ever sees it. `recompact_compound_set` therefore had
nothing left to decide from and always guessed "compound".

## The fix

Stop destroying the operator instead of tuning the guess. `$state` / `$state.raw` /
`$state.eager` are now lowered to their bare initializer **before** the client transform
runs. State bindings are then not reactive on this path, no `$.set` / `$.get` is emitted
for them, and no assignment has to be reconstructed.

This reuses the existing scope-aware `rune_call_ast::transform_rune_calls_combined`, the same
locator the component server path already uses, so shadowing (`function g($state) { $state(1) }`)
and rune-shaped text inside string literals are resolved structurally.

`#2317` lowered `$state` **declarations** on this path; this extends it to the **assignment**
path it did not reach.

## Why rewriting *every* call is the upstream algorithm, not over-reach

The pass matches any unshadowed `$state` / `$state.raw` / `$state.eager` call, not only
declarator initialisers. That is deliberate, and it mirrors two upstream visitors, not one:

- `3-transform/server/visitors/VariableDeclaration.js:136-138` — a state rune in a declarator
  emits `b.declarator(declarator.id, value)`, i.e. just the first argument.
- `3-transform/server/visitors/CallExpression.js:39-50` — **wherever else** the call occurs,
  `$state` / `$state.raw` return `node.arguments[0] ?? b.void0` and `$state.eager` returns
  `node.arguments[0]`.

The same file (`:52`) maps `$state.snapshot` to `$.snapshot` rather than stripping it, which is why
`$state.snapshot` is excluded here: that exclusion is upstream behaviour, not a carve-out.

## Tests

Two of the six fail before the fix; four are pins against the pre-pass over-reaching, green
on both sides and labelled as such in the file. This split was measured by re-running the
suite against the unmodified `server/mod.rs`, not reasoned about.

**Repros (fail on `main`)**
- `runs = runs + 1` / `runs = runs - 2` contracted to compound form
- `$state.eager(0)` emitted **verbatim** — a call to an undefined `$state` — in a
  `.svelte.js` module on the server

**Pins (green on both sides)**
- `plain = plain + 1` (non-state binding) and `runs = 1 + runs` (state on the right) unchanged
- `runs += 1` / `runs *= 2` genuinely compound, still preserved
- `$state.snapshot` survives as `$.snapshot`
- a class `$state` field stays a plain public field, not privatized
- `$derived` still wraps, reading the now-plain state binding

## Known gap, pre-existing and untouched

`function g($state) { return $state(1); }` in a `.svelte.js` module still emits `return 1;`
on the server. The new pass resolves that shadowing correctly; the *downstream* client
module transform rewrites it textually afterwards. Same for a rune-shaped string literal.
Both are the string-based module path, out of scope here.

### DIFF
--- a/.changeset/module-server-state-assignment.md
+++ b/.changeset/module-server-state-assignment.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): stop contracting plain `$state` assignments in server modules
--- a/compatibility/known-failures.md
+++ b/compatibility/known-failures.md
@@ -47,13 +47,16 @@ and #2343 (the spurious `$.set` proxy flag for a `BinaryExpression`). #2307
 (spurious `/* @__PURE__ */`) is comment-only, so the AST-structural comparator
 does not see it at all; it burns down with the esrap comment epic (#2336).
 
-## Server (`known-failures.server.json`, 1 entry)
+## Server (`known-failures.server.json`, 0 entries)
 
-The one entry is #2308, from the `runed` / `svelte-toolbelt` enrolment:
-`watch.test.svelte.ts` writes `runs = runs + 1` and rsvelte **contracts** it to
+The last entry was #2308, from the `runed` / `svelte-toolbelt` enrolment:
+`watch.test.svelte.ts` writes `runs = runs + 1` and rsvelte **contracted** it to
 `runs += 1` (that direction, not the reverse). The `.svelte.(js|ts)` server path
-round-trips through the client transform, which rewrites the assignment, and the
-server printer has no way back to the source form.
+round-trips through the client transform, which rewrote the assignment, so the
+operator was already gone before the server printer ran. Fixed by lowering
+`$state` to its bare initializer *before* the client transform, so state
+bindings on this path are never signal-wrapped and nothing has to be
+reconstructed.
 
 Its previous sole entry was the SSR half of the same #2031 divergence (the extra
 `<!---->` the dynamic form pushes), fixed by the same change.
--- a/compatibility/known-failures.server.json
+++ b/compatibility/known-failures.server.json
@@ -1,3 +1,1 @@
-[
-	"runed/packages/runed/src/lib/utilities/watch/watch.test.svelte.ts"
-]
+[]
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/mod.rs
@@ -109,6 +109,17 @@ pub fn transform_server_module(
     // (mirrors the instance-script path + upstream server CallExpression visitor).
     let source_without_effects = source_without_effects.replace("$effect.tracking()", "false");
 
+    // Lower `$state` to its bare initializer BEFORE the client transform, the way
+    // upstream's server `VariableDeclaration` visitor does. Otherwise the client
+    // transform signal-wraps the binding and collapses both `x = x + 1` and
+    // `x += 1` to the same `$.set(x, $.get(x) + 1)`, losing the operator that
+    // `post_process_for_server` would have to restore.
+    let source_without_effects = rune_call_ast::transform_rune_calls_combined(
+        &source_without_effects,
+        &["$state(", "$state.raw(", "$state.eager("],
+    )
+    .unwrap_or(source_without_effects);
+
     // Transform rune calls using the same infrastructure as client modules.
     let transformed =
         super::client::transform_module_source_for_module(&source_without_effects, analysis, false);
--- a/crates/rsvelte_core/tests/module_server_state_assignment_2308.rs
+++ b/crates/rsvelte_core/tests/module_server_state_assignment_2308.rs
@@ -65,6 +65,10 @@ fn plain_assignment_to_state_is_not_contracted() {
     );
 }
 
+/// Regression pin, not a repro: this shape was already green before the fix,
+/// because the client transform's `$.set` folded straight back to a compound
+/// assignment. It only guards against the fix over-reaching in the other
+/// direction.
 #[test]
 fn genuine_compound_assignment_to_state_is_preserved() {
     let src = r#"export function t() {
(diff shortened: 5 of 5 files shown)


baseballyama:
**LGTM with nits** — 0 blocking findings. The mechanism is upstream-faithful (more so than the PR description claims); the gaps are all in what the tests pin.

Read-only review: I did not build or run anything, so everything below is from reading the code at `55a2d94` against `submodules/svelte`. CI was still pending when I looked.

---

### 1. Upstream fidelity — correct, and the code comment undersells it

`server/mod.rs:112-116` says the pre-lowering mirrors "upstream's server `VariableDeclaration` visitor". That visitor only covers a declarator *init* (`get_rune(declarator.init)`), which would make a blanket rewrite of *every* `$state(...)` call an over-reach. It isn't, because the actual counterpart is the server **`CallExpression`** visitor:

```
3-transform/server/visitors/CallExpression.js:39
  if (rune === '$state' || rune === '$state.raw') {
      return node.arguments[0] ? context.visit(node.arguments[0]) : b.void0;
  }
  ...:48
  if (rune === '$state.eager') { return context.visit(node.arguments[0]); }
```

That strips the wrapper wherever the call occurs, which is exactly what `transform_rune_calls_combined` does. Three consequences worth recording:

- `$state()` with no args → `emit_rune_replacement("")` → `void 0`, matching `b.void0`. ✓
- `$state.snapshot` is correctly **absent** from the prefix list — upstream `CallExpression.js:52` emits `$.snapshot(...)`, and `RuneSpec::from_prefix("$state(")` builds `member_name: None`, so `matching_spec` requires an `Expression::Identifier` callee and a `StaticMemberExpression` callee can never match it (`rune_call_ast.rs:150-163`). `raw`/`eager` match on exact property name. ✓
- `AssignmentExpression.js` returns `null` for a plain (non-store, non-derived) state binding, i.e. the assignment is emitted verbatim — which is precisely the invariant #2308 wanted. ✓

Please cite `CallExpression.js` in the comment alongside `VariableDeclaration.js`; the next reader will otherwise conclude the pass over-reaches.

### 2. `genuine_compound_assignment_to_state_is_preserved` does not discriminate

`crates/rsvelte_core/tests/module_server_state_assignment_2308.rs:70-88`. On the **unfixed** code, `runs += 1` lowers to `$.set(runs, $.get(runs) + 1)`, `post_process_for_server` collapses the read to `runs`, and `recompact_compound_set` (`server/mod.rs:376`) folds it back to `runs += 1`. Same for `runs *= 2` via the `("*", "*=")` entry. So both assertions pass against `main`.

That is fine as a *non-regression control* — it is the right thing to pin — but it is not evidence the fix works, and the PR body should say which test carries the load. Only `plain_assignment_to_state_is_not_contracted` does.

Same file, same test, minor: `assert!(!out.contains("$.set(") && !out.contains("$.get("))` (line 60) also cannot fail on the unfixed path, because `post_process_for_server` rewrites every `$.set(`/`$.get(` before the string is returned. Harmless, just not load-bearing.

### 3. No negative controls for the shapes the new pre-pass could eat

The pass now rewrites every unshadowed `$state(`/`$state.raw(`/`$state.eager(` call in a server module. I traced each of the following and believe all four are correct today, but nothing in the PR pins them, and the whole point of a blanket pre-pass is that it can start eating neighbours:

- `let s = $state.snapshot(x)` in a `.svelte.js` — must still reach the `$.snapshot` handling at `server/mod.rs:144-148`. (Safe by the callee-shape argument in §1.)
- `$state.raw(...)` / `$state.eager(...)` at `compile_module` level — both are in the prefix list and neither has a test above the `rune_call_ast` unit level.
- A class with `$state` fields inside a `.svelte.js` module. The new pass runs *after* `transform_class_fields_server` (`server/mod.rs:106`), so it now also strips anything that pass left behind — including on that pass's bail path. Worth one fixture.
- A `$derived` that reads a `$state` in the same module. This is the read side that flips: with the state no longer signal-wrapped, `$derived(n * 2)` becomes `$.derived(() => n * 2)` over a plain `n` instead of routing through `$.get(n)` → `n()`. That is what upstream emits, but it is the interaction most likely to move output, and it is untested here.

### 4. The fallback path keeps the old guess, silently and untested

`server/mod.rs:117-121` uses `.unwrap_or(source_without_effects)`. `transform_rune_calls_combined` returns `None` both when nothing matched *and* when the (already twice-rewritten) text fails to parse as `SourceType::mjs()`. On the parse-failure branch the module goes down the old path where `recompact_compound_set` still guesses "compound" — i.e. #2308 reproduces. That is a reasonable design, but it deserves the one-line "why" the repo's comment rules ask for, and no test exercises it.

Related: with plain `$state` gone, `recompact_compound_set` and `has_top_level_binary_operator` are reachable **only** through that fallback — the `$.set(` rewrite at `server/mod.rs:559-602` takes the member-expression branch and then the `derived_names` branch first, and after this change nothing else produces a bare-identifier `$.set`. Not asking you to delete them; a note would stop the next reader from treating them as live.

### 5. What I checked and found clean

- TS: not a concern. `compile_module` (`compiler/mod.rs:865`) parses with `is_typescript: false` by design, mirroring upstream `analyze_module`, so the source reaching `SourceType::mjs()` here is already plain JS. The `.svelte.ts` file that motivated the issue is stripped by the caller.
- `$state` shadowed by a local binding — handled by `is_bound` (`rune_call_ast.rs:134-143`), matching `get_rune` returning null.
- Trailing comma: `emit_rune_replacement`'s plain branch does `trim_end().trim_end_matches(',').trim_end()`, so `$state(0,)` → `0`, not `0,`.
- Placement over-reach (`let x = c ? $state(0) : 1`) is unreachable: Phase 2 raises `state_invalid_placement` (`2_analyze/visitors/call_expression.rs:801`).
- `$derived`, `$effect`, `$effect.tracking()` and `$props` are untouched by the prefix list.

Issue #2336, opened 2026-08-06
epic: port esrap's comment-cursor discipline (unblocks #2262 and #2307)


baseballyama:
Two open issues turned out to be the same machinery, diagnosed with instrumented traces against the vendored `esrap@2.2.12`. Neither can be fixed alone; this epic records the plan so the analysis is not stranded in issue comments.

## The mechanism

esrap emits comments from a flat `analysis.comments` array through a **single monotonic `comment_index`**, flushed in the `_` wildcard (`languages/ts/index.js:1016-1025`) and re-synced **only** in `body()` (Program / BlockStatement) via `reset_comment_index`. The compiler prints with `print(program, ts({ comments: analysis.comments }))` (`phases/3-transform/index.js:39`, `:98`) and never passes `getLeadingComments` — that option exists only on the public `print()` API at `print/index.js:28-29`. So `leadingComments` plays no part in compiler output at all.

**The decisive line, which both earlier analyses missed:** `reset_comment_index` on a node with **no `loc`** sets `comment_index = comments.length` — it *parks the cursor past the end*. `crates/rsvelte_esrap/src/printer.rs:655` returns early instead. That single divergence governs both issues, with opposite signs:

- **#2262** — the cursor *rewinds* on a synthesized subtree's spans → comments duplicated.
- **#2307** — the cursor sits where official's does not → an extra emission.

## Why the one-line fix does not work

Porting the exhaust rule alone passes all 73 `rsvelte_esrap` tests and then **silently drops comments across every component**: rsvelte's synthesized wrapper blocks (component body, the `$$renderer.component((…) => {…})` arrow body) have no real span, whereas official's *do* — Svelte re-types the instance `Program` into the component-body `BlockStatement` and carries its `loc` across. Confirmed by instrumentation: `RESET BlockStatement {line:1,col:0} idx=3` sends the cursor back to 0. An agent implemented this, measured the regression, and reverted rather than baseline it.

> **Scope update (verified against `origin/main` @ `88b8d2b7`; see the comment thread for the
> measurements).** This section is out of date in two ways and should be read with the
> amendments below.
>
> - **Consumers: four, not two.** #2368 and #2375 were split into this epic after it was
>   filed. All four still reproduce; both code sites named as decisive are unchanged.
> - **A fifth change is needed, and the cursor work does not subsume it.** #2375's
>   *mechanism B* is not a cursor problem at all: `expression_converter.rs:476-484` builds
[...]

PR #2365, opened 2026-08-06
fix(compiler): treat an ASI boundary as a `$:` statement start (#2355)

### LINKED ISSUE #2355: Client keeps a comment inside a `$:` block body that official drops from `legacy_pre_effect`
Split out of the #2351 investigation. It is **not** the SIGSEGV and it is not a delimiter
problem: it reproduces with a delimiter-free `/* c */`, which compiles fine.

## Repro

```svelte
<script>
	let bar
	$: {
		/* c */
		bar = []
	}
</script>
```

`generate: 'client'`. rsvelte keeps `/* c */` inside the generated
`$.legacy_pre_effect(...)` body; the official compiler drops it. Same for a `// c` line
comment. Purely cosmetic — the emitted code is otherwise equivalent — but it is a
byte-parity divergence, so it blocks the corpus/matrix gates for any entry with this shape.

## Why upstream drops it

Upstream replaces the reactive statement with a synthesized `$.legacy_pre_effect(...)`
call. Comments *inside* the statement travel into a node that no longer exists in the
source, so they have nowhere to be re-homed and are simply lost.

## What is NOT the cause

The obvious suspect is
`crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs`'s
`strip_reactive_statement_comments`, on the theory that it only handles comments
*trailing* a `$:` statement. **That theory is wrong** — the function already has passing
unit tests for exactly this shape:

```rust
// formatting.rs, drops_comments_inside_a_reactive_block
assert_eq!(
    strip_reactive_statement_comments("$: {\n\t// inner\n\ty = 1;\n}\n"),
    "$: {\n\t\n\ty = 1;\n}\n"
);
// drops_comments_inside_a_reactive_if_block covers the `/* */` form
```

So at the string level the stripper does the right thing. The question for whoever picks
this up is **why that stripper is not the path the comment takes**. The likeliest answer:
client codegen defaults to `js_ast::to_oxc` → `rsvelte_esrap` and only falls back to the
text printer for comment-bearing programs, so a comment attached to the reactive block in
the AST reaches the output without ever passing through the text pass that would have
removed it. Confirm which of the two printers ran for this input before changing anything —
fixing `strip_reactive_statement_comments` would be a no-op if it never sees the comment.

## Scope

Independent of #2351 (SIGSEGV) and #2347 (`$props()` declarator comment). Fixing #2351
does not fix this, and this does not need #2351.

baseballyama:
Fixes #2355.

## Root cause

Not the printer. `strip_reactive_statement_comments` **is** the path the comment takes —
the theory in the issue that it never sees the comment is wrong; what is wrong is that
the scanner never recognizes the statement.

`reactive_statement_spans` (`3_transform/client/formatting.rs`) only accepted a `$:` as a
reactive statement when the preceding code byte was one of `None | ';' | '{' | '}'`:

```rust
&& matches!(scan.last_code, None | Some(b';') | Some(b'{') | Some(b'}'))
```

The repro's predecessor is `let bar` with no semicolon, so `last_code` is `r` — the guard
fails, no span is recorded, and nothing is stripped. Every existing unit test for the
stripper happens to terminate its predecessor with `;`, which is why they all pass while
the repro fails.

Automatic semicolon insertion makes a line terminator after a token that can end an
expression a statement boundary too, so the guard now also accepts that.

## Verification

Divergence confirmed to survive the corpus normalizer: oxfmt 0.62 preserves the comment in
the `expected` output, so the gate sees it (checked against `main` at `8c7851ca`, i.e. after
the 0.61 → 0.62 bump).

New tests (`crates/rsvelte_core/tests/legacy_pre_effect_comment_2355.rs`) assert the official
compiler's byte-for-byte output for the `/* c */` and `// c` forms, plus the `;`-terminated
shape that already worked; a unit test in `formatting.rs` covers the scanner directly and a
negative one pins that a conditional consequent (`c ?` / newline / `$: 1`) is still not
mistaken for a label.

## Known divergence left in place

When a statement **survives after** the reactive one, upstream re-homes the comment
positionally onto that successor (and, for a `$: if (…) {}`, prints it a second time inside
the effect, because the `if` block still has a source location). rsvelte's text stripper
cannot move a comment, so that shape stays divergent — as it already was for the
`;`-terminated form. Out of scope here.

### DIFF
--- a/.changeset/olive-pianos-smoke.md
+++ b/.changeset/olive-pianos-smoke.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): recognize a `$:` label after an ASI statement boundary when stripping reactive-statement comments
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs
@@ -276,6 +276,13 @@ fn reactive_statement_spans(source: &str) -> Vec<(usize, usize)> {
             after_last_code = end;
             continue;
         }
+        // Bracket depth is what confines the scan to the script's top level;
+        // `reactive_statement_end` maintains it only while inside a statement.
+        match c {
+            b'{' | b'(' | b'[' => scan.depth += 1,
+            b'}' | b')' | b']' => scan.depth = scan.depth.saturating_sub(1),
+            _ => {}
+        }
         scan.note_code(c);
         i += 1;
         after_last_code = i;
@@ -1331,6 +1338,51 @@ mod reactive_comment_tests {
         assert_eq!(strip_reactive_statement_comments(source), source);
     }
 
+    /// The ASI accept side, on the other token classes that can end an
+    /// expression — a bare `?` is not the only byte that has to be rejected.
+    #[test]
+    fn any_token_that_can_end_an_expression_opens_a_statement() {
+        for predecessor in ["f()", "xs[0]", "y", "1", "_x", "$x", "'s'", "`t`"] {
+            let source = format!("{predecessor}\n$: {{\n\t// inner\n\ty = 1;\n}}\n");
+            let expected = format!("{predecessor}\n$: {{\n\t\n\ty = 1;\n}}\n");
+            assert_eq!(
+                strip_reactive_statement_comments(&source),
+                expected,
+                "predecessor: {predecessor}"
+            );
+        }
+    }
+
+    /// The ASI reject side: an operator cannot end an expression, so the next
+    /// line continues it rather than starting a labeled statement.
+    #[test]
+    fn an_operator_does_not_open_a_statement() {
+        for predecessor in ["let y = c ?", "let y = a +", "let y =", "let y = a,"] {
+            let source = format!("{predecessor}\n/* keep */\n$: 1;\n");
+            assert_eq!(
+                strip_reactive_statement_comments(&source),
+                source,
+                "predecessor: {predecessor}"
+            );
+        }
+    }
+
+    /// The newline term: without a line terminator there is no ASI, so the
+    /// `$:` cannot be a statement start however its predecessor ends.
+    #[test]
+    fn a_same_line_predecessor_does_not_open_a_statement() {
+        let source = "let y = a $: /* keep */ 1;\n";
+        assert_eq!(strip_reactive_statement_comments(source), source);
+    }
+
+    /// `$` is an ordinary object key, and `,` sits at bracket depth — upstream
+    /// only treats a `$:` at the script's top level as reactive.
+    #[test]
+    fn an_object_literal_key_is_not_a_reactive_statement() {
+        let source = "let o = {\n\ta: 1,\n\t/* keep */\n\t$: 2\n};\n";
+        assert_eq!(strip_reactive_statement_comments(source), source);
+    }
+
     #[test]
     fn drops_comments_inside_a_reactive_if_block() {
         let out = strip_reactive_statement_comments(
@@ -1357,9 +1409,12 @@ mod reactive_comment_tests {
         assert_kept("let a = 'a // not a comment';\n// kept\nlet b = 1;\n");
     }
 
+    /// Upstream bails on `context.path.length > 1`, so even a `$:` — not just a
+    /// label that could never be confused for one — is plain inside a function.
     #[test]
     fn a_label_inside_a_function_body_is_not_a_reactive_statement() {
         assert_kept("function f() {\n\t// kept\n\tlab: for (;;) break lab;\n}\n");
+        assert_kept("function f() {\n\tlet a = 1\n\t$: {\n\t\t/* keep */\n\t\ta = 2;\n\t}\n}\n");
     }
 
     #[test]
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs
@@ -267,7 +267,7 @@ fn reactive_statement_spans(source: &str) -> Vec<(usize, usize)> {
         if c == b'$'
             && bytes.get(i + 1) == Some(&b':')
             && scan.depth == 0
-            && matches!(scan.last_code, None | Some(b';') | Some(b'{') | Some(b'}'))
+            && statement_can_begin(&scan, bytes, after_last_code, i)
         {
             let end = reactive_statement_end(source, &mut scan, i + 2);
             let end = extend_over_trailing_comment(source, end);
@@ -289,6 +289,22 @@ fn reactive_statement_spans(source: &str) -> Vec<(usize, usize)> {
             (if has_successor { label } else { leading }, end)
         })
         .collect()
+}
+
+/// Whether a statement can begin at `at`, `after_last_code` being the byte just
+/// past the preceding code token. Besides the explicit boundaries, automatic
+/// semicolon insertion makes a line terminator after a token that can end an
+/// expression one too — `let bar` followed by a `$:` line is a labeled
+/// statement just as much as `let bar;` is.
+fn statement_can_begin(scan: &JsScan, bytes: &[u8], after_last_code: usize, at: usize) -> bool {
+    match scan.last_code {
+        None | Some(b';') | Some(b'{') | Some(b'}') => true,
+        Some(c) => {
+            (c.is_ascii_alphanumeric()
+                || matches!(c, b'_' | b'$' | b')' | b']' | b'\'' | b'"' | b'`'))
+                && bytes[after_last_code..at].contains(&b'\n')
+        }
+    }
 }
 
 /// Byte just past the end of the reactive statement whose body starts at
@@ -1301,6 +1317,18 @@ mod reactive_comment_tests {
     fn drops_comments_inside_a_reactive_block() {
         let out = strip_reactive_statement_comments("$: {\n\t// inner\n\ty = 1;\n}\n");
         assert_eq!(out, "$: {\n\t\n\ty = 1;\n}\n");
+    }
+
+    #[test]
+    fn drops_comments_after_a_semicolon_less_predecessor() {
+        let out = strip_reactive_statement_comments("let y\n$: {\n\t// inner\n\ty = 1;\n}\n");
+        assert_eq!(out, "let y\n$: {\n\t\n\ty = 1;\n}\n");
+    }
+
+    #[test]
+    fn a_conditional_consequent_is_not_a_reactive_statement() {
+        let source = "let y = c ?\n/* keep */\n$: 1;\n";
+        assert_eq!(strip_reactive_statement_comments(source), source);
     }
 
     #[test]
--- a/crates/rsvelte_core/tests/legacy_pre_effect_comment_2355.rs
+++ b/crates/rsvelte_core/tests/legacy_pre_effect_comment_2355.rs
@@ -1,0 +1,43 @@
+//! A comment inside a `$:` statement body is dropped: upstream rewrites the
+//! reactive statement into a synthesized `$.legacy_pre_effect(...)` call, so
+//! comments attached to the original statement have no node to re-home onto.
+//!
+//! Expected outputs are the official Svelte compiler's, verbatim.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+fn client(name: &str, source: &str) -> String {
+    compile(
+        source,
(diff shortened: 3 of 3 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. The diagnosis is right and the widening is sound in the ASI direction; the gaps are that the new guard is a two-term conjunction and the tests pin one value of one term.

Read-only review at `f3ed947` — no build, no test run.

---

### 1. The newline conjunct is not pinned by any test

`formatting.rs:299-308`. The new arm is a conjunction:

```rust
(c.is_ascii_alphanumeric() || matches!(c, b'_' | b'$' | b')' | b']' | b'\'' | b'"' | b'`'))
    && bytes[after_last_code..at].contains(&b'\n')
```

Delete the whole `&& … contains(&b'\n')` term and **all five new tests stay green**:

- `drops_comments_after_a_semicolon_less_predecessor` (`"let y\n$: {…"`) has a newline, so it passes either way;
- `a_conditional_consequent_is_not_a_reactive_statement` is rejected by the *byte-class* term (`last_code` is `?`), so it never reaches the newline term;
- all three tests in `legacy_pre_effect_comment_2355.rs` have newlines.

Same for the byte-class term in the other direction: replace the whole `matches!(…)` list with `c != b'?'` and every test still passes. So the suite pins the accept side at exactly one value (`y`, an identifier tail) and the reject side at exactly one value (`?`), out of the eight bytes the list accepts and the nineteen `continues_statement` rejects.

I could not construct *valid* JS that reaches the newline term (a same-line `$:` after an expression-ending token is a syntax error, and ASI by definition needs a line terminator), which is itself worth recording: either the conjunct is load-bearing and deserves a test, or it is unreachable by construction and deserves a one-line note saying so. Right now it is neither.

### 2. The realistic must-reject value is `,`, and it is untested

The negative control the suite is missing is not the ternary — it is an object literal, which is ordinary code:

```js
let o = {
	a: 1,
	$: 2
}
```

The byte before `$` is `,`, so the byte-class term rejects it and the source is left alone. Correct — and load-bearing, since `$` is a perfectly common key (jQuery, i18n maps). Please add it; it exercises the term at a value `?` does not, and it is the shape a future widening is most likely to break.

While you are there: `)` is the accepted byte most likely to appear in real legacy Svelte (`let x = f()` / newline / `$: …`) and is currently untested.

### 3. `scan.depth == 0` at `:269` is vacuous, and the widening now leans on it

`JsScan::depth` is mutated **only** inside `reactive_statement_end` (`formatting.rs:340-348`), which restores it to `base_depth` before returning. `note_code` never touches it and the main loop of `reactive_statement_spans` never does either. So `scan.depth == 0` is always true at `:269`, and the comment at `:265-266` — "A `$:` label only starts a reactive statement at the top level of the script" — is not what the code enforces.

That matters here because the guard you widened was the only thing keeping the scanner out of function bodies:

```svelte
<script>
	function f() {
		let a = 1
		$: {
			/* keep */
			a = 2
		}
	}
</script>
```

Before: `last_code` is `1` → rejected → `/* keep */` survives, matching upstream (a label inside a function is not a reactive statement and is never rewritten into `legacy_pre_effect`). After: accepted → the comment is stripped.

Two things bound this: the class already existed (`function f() {` / newline / `$: {…}` hits the `Some(b'{')` arm and was accepted before this PR), and `strip_reactive_statement_comments` only runs in **legacy, non-runes** mode (`client/mod.rs:4492-4497`), so the blast radius is legacy components with a `$:` label nested inside a function — rare. Not blocking. But please either make the depth guard real or fix the comment, so the next widening does not inherit a protection that is not there.

### 4. Two hand-maintained ASI byte lists, only one updated

`continues_statement` (`:456-480`) already encodes "this token cannot end a statement", and the new `statement_can_begin` (`:299`) encodes the near-complement ~150 lines away. They are consistent today — I checked, no byte appears in both — but they are now two places to keep in sync, and only one was touched. Deriving the new arm from `continues_statement` (it cannot be a pure complement, since `None` must map to `true` on one side and `true` on the other) would keep them locked together.

Related under-reach, deliberate-looking but worth confirming: postfix `++`/`--` (`i++` / newline / `$: 1`) is still rejected because `+`/`-` are not in the accept list. That is an ASI boundary upstream honours.

### 5. Stopgap or deepening a doomed path? — stopgap, and the right kind

Given #2368's conclusion that `strip_reactive_statement_comments` should be **deleted** once #2336 lands, the question is what this PR leaves behind when that happens. Answer: the three tests in `crates/rsvelte_core/tests/legacy_pre_effect_comment_2355.rs` assert the **official compiler's byte-for-byte output**, so they survive the deletion intact and become free acceptance criteria for #2336's cursor-based replacement. Only the two unit tests inside `formatting.rs` die with the function. That is the right ratio, and the change is one small helper rather than a new special case per shape.

It also does not repeat the mistake #2368 calls out: `drops_comments_inside_a_reactive_if_block` pins a divergence from upstream, and neither new unit test does that — `a_conditional_consequent_is_not_a_reactive_statement` pins an identity and `drops_comments_after_a_semicolon_less_predecessor` pins behaviour that matches upstream.

Suggestion: say in #2336 that `legacy_pre_effect_comment_2355.rs` must keep passing after the deletion, so nobody re-derives those expectations.

### 6. What I checked and found clean

- `bytes[after_last_code..at]` cannot panic or slice mid-char: `after_last_code` is only ever set to a value `<= i` (`next` after a string/comment, `i + 1` after a code byte, `end` after a reactive span, where `i` is then set to the same value), the slice is over `bytes` not `source`, and the search byte is ASCII.
- The new arm's accept set and `continues_statement`'s continue set are disjoint, so `statement_can_begin` and `reactive_statement_end` cannot contradict each other on the same byte.
- The `/* c */` and `// c` integration tests differ in comment kind, not just position — the two comment kinds do take different paths through `JsScan::step` (`:605` vs `:613`).
- The PR body's note that the divergence survives oxfmt 0.62 normalization is the right check to have run, given how recently the normalizer was silently doing nothing.

baseballyama:
Both nits addressed in `435be46a`; rebased onto `88b8d2b7` so clippy is verifiable on this branch (the earlier base predated #2363).

## Nit 2 — decided: make the depth check real

Confirmed the claim independently: `JsScan::depth` is mutated only at `formatting.rs:340-348`, inside `reactive_statement_end`, which restores it to `base_depth` before returning. Nothing else touches it, so `scan.depth == 0` was always true.

**Chose to make the check real rather than correct the comment**, because top-level-only is what upstream does — `visitors/LabeledStatement.js:12` bails on `context.path.length > 1`, so a `$:` inside a function body is an ordinary label whose comments are kept. Weakening the comment would have documented a behaviour that diverges from upstream; the widening was a genuine bug, not a deliberate reach. The scan now tracks bracket depth in its main loop, and the comment is accurate.

The reviewer's example is now pinned:

```js
function f() { let a = 1
$: { /* keep */ … } }
```

## Nit 1 — each term now has a negative control

Added accept-side coverage across the token classes (`f()`, `xs[0]`, `y`, `1`, `_x`, `$x`, `'s'`, `` `t` ``), reject-side coverage across operators (`?`, `+`, `=`, `,`), a no-newline control, an object-literal control (`{ a: 1, /* keep */ $: 2 }`), and the nested-function control above.

Verified by re-running the reviewer's two mutants plus one for the term I changed — **all three now die**:

| mutant | before | after |
|---|---|---|
| drop `&& …contains(&b'\n')` | green | **FAILED** |
| accept list → `c != b'?'` | green | **FAILED** |
| `scan.depth == 0` → `true` | green (vacuous) | **FAILED** |

## One more instance of the same defect, found while doing this

`a_label_inside_a_function_body_is_not_a_reactive_statement` already existed — and asserted on `lab: for (;;) break lab;`. A label named `lab` can never be mistaken for a reactive statement, so the test could not fail whatever the guard did, despite its name promising exactly this coverage. Extended it with the `$:` case rather than adding a second test under a new name. Worth adding to the #1791 inventory as the "name promises coverage the input cannot deliver" variant.

Green: `--lib` 1409, `legacy_pre_effect_comment_2355` 3/3, and the seven neighbouring comment/reactive suites (`client_script_comments_pin`, `client_transforms_pin`, `reactive_block_comment_delimiter_2351`, `reactive_label_comment_ownership`, `reactive_cycle_block_body`, `svelte_ignore_comment_attachment_2256`, `comment_whitespace_run_1975`). Clippy clean, no suppressions.

baseballyama:
**LGTM** — 0 blocking. The upstream claim holds, making `scan.depth` real was the right call rather than the comment fix I suggested, and all three mutants die. One correction to which test kills which mutant.

Re-reviewed at `435be46a` (rebased onto `88b8d2b7`) against my earlier comment. Read-only, no build.

---

### The load-bearing claim is correct

`3-transform/client/visitors/LabeledStatement.js:12`:

```js
if (context.state.analysis.runes || context.path.length > 1 || node.label.name !== '$') {
    context.next();
    return;
}
```

A top-level statement's `path` is `[Program]`, length 1, so `path.length > 1` is exactly "not a direct child of the Program". A `$:` inside a function body is therefore never rewritten into `$.legacy_pre_effect` — it stays an ordinary labeled statement and keeps its comments. So top-level-only is the **correct** behaviour, not merely what the comment claimed, and enforcing it beats correcting the comment. Taking the harder option was right.

### The depth fix itself checks out

The new `match` at `:279-284` runs after the `$:` branch and before `note_code`, so the bytes consumed by `reactive_statement_end` are not double-counted (`i = end` skips them, and that function restores `scan.depth` to `base_depth` before returning). Consistency between the two is preserved.

Three things I checked that could have broken:

- **`Some(b'{')` no longer accepts inside a function.** After `{` the depth is 1, so `function f() {\n$: {…}\n}` — accepted before this PR via the `last_code == '{'` arm — is now rejected. That is the intended change and it matches `path.length > 1`.
- **Balanced constructs still leave depth at 0 at the top level.** `if (a) { … }` → `(`/`)`/`{`/`}` net 0; `const o = { a: 1 };` → net 0. A following `$:` is still seen.
- **Template interpolations do not leak depth.** `JsScan::step` returns `Some` for `${` (`:567`) and for the interpolation's own braces (`:582-596`), so the new match never sees them; ordinary bytes inside an interpolation do reach it, but `(`/`[` there are balanced by their own closers. And a `$` inside a string returns `Some(i + 1)` from the `in_string` branch, so the `$:` check cannot fire inside a literal.
- `saturating_sub` clamps at 0 rather than going negative, which fails toward "still at top level" — the permissive direction, and the right one for a text pass that may see an unbalanced fragment.

### Mutant kills — confirmed, with one re-attribution

| mutant | dies to |
|---|---|
| drop `&& bytes[after_last_code..at].contains(&b'\n')` | `a_same_line_predecessor_does_not_open_a_statement` — `"let y = a $: /* keep */ 1;\n"` has `a` (accepted byte class) with no newline, so only the newline term rejects it |
| replace the byte-class list with `c != b'?'` | `an_operator_does_not_open_a_statement` — `+`, `=`, `,` would all pass `c != b'?'` and all have a newline |
| remove the `scan.depth` mutation | the **second assertion in `a_label_inside_a_function_body_is_not_a_reactive_statement`** — `"function f() {\n\tlet a = 1\n\t$: {…}\n}"` has `1` (accepted) plus a newline, so depth is the only term that rejects it |

All three die. My earlier finding — that deleting the newline conjunct or flattening the byte list left every test green — is closed.

**One correction.** The doc comment on `an_object_literal_key_is_not_a_reactive_statement` says "`,` sits at bracket depth — upstream only treats a `$:` at the script's top level as reactive", which reads as though that test isolates the depth term. It does not: for `let o = {\n\ta: 1,\n\t/* keep */\n\t$: 2\n};` the predecessor byte is `,`, which the byte-class term already rejects, **and** the depth is 1 — both terms reject it, so it isolates neither. It is a good end-to-end regression test for a realistic shape (a `$` object key is ordinary code), and worth keeping; the comment should just say that rather than name a mechanism it does not exercise. Given that the point of the other two tests is term isolation, the distinction is worth keeping straight.

### The rest of my earlier comment

- `any_token_that_can_end_an_expression_opens_a_statement` covers all eight accepted classes including `f()`, `xs[0]`, `'s'` and `` `t` ``, which also happens to exercise the new depth tracking's balance on `()` and `[]`. Closes the "one accept value tested" finding.
- The stopgap assessment is unchanged and still good: the three tests in `legacy_pre_effect_comment_2355.rs` assert official byte-for-byte output, so they survive the eventual deletion of `strip_reactive_statement_comments` and become #2336's acceptance criteria. Worth saying so in #2336 explicitly.
- The two hand-maintained ASI byte lists (`continues_statement` at `:456`, `statement_can_begin` at `:301`) remain disjoint and consistent. Still two places to keep in sync; still only a maintenance nit.

Issue #2357, opened 2026-08-06
epic: route Phase-3's ~200 hand-rolled bracket scans through skip_opaque


baseballyama:
## Summary

Phase 3 still contains ~200 hand-rolled bracket-depth scanners over generated JS **text**, and only 2 of the 33 files holding them route any scan through the shared lexer `crates/rsvelte_core/src/compiler/phases/3_transform/shared/js_scan.rs::skip_opaque`. A scan without lexical awareness moves its depth counter on a `(`, `[`, `{`, `)`, `]` or `}` that sits inside a comment, string, template literal or regex.

This class has now produced, in three days:

- **#2351 — a SIGSEGV.** `find_assignment_position` counted characters blind; a `}` inside a comment returned depth to 0, the resulting unbalanced `lhs` fragment made `extract_destructure_targets` recurse on an identical string, and the stack was exhausted. Fixed in #2354.
- **#2347 — silently swallowed codegen.** A trailing `// c` in a `$props()` pattern became part of the binding name, so `= $.rest_props(...)` was appended *inside the comment*. Fixed in #2356.
- **#2328 — five statement shapes silently unsplit.** `split_top_level_statements` ended a statement at a top-level `}`/`;` with no continuation check. Its own report recorded that the bug class was **not** closed.
- **#2253** and the class-transform sites, fixed one at a time.

Each was found by a user hitting it, not by us. That is the argument for a sweep: the same defect is being rediscovered from the outside, one crash at a time.

## Inventory

Proxy metric is `depth += 1` / `depth -= 1` sites. Columns: sites / `skip_opaque` uses / has its own comment state / has its own string state.

| sites | skip_opaque | comments | strings | file |
|---:|---:|:--:|:--:|---|
| 84 | 1 | yes | yes | `server/transform_script.rs` |
| 62 | 0 | **no** | yes | `server/transform_legacy.rs` |
| 54 | 0 | yes | yes | `client/expression_utils.rs` |
| 40 | 0 | **no** | yes | `client/state_transforms.rs` |
| 38 | 0 | yes | yes | `client/props_transforms.rs` |
| 30 | 0 | **no** | yes | `server/helpers.rs` |
| 26 | 0 | **no** | yes | `shared/async_body.rs` |
| 22 | 0 | yes | yes | `client/mod.rs` |
| 18 | 0 | **no** | yes | `client/destructure_transforms.rs` |
| 12 | 0 | yes | yes | `server/transform_store.rs` |
[...]

PR #2366, opened 2026-08-06
fix(compiler): read legacy SSR bracket scanners lexically (#2357)

baseballyama:
Batch 2 of #2357: `server/transform_legacy.rs`.

The legacy SSR line scanners counted `{}`/`()`/`[]`/`,`/`;`/`=` without skipping
comments — most of them tracked string literals by hand but no comment kind at
all, and three (`find_assignment_eq`, `split_property_key_value_ssr`, and the
seven per-line `for c in trimmed.chars()` brace counters) tracked nothing. A
delimiter inside a `//` or `/* */` comment therefore moved the depth counter:
`$: { b = a; // }` closes the reactive block a line early, `export let x = 1 // done;`
reads as terminated, `{ a /* } */, b }` ends the destructuring pattern inside the
comment.

Every converted scanner now walks `js_scan::code_bytes`, the iterator added in #2363.

**15 unit tests, each written and confirmed failing before the fix** (`cargo test
-p rsvelte_core --lib transform_legacy::tests`), covering both comment kinds and
string literals per scanner.

Two scanners needed more than a `code_bytes` swap:

- `export_let_declaration_seems_complete` used its `in_string` flag *after* the
  loop to detect an unterminated literal (a template continuing on the next
  line). `code_bytes` cannot report that, so it keeps a `skip_opaque` loop plus
  `opaque_run_is_unterminated`. Its trailing-operator check also now runs past a
  trailing comment, so `= [1] /* ] */` no longer reads as ending in `/`.
- The multi-line `$:` collector threads template-literal state *across* lines and
  deliberately does not descend into `${…}` (the constraint recorded in #2363).
  It became `scan_reactive_line`, which keeps the cross-line backtick state and
  the opaque-`${…}` behaviour while gaining comment and string awareness.

Skipped, with reasons, in the report to the epic: the `Vec<char>`-indexed
scanners (`extract_simple_assignments`, `extract_reactive_rhs_identifiers`,
`extract_store_set_targets`) and `find_trailing_comment_start`, which already
looks for comments by construction.

## Reachability: this file cannot be reached by a `$:` in a shipping compile

Establishing this **before** the defect descriptions, because they were written when I believed
otherwise and the urgency they convey is wrong.

`reorder_reactive_statements_after_functions` has one non-test caller chain, and it terminates in
`{@const}` lowering:

```
transform_script.rs:266  ->  transform_script_content_with_imports_and_derived  (one caller)
                         ->  ast/visitors/declaration_tag.rs:346
```

`declaration_tag.rs:328-332` builds that input from a `{@const}` body with a `;` appended, so the
input is a **declaration**. A `$:` label is not a declaration and cannot appear there. This does
not depend on a probe or on which tests exist — it is the caller graph plus the shape of the
terminus. It also agrees with `AGENTS.md`: server SSR is pure-AST and the legacy text generator is
deleted, so what survives here is whatever `{@const}` lowering still calls.

**So every "emits output that does not parse" below means "would, if the path were revived."** The
code is `pub(crate)`, compiled, and reachable from tests; the defects are real and the fixes are
right. They are not user-facing today. Deletion of the unreachable machinery is #2397.

## Review round: two exits, two different histories

The last commit fixes **two** things, and they do **not** have the same relationship to `main`.
Splitting them because a body that calls the whole commit a divergence would justify reverting
the half that is mandatory.

Both multi-line `$:` accumulation loops have **three exits each — six total, two wrong**. The
tail break consulted the threaded template state; the *leading* guard did not, so a blank line,
a `$:`, a `function ` or a `//` occurring as literal text inside a template literal ended the
statement there.

### The gate term — **restores parity this branch broke** (mandatory)

For `` `hello ${ `` at the end of a line, `88b8d2b7`'s naive `for c in next_trimmed.chars()`
counter counts the `{` of `${`, depth becomes 1, the gate does not fire, and the statement is
collected whole — **`main` is correct here.** Batch 2 made the backtick run opaque, so that `{`
stopped counting while the gate still had no `!acc_template`, and the statement started
splitting. That is a batch-2 regression and this term repairs it. The two pins covering it are
labelled in the source as having this change's own pre-fix state as their control, precisely
because the code they replaced passes them.

### The leading-guard term — **fixes a bug `main` shares** (an improvement)

For the four boundary tokens as literal text inside a template, `main` is equally broken:
`88b8d2b7`'s continuation loop gives `` `hello `` no bracket contribution at all, depth stays 0
and the gate fires at iteration 1 — the same truncation, one exit earlier. Batch 2 could not
have introduced this; that loop carried no template state on `main`. On a strict no-regression
standard this term was not required to unblock the PR. It ships anyway because the alternative
is knowingly emitting a statement-splitter:

```js
$: msg =
`hello

${name}`;
```

became `` $: msg =\n`hello `` plus a dangling `` ${name}`; `` — output that does not parse, in a
path no shipping compile enters.

The evidence that this was an oversight rather than a decision is the sibling branch at `:998`:
it holds its template state in the *loop condition* rather than in a break-set, and it was
already correct.

### Safety argument

- **Structural, not measured**: `!acc_template && (…)` reduces to the original condition whenever
  the flag is false, so every non-template input takes a bit-identical path **by construction**.
  The byte-identical control on `$: msg =\nfoo +\nbar;` confirms that rather than establishing it.
- **No gate regression**, and now doubly so: an unterminated template cannot normalize to MATCH
  because it does not parse, *and* no corpus input reaches this code at all. That makes the claim
  safe and correspondingly weak as a motivation — it was never evidence of user impact, and there
  is no user impact to be evidence of.

Pinned as `template_text_is_not_a_statement_boundary` across all four boundary tokens, confirmed
failing on `65ce783b`.

### The `//` break-set asymmetry was left alone deliberately, and now has a mechanism

Three shapes built to exercise the difference all came back **IDENTICAL** to `main`. The reason
is now known rather than merely observed: the two loops genuinely *do* disagree — the decision
loop scans through `// c` and breaks at the gate one line later, the collection loop breaks
before it — but `needs` is `true` either way, because a comment is not a statement. **Real
disagreement, unobservable outcome.** Unifying them still needs a repro; filed as #2387 with the
mechanism recorded.

The comment at `:848` asserted the sets match, which is false. It is now a conditional that holds
as written.

A **separate, pre-existing** defect in the same `//` entry — `$: total =` / `// running sum` /
`a + b;` truncating to a dangling `$: total =` — is out of scope here and filed on its own.

Closes nothing on its own; #2357 stays open for batches 3+.

### DIFF
--- a/.changeset/scan-sweep-batch2.md
+++ b/.changeset/scan-sweep-batch2.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): read legacy SSR bracket scanners lexically
+
+The `export let` / reactive `$:` line scanners in the server transform counted
+brackets, commas, semicolons and `=` without skipping comments, so a delimiter
+inside a `//` or `/* */` comment moved the depth counter — splitting a `$:` block
+at a `}` that lived in a comment, or truncating a declarator at a commented-out
+`;`. They now walk only the code bytes.
--- a/.changeset/scan-sweep-batch4.md
+++ b/.changeset/scan-sweep-batch4.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): fix byte/char index mix in the legacy SSR store-set scan
+
+`extract_store_set_targets` fed a byte offset from a `memmem` match into a
+`Vec<char>`, so any non-ASCII before a `$.store_set(` call made it read the
+store name from the wrong position and record a truncated dependency.
+`extract_simple_assignments` alongside it now skips comments and regex
+literals instead of reading assignments out of them.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -2223,6 +2223,7 @@ mod tests {
         );
     }
 
+    // Control is this change's own pre-fix state; the code it replaced passes too.
     #[test]
     fn interpolation_brace_does_not_end_the_reactive_block_early() {
         let script = "let name = 'w';\n$: b = 1;\n$: { msg = `hello ${\nname\n}`;\n}\n";
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -843,8 +843,8 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
         let mut needs = false;
         let mut in_reactive_multiline = false;
         let mut reactive_depth: i32 = 0;
-        // Must match the collection loop below, or the two disagree on where a
-        // reactive block ends and this scan mis-routes.
+        // Threaded like the collection loop below; if the two disagree on where a
+        // block ends, this scan mis-routes.
         let mut in_template = false;
         let mut i = 0;
         while i < lines.len() {
@@ -904,7 +904,12 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
                         i += 1;
                         while i < lines.len() {
                             let nt = lines[i].trim();
-                            if nt.is_empty() || nt.starts_with("$:") || nt.starts_with("function ")
+                            // Inside a template literal these are literal text, not
+                            // statement boundaries.
+                            if !acc_template
+                                && (nt.is_empty()
+                                    || nt.starts_with("$:")
+                                    || nt.starts_with("function "))
                             {
                                 break;
                             }
@@ -1046,10 +1051,13 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
                     while i < lines.len() {
                         let next = lines[i];
                         let next_trimmed = next.trim();
-                        if next_trimmed.is_empty()
-                            || next_trimmed.starts_with("$:")
-                            || next_trimmed.starts_with("function ")
-                            || next_trimmed.starts_with("//")
+                        // Inside a template literal these are literal text, not
+                        // statement boundaries.
+                        if !acc_template
+                            && (next_trimmed.is_empty()
+                                || next_trimmed.starts_with("$:")
+                                || next_trimmed.starts_with("function ")
+                                || next_trimmed.starts_with("//"))
                         {
                             break;
                         }
@@ -2228,6 +2236,26 @@ mod tests {
             split_destructuring_properties_ssr("a // , b"),
             vec!["a // , b"]
         );
+    }
+
+    #[test]
+    fn template_text_is_not_a_statement_boundary() {
+        // A blank line, `$:`, `function ` or `//` inside a template literal is
+        // literal text; the loops' first exit used to treat it as a boundary.
+        for body in [
+            "\n",
+            "$: not a statement\n",
+            "// not a comment\n",
+            "function nope() {}\n",
+        ] {
+            let script =
+                format!("let a = 1;\n$: msg =\n`hello\n{body}${{name}}`;\nlet after = 2;\n");
+            let out = reorder_reactive_statements_after_functions(&script);
+            assert!(
+                out.contains(&format!("$: msg =\n`hello\n{body}${{name}}`;")),
+                "statement split on {body:?}: {out}"
+            );
+        }
     }
 
     // Control is this change's own pre-fix state; the code it replaced passes too.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -1341,21 +1341,15 @@ fn extract_store_set_targets(code: &str, vars: &mut Vec<String>) {
     let mut search_from = 0;
     while let Some(pos) = finder.find(&code.as_bytes()[search_from..]) {
         let abs_pos = search_from + pos;
-        let after_call = abs_pos + 12; // "$.store_set(".len()
-        // Read the first argument (store name)
-        let mut j = after_call;
-        let chars: Vec<char> = code.chars().collect();
-        while j < chars.len() && (chars[j] == ' ' || chars[j] == '\t') {
-            j += 1;
-        }
-        let name_start = j;
-        while j < chars.len() && (chars[j].is_alphanumeric() || chars[j] == '_' || chars[j] == '$')
-        {
-            j += 1;
-        }
-        if j > name_start {
-            let store_name: String = chars[name_start..j].iter().collect();
-            let store_sub = format!("${}", store_name);
+        // `memmem` yields a byte offset; the old code fed it to a `Vec<char>`,
+        // mis-slicing the name whenever non-ASCII preceded the call.
+        let rest = code[abs_pos + 12..].trim_start_matches([' ', '\t']);
+        let store_name: String = rest
+            .chars()
+            .take_while(|c| c.is_alphanumeric() || *c == '_' || *c == '$')
+            .collect();
+        if !store_name.is_empty() {
+            let store_sub = format!("${store_name}");
             if !vars.contains(&store_sub) {
                 vars.push(store_sub);
             }
@@ -1369,67 +1363,38 @@ fn extract_store_set_targets(code: &str, vars: &mut Vec<String>) {
 /// to find variable assignments that indicate the reactive statement modifies a variable.
 fn extract_simple_assignments(code: &str) -> Vec<String> {
     let mut vars = Vec::new();
-    // Find patterns like `identifier =` (not `==`), `identifier++`, `identifier--`,
-    // `++identifier`, `--identifier`
-    let chars: Vec<char> = code.chars().collect();
-    let len = chars.len();
+    let bytes = code.as_bytes();
     let mut i = 0;
-    let mut in_string = false;
-    let mut string_char = ' ';
+    let mut prev: Option<u8> = None;
 
-    while i < len {
-        let c = chars[i];
-
-        // Track string literals to avoid matching inside them
-        if (c == '\'' || c == '"' || c == '`') && !in_string {
-            in_string = true;
-            string_char = c;
-            i += 1;
-            continue;
-        }
-        if in_string {
-            if c == string_char && (i == 0 || chars[i - 1] != '\\') {
-                in_string = false;
+    while i < bytes.len() {
+        if let Some((next, is_comment)) = skip_opaque(bytes, i, prev) {
+            if !is_comment {
+                prev = Some(b'x');
             }
-            i += 1;
+            i = next;
             continue;
         }
 
-        // Check for `++identifier` and `--identifier` prefix operators
-        if i + 2 < len
-            && ((chars[i] == '+' && chars[i + 1] == '+')
-                || (chars[i] == '-' && chars[i + 1] == '-'))
+        // Prefix `++x` / `--x`
+        if (bytes[i] == b'+' && bytes.get(i + 1) == Some(&b'+'))
+            || (bytes[i] == b'-' && bytes.get(i + 1) == Some(&b'-'))
         {
-            let op_end = i + 2;
-            // Skip whitespace after operator
-            let mut j = op_end;
-            while j < len && chars[j] == ' ' {
+            let mut j = i + 2;
+            while bytes.get(j) == Some(&b' ') {
                 j += 1;
             }
-            // Read identifier
-            if j < len && (chars[j].is_alphabetic() || chars[j] == '_' || chars[j] == '$') {
-                let start = j;
-                while j < len && (chars[j].is_alphanumeric() || chars[j] == '_' || chars[j] == '$')
-                {
-                    j += 1;
-                }
-                let ident: String = chars[start..j].iter().collect();
+            if let Some((ident, end)) = read_identifier(code, j) {
                 if !is_reactive_keyword(&ident) && !vars.contains(&ident) {
                     vars.push(ident);
                 }
-                i = j;
+                prev = Some(b'x');
+                i = end;
                 continue;
             }
         }
 
-        if c.is_alphabetic() || c == '_' || c == '$' {
-            // Read identifier
-            let start = i;
-            while i < len && (chars[i].is_alphanumeric() || chars[i] == '_' || chars[i] == '$') {
-                i += 1;
-            }
-            let ident: String = chars[start..i].iter().collect();
-
+        if let Some((ident, end)) = read_identifier(code, i) {
             // A member property (`foo.x = …` / `foo.x += …` / `foo.x++`) is not
             // a declared variable: the assignment mutates the *base object*, not the
             // property. Recording the property would create a false reactive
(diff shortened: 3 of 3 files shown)


baseballyama:
**LGTM with nits** — 0 blocking findings, but finding 1 is a real behaviour change I could not confirm end-to-end (read-only review, no `cargo`), so please sanity-check it before merging.

Reviewed at `d210726`. The index-unit audit and the test-discrimination audit both came back clean except where noted; details in §6/§7 so the coverage is on the record.

---

### 1. The `${…}` regression shape is present — in the five scanners that did *not* get template state

`transform_legacy.rs:825` `bracket_depth_delta` and `transform_legacy.rs:783` `scan_reactive_line` are not equivalent, and the five `bracket_depth_delta` call sites used to count `${`'s brace.

`scan_reactive_line` is fine: it threads `in_template` across lines and treats the template body as opaque, which is what the old raw loop at the same site did (backtick toggled `in_template_literal`, brackets counted only when false). The `reorder_reactive_statements_after_functions` collector at `:997`/`:1005` is unchanged in behaviour. ✓

The other five sites — `:862`, `:872`, `:916` (inside `needs_reorder`) and `:1170`, `:1176` (`sort_reactive_in_place`) — replaced `for c in trimmed.chars() { '{'|'('|'[' => +1, … }`, which counted **every** brace in the line including inside a template literal, with `bracket_depth_delta`, which routes through `code_bytes` → `skip_opaque`. An unterminated backtick makes `skip_opaque` (`js_scan.rs:71-83`) return `Some((bytes.len(), false))`, so the rest of the line is swallowed and the interpolation's brace is never seen.

Concrete input:

    $: {
    	msg = `hello ${
    		name
    	}`;
    }

`needs_reorder` before this PR: `$: {` → +1, `` msg = `hello ${ `` → +1 (depth 2), `name` → 0, `` }`; `` → −1 (depth 1), `}` → 0. The whole block is consumed as one reactive statement and `needs` stays false.

After: line 2 contributes 0, so depth is 1 not 2; line 4's `}` is now counted at depth 1 → 0, and `in_reactive_multiline` clears **one line early**. The trailing `}` then hits `:958` (`saw_reactive && !trimmed.is_empty()`) and flips `needs = true`. The reorder branch is taken where the in-place branch used to be, i.e. reactive statements get moved to the end of the script instead of only topologically sorted among themselves. The statement itself survives (the reorder branch uses the correct `scan_reactive_line` collector), so I expect an ordering/placement difference rather than a mangled statement — but I could not run the compiler to name the emitted bytes, hence "please check" rather than "changes requested".

The simpler variant hits `sort_reactive_in_place` the same way:

    $: msg = `a ${
    	b
    }`;

`:1170` sees delta 0 and the trailing `{` is not in the continuation-operator list at `:1182-1199`, so the statement is closed after line 1.

If the intent is that only `scan_reactive_line` knows about templates, that is defensible — but then the three `bracket_depth_delta` sites that walk `$:` continuation lines should share it (or the PR should record why the divergence is safe). Note the string-literal half of this change is a strict improvement in the same five places (`$: s = '{';` no longer opens a phantom block), so it is only the template case that moves backwards.

### 2. Comment awareness is per-line only, which the changeset overstates

`.changeset/scan-sweep-batch2.md` says "a delimiter inside a `//` or `/* */` comment moved the depth counter … They now walk only the code bytes." True within a line. Across lines it still happens, because `scan_reactive_line` threads only backtick state and `bracket_depth_delta` threads nothing:

    $: {
    	/* }
    	*/
    	b = a;
    }

The `}` on line 2 is inside an unterminated block comment; `skip_opaque` consumes to end-of-line, then line 3 starts a fresh scan with no comment state. Same as before the PR — not a regression, but the new test `reactive_block_brace_in_comment_does_not_close_the_block` uses a single-line `//`, so nothing distinguishes "fixed" from "fixed for one comment kind on one line". Worth a sentence in the changeset.

### 3. One test assertion is vacuous

`transform_legacy.rs` tests, `declaration_completeness_ignores_comment_brackets`:

    assert!(!export_let_declaration_seems_complete("x = [1 /* ] */"));

On the pre-PR body this returns `false` too — the `]` inside the comment balanced the `[`, so the depth check passed, and the function then fell into `trimmed.ends_with('/')` (`:141`). Both sides return `false` for different reasons, so this line cannot fail against the unfixed code.

The other four assertions in that test do discriminate (`"x = [1 // ]"` → old `true`; `"x = [1] /* ] */"` → old `false` via `bracket_depth == -1`; `"x = 1 /* open"` → old `true`). Consider swapping the vacuous one for a case where the trailing-operator rule cannot rescue it, e.g. `"x = [1 /* ] */ , 2"`.

### 4. Orphaned doc comment

`transform_legacy.rs:760-767` is the doc block for `reorder_reactive_statements_after_functions` ("Reorder legacy reactive `$:` statements in SSR script…"). `template_literal_close` was inserted directly after it at `:768`, so that block now documents `template_literal_close`, and `reorder_reactive_statements_after_functions` at `:837` has no doc comment at all. Move the new helper below it.

### 5. Unrelated change in the diff

`1_parse/read/expression.rs:7051` — hoisting `is_estree_node(value)` into a match guard is behaviour-preserving (a `Value::Object` that fails the guard falls past the `Value::Array` arm into `_ => {}`, which is what the old empty `if` did), but it is scope creep in a scanner PR and makes the diff harder to bisect.

### 6. New regex heuristic — noting the amplifier, not asking for a change

`code_bytes` inherits `slash_starts_regex`, so a `/` preceded by an operator now opens a regex. These scanners are fed single **trimmed lines with no `\n`**, which means `skip_opaque`'s deliberate bail — `b'\n' => return None`, "an unterminated regex is really a division" (`js_scan.rs:117-119`) — can never fire here, and an accidental regex swallows the remainder of the line instead. `export let x = i++ / 2;` makes `has_top_level_semicolon` return `false` (the `;` is inside the pseudo-regex) and the declaration reads as incomplete. Pathological input; the same heuristic fixes the far more likely `export let re = /,/, b = 2;`, which the old `split_declarators` split inside the regex. Net win, recorded for completeness.

### 7. Byte/char index audit — clean

Checked every converted scanner for the batch-1 unit mismatch:

- All indices are byte offsets from `code_bytes`, all matched delimiters are ASCII, so `s[..i]`, `&s[start..i]`, `Some(i + 1)` and `s[i + 1..]` stay on char boundaries even with non-ASCII in the line. Continuation bytes (≥ 0x80) never match an ASCII delimiter arm.
- `export_let_declaration_seems_complete`'s `last_code_end` (`:104`/`:121`) is always a boundary: it is either `skip_opaque`'s `next` (past an ASCII quote/`*/`) or `i + 1` on the last byte of a multi-byte char (continuation bytes are not `is_ascii_whitespace`, so each one advances it).
- `opaque_run_is_unterminated`'s `bytes[i + 2..]` (`:189`) cannot panic — the arm is guarded by `bytes.get(i + 1) == Some(&b'*')`, so `i + 2 <= len` and the worst case is the empty slice.
- `split_logical_expression`'s `let Some(&next) = bytes.get(i + 1) else { continue }` faithfully reproduces the old `(0..len.saturating_sub(1)).rev()` bound — the last byte contributed no depth change before either. `find_arrow_at_depth_zero` now *does* visit the last byte, but that is the final iteration and `bytes.get(i + 1)` is `None` there, so the return value is unchanged.
- `find_assignment_eq`'s `skip_until = i + 2` is equivalent to the old `i += 2`, and the two bytes it skips (`=`/`>`) carry no depth.
- No `depth` is used as an index anywhere, so a negative depth cannot underflow a slice.

The remaining 14 tests all discriminate — I checked each against the pre-PR function body. `assignment_eq_ignores_comments_and_strings`'s `find_assignment_eq("'=' + x")` is a particularly good one: that scanner tracked *nothing* before, so it returned `Some(1)`.

baseballyama:
**changes requested** — 1 blocking. The fix is correct at 5 of the 7 converted sites but **spot-correct at the other 2**: both continuation loops thread `acc_template` and then never read it, so the same regression survives — and at the collection site it now tears a reactive statement apart rather than merely re-routing the decision.

Re-reviewed `8a2836f7` + `b1f900e2` on top of my earlier comment. Read-only, no build.

---

### 1. BLOCKING — `acc_template` is written and never consulted, in both continuation loops

`transform_legacy.rs:938`:

```rust
if !is_cont && !following_starts && acc_depth <= 0 { break; }
```

`transform_legacy.rs:1084`:

```rust
if !next_is_continuation && !following_starts_cont && accumulated_depth <= 0 { break; }
```

Neither mentions `acc_template`, while the collection loop it is modelled on correctly uses `depth > 0 || in_template_literal` (`:995`). Since `scan_reactive_line` **returns early** on an unterminated backtick (`:801-804`) without counting anything on that line, an interpolation-opening continuation line contributes `0` — which is exactly the −1/0 asymmetry `8a2836f7` set out to remove, just moved one branch over.

Concrete input, four lines:

```
$: msg =
	`hello ${
name
}`;
```

- `needs_reorder`: `$: msg =` → depth 0, `in_template` false, ends with `=` → continuation loop, `acc_depth = 0`, `acc_template = false`. Line 2 `` `hello ${ `` → `scan_reactive_line` sets `acc_template = true` and returns, `acc_depth` stays 0. `nl` is `{`, not in the continuation set; the next line `name` starts with `n`, not in the follow set; `acc_depth <= 0` → **break**. Line 3 `name` then reads as non-reactive content after a `$:` → `needs = true`.
- On `main` the raw counter scored `${`'s `{` as +1, so `acc_depth` was 1, the loop ran on through `` }`; ``, and `needs` stayed **false**.

That alone would only change the routing. The second site makes it worse: `needs_reorder = true` sends the same script into the collection path, whose continuation accumulator (`:1042`/`:1055`) breaks at the identical point. `stmt_lines` ends up as `["$: msg =", "\t`hello ${"]`, and `name` / `` }`; `` fall through to `non_reactive_lines` — which the assembler emits **before** the reactive statements (`:1118-1131`). The statement's own body is split and reordered around it. On `main` the whole thing stays one statement.

Fix is one term at each site: `&& !acc_template`, mirroring `:995`. Please also add a pin for the continuation shape — the existing `interpolation_brace_does_not_end_the_reactive_block_early` uses a `$: {` block, which routes through the *multiline* branch and cannot reach either of these loops.

While you are there: a `&mut` out-parameter that is never read is invisible to `clippy`, which is why this survived. If `scan_reactive_line` returned the template state instead of mutating it through a reference, both sites would have failed to compile.

### 2. The new pin does discriminate, and its stated control is right

I checked it rather than taking the note. For `"let name = 'w';\n$: b = 1;\n$: { msg = `hello ${\nname\n}`;\n}\n"`:

- **`main`**: raw counting scores the block's `{` **and** the interpolation's `{` (+2), the closing `}` on line 4 (−1 → 1), the block's `}` on line 5 (−1 → 0). `needs = false` → `sort_reactive_in_place` → output equals the input. Passes.
- **`8a2836f7~1`** (the intermediate this PR created): `bracket_depth_delta` skips `${` but counts line 4's `}`, so `reactive_depth` hits 0 one line early, line 5's `}` reads as non-reactive content, `needs = true` → the reorder path — which unconditionally emits a blank line before the reactive block (`:1118`, `result.push('\n')`). Output gains a blank line after `let name = 'w';` and no longer equals `script.trim_end()`. **Fails.**
- **head**: `in_template` is true entering line 4, `template_literal_close` consumes through the backtick and the `}` is never counted, so the block closes on line 5. `needs = false`. Passes.

So the comment at `:2188` — *"Control is this change's own pre-fix state; the code it replaced passes too"* — is accurate, and worth keeping: a reviewer running the usual "does it fail on main?" check would otherwise read this pin as vacuous. Your account of the five identical probes is also consistent with the mechanism: with content after the block both branches set `needs = true` and converge, so the blank-line difference never appears.

### 3. The replaced assertion now discriminates

`assert!(export_let_declaration_seems_complete("x = [1 /* ] */ ]"))` (`:2096`). On `main`: `[` → +1, the `]` inside the comment → 0, the real `]` → **−1**, so `bracket_depth != 0` and it returns `false`. On head: the comment is skipped, `[`/`]` balance to 0, `last_code_end` lands past the final `]`, the trailing-operator check sees `]` → returns `true`. Old `false` vs new `true` — the vacuity I flagged (both sides returning `false`, old via `ends_with('/')`) is gone, and the input is semantically complete (`[1]` with a comment). Good replacement.

### 4. Completeness of the rest — checked, clean

- `bracket_depth_delta` is **gone**, not merely unused: `grep` over the file at head returns zero hits, and all 7 former call sites now call `scan_reactive_line`. No other caller could have depended on the accidental `+1/−1` cancellation.
- Cross-line state initialisation at the 5 sites I could reach:
  - `:852` (`in_reactive_multiline`) uses the function-scoped `in_template` and correctly gates on `reactive_depth <= 0 && !in_template`.
  - `:864` resets `in_template = false` before scanning each new `$:` line — right, a new statement must not inherit state.
  - `:909` seeds `acc_template = in_template`, which is provably `false` there (the branch is the `else` of `depth > 0 || in_template`), so it is a defensive no-op rather than a bug.
  - `:1042` seeds a fresh `acc_template = false` and re-scans `trimmed`; that re-derives the same value the `:990` scan already produced, since that branch is likewise the `else` of `|| in_template_literal`. Redundant, not wrong.
  - `:1167` (`sort_reactive_in_place`) is fresh per statement with the loop gated on `depth > 0 || in_template`. Correct.
- Threading `in_template` is also a strict improvement beyond the regression: a multi-line template with **no** interpolation at all — `` $: s = `a `` / `` b`; `` — was invisible to the old raw counter (delta 0 on both lines) and is now collected.
- Unchanged shared limitation, not a regression: `template_literal_close` (`:769`) finds the first unescaped backtick and does not skip `${…}`, so a nested backtick inside an interpolation split across lines still mis-closes. `main`'s backtick toggle had the same hole.

baseballyama:
Re-reviewed at `65ce783b`. **1 blocking.** The design change is the right one and it did its job — it turned a silent dead write into a compile error and surfaced a second one neither of us had named. But the completeness claim in the commit headline ("consume the reactive scanner's template state at **every** site") does not hold: both accumulation loops still have a second exit that ignores the template state, and reaching it produces the same non-parsing output you reported for the previous state.

Read-only review — no build, no test run. Everything below is derived from the code at head; the traces are line-by-line and I have marked the one step I could not check.

---

### Verified first: the new pin is a true repro

Traced `continuation_interpolation_keeps_the_statement_whole` against `b1f900e2`. `$: msg =` ends in `=`, so it takes the continuation branch at `:1039`; the first collected line `` `hello ${ `` leaves `accumulated_depth == 0` (the `{` is inside the opaque backtick run) and sets `acc_template`; the trailing `{` is not in the continuation-char set and `name` does not start with one, so on `b1f900e2` the gate `!next_is_continuation && !following_starts_cont && accumulated_depth <= 0` is **true** and the statement breaks after `` `hello ${ `` — an unterminated template literal. At `65ce783b` the added `&& !acc_template` blocks that break, and the following two lines close the interpolation and the template, so the whole statement is collected. **The pin fails on the stated control and passes at head, for the stated reason.** ✓

### Verified: the return-value form is complete *as far as the compiler can see*

Nine call sites (`:855, :866, :912, :993, :1002, :1045, :1059, :1174, :1180`). At every one the tuple is destructured into a binding that some subsequent condition consults — there is no site that binds the template half and never reads it, which is the exact defect of the previous round. `#[must_use]` plus returning the state genuinely retires "written and never read".

**But that is a weaker property than it looks, and the gap is where this PR's remaining bug lives.** Returning the state guarantees the value is *bound* and read *somewhere*. It cannot see "consulted at one decision and not at a sibling decision in the same loop" — the binding is live either way, so nothing warns. That residual class is invisible to exactly the same degree the old one was visible, and it is what remains here.

### Blocking: both accumulation loops have a second exit that ignores `acc_template`

The new gate term was added to the "does this line look complete?" exit. Both loops also have a *boundary* exit at the top, before the scan, and it is unconditional:

- `:907-910` — `if nt.is_empty() || nt.starts_with("$:") || nt.starts_with("function ") { break; }`
- `:1049-1055` — the same three plus `|| next_trimmed.starts_with("//")`, and it breaks **before** `stmt_lines.push(next)` at `:1056`

Inside a template literal those are not boundaries — they are literal text. A blank line, a line reading `function foo() {}`, a line starting `//`, are all just characters between backticks.

Repro, which differs from your own pin only in what appears on the third line:

```
let a = 1;
$: msg =
`hello

${name}`;
let after = 2;
```

Trace at `65ce783b`, collection pass: `$: msg =` → `depth == 0`, no template → `is_continuation` (ends `=`) → the loop at `:1046` with `(0, false)`. Iteration 1 takes `` `hello ``, pushes it, and `scan_reactive_line` returns `(0, true)` — unterminated backtick. The `:1088` gate now correctly declines to break, **which is your fix working**. Iteration 2 then reads the blank line, hits `next_trimmed.is_empty()` at `:1049`, and breaks *before pushing*.

`stmt_lines` is `["$: msg =", "`hello"]`. The emitted reactive statement ends inside an unterminated template literal, and `${name}`;` is left behind to be emitted as ordinary code. **The output does not parse** — the same severity, the same mechanism, one door over.

Two things make this the completeness question rather than a nit:

1. **The sibling shape is untested precisely because the pin picked the other door.** Both inputs open the template on the continuation line; the pin's third line is `name`, the repro's is empty. The gate term fixes "stop because the statement looks complete"; nothing fixes "stop because the line looks like a boundary".
2. **The first branch of the same `if` has no such hole.** When the `$:` line itself opens the template (`` $: msg = `hello ``), collection runs through `while i < lines.len() && (depth > 0 || in_template_literal)` at `:998` with no break-set at all, so blank lines inside the template are collected correctly. The two paths through the same statement type disagree, which is why this reads as an oversight rather than a decision.

Suggested fix, which keeps the state consumed at *both* exits:

```rust
if !acc_template
    && (nt.is_empty() || nt.starts_with("$:") || nt.starts_with("function "))
{
    break;
}
```

and the four-way equivalent at `:1049`. The runaway risk (a stray backtick swallowing the rest of the block) is the same one `:998`'s loop already accepts, so this does not introduce a new failure mode. Please pin the blank-line variant above; `function `/`//` inside a template are the same hole and one pin per loop is enough.

### Secondary, pre-existing, not blocking — the two loops' break-sets already disagree

`:848-849` states the invariant outright: *"Must match the collection loop below, or the two disagree on where a reactive block ends and this scan mis-routes."* They do not match. The `needs_reorder` scan breaks on three prefixes; the collection loop breaks on four — it adds `//`. So a comment line inside a `$:` continuation:

```
$: total =
// running sum
a + b;
```

is scanned through by the first loop and terminates the second. I did **not** establish that this changes emitted output today — that needs a run I cannot do — but a stated invariant that is false on inspection is worth either restoring or rewriting, and it is adjacent enough to the fix above that fixing both break-sets at once is cheaper than twice.

### Worth keeping from this round

The `&mut` → return change is the reusable half, and it is worth stating precisely what it bought, because the boundary is instructive: it converts *never consulted* into a compile error, and leaves *inconsistently consulted* exactly as invisible as before. The first is what bit the previous round; the second is what is left. Threading state through a loop with more than one exit needs the exits enumerated, not just the state returned.

baseballyama:
The `//` break-set asymmetry flagged in review is filed as #2387 rather than carried here, with the three IDENTICAL probes and the three shapes they did *not* cover written down so the next attempt starts from a gap, not from zero. This PR corrects only the comment at `:848`, which asserted a correspondence that does not hold.

baseballyama:
Re-reviewed at `d00b34c2`. **0 blocking.** The fix is at both sites and the enumeration is the right method. I reproduced the attribution independently rather than accepting it, and it holds — **but it is shape-dependent in a way the current framing loses**, and the distinction matters because it decides whether this PR is required or optional.

Read-only — no build, no test run. Everything below is derived from the code at `88b8d2b7`, `b1f900e2` and `d00b34c2`.

### The fix itself

`!acc_template && (…)` at both `:907` and `:1051`, wrapping the whole boundary set at each. That is the complete form: the leading guard and the trailing gate now consult the same state, and no third exit exists in either loop. The new `template_text_is_not_a_statement_boundary` covers all four boundary tokens through one parameterised body, which is the right granularity — the four share one mechanism, so four separate tests would have added length, not coverage.

### Reproducing the attribution: `main` is broken for the *leading-guard* shapes, and correct for the *gate* shape

I traced the two exits against `88b8d2b7`, whose continuation loop (`:1163-1190`) counts brackets with a bare `for c in next_trimmed.chars()` — no string, comment or template awareness at all.

**Leading-guard shapes (this PR's four): `main` is equally broken.** Take `` $: msg = `` / `` `hello `` / blank / `` ${name}`; ``. On `88b8d2b7`, `` `hello `` contributes no brackets, so `accumulated_depth` stays 0; `next_last` is `o` and the following line is blank, so the gate `!next_is_continuation && !following_starts_cont && accumulated_depth <= 0` fires and the loop breaks **at iteration 1** — before it ever reaches the blank line. Same truncation, same unterminated template literal, one exit earlier. Your correction is right: for these shapes this PR is an improvement over `main`, not a restoration.

**The gate shape is the opposite, and it is the one your own earlier pin covers.** `` $: msg = `` / `` `hello ${ `` / `name` / `` }`; ``. On `88b8d2b7` the naive char loop counts the `{` of `${` as a bracket, so `accumulated_depth` becomes 1, the gate does not fire, and the statement is collected **whole and correct**. On `b1f900e2` the backtick run became opaque, so that `{` stopped being counted while the gate still had no `!acc_template` term — depth 0, break, unterminated template. That shape *is* a batch-2 regression, `main` *is* correct on it, and `continuation_interpolation_keeps_the_statement_whole` is a genuine parity restoration.

So the accurate statement is not "this was not required" but: **the two exits have different histories.** The gate term restored parity that batch 2 broke; the leading-guard term fixes a bug `main` shares. Both belong in this PR, but only the first was mandatory under a strict no-regression standard. Worth saying that way in the PR body, because "`main` emits the same unterminated template" reads as covering the pin too, and for the pin's shape it is false.

### The two-sided safety argument

**(b) is stronger than measured — it is structural.** `!acc_template && (…)` reduces to the original condition whenever `acc_template` is false, so every input that never opens a template literal in a `$:` continuation takes a bit-identical path. That is a proof, not a sample; the byte-identical control confirms it rather than establishing it. Worth stating in that form, since a control can only cover the inputs it ran.

**(a) is true in the form that matters and overstated in the form written.** "No corpus entry can be passing on it" holds: an unterminated template literal cannot survive the comparison — it does not parse, so it cannot be normalised into a MATCH, and the ratchets are at 0/1. That is the claim the gate needs.

But it rests on rsvelte's output over corpus inputs, not on anything about official. "Official never emits that shape" is neither needed nor establishable — official does not produce this text at all; rsvelte does, from ordinary Svelte 4 input. And the corpus statement says nothing about the wild: a blank line inside a template literal in a `$:` continuation is unremarkable code that no corpus repository happens to contain. So the safety argument supports **"no gate regression"**, and should not be read as **"no user impact"** — the latter is exactly the inference the #2281 note warns against, since the collected corpus samples the marginal distribution of published code and every shape here is an interaction.

### The `//` break-set: I agree with not unifying, and I did not find the divergence — I found something adjacent

Your probes are right that the *set difference* is not observable, and I can say why rather than only reproducing it. For `` $: total = `` / `// c` / `a + b;`, the decision loop scans through `// c` (a comment, depth 0) and then breaks at the gate with the cursor on `a + b;`; the collection loop breaks *before* `// c`. The two loops genuinely disagree on where the block ends — but `needs` comes out `true` either way, because a comment line is not a statement and `a + b;` follows a reactive statement in both readings. The disagreement is real and unobservable, which is exactly what three IDENTICAL probes should look like. Correcting the comment to a true conditional rather than forcing the sets together was the right call.

**What I did find is a defect in the `//` entry itself, independent of any disagreement.** At `d00b34c2`, `acc_template` is false for:

```
let a = 1;
$: total =
// running sum
a + b;
let after = 2;
```

so `next_trimmed.starts_with("//")` still fires at `:1054` and `stmt_lines` truncates to `["$: total ="]` — a dangling `=`, with `// running sum` and `a + b;` reclassified as non-reactive and emitted before it. A comment between `=` and its operand is ordinary code. This is pre-existing (`88b8d2b7`'s loop has the same four-way set), out of scope here, and I could not run it — but it is the same family as the four shapes you just fixed, differing only in that the boundary token is outside a template rather than inside one. Worth an issue rather than a hunk in this PR; one command against the fixture confirms or kills it.

### Method note

Enumerating the exits and reporting the count — six, two wrong — rather than fixing the one the repro reaches is what made this checkable, and it produced the second finding on the first build. That is the generalisable half: returning the state told you it was consulted *somewhere*, and only enumerating the exits told you *where it was not*.

baseballyama:
Rebased onto `origin/main` (`3aeac434`) after #2394 landed in the same file. Head `8292661f` → **`4e2fc94d`**, `--force-with-lease`.

**One conflict, and its `main` side was a real semantic change.** #2394 added the `!is_member_prop` guard to `extract_simple_assignments`; the conflict itself was in `mod tests` (both sides append), and the production hunk auto-merged with all four `is_member_prop` uses intact — verified by count against `origin/main`.

**The tests were re-checked for discriminating power, not just for passing.** A green suite proves the fix still works; only the pre-fix comparison proves the tests still prove it. Removing both `!acc_template` guards on the rebased tree and re-running each of the four template shapes independently:

| body inside the template | guards removed | guards present |
|---|---|---|
| blank line | split | intact |
| `$: not a statement` | split | intact |
| `// not a comment` | split | intact |
| `function nope() {}` | split | intact |

All four still discriminate, so none of them survived the rebase as a passing-for-another-reason test. Restored and re-verified afterwards.

23 `transform_legacy` tests pass (batch 2's 18 + #2394's 4 + one adjacent pin), bare clippy clean, rustfmt clean. `git diff --name-only $(git merge-base origin/main HEAD) HEAD` → the two expected files, single author.

The unreachability re-label is unchanged: this code is still not reachable by a `$:` in a shipping compile, and #2394 is where that was established.

Issue #800, opened 2026-06-08
`--tsgo`: same-name `Foo.svelte.ts` companion breaks resolution of `./Foo.svelte` (loses default + `<script module>` named exports)


baseballyama:
## Summary

When a component `Foo.svelte` (with `<script module> export const …` **and** a default export) has a **same-name companion `Foo.svelte.ts`**, the overlay mis-resolves `import … from './Foo.svelte'`: it reports the module-context named exports as "not exported", **and loses the default export too** ("has no default export"). Both the companion file and a barrel re-export break. The official `svelte-check` resolves all of it (0 errors).

Without the companion, the same `<script module> export const` import resolves correctly — so the trigger is specifically the presence of a same-name `Foo.svelte.ts` alongside `Foo.svelte`. (Distinct from #751 — that was importing *from* the companion; this is the companion/barrel importing the **component's** module exports + default.)

## Minimal repro

```svelte
<!-- H.svelte -->
<script module lang="ts">export const ATTR = 'data-x' as const;</script>
<script lang="ts">let { n }: { n: number } = $props();</script>
<span data-x={ATTR}>{n}</span>
```
```ts
// H.svelte.ts  (same-name companion)
import H, { ATTR } from './H.svelte';
export const useAttr = (): string => ATTR;
export const Comp = H;
```
```ts
// idx.ts  (barrel)
export { default as H, ATTR } from './H.svelte';
```

Official `svelte-check`: **0 errors**.
rsvelte `--tsgo`:
```
H.svelte.ts:1  Circular definition of import alias 'ATTR'.
H.svelte.ts:1  Module './H.svelte' declares 'ATTR' locally, but it is not exported.
H.svelte       has no default export.
idx.ts:1       Module './H.svelte' has no exported member 'default'.
idx.ts:1       Module './H.svelte' declares 'ATTR' locally, but it is not exported.
```

Removing `H.svelte.ts` makes all of these disappear.

## Likely cause

With a same-name `Foo.svelte.ts` present, the resolver seems to merge/shadow `./Foo.svelte` with the companion's module record (or treats the companion as the source of truth for `./Foo.svelte`), so the component overlay's `default` export and its `<script module>` named exports are no longer visible. `./Foo.svelte` should always resolve to the **component** overlay (default + `<script module>` exports); `./Foo.svelte.ts` / `./Foo.svelte.js` is the separate companion module.

[...]

PR #2367, opened 2026-08-06
test: fail loudly when a CI-provided prerequisite is missing (#1791)

baseballyama:
Part of #1791 (「実は何も検証していないテスト」の棚卸しと修正).

## The hole

`crates/*/tests/**` contains **30 early-return skip sites** across **12 files** that bail before
any assertion when a prerequisite (a submodule, `node`, a TypeScript compiler) is missing. libtest
reports a body that returns early as **PASS**, not as a skip.

Two of those files already carried the correct guard —
`svelte_check_shim_drift.rs:61` and `svelte_check_golden.rs:199` — and one of them says in its own
comment why:

> A silent skip here is exactly how the TypeScript half of this test greened for every
> #1883-#1889 report; on CI a missing compiler means the job is misconfigured, not that the
> assertions are unavailable.

That guard was never applied to the other ten files. This PR applies it.

## What decides which sites are safe to harden

`ci.yml:186` (the sharded `test` job) checks out exactly
`svelte language-tools eslint-plugin-svelte esrap`, installs Node, and exposes a TS compiler via
`TSGO_BIN`. Those prerequisites are unconditional for every job that runs these suites, so their
absence under `CI` is a misconfiguration and must be loud.

`submodules/svelte.dev` is **explicitly excluded** there (ci.yml:180-184 — it only exists in the
dedicated `test-fmt-corpus` job), so the two formatter-corpus suites correctly keep their soft
skip and are **not** touched.

## Fixed here — 20 sites / 9 files

| File | Sites | Prerequisite |
|---|---|---|
| `crates/rsvelte_check/tests/svelte_check_warning_filter.rs` | 9 | `node` |
| `crates/rsvelte_check/tests/svelte_check_golden.rs` | 3 | `language-tools` + fixture dir names |
| `crates/rsvelte_check/tests/svelte_check_companion_800.rs` | 2 | `tsgo` / `tsc` |
| `crates/rsvelte_fmt/tests/cli/tailwind.rs` | 1 | `node` |
| `crates/rsvelte_fmt/tests/cli/daemon.rs` | 2 | `node` |
| `crates/rsvelte_fmt/tests/embed.rs` | 2 | `node` |
| `crates/rsvelte_esrap/tests/esrap_samples.rs` | 1 | `esrap` submodule |
| `crates/rsvelte_lint/tests/eslint_plugin_oracle.rs` | 1 | `eslint-plugin-svelte` submodule |
| `crates/rsvelte_projection/tests/svelte2tsx_fixtures.rs` | 1 | `language-tools` submodule |

Where several sites share one helper (`node_available`, `node_runnable`, `node_bin`,
`fixture_root`) the assertion goes **in the helper**, so the guard cannot be forgotten at a future
call site.

`svelte_check_golden.rs` also gains `require_fixture()`. Its three tests hardcode the upstream
directory names `test-success` / `test-error`; that is precisely issue item 4's failure mode
(「upstream のリネーム1つでカテゴリが無言でゼロになる」), which is what produced `Sourcemaps 0/0`
in #1787. An upstream rename now fails on CI instead of silently zeroing the suite.

## Demonstration that the guards actually fire

Measured, not assumed. The "deliberately broken code" for this class is a broken *environment*,
since that is the exact condition being asserted against — and **two of these suites turned out to
be vacuously green on this machine already**, which is the finding rather than a contrivance.

`submodules/esrap` is initialised here but has no `test/samples`; `submodules/language-tools` has no
`packages/svelte-check` at all. Both suites reported PASS in 0.00s before this change:

```
$ cargo test -p rsvelte_esrap --test esrap_samples
test esrap_samples_match ... ok            # 0.00s, asserted nothing

$ ./svelte_check_golden-<hash>
test result: ok. 3 passed; 0 failed        # 0.00s, asserted nothing
```

With `CI=1` they now fail by name:

```
$ CI=1 cargo test -p rsvelte_esrap --test esrap_samples
panicked at crates/rsvelte_esrap/tests/esrap_samples.rs:117:
submodules/esrap is not checked out while running under CI — the sample-parity
assertions would be silently skipped.
test result: FAILED. 0 passed; 1 failed

$ CI=1 ./svelte_check_golden-<hash>
test result: FAILED. 0 passed; 3 failed
```

For `svelte_check_warning_filter` (node on `$PATH`) a three-value control separates "the guard
fires" from "the guard fires for the right reason". The middle row is the old behaviour:

| `node` | `CI` | result |
|---|---|---|
| present | unset | 11 passed |
| **absent** | **unset** | **11 passed** ← pre-change behaviour: vacuous green |
| **absent** | **`CI=1`** | **9 failed** ← equals the 9 skip sites exactly |

The bottom row failing exactly 9 of 11 (the other two tests call `apply()` in-process and need no
Node) confirms the assertion is attached to the skip sites and not firing globally. The middle row
confirms local runs on a machine without the toolchain are unaffected.

`cargo clippy -p rsvelte_check -p rsvelte_esrap -p rsvelte_fmt -p rsvelte_lint -p rsvelte_projection
--all-targets --all-features -- -D warnings` is clean, with no `-A` suppressions.

## Deliberately NOT fixed (itemised for follow-ups on #1791)

1. **`crates/rsvelte_fmt/tests/svelte_dev_markdown.rs:114`** and
   **`crates/rsvelte_formatter/tests/svelte_dev_corpus.rs:196`** — gated on `submodules/svelte.dev`,
   which the main `test` job deliberately does not check out. A blanket `CI` assert would be wrong;
   these need a job-scoped variable (e.g. assert on `FMT_CORPUS_OXFMT` being set) instead. **This is
   the highest-value remaining item** — it is the only one where the prerequisite is genuinely
   absent in some CI jobs and present in others, which is the condition under which a wrong guard
   is worse than none.
2. **`crates/rsvelte_fmt/tests/cli/tailwind.rs:204,239,470,503`** — gated on
   `RSVELTE_FMT_TW_TEST_DIR`, set only at `ci.yml:613`. Same job-scoped shape as (1). The `node`
   half of this file is fixed here; the Tailwind-env half is not.
3. **`STRICT_S2TSX_FIXTURES` / `STRICT_SOURCEMAPS`** — opt-in strictness flags that appear in no
   workflow, `package.json`, or script. These are not vacuous-pass bugs (the non-strict path still
   ratchets), but they are dead switches worth either wiring up or deleting.

## False positive I am retracting

`crates/rsvelte_projection/tests/svelte2tsx_fixtures.rs:36` reads its ratchet with a
`let Ok(text) = ... else { return BTreeSet::new() }`. This looks like the `unwrap_or_default()`
shape from issue item 7, but it is **not** a vacuous pass: an empty baseline makes every known
failure count as a *new* regression, so a missing ratchet fails loudly rather than greening. Only
the skip at line 71 in that file was a real defect.

Also worth recording as a genuine negative: a full sweep of `scripts/**/*.mjs` (70 files, ~25 of
them actual gates) found **no** surviving zero-match-passes-silently harness. PRs #2234 and #2245
closed that population, and the remaining `?? []` baseline reads all fail in the strict direction.

## Scope

`test(...)` only — no shipped behaviour changes, so no changeset. Nine test files, +97/-10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,6 +291,12 @@ jobs:
         run: bash scripts/ci/run-test-shard.sh ${{ matrix.shard }} 3
         timeout-minutes: 30
         env:
+          # This job checked the submodules out above and supplies node and a
+          # TS compiler, so the guarded suites must fail rather than skip when
+          # one is missing. Only a job that actually supplies them may set this:
+          # `publish-crates.yml` runs the same test binaries with no submodules
+          # and must keep skipping, which is why the guards cannot key on `CI`.
+          RSVELTE_REQUIRE_PREREQS: '1'
           # Some fixtures (notably the compatibility_report harness) walk
           # deeply nested ASTs and overflow the default 8 MB thread stack.
           RUST_MIN_STACK: '33554432'
@@ -611,6 +617,8 @@ jobs:
         env:
           OXFMT_BIN: ${{ runner.temp }}/fmt-oracle/node_modules/.bin/oxfmt
           RSVELTE_FMT_TW_TEST_DIR: ${{ runner.temp }}/tw-oracle
+          # `--test cli` also carries node-gated tests; this job has node.
+          RSVELTE_REQUIRE_PREREQS: '1'
 
   language-server:
     name: Language server
--- a/crates/rsvelte_check/tests/svelte_check_companion_800.rs
+++ b/crates/rsvelte_check/tests/svelte_check_companion_800.rs
@@ -72,6 +72,13 @@ fn run_check(dir: &Path) -> Vec<String> {
 #[test]
 fn companion_svelte_ts_does_not_shadow_component_module() {
     if find_compiler(&PathBuf::from("."), true).is_err() {
+        // CI provides a TypeScript compiler via TSGO_BIN, so a miss there means
+        // the job is misconfigured rather than that the check is unavailable.
+        assert!(
+            std::env::var_os("CI").is_none(),
+            "no `tsgo` / `tsc` binary found while running under CI — the \
+             companion-module assertions would be silently skipped."
+        );
         eprintln!("skip #800: no tsgo/tsc found");
         return;
     }
@@ -131,6 +138,13 @@ fn companion_svelte_ts_does_not_shadow_component_module() {
 #[test]
 fn no_companion_means_no_augmentation() {
     if find_compiler(&PathBuf::from("."), true).is_err() {
+        // CI provides a TypeScript compiler via TSGO_BIN, so a miss there means
+        // the job is misconfigured rather than that the check is unavailable.
+        assert!(
+            std::env::var_os("CI").is_none(),
+            "no `tsgo` / `tsc` binary found while running under CI — the \
+             companion-module assertions would be silently skipped."
+        );
         eprintln!("skip #800: no tsgo/tsc found");
         return;
     }
--- a/crates/rsvelte_check/tests/svelte_check_companion_800.rs
+++ b/crates/rsvelte_check/tests/svelte_check_companion_800.rs
@@ -9,7 +9,7 @@
 //! exports. The same fixture pins #751 (named imports resolved *from* the
 //! companion via `./Foo.svelte.js`) so neither direction regresses.
 //!
-//! Real-tsc/tsgo e2e; skipped when no tsgo/tsc is found.
+//! Real-compiler e2e; skipped without TSGO_BIN or @typescript/native-preview.
 
 use std::fs;
 use std::path::{Path, PathBuf};
@@ -72,14 +72,15 @@ fn run_check(dir: &Path) -> Vec<String> {
 #[test]
 fn companion_svelte_ts_does_not_shadow_component_module() {
     if find_compiler(&PathBuf::from("."), true).is_err() {
-        // CI provides a TypeScript compiler via TSGO_BIN, so a miss there means
-        // the job is misconfigured rather than that the check is unavailable.
+        // Only a job that promised a TypeScript compiler may fail on its absence.
         assert!(
-            std::env::var_os("CI").is_none(),
-            "no `tsgo` / `tsc` binary found while running under CI — the \
-             companion-module assertions would be silently skipped."
+            std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_none(),
+            "no TypeScript 7 native compiler in a job that declares \
+             RSVELTE_REQUIRE_PREREQS — set TSGO_BIN, or install \
+             @typescript/native-preview. The companion-module assertions \
+             would be silently skipped."
         );
-        eprintln!("skip #800: no tsgo/tsc found");
+        eprintln!("skip #800: no TSGO_BIN and no @typescript/native-preview");
         return;
     }
     let dir = std::env::temp_dir().join(format!("rsvelte_800_{}", std::process::id()));
@@ -138,14 +139,15 @@ fn companion_svelte_ts_does_not_shadow_component_module() {
 #[test]
 fn no_companion_means_no_augmentation() {
     if find_compiler(&PathBuf::from("."), true).is_err() {
-        // CI provides a TypeScript compiler via TSGO_BIN, so a miss there means
-        // the job is misconfigured rather than that the check is unavailable.
+        // Only a job that promised a TypeScript compiler may fail on its absence.
         assert!(
-            std::env::var_os("CI").is_none(),
-            "no `tsgo` / `tsc` binary found while running under CI — the \
-             companion-module assertions would be silently skipped."
+            std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_none(),
+            "no TypeScript 7 native compiler in a job that declares \
+             RSVELTE_REQUIRE_PREREQS — set TSGO_BIN, or install \
+             @typescript/native-preview. The companion-module assertions \
+             would be silently skipped."
         );
-        eprintln!("skip #800: no tsgo/tsc found");
+        eprintln!("skip #800: no TSGO_BIN and no @typescript/native-preview");
         return;
     }
     let dir = std::env::temp_dir().join(format!("rsvelte_800_solo_{}", std::process::id()));
--- a/crates/rsvelte_check/tests/svelte_check_golden.rs
+++ b/crates/rsvelte_check/tests/svelte_check_golden.rs
@@ -13,13 +13,14 @@
 //!
 //!   * The Svelte-side assertions ("the Svelte compile is clean") run
 //!     unconditionally and are the part this test always enforces.
-//!   * The full TypeScript assertions only run when a `tsgo` / `tsc`
-//!     binary can be located via `find_compiler` — otherwise they're
-//!     skipped with a printed notice, so the test stays runnable on a
-//!     machine without a TS toolchain. Under `CI` that same condition is
-//!     a hard failure instead: a CI job that cannot type-check is
-//!     misconfigured, and a silent skip there is indistinguishable from
-//!     a pass.
+//!   * The full TypeScript assertions only run when `find_compiler` locates
+//!     a compiler — `$TSGO_BIN`, else `@typescript/native[-preview]`, since
+//!     these call it with `prefer_tsgo` and it never falls back to `tsc` on
+//!     `$PATH`. Otherwise they're skipped with a printed notice, so the test
+//!     stays runnable on a machine without a TS toolchain. In a job that
+//!     declares `RSVELTE_REQUIRE_PREREQS` the same condition is a hard
+//!     failure: that job promised a compiler, and a silent skip there is
+//!     indistinguishable from a pass.
 //!
 //! Run with:
 //!     cargo test --release --test svelte_check_golden -- --nocapture
@@ -38,11 +39,10 @@ fn fixture_root() -> Option<PathBuf> {
         .join("language-tools")
         .join("packages")
         .join("svelte-check");
-    // CI checks this submodule out unconditionally, so absence there means the
-    // job is misconfigured rather than that the fixtures are unavailable.
+    // Only a job that promised this submodule may fail on its absence.
     assert!(
-        p.exists() || std::env::var_os("CI").is_none(),
-        "submodules/language-tools is not checked out while running under CI — \
+        p.exists() || std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_none(),
+        "submodules/language-tools is not checked out in a job that declares RSVELTE_REQUIRE_PREREQS — \
          the svelte-check golden assertions would be silently skipped."
     );
     if p.exists() { Some(p) } else { None }
@@ -53,8 +53,8 @@ fn fixture_root() -> Option<PathBuf> {
 fn require_fixture(workspace: &Path) -> bool {
     let exists = workspace.exists();
     assert!(
-        exists || std::env::var_os("CI").is_none(),
-        "fixture workspace {} is missing while running under CI — it was \
+        exists || std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_none(),
+        "fixture workspace {} is missing in a job that declares RSVELTE_REQUIRE_PREREQS — it was \
          probably renamed upstream; update this test instead of skipping it.",
         workspace.display()
     );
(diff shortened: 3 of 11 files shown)


baseballyama:
**changes requested** — 1 blocking. Two of the guards will fail the crates.io release workflow, which does not run on this PR, so CI here will be green either way.

Read-only review — no build, no test run. I checked every guard against every workflow that invokes its test binary, not only against `ci.yml`'s `Test` job.

---

### 1. BLOCKING — `publish-crates.yml` runs two of the guarded binaries with no submodules checked out

`.github/workflows/publish-crates.yml:94` and `:96`, in the `preflight` job:

```yaml
- name: Test every public feature on the declared MSRV
  run: |
    cargo test --locked -p rsvelte_esrap --all-features
    cargo test --locked -p rsvelte_core --all-features --lib
    cargo test --locked -p rsvelte_projection --all-features
    cargo test --locked -p rsvelte --all-features
```

That job's checkout is `publish-crates.yml:35-37` — `actions/checkout` with only `fetch-depth: 0` and **no `submodules:` key**, i.e. no submodules at all.

`-p rsvelte_esrap --all-features` builds and runs every target of the crate, including `tests/esrap_samples.rs`. With `submodules/esrap` absent, `samples_dir()` returns `None` (`dir.is_dir().then_some(dir)`), the `let … else` fires, and the new assertion at `esrap_samples.rs:115-119` evaluates `std::env::var_os("CI").is_none()` — which is `false` in every GitHub Actions job, since the runner sets `CI=true` unconditionally. The job fails with *"submodules/esrap is not checked out while running under CI"*.

`-p rsvelte_projection --all-features` does the same through `tests/svelte2tsx_fixtures.rs`: `iter_svelte2tsx_outcomes()` returns `None` without `submodules/language-tools`, and the new assertion at `svelte2tsx_fixtures.rs:71-77` fires.

Both are true today and neither shows up here, because `publish-crates.yml` is `workflow_dispatch`-only and gated on `refs/tags/rsvelte-crates-v*`. So this lands green and detonates at the next crates.io publish — the exact "PR green, main/release explodes" shape the release-gate hardening work was about.

`crates-io-packages.yml:85-88` is **not** affected: it uses `--lib` for `rsvelte_esrap` and `rsvelte_projection`, and the `rsvelte` facade's only test binaries (`facade_projection_api.rs`, `runtime_api.rs`) are unguarded.

### 2. `CI` is the wrong selector, and fixing that fixes finding 1

The guards say "under CI" but mean "in a job that promised to supply this prerequisite". Those are different sets, and `publish-crates.yml` is the counterexample. A repo-specific variable set only by the jobs that actually provide the prerequisite — e.g. `RSVELTE_REQUIRE_PREREQS: 1` in `ci.yml`'s `Test`, `test-fmt-corpus`, and any future job — makes the guard say what it means, keeps `publish-crates` skipping as it does today, and removes the second-order hazard below.

Second-order hazard: plenty of local setups export `CI=true` (some editors, `act`, container images). A contributor without submodules then gets nine hard failures instead of nine skips — which is precisely how a guard gets disabled wholesale rather than fixed.

Minor, same area: `var_os("CI").is_none()` treats `CI=` (set but empty) as "on CI". `is_none_or(|v| v.is_empty())` is the usual form.

### 3. The `find_compiler(".", true)` root is *not* the defect — but the guard is coupled to the wrong thing

`svelte_check_companion_800.rs:74` / `:140` resolve with a cwd-relative root while `svelte_check_golden.rs:199` uses the fixture workspace. In CI that distinction is invisible: `find_compiler` returns `$TSGO_BIN` before `workspace` is ever read —

```
crates/rsvelte_check/src/svelte_check/tsgo.rs:101-108
    if let Ok(explicit) = std::env::var("TSGO_BIN") && !explicit.is_empty() {
        return Ok(TsgoBinary { program: explicit, args_prefix: Vec::new() });
    }
```

— and `ci.yml:304` sets `TSGO_BIN` for the `Test` job. So both call sites behave identically there and the guard will not fire. The root only matters locally.

What *is* worth changing is the coupling. `TSGO_BIN` exists in `ci.yml` for a different reason entirely — the comment at `ci.yml:296-303` says it is there to **pin TS 5.x** so the runner's TS 6 pre-release does not bail with TS5107. The moment someone removes that pin (say, when the runner image stops shipping TS 6), these two tests flip from skipping to failing, for a reason unrelated to what the guard claims to protect. Assert on the prerequisite you actually depend on.

Two smaller things in the same code:

- The messages ("no `tsgo` / `tsc` binary found", and the pre-existing `eprintln!("skip #800: no tsgo/tsc found")`) are wrong for this call. With `prefer_tsgo = true` and no `TSGO_BIN`, `find_compiler` takes the `resolve_ts7_native(workspace)` branch (`tsgo.rs:109-114`) and returns `Err(Ts7NotFound)` **without ever trying `tsc` or `$PATH`**. A developer with a working `tsc` still gets the skip, and now a misleading assertion text to go with it.
- Passing the fixture workspace instead of `"."` — matching `golden.rs:199` — would make the guard test the resolution path it names, and costs nothing.

### 4. The `svelte2tsx_fixtures.rs:36` retraction is correct

`read_baseline()` returns an **empty** `BTreeSet` when `compatibility/svelte2tsx-fixtures-known-failures.json` is missing or unreadable, and *panics* when it is present but not a JSON string array. So a lost baseline either fails loudly or makes every currently-known failure an unlisted new failure — strict in both directions, and the two-sided ratchet's other half ("a listed entry that now passes") is vacuous on an empty set. Retracting that one was right.

### 5. Guards I checked and found correctly supplied

All nine guarded binaries run in `ci.yml`'s `Test` job: `scripts/ci/run-test-shard.sh:56` excludes only `compile_module_thread_safety`, `svelte_dev_corpus`, `svelte_dev_markdown` from the binary partition, and none of those is guarded here. That job supplies every prerequisite the guards assert:

- `ci.yml:186` checks out `svelte language-tools eslint-plugin-svelte esrap` unconditionally → covers `svelte_check_golden.rs:41-56` (`language-tools`), `esrap_samples.rs` (`esrap`), `eslint_plugin_oracle.rs` (`eslint-plugin-svelte`), `svelte2tsx_fixtures.rs` (`language-tools`).
- `ci.yml:193` (`./.github/actions/setup-node-pnpm`) supplies `node` → covers `svelte_check_warning_filter.rs:node_available`, `rsvelte_fmt/tests/cli/daemon.rs:node_bin`, `cli/tailwind.rs:node_runnable`, `embed.rs:node_runnable`.
- `ci.yml:304` supplies `TSGO_BIN` → covers the two `svelte_check_companion_800.rs` guards (with the caveat in §3).

The one binary that also runs outside the shards, `ci.yml:609` (`-p rsvelte_fmt --test cli sort_tailwindcss` in `test-fmt-corpus`), reaches `cli/tailwind.rs:node_runnable`; that job runs `pnpm run generate-fmt-corpus`, so `node` is present. Fine.

`require_fixture` (`svelte_check_golden.rs:62-71`) is passed a file, not a workspace, at the `tsconfig` call site (`:207`), so the message *"fixture workspace … was probably renamed upstream"* is slightly off there. Cosmetic. The `!require_fixture(&workspace) || !require_fixture(&tsconfig)` short-circuit is correct: a missing workspace asserts before the second call is reached.

### 6. Nit — nine failures for one cause

Putting `assert!` inside shared helpers means one missing `node` produces nine identical failures in `svelte_check_warning_filter.rs`. A single `#[test] fn prerequisites_present()` per suite would fail once, name the cause, and leave the real tests skipping — easier to read in a CI log and much harder to "fix" by deleting one assertion at a time.

baseballyama:
Confirmed the blocking finding, reproduced it, and fixed it at the root as proposed. Pushed as `f038915f`.

## The finding is real

I verified it rather than taking it on trust:

- `publish-crates.yml:35-37` — checkout has only `fetch-depth: 0`, **no `submodules:` key**.
- `:94` `cargo test --locked -p rsvelte_esrap --all-features` builds and runs `tests/esrap_samples.rs`.
- `:96` `-p rsvelte_projection --all-features` builds and runs `tests/svelte2tsx_fixtures.rs`.
- GitHub Actions sets `CI=true` in every job.

The old guard's trigger was `CI` alone, so in that job it is satisfied while the submodules are
absent — the assert fires and the release preflight fails. `workflow_dispatch`-only and tag-gated,
so no PR can catch it.

I also swept the other invocations for the same exposure and found none:
`crates-io-packages.yml:85-88` and all four `cargo test` calls in
`scripts/ci/verify-registry-dependency-surface.sh:67-73` are `--lib`, which does not build the
integration binaries.

## Fix: assert on the job's promise, not on the runner

`CI` answers "am I on a runner". The question is "did **this** job promise me submodules / node /
a TS compiler". So the guards now key on `RSVELTE_REQUIRE_PREREQS`, set only by jobs that actually
supply the prerequisites:

- `ci.yml` test-shard `Run tests` step — covers all nine guarded binaries.
- `ci.yml` `sort_tailwindcss` step — `--test cli` also carries node-gated tests.
- `publish-crates.yml` — deliberately **not** set, so it keeps skipping.

Applied to all 22 sites including the two that predated this PR (`shim_drift.rs`,
`golden.rs`), so the class is retired rather than patched. I did not add `submodules:` to the
publish workflow: that fixes the symptom and leaves the next job that forgets the key armed.

On unifying with #2371: it is the same rule at a finer grain. `FMT_CORPUS_OXFMT` and
`RSVELTE_FMT_TW_TEST_DIR` are *also* "the signal the owning job sets" — they just additionally
name the tool, so they can catch declared-but-unusable, which a boolean cannot. I have kept them
rather than coarsening them, and the comments now state the one rule.

## Verification — three regimes, per binary

The middle row is the one that was broken:

| regime | `CI` | `RSVELTE_REQUIRE_PREREQS` | required | `esrap_samples` | `svelte2tsx_fixtures` |
|---|---|---|---|---|---|
| local dev | – | – | pass | ok | ok |
| **publish preflight** | **true** | **–** | **pass** | **ok** | **ok** |
| ci.yml test shard | true | 1 | fail | FAILED | FAILED |

Row 2 is the regression this fixes; row 3 confirms the guard still fires where it must, naming the
missing submodule. Under the old condition row 2 was identical to row 3 by construction, since
`CI` alone was the trigger.

`svelte_check_warning_filter` also passes 11/11 in the publish regime.
`cargo clippy -p rsvelte_check -p rsvelte_esrap -p rsvelte_fmt -p rsvelte_lint -p rsvelte_projection
--all-targets --all-features -- -D warnings` is clean, no `-A`. `ci.yml` parses.

## Secondary finding: the diagnostics were false

Verified and fixed. `find_compiler(workspace, prefer_tsgo: true)` resolves `$TSGO_BIN`, then
`resolve_ts7_native`, and returns `Err(Ts7NotFound)` — it **never** reaches the `tsc` / `$PATH`
branch (`tsgo.rs:101-115`). Both call sites pass `true`, so "no `tsgo` / `tsc` binary found" named
a fallback that cannot happen. Messages, `eprintln!`s and the module doc now say "no `TSGO_BIN` and
no `@typescript/native-preview`". A wrong diagnostic is a small version of the same disease.

The `TSGO_BIN`-as-version-pin coupling is real and I have **not** fixed it here — `ci.yml:296-303`
sets it to pin TS 5.x, and these tests read it as an availability signal, so removing the pin flips
them from skipping to failing. That is a separate change with its own risk; noting it for a
follow-up rather than folding it into a test-only PR.

## Not changed

The `svelte2tsx_fixtures.rs:36` retraction stands, and thanks for confirming it independently.

baseballyama:
Re-reviewed at `f038915f`. **The blocking `publish-crates.yml` finding is resolved, and resolved at the root. LGTM — 0 blocking.**

Keying on "did *this job* promise me prerequisites" instead of "am I on a runner" is the correct predicate, and declining to paper over it with `submodules:` in the publish workflow was the right instinct: that fixes one symptom and leaves the next job that forgets the key armed.

I verified the sweep by measurement at head rather than by reading the diff, because a diff cannot show what was left behind.

**The class is retired, not patched.** `git grep 'var_os("CI")' f038915f -- crates` returns **nothing**; so does a search for `GITHUB_ACTIONS`. There is no surviving guard on the old predicate.

**Site count — one thing to fix in the PR body.** At head there are **13** reads of `RSVELTE_REQUIRE_PREREQS` in guards, across 10 files: `companion_800` ×2, `golden` ×3, `shim_drift` ×1, `warning_filter` ×1, `esrap_samples` ×1, `daemon` ×1, `tailwind` ×1, `embed` ×1, `eslint_plugin_oracle` ×1, `svelte2tsx_fixtures` ×1. Those 13 guards cover the 22 *skip sites* the body counts — `warning_filter`'s single helper alone guards nine of them, as your own three-value table shows. Both numbers are right, but they appear in the same PR with no bridge between them and a reader cannot reconcile "22 sites" with 13 assertions. One clause naming the unit ("22 skip sites, guarded at 13 points, because a helper covers all callers") makes the completeness claim auditable instead of asking for trust.

**Setters — exactly two, and both are justified.** `git grep RSVELTE_REQUIRE_PREREQS f038915f -- .github` lists `ci.yml` only, at `:299` (test-shard `Run tests`) and `:621` (`sort_tailwindcss`). No other workflow sets it, so `publish-crates.yml` keeps skipping — which is the fix.

**I checked the direction that actually breaks CI**, i.e. whether a setter job can fail to supply something its guards demand. It cannot:

- the shard job's checkout inits `svelte language-tools eslint-plugin-svelte esrap` (`ci.yml:186`) — exactly the four submodules the guards name;
- `setup-node-pnpm` supplies node for the three node guards; `TSGO_BIN` is pinned to the cached Svelte build for the tsc guards; fixtures are generated in-job;
- the Tailwind step has node and installs the oracle.

**And the reverse — that the guards actually execute where the var is set.** `run-test-shard.sh:56` excludes only `compile_module_thread_safety`, `svelte_dev_corpus` and `svelte_dev_markdown`; none of the 10 guarded binaries is on that list, so all 13 guards really run under the var rather than being nominally covered by a job that never compiles them.

One non-blocking follow-up, stated as a distinction rather than a request. `publish-crates.yml` now correctly keeps skipping — but that means the release preflight's `-p rsvelte_esrap` / `-p rsvelte_projection` runs are still green *without checking anything*, which is where this started. Root-causing the guard and giving the publish workflow its submodules are not alternatives; the first stops the wrong failure, the second is what would make the preflight mean something. Worth an issue so it does not get closed along with this PR.

PR #2369, opened 2026-08-06
fix(compiler): route client/state_transforms delimiter scans through the shared lexer (#2357)

baseballyama:
Batch 3 of #2357: `client/state_transforms.rs`.

Based on `main` after #2363 landed, so `code_bytes` is available.

## Re-derived inventory (the epic's table is stale for this file)

The table says 40 sites / 0 `skip_opaque`. Reading it instead of grepping: the file
already imports `skip_opaque` and uses it in `find_assignment_position`, and the 40
`depth ±= 1` sites are **11 scanners**, not 40 candidates — `(`/`[`/`{` on one arm and
`)`/`]`/`}` on the next is two "sites" for one scan. Of those:

| scanner | disposition |
|---|---|
| `find_assignment_position` | already converted (batch 0) |
| `strip_function_scopes_that_shadow` — 4 scans | **converted** |
| `body_references_identifier_recursive` — 2 scans | **converted** |
| `body_references_identifier_in_statements` | **converted** |
| `transform_legacy_destructure_declarations` pattern scan | **converted** + char/byte fix |
| `split_multi_declarator` — 2 scans | **converted** (rewritten to cut points) |
| `strip_string_literal_text` | **comment + regex carve-out only** — see below |
| `find_colon_at_depth0` | **skipped** — see "still open" |
| `is_inside_string_literal` | **skipped** — answers the inverse question, `${…}` descent |
| `strip_object_property_keys` | **skipped** — no depth scan; blanking a comment cannot change the answer |
| `transform_legacy_state_declarations` TS-annotation scan | **skipped** — counts `<`/`>` as depth for TS generics; a separate change |

## Repro-first: 8 candidates written, 8 failed, 8 fixed

Nothing here was converted defensively. Every conversion has a test that fails on the
pre-fix code:

- an apostrophe in `// don't` opened a string literal and blanked every read after it
- a `'` inside a **regex literal** (`raw.replace(/'/g, …)`) did the same, through the other
  `/`-led opaque run
- `}` in a block comment, and in a line comment, closed a shadowing function body early
  (two tests — the block-body and arrow-body scanners are separate)
- `{` in a comment raised the depth so a real `;` did not split, folding two statements
  into one and reading the identifier off the merged RHS
- `,` in a comment split a single declarator into two
- `}` in a comment ended a destructuring pattern early, so the declaration was left
  untransformed
- **a panic**: `let { ああ } = obj;` — the pattern scan enumerated `chars()` and sliced
  the result as bytes (`end byte index 6 is not a char boundary`). This is the exact
  latent-panic class batch 1 flagged as constraint 2.

Two of the eight were retargeted mid-way: my first versions asserted an outcome the
scanner does not control (a comment surviving into the regex match, and a var-name
extractor that keeps comment text in the name). Both retargeted versions were
re-confirmed failing against the unmodified file before being kept.

## `${…}` carve-out

`strip_string_literal_text` is the one scanner here whose job is to descend into `${…}` —
`skip_opaque` would make the interpolation opaque and change output. It keeps its
hand-rolled backtick branch, and delegates the two `/`-led opaque runs — comments **and
regex literals** — to `skip_opaque` restricted to a leading `/`, in both the outer loop
and the interpolation-depth loop. `prev` is tracked so a division is not read as a regex
(pinned by a test: `a = n / 2` was green before and stays green). Comments are skipped,
not blanked, which is the smallest change that fixes the bug.

## Still open in this file

`find_colon_at_depth0` — **closed by #2376 while this PR was in review**, with a stronger
input than I could build: it panics through the *real* caller,
`check_identifier_in_statement("cond ? component = \"ああa\" : component = other", …)`,
which settles the reachability I had left open rather than arguing it. #2378 remains open
as the class-wide item (char indices escaping into byte slices), not the one function.

Also worth adding to the epic: **comments survive into `body_references_identifier`'s
regex match**. `strip_string_literal_text` blanks string and template *text* but not
comment bodies, so `$: y = a; // count twice` counts as a read of `count`. That is a false
positive in the **seam**, not scanner blindness — the same category as #2347 — and it is
what blocked two of my first test formulations.

### DIFF
--- a/.changeset/phase3-state-transforms-lexer-sweep.md
+++ b/.changeset/phase3-state-transforms-lexer-sweep.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): route client/state_transforms delimiter scans through the shared lexer
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/state_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/state_transforms.rs
@@ -13,7 +13,9 @@ use super::rune_transforms::{
 };
 use super::{STATE_TMP_COUNTER, get_or_compile_regex};
 use crate::compiler::phases::phase2_analyze::scope::DeclarationKind;
-use crate::compiler::phases::phase3_transform::shared::js_scan::skip_opaque;
+use crate::compiler::phases::phase3_transform::shared::js_scan::{
+    code_bytes, code_bytes_from, skip_opaque,
+};
 
 // ---------------------------------------------------------------------------
 // Identifier reference detection (lines 7653-8602 of mod.rs)
@@ -104,6 +106,29 @@ pub(super) fn body_references_identifier(body: &str, identifier: &str) -> bool {
 ///
 /// This prevents false identifier matches inside literal text, e.g., `<circle>` in
 /// a template literal won't match the variable name `circle`.
+/// Byte just past the comment starting at `i`, if one starts there. `skip_opaque`
+/// cannot be used here: this scanner's whole job is to descend into string and
+/// template literals, which `skip_opaque` treats as opaque.
+fn skip_comment(bytes: &[u8], i: usize) -> Option<usize> {
+    match (bytes[i], bytes.get(i + 1)) {
+        (b'/', Some(b'/')) => {
+            let mut j = i + 2;
+            while j < bytes.len() && bytes[j] != b'\n' {
+                j += 1;
+            }
+            Some(j)
+        }
+        (b'/', Some(b'*')) => {
+            let mut j = i + 2;
+            while j + 1 < bytes.len() && !(bytes[j] == b'*' && bytes[j + 1] == b'/') {
+                j += 1;
+            }
+            Some((j + 2).min(bytes.len()))
+        }
+        _ => None,
+    }
+}
+
 pub(super) fn strip_string_literal_text(code: &str) -> String {
     // Fast path: if no string delimiters exist, return as-is
     // Uses memchr3 for SIMD-accelerated search of all three delimiters at once
@@ -118,6 +143,12 @@ pub(super) fn strip_string_literal_text(code: &str) -> String {
     let mut i = 0;
 
     while i < len {
+        // A quote inside a comment (`// don't`) must not open a string literal
+        // and blank out every live read that follows it.
+        if let Some(next) = skip_comment(bytes, i) {
+            i = next;
+            continue;
+        }
         match bytes[i] {
             // Handle single/double-quoted strings
             b'\'' | b'"' => {
@@ -151,6 +182,10 @@ pub(super) fn strip_string_literal_text(code: &str) -> String {
                         // Find matching `}` - track depth
                         let mut depth = 1;
                         while i < len && depth > 0 {
+                            if let Some(next) = skip_comment(bytes, i) {
+                                i = next;
+                                continue;
+                            }
                             match bytes[i] {
                                 b'{' => depth += 1,
                                 b'}' => {
@@ -347,15 +382,14 @@ pub(super) fn strip_function_scopes_that_shadow(body: &str, identifier: &str) ->
             }
 
             // Find the opening brace of the function body
-            let after_pat = &result[after_ident..];
             let mut found_paren_close = false;
             let mut brace_start = None;
             let mut depth = 1; // We're inside the opening (
-            for (i, ch) in after_pat.char_indices() {
+            for (i, ch) in code_bytes_from(result.as_bytes(), after_ident) {
                 if !found_paren_close {
                     match ch {
-                        '(' => depth += 1,
-                        ')' => {
+                        b'(' => depth += 1,
+                        b')' => {
                             depth -= 1;
                             if depth == 0 {
                                 found_paren_close = true;
@@ -363,10 +397,10 @@ pub(super) fn strip_function_scopes_that_shadow(body: &str, identifier: &str) ->
                         }
                         _ => {}
                     }
-                } else if ch == '{' {
-                    brace_start = Some(after_ident + i);
+                } else if ch == b'{' {
+                    brace_start = Some(i);
                     break;
-                } else if !ch.is_whitespace() {
+                } else if !ch.is_ascii_whitespace() {
                     break;
                 }
             }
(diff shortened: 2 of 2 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. All seven tests discriminate, the index-unit conversions are correct including the relative→absolute ones, and the `${…}` carve-out is complete for the kinds it claims. Two gaps below.

Read-only review at `676279e` — no build, no test run.

---

### 1. The carve-out fixes one of the two ways a quote reaches non-string code

`strip_string_literal_text` (`state_transforms.rs:132`) now skips comments, but neither the outer loop nor the interpolation loop knows about **regex literals** — and a regex is the other place an unpaired `'` lives in real code:

```js
$: escaped = raw.replace(/'/g, '&#39;');
$: b = width;
```

The `'` inside `/'/g` opens a string literal, and everything up to the next `'` is blanked — so `body_references_identifier` loses reads that follow. That is byte-for-byte the failure mode the new `apostrophe_in_a_line_comment_does_not_blank_the_code_after_it` test pins for `// don't`, just reached through a regex instead of a comment. `str.replace(/'/g, …)` for HTML escaping is not an exotic input.

The interpolation loop is otherwise complete — it handles `'`/`"`, nested backticks, `${…}` depth and escapes, and now comments — so regex is the single remaining kind, and adding it is the same shape as `skip_comment`. Worth doing here rather than as a follow-up, because the carve-out is the one scanner in this file that will never get `skip_opaque`.

(`skip_comment` itself is safe against the reverse confusion: it matches only `//` and `/*`, and `//` can never be an empty regex in JS.)

### 2. `find_colon_at_depth0` — the disclosure understates it

The PR says it "is not reached by any input I could construct". At the function level it is, and the input is small. `find_colon_at_depth0` (`state_transforms.rs:899`) builds `let chars: Vec<char>` and returns `i`, a **char** index; the only caller slices bytes (`state_transforms.rs:731-733`):

```rust
if let Some(colon_pos) = find_colon_at_depth0(rhs) {
    let true_rhs = rhs[..colon_pos].trim();
    let false_branch = rhs[colon_pos + 1..].trim();
```

With `rhs = "'日日' : b = 2"` the colon is at char index 5 and byte index 9; bytes 4-6 are the second `日`, so `rhs[..5]` panics with *"byte index 5 is not a char boundary"* — the identical message class the destructure-pattern fix in this PR removes. One non-ASCII character is enough to produce a silent **mis-slice** (`'日' : b = 2` → `true_rhs` becomes `'日`, `false_branch` becomes `: b = 2`); two are enough to panic.

I have not verified end-to-end reachability through the compiler either — the call site needs the legacy `cond ? a = X : b = Y` shape to route here — so this is the same limitation you flagged, not a correction of it. What is worth changing is the framing: the input exists and is one line, so the follow-up does not have to start from scratch. Please put that string in the issue.

### 3. Index-unit audit — clean, including the two easy-to-get-wrong conversions

Checked every touched scanner and its call sites:

- **`char_indices()` → `code_bytes` is unit-neutral**: `char_indices` already yields byte offsets. That covers `body_references_identifier_recursive` (`:588`, `:615`), `body_references_identifier_in_statements` (`:669`, `content[start..i]`), and the four scans in `strip_function_scopes_that_shadow`.
- **Relative → absolute was handled**: `strip_function_scopes_that_shadow` previously iterated a *subslice* and added the base (`after_ident + i`, `brace_pos + 1 + i + 1`); `code_bytes_from` yields absolute offsets and the diff drops the additions in all four places. Correct, and this is the conversion most likely to go wrong silently.
- **`Vec<char>` + `enumerate()` → byte cuts**: `split_multi_declarator` is the one that was genuinely char-indexed, and the rewrite to `cuts: Vec<(usize, u8)>` + `&rest[start..i]` is consistently byte-based with cut points on ASCII `,`/`;`. The two-pass → one-pass restructure preserves semantics: the old first pass broke at the first top-level `,` and the second re-derived the same points; the `depth > 0` saturating decrement matches the old first pass; declarator text is still copied verbatim, so comments and literals inside a declarator are unchanged.
- **The panic fix**: `transform_legacy_destructure_declarations` (`:1266-1284`) now feeds a byte `pattern_end` into `&rest[..=pattern_end]` / `rest[pattern_end + 1..]`, both on an ASCII `}`/`]`/`)`. Fixed.
- `code_bytes_from(..., start)` begins with `prev = None`, so a `/` at `start` reads as a regex. At all four call sites `start` is just past `(`, `{`, or an identifier's params, where a regex is the correct reading (`x => /re/.test(x)`, `{ /re/.test(s) }`). No mis-read.

### 4. Test discrimination — all seven fail against the unfixed file

I re-derived each against the pre-PR bodies rather than taking the claim:

| test | why the old code fails it |
|---|---|
| `apostrophe_in_a_line_comment_…` | `'` in `don't` opens a string, blanks to EOF, `width` disappears |
| `brace_in_a_block_comment_…` | `}` in `/* } */` closes the arrow body early, `return a` escapes the shadow |
| `brace_in_a_line_comment_…` | same via the `function (a) {` scanner — correctly a separate test, different code path |
| `brace_in_a_comment_does_not_merge_two_statements` | `{` raises depth, the `;` no longer splits, `x` is read off the merged RHS |
| `comma_in_a_comment_does_not_split_declarators` | `,` in the comment sets `has_top_level_comma`, returns `Some` instead of `None` |
| `non_ascii_identifier_in_a_destructure_pattern_does_not_panic` | panics |
| `brace_in_a_comment_does_not_end_a_destructure_pattern` | `pattern_end` lands in the comment, `after_pattern` starts `*/`, statement returned unchanged |

The two retargeted replacements are the ones that now assert through `split_multi_declarator` and `transform_legacy_destructure_declarations` rather than through a regex match — both assert a value the scanner does control. Retargeting was the right call.

Note the block-comment and line-comment shadowing tests look duplicative but are not: they exercise the block-body and arrow-body brace matchers, which are separate loops (`:408` and `:496`). Worth a one-line comment saying so, or a future cleanup will merge them.

### 5. Nit

The skipped list is well argued, but `strip_object_property_keys`'s justification ("blanking a comment cannot change the answer") is the only one that is an argument rather than an observation. If it holds it is worth one line *in the code*, since the next sweep will re-derive it.

baseballyama:
Re-reviewed at `da5148a0`. **LGTM — 0 blocking.** Deleting the hand-rolled scanner instead of teaching it a regex arm is the right call, and I checked the three things that could have gone wrong rather than the one the repro exercises.

**Both loops, not just the repro's.** The outer loop and the `${…}` loop have the identical shape: helper first, then `if !bytes[i].is_ascii_whitespace() { prev = Some(bytes[i]) }` *before* the match. Same order, same reset rule, separate cursor (`inner_prev`) per loop. Nothing is threaded from one into the other, so there is no site to miss. `inner_prev` starting at `None` is right rather than incidental: `${` opens expression position, so `${/re/.test(x)}` must read as a regex.

**The `${…}` carve-out is intact, and now for a checkable reason.** `skip_slash_run` returns `None` unless `bytes[i] == b'/'`, so `'`, `"` and `` ` `` never reach `skip_opaque` and always fall through to the descending branches. The old comment *asserted* that `skip_opaque` was unusable here; the new code turns that into a one-line precondition and then delegates. That is the better shape — the invariant is now enforced at the boundary instead of being restated in prose.

**Division vs. regex is right in both loops.** Because `prev` is assigned before the match, a quote byte leaves `prev` as `'` / `"` / `` ` ``, all three of which are in `slash_starts_regex`'s exclusion list — so `"a" / b` divides. `prev = Some(b'x')` after a non-comment run models "a regex literal is a value"; `x` is alnum, so it excludes too. A comment deliberately does not update `prev`, which is correct: a comment is not a value and must not change what the next `/` means.

**The failure mode is strictly better than before, which is worth stating.** If the disambiguation ever loses, `skip_slash_run` *skips* bytes rather than blanking them, and `skip_opaque`'s regex arm bails on `b'\n'`. So a division misread as a regex costs at most the rest of one line. The bug this PR fixes blanked everything from the apostrophe to EOF.

**The two tests are correctly labelled, which matters here.** `apostrophe_in_a_regex_literal_…` is the repro: in `replace(/'/g, "&#39;")` the `'` has no partner anywhere in the input (`&#39;` is entity text, not an apostrophe), so on the parent the string scan runs to the end and `width` is blanked. `a_division_after_a_value_…` does **not** discriminate against `676279e` — that revision had no regex arm at all, so `n / 2` was already left alone — and the in-code comment says exactly that ("Negative control for the regex arm"). Counting it as a second repro would have been the defect; calling it a negative control is the fix.

**Doc comment.** Reattached — `skip_slash_run` now sits above the `/// Strip text content from string literals…` block, which directly precedes `strip_string_literal_text`. Confirmed by reading the file at head, not from the diff.

Two non-blocking nits, both fine to ignore:

- `skip_slash_run` indexes `bytes[i]` unchecked. Every caller is inside `while i < len` so it cannot panic today, but `bytes.get(i) != Some(&b'/')` would make the helper safe standalone at no cost — and the byte-vs-index class has already bitten this sweep twice.
- `prev = Some(b'x')` is a sentinel standing for "the previous token was a value". It is correct and the surrounding code makes it readable, but the one thing a future edit could get wrong is picking a sentinel that `slash_starts_regex` does *not* exclude. A named `const PREV_VALUE: u8 = b'x';` would pin that down.

PR #2370, opened 2026-08-06
fix(compiler): fix byte/char index mix in the legacy SSR store-set scan (#2357)

baseballyama:
Batch 4 of #2357. **Stacked on #2366** (base `fix/2357-scan-sweep-batch2`), not on `main`: both PRs edit `transform_legacy.rs` and both append to its `#[cfg(test)] mod tests`, so branching from `main` would have produced a conflict *and* two `mod tests` blocks in one file. GitHub will retarget this to `main` when #2366 merges. Review only the top commit, `b970b7c1`.

## 1. A live defect, not a conversion — `extract_store_set_targets`

This one is worth reading on its own; it is not part of the lexical sweep.

`memmem::Finder::find` searches `code.as_bytes()` and returns a **byte** offset. The function fed `abs_pos + 12` straight into a `Vec<char>`, so as soon as any non-ASCII appeared *before* a `$.store_set(` call, the read started too far right and recorded a truncated store name:

```rust
extract_store_set_targets("{ const t = '日'; $.store_set(count, t); }", &mut vars);
// before: ["$unt"]      after: ["$count"]
```

`'日'` is 3 bytes / 1 char, so the byte offset ran 2 positions ahead and the name was read from the middle of `count`. Confirmed failing before the fix.

**Reachability.** The caller chain is `sort_reactive_statements_topologically` → `extract_reactive_lhs_vars` → here, and the input is the text of one `$:` statement *after* the store transform. So it needs: legacy mode, a store assignment inside a `$:` statement, and non-ASCII earlier in that same statement — e.g. `$: { const t = '日'; $a = t.length; }`. The consequence is a wrong name in the declared-variable set, which drops a real dependency edge and lets the topological sort leave two `$:` statements in an order the dependency should have forced. I did **not** build the end-to-end component that makes the sort actually swap, so I am claiming the mis-slice and the reachable path, not a demonstrated output diff. That is the remaining work if you want it pinned to a fixture.

The fix drops the `Vec<char>` entirely and slices the `&str` from the byte offset — `$.store_set(` is ASCII so that offset is always a char boundary — keeping Unicode identifier support via `chars()`.

## 2. `extract_simple_assignments` — whole-function rewrite

Char-indexed throughout, and it tracked string literals but no comment kind and no regex. Now byte-indexed and driven by `skip_opaque`. Repro confirmed failing first: `extract_simple_assignments("// a = 1")` returned `["a"]`, inventing a reactive dependency out of a commented-out assignment. Regex literals are now skipped too, which matters for `/a=b/.test(s)` — previously read as an assignment to `a`.

Behaviour deliberately preserved, including one asymmetry I did not touch: the compound-assignment branch (`+=`, `-=`, …) does not check `is_member_prop`, while the plain `=` and `++`/`--` branches do. That looks wrong but changing it is a semantic change, not this PR's business.

## 3. `extract_reactive_rhs_identifiers` — deliberately NOT converted

Per the scoping: it is a hand-rolled four-state scanner that **deliberately descends into `${…}` interpolations**, which is precisely where `skip_opaque`'s all-or-nothing template handling would change output. Unit tests cannot tell me whether that descent matters on real input, and the corpus is not available here. Left alone, with the measurement that would settle it written up in the report to the epic.

## Tests

18/18 in `cargo test -p rsvelte_core --lib transform_legacy::tests`. Two of the three new tests were confirmed failing before the fix; the third (`simple_assignments_survive_non_ascii`) **passed beforehand** and is a regression guard for the char→byte index rewrite, not a repro — flagged so the count is not read as stronger than it is.

`cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` clean, no `-A`. `--test ssr` / `--test runtime` cannot run here (`Fixtures not found`).

### DIFF
--- a/.changeset/scan-sweep-batch4.md
+++ b/.changeset/scan-sweep-batch4.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): fix byte/char index mix in the legacy SSR store-set scan
+
+`extract_store_set_targets` fed a byte offset from a `memmem` match into a
+`Vec<char>`, so any non-ASCII before a `$.store_set(` call made it read the
+store name from the wrong position and record a truncated dependency.
+`extract_simple_assignments` alongside it now skips comments and regex
+literals instead of reading assignments out of them.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -1341,21 +1341,15 @@ fn extract_store_set_targets(code: &str, vars: &mut Vec<String>) {
     let mut search_from = 0;
     while let Some(pos) = finder.find(&code.as_bytes()[search_from..]) {
         let abs_pos = search_from + pos;
-        let after_call = abs_pos + 12; // "$.store_set(".len()
-        // Read the first argument (store name)
-        let mut j = after_call;
-        let chars: Vec<char> = code.chars().collect();
-        while j < chars.len() && (chars[j] == ' ' || chars[j] == '\t') {
-            j += 1;
-        }
-        let name_start = j;
-        while j < chars.len() && (chars[j].is_alphanumeric() || chars[j] == '_' || chars[j] == '$')
-        {
-            j += 1;
-        }
-        if j > name_start {
-            let store_name: String = chars[name_start..j].iter().collect();
-            let store_sub = format!("${}", store_name);
+        // `memmem` yields a byte offset; the old code fed it to a `Vec<char>`,
+        // mis-slicing the name whenever non-ASCII preceded the call.
+        let rest = code[abs_pos + 12..].trim_start_matches([' ', '\t']);
+        let store_name: String = rest
+            .chars()
+            .take_while(|c| c.is_alphanumeric() || *c == '_' || *c == '$')
+            .collect();
+        if !store_name.is_empty() {
+            let store_sub = format!("${store_name}");
             if !vars.contains(&store_sub) {
                 vars.push(store_sub);
             }
@@ -1369,67 +1363,38 @@ fn extract_store_set_targets(code: &str, vars: &mut Vec<String>) {
 /// to find variable assignments that indicate the reactive statement modifies a variable.
 fn extract_simple_assignments(code: &str) -> Vec<String> {
     let mut vars = Vec::new();
-    // Find patterns like `identifier =` (not `==`), `identifier++`, `identifier--`,
-    // `++identifier`, `--identifier`
-    let chars: Vec<char> = code.chars().collect();
-    let len = chars.len();
+    let bytes = code.as_bytes();
     let mut i = 0;
-    let mut in_string = false;
-    let mut string_char = ' ';
+    let mut prev: Option<u8> = None;
 
-    while i < len {
-        let c = chars[i];
-
-        // Track string literals to avoid matching inside them
-        if (c == '\'' || c == '"' || c == '`') && !in_string {
-            in_string = true;
-            string_char = c;
-            i += 1;
-            continue;
-        }
-        if in_string {
-            if c == string_char && (i == 0 || chars[i - 1] != '\\') {
-                in_string = false;
+    while i < bytes.len() {
+        if let Some((next, is_comment)) = skip_opaque(bytes, i, prev) {
+            if !is_comment {
+                prev = Some(b'x');
             }
-            i += 1;
+            i = next;
             continue;
         }
 
-        // Check for `++identifier` and `--identifier` prefix operators
-        if i + 2 < len
-            && ((chars[i] == '+' && chars[i + 1] == '+')
-                || (chars[i] == '-' && chars[i + 1] == '-'))
+        // Prefix `++x` / `--x`
+        if (bytes[i] == b'+' && bytes.get(i + 1) == Some(&b'+'))
+            || (bytes[i] == b'-' && bytes.get(i + 1) == Some(&b'-'))
         {
-            let op_end = i + 2;
-            // Skip whitespace after operator
-            let mut j = op_end;
-            while j < len && chars[j] == ' ' {
+            let mut j = i + 2;
+            while bytes.get(j) == Some(&b' ') {
                 j += 1;
             }
-            // Read identifier
-            if j < len && (chars[j].is_alphabetic() || chars[j] == '_' || chars[j] == '$') {
-                let start = j;
-                while j < len && (chars[j].is_alphanumeric() || chars[j] == '_' || chars[j] == '$')
-                {
-                    j += 1;
-                }
-                let ident: String = chars[start..j].iter().collect();
+            if let Some((ident, end)) = read_identifier(code, j) {
                 if !is_reactive_keyword(&ident) && !vars.contains(&ident) {
                     vars.push(ident);
                 }
-                i = j;
+                prev = Some(b'x');
+                i = end;
                 continue;
             }
         }
 
-        if c.is_alphabetic() || c == '_' || c == '$' {
-            // Read identifier
-            let start = i;
-            while i < len && (chars[i].is_alphanumeric() || chars[i] == '_' || chars[i] == '$') {
-                i += 1;
-            }
-            let ident: String = chars[start..i].iter().collect();
-
+        if let Some((ident, end)) = read_identifier(code, i) {
             // A member property (`foo.x = …` / `foo.x += …` / `foo.x++`) is not
             // a declared variable: the assignment mutates the *base object*, not the
             // property. Recording the property would create a false reactive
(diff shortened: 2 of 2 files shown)


baseballyama:
**LGTM** — 0 blocking. Reviewed the top commit `b970b7c1` only, against base `fix/2357-scan-sweep-batch2`. The index fix is real and its test discriminates, the rewrite is behaviour-preserving where it claims to be, and the self-classification of the third test as a preservation guard is **correct** — I re-derived it rather than taking it.

Read-only review — no build, no test run.

---

### Verified, not assumed

**The `extract_store_set_targets` mis-slice and its repro.** `memmem` yields a byte offset; the old code indexed `Vec<char>` with it. For the test input `{ const t = '日'; $.store_set(count, t); }`: `$` is at byte 19 / char 17, so `after_call` was byte 31 while the name starts at char 29 — the old scan begins at `chars[31]`, two positions into `count`, and yields `$unt`. New code slices `code[abs_pos + 12..]`, and `abs_pos + 12` is always a char boundary because `$.store_set(` is ASCII and a match on an ASCII needle in valid UTF-8 cannot start mid-character. `search_from = abs_pos + 1` indexes `code.as_bytes()`, which has no boundary requirement. **The test fails against the unfixed code.** ✓

**The reachability claim, and its limit.** `sort_reactive_statements_topologically` → `extract_reactive_lhs_vars` → here, on the post-store-transform text of one `$:` statement, is the chain, and the stated preconditions (legacy mode, a store assignment inside `$:`, non-ASCII earlier in the same statement) are the right ones. I also confirmed the consequence is *dropping* an edge rather than corrupting one in the common case: for a one-character store name like your `$a` example, the overshoot lands on `,` or whitespace, the `take_while` yields nothing, and **nothing is pushed at all** — so the dependency silently disappears. You are right not to claim an output diff, and I am not upgrading it.

One cheap way to close the gap without building a component: `extract_reactive_lhs_vars` is private in the same module as `mod tests`, so a unit test one level up — feed it a `$:` statement whose post-transform text contains `'日'` before `$.store_set(a, …)` and assert `$a` is in the returned set — would pin the *reachable* claim without needing the topological sort to actually swap. That is a strictly smaller step than a fixture and would move this from "mis-slice + plausible path" to "mis-slice + demonstrated wrong dependency set".

**The preservation-guard classification is right.** `simple_assignments_survive_non_ascii` passes on the old code, and the reason is worth stating precisely: the old `extract_simple_assignments` was *internally consistent* — `Vec<char>` throughout, `i` never used as a byte slice index, `chars[start - 1] == '.'` on the same axis. There was no byte/char mix in that function to begin with; the only mix in this file was in `extract_store_set_targets`. So the test guards the char→byte rewrite you just performed and nothing else. Flagging it explicitly so the count is not read as 3/3 was the right call, and it is the classification the repo's current standard asks for.

**The two repros that do discriminate.** `extract_simple_assignments("// a = 1")` → old `["a"]` (`/` is not alphabetic, falls through, `a` reads as an identifier, `=` follows) vs new `[]`. `"/* a = 1 */ b = 2"` → old `["a", "b"]` vs new `["b"]`. Both fail first. ✓

**Behaviour preservation in the rewrite.** Checked arm by arm: the `=` branch's nested `if`s collapse into one `&&` chain with identical semantics; the compound branch keeps its `bytes.get(j + 2) != Some(&b'=')` guard; the postfix branch still advances `end + 2`; the fall-through still resumes at `end` without an extra `+= 1`, matching the old loop. The `is_member_prop` byte check is equivalent to the old char check, including for a multi-byte predecessor (a continuation byte is never `.`).

**`read_identifier` is boundary-safe by construction**, which matters because `i` walks byte-by-byte through non-identifier text and *can* land mid-character: for `"— a = 1"` the loop calls `read_identifier(code, 1)` and `code.get(1..)` returns `None`, so the `?` bails instead of panicking. That is the right defensive shape and it is worth a one-line comment saying the `get` is load-bearing, not stylistic — a future refactor to `&code[at..]` would reintroduce exactly the panic class this PR is cleaning up.

### Nits

1. **`extract_store_set_targets` is still not lexical**, in a commit that makes its neighbour lexical. `$: { const s = "$.store_set(x,"; $a = 1 }` still records `$x` from inside a string literal. The changeset scopes this PR to the index bug for that function, which is defensible, but the function was open on the table and this is the exact class the epic exists for — say in the changeset that it was left, so batch 5 does not have to rediscover it.

2. The `prev` guard in the `=` branch (`!matches!(j.checked_sub(1).map(|k| bytes[k]), Some(b'!' | b'<' | …))`) looks unreachable in both the old and new code: `j` is `end` plus ASCII spaces, so `bytes[j - 1]` is either a space or the identifier's last byte, never an operator — the `x != 1` case it appears to guard exits at `bytes.get(j) == Some(&b'=')` being false. Faithfully preserved, so not a change request; worth a note if you touch it again.

3. You already flag the `is_member_prop` asymmetry on the compound branch (`$: foo.x += count` records `x`, the plain `=` branch would not). Agreed it is out of scope here — noting only that the comment explaining *why* member properties must not be recorded sits directly above the branch that ignores it, which is how it will get re-litigated.

4. Keyword-before-regex remains ambiguous for `slash_starts_regex`: after an identifier you set `prev = Some(b'x')`, so `return /re/.test(s)` reads the `/` as division. Shared-lexer limitation, not this PR's, and it now behaves *better* than batch 2 for `i++ / 2` (postfix sets `prev = 'x'`, so that is a division as it should be).

### Stacking

Base `fix/2357-scan-sweep-batch2` is still open with review comments outstanding, including a question about `bracket_depth_delta` and `${…}` in `reorder_reactive_statements_after_functions` / `sort_reactive_in_place`. That is not a conflict with this PR, but the two touch **opposite ends of the same `$:` pipeline**: #2366 changes which source lines get grouped into one reactive statement, this PR changes which variables are extracted from that statement's text. A wrong grouping in #2366 feeds wrong text to `extract_reactive_lhs_vars` here. Worth running the corpus once on the merged pair rather than on each alone.

baseballyama:
Re-checked at `b970b7c1`. **LGTM stands — 0 blocking.** One correction to the premise, which is worth stating because it changes what has been tested rather than what is in the diff.

**There was no rebase.** `b970b7c1`'s parent is `dfee2a40` — batch2's *first* commit. None of `8a2836f7`, `b1f900e2`, `65ce783b` is an ancestor of this branch, so batch4 still sits on batch2 as it was three commits ago. The head is also byte-identical to what I reviewed earlier (same SHA, committed 19:58 JST, reviewed 20:18), so nothing in the diff has moved and my earlier LGTM applies unchanged.

**The merge is nonetheless clean** — GitHub reports `MERGEABLE / CLEAN`, and the line ranges say why:

- the code hunks (`extract_store_set_targets` / `extract_simple_assignments`, `:1327-1500`) are disjoint from everything batch2 touches (`scan_reactive_line` and its callers, `:780-1200`). No semantic contact at all.
- the only shared region is `mod tests`: batch2 inserts `continuation_interpolation_keeps_the_statement_whole` near old `:2192`, batch4 inserts at old `:2188+6`. Close, but git resolves them — the two insertions land at distinguishable anchors.

**The one real consequence of the stale base**: every CI run on this PR exercised a tree *without* batch2's template-state fix, and the combination has never been built or tested together. That is normally fine for a stacked PR — the merge into batch2 re-runs — but batch2 has an open blocking finding from this round (`65ce783b` still splits a statement on a blank line inside a template literal, see my comment there), so batch2's head will move again before this lands. Re-point at batch2's *final* head before merging rather than at whatever it is now; a rebase performed today would be discarded anyway.

Nothing to change in this PR.

baseballyama:
Re-pointed onto batch 2's final head: `dfee2a40` → `d00b34c2` (rebase + `--force-with-lease`, now `7bf12647`).

This combination had never been built. `dfee2a40` was batch 2's head when this branch was cut; batch 2 has since gained four commits from review, including the template-state threading.

**Conflict and how it was resolved.** One hunk, in `mod tests` — both branches append tests at the same point, so git could not order them. Resolved by keeping both sides in full; no production hunk conflicted, and the diff this branch introduces over the merge-base is exactly three files:

```
.changeset/scan-sweep-batch2.md
.changeset/scan-sweep-batch4.md
crates/.../3_transform/server/transform_legacy.rs
```

**Verified on the combined tree**, not replayed from either side:

- `cargo test -p rsvelte_core --lib transform_legacy` — 22 passed, 0 failed (batch 2's 18 + batch 4's 3 + one adjacent `state_transforms` pin).
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — clean, no `-A` suppressions.
- `rustfmt --edition 2024 --check` on the merged file — clean.
- `git log --oneline <merge-base>..HEAD` — six commits, all batch 2 or batch 4, no foreign authorship.

CI now gets a genuine run of batch2 + batch4 rather than two independent green runs that were never combined.

baseballyama:
Re-pointed onto batch 2's new head after its rebase: parent `d00b34c2` → **`4e2fc94d`**, now `8a3045df` (`--force-with-lease`).

**Two conflicts, and the second is the one worth reading.** Batch 4 rewrites `extract_simple_assignments` from `Vec<char>` to byte indexing; #2394 (merged since) added a `!is_member_prop` guard to the same function, including its **compound-assignment** branch. Batch 4's version of that branch predates #2394 and therefore lacks the guard — **resolving in batch 4's favour would have silently reverted a merged fix.**

Resolved as batch 4's byte-based form **plus** `&& !is_member_prop`, and the comment kept in #2394's wording (which names `foo.x += …`, the case #2394 added) rather than batch 4's older text.

**Verified the guard is load-bearing, not just present.** Removing `&& !is_member_prop` from the merged form fails exactly the two #2394 tests:

```
member_assignment_targets_are_not_recorded ... FAILED
compound_member_assignment_does_not_reorder ... FAILED
```

Restored and re-verified. Keeping the text without the behaviour would have looked identical in a diff review.

26 `transform_legacy` tests pass on the combined tree, bare clippy clean, rustfmt clean. `git diff --name-only $(git merge-base origin/main HEAD) HEAD` → two changesets plus `transform_legacy.rs`, single author, no foreign commits.

PR #2371, opened 2026-08-06
test: gate the fmt-corpus skips on the job that supplies them (#1791)

baseballyama:
Follow-up to #2367, closing the six sites that PR deliberately left alone. Part of #1791.

## Why these six could not take the `CI` guard

#2367 hardened 20 skip sites with `assert!(prereq || env CI is_none)`. That guard is only correct
where the prerequisite is supplied by *every* job that runs the suite. These six are not:

- `ci.yml:186` (sharded `test` job) checks out `svelte language-tools eslint-plugin-svelte esrap`
  and **explicitly excludes `svelte.dev`** — it lives only in `test-fmt-corpus`.
- `RSVELTE_FMT_TW_TEST_DIR` is set at `ci.yml:613`, in one step of one job.

A blanket `CI` assert would fail jobs that are *correctly* not supplying the prerequisite —
converting real signal into noise that someone eventually disables wholesale.

## The fix: key the guard on the variable the owning step sets

| Suite | Signal | Set at |
|---|---|---|
| `svelte_dev_corpus.rs`, `svelte_dev_markdown.rs` (3 sites each) | `FMT_CORPUS_OXFMT` | `ci.yml:603` |
| `cli/tailwind.rs` (4 sites, via `tw_fixture`) | `RSVELTE_FMT_TW_TEST_DIR` | `ci.yml:613` |

Absence still skips; **presence-plus-unusable now fails**. So the `test` shards stay green while
the job that is supposed to have the corpus fails loudly when it does not. The Tailwind guard sits
inside `tw_fixture`, which all four call sites funnel through, so it cannot be forgotten later.

Note the guard is strictly stronger than a presence check: `FMT_CORPUS_OXFMT` pointing at a
*broken* oxfmt, or `RSVELTE_FMT_TW_TEST_DIR` pointing at a directory missing one of the two
packages, now fails instead of skipping. That is the failure mode a cached-oracle drift produces,
and `ci.yml:537` records that it has already happened once ("a hardcoded pin silently drifts from
package.json — it did").

## Second commit: a real bug the guard exposed

Wiring the guard up made the corpus suite fail with the *wrong* message, which turned out to be a
latent defect rather than a bad assertion.

`svelte_dev_short_sha` ran `git -C submodules/svelte.dev rev-parse HEAD`. An uninitialised
submodule is an **empty directory**, and `git -C` in an empty directory walks up to the
superproject and succeeds — returning the superproject's HEAD. Observed directly: it returned
`2665d4e824ba`, which is this branch's own commit, while the recorded submodule SHA is `7cca42dd`.

Consequences, both invisible before:
- the `else` branch ("svelte.dev submodule not checked out") was **unreachable** in the normal
  uninitialised state — the documented skip could never fire;
- the suite instead proceeded with a valid-looking wrong SHA, looked under
  `fixtures/fmt-corpus/<superproject-sha>/`, found nothing, and skipped one step later blaming
  missing fixtures.

Both files now require `submodules/svelte.dev/.git` to exist first. After the fix the correct guard
fires and names the actual cause.

This is the same shape as the rest of #1791 — a check that reported the wrong thing rather than
failing — and it was only observable because something finally asserted on the outcome.

## Demonstration

Measured per suite, with the negative control first. `submodules/svelte.dev` is genuinely absent in
this checkout, so no contrivance was needed.

| Suite | Signal unset (control) | Signal set, prerequisite absent |
|---|---|---|
| `svelte_dev_corpus` | 1 passed (soft skip) | **FAILED** — names the submodule |
| `svelte_dev_markdown` | 1 passed (soft skip) | **FAILED** — names the submodule |
| `cli/tailwind` (`sort_tailwindcss`) | 11 passed | **4 failed, 7 passed** |

The Tailwind row is the discriminating one: exactly the 4 gated tests fail and the 7 that do not
depend on the oracle keep passing, so the assertion is attached to the skip sites rather than
firing globally.

Before the second commit, the `FMT_CORPUS_OXFMT`-set runs failed with "no markdown fixtures at
fixtures/fmt-corpus/2665d4e824ba"; after it they fail with "submodules/svelte.dev is not checked
out". Same pass/fail, correct diagnosis.

## Scope

`test(...)` only, no shipped behaviour change, so no changeset. Three files. `rustfmt --check`
clean; `cargo build --tests` clean for both crates.

Category 1 of #1791 is now closed: 30 of 30 skip sites carry a guard. The remaining follow-up on
the issue is the **axis-B** work (asserts that run but cannot distinguish working from broken),
which is a different and larger problem.

## Conflict note

Both this and #2367 touch `crates/rsvelte_fmt/tests/cli/tailwind.rs` — #2367 adds a `CI` assert
inside `node_runnable`, this adds `tw_oracle_declared` + guards inside `tw_fixture`. Different
hunks; they should auto-merge. This branch is based on `origin/main`, not on #2367, so the two are
independent and can land in either order.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -601,6 +601,11 @@ jobs:
         env:
           RUST_MIN_STACK: '33554432'
           FMT_CORPUS_OXFMT: ${{ runner.temp }}/fmt-oracle/node_modules/.bin/oxfmt
+          # This job checked svelte.dev out and generated the corpus, so a
+          # missing prerequisite here is a misconfigured job. `FMT_CORPUS_OXFMT`
+          # alone cannot say that — it is a user-facing knob, and gating on it
+          # would make a contributor's local export fail their test run.
+          RSVELTE_REQUIRE_PREREQS: '1'
 
       - name: Run sortTailwindcss JS-parity tests (custom config vs. oxfmt)
         # Asserts rsvelte-fmt's JS-sorted output is byte-identical to the real
--- a/crates/rsvelte_fmt/tests/cli/tailwind.rs
+++ b/crates/rsvelte_fmt/tests/cli/tailwind.rs
@@ -147,11 +147,15 @@ fn node_runnable() -> bool {
         .unwrap_or(false)
 }
 
-/// `RSVELTE_FMT_TW_TEST_DIR` is set by exactly the CI step that installs the
-/// Tailwind oracle, so it — not `CI` — marks the run that must not skip. Every
-/// other job legitimately lacks the oracle and has to stay green.
+/// May this run fail on a missing Tailwind oracle? Both conditions are needed:
+/// `RSVELTE_REQUIRE_PREREQS` says the job promised its prerequisites, and
+/// `RSVELTE_FMT_TW_TEST_DIR` says the oracle is in scope for it. This binary
+/// also runs in the sharded `test` job, which sets the former but installs no
+/// oracle — and the latter is a user-facing knob, so either alone turns a
+/// legitimate skip into a failure.
 fn tw_oracle_declared() -> bool {
-    std::env::var_os("RSVELTE_FMT_TW_TEST_DIR").is_some()
+    std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_some()
+        && std::env::var_os("RSVELTE_FMT_TW_TEST_DIR").is_some()
 }
 
 /// Build a fixture project (custom v4 stylesheet + `node_modules` symlinked from
--- a/crates/rsvelte_fmt/tests/cli/tailwind.rs
+++ b/crates/rsvelte_fmt/tests/cli/tailwind.rs
@@ -147,14 +147,35 @@ fn node_runnable() -> bool {
         .unwrap_or(false)
 }
 
+/// `RSVELTE_FMT_TW_TEST_DIR` is set by exactly the CI step that installs the
+/// Tailwind oracle, so it — not `CI` — marks the run that must not skip. Every
+/// other job legitimately lacks the oracle and has to stay green.
+fn tw_oracle_declared() -> bool {
+    std::env::var_os("RSVELTE_FMT_TW_TEST_DIR").is_some()
+}
+
 /// Build a fixture project (custom v4 stylesheet + `node_modules` symlinked from
 /// the resolved Tailwind env) and return `(dir, sidecar_in_dir, oxfmt)`. The
 /// sidecar is copied into the fixture so Node resolves the plugin from there.
 #[cfg(unix)]
 fn tw_fixture(css: &str) -> Option<(PathBuf, PathBuf, PathBuf)> {
-    let tw_dir = tw_test_dir()?;
+    let Some(tw_dir) = tw_test_dir() else {
+        assert!(
+            !tw_oracle_declared(),
+            "RSVELTE_FMT_TW_TEST_DIR is set but does not contain both \
+             node_modules/tailwindcss and node_modules/prettier-plugin-tailwindcss \
+             — the JS-parity assertions would be silently skipped."
+        );
+        return None;
+    };
     let oxfmt = real_oxfmt_bin();
     if !real_oxfmt_runnable(&oxfmt) || !node_runnable() || !sidecar_script().is_file() {
+        assert!(
+            !tw_oracle_declared(),
+            "the Tailwind oracle is declared via RSVELTE_FMT_TW_TEST_DIR but \
+             oxfmt / node / the sidecar script is unusable — the JS-parity \
+             assertions would be silently skipped."
+        );
         return None;
     }
     let dir = tempdir();
--- a/crates/rsvelte_fmt/tests/svelte_dev_markdown.rs
+++ b/crates/rsvelte_fmt/tests/svelte_dev_markdown.rs
@@ -43,6 +43,13 @@ fn svelte_dev_short_sha(root: &Path) -> Option<String> {
     (sha.len() >= 12).then(|| sha[..12].to_string())
 }
 
+/// `FMT_CORPUS_OXFMT` is set by exactly the CI step that checks out svelte.dev
+/// and generates the corpus, so it — not `CI` — marks the run that must not skip.
+/// The sharded `test` job deliberately omits svelte.dev and has to stay green.
+fn in_corpus_job() -> bool {
+    std::env::var_os("FMT_CORPUS_OXFMT").is_some()
+}
+
 fn oxfmt_bin() -> PathBuf {
     if let Ok(p) = std::env::var("FMT_CORPUS_OXFMT") {
         return PathBuf::from(p);
@@ -111,12 +118,24 @@ fn svelte_dev_markdown_cli_parity() {
     let root = repo_root();
 
     let Some(short_sha) = svelte_dev_short_sha(&root) else {
+        assert!(
+            !in_corpus_job(),
+            "submodules/svelte.dev is not checked out in the job that sets \
+             FMT_CORPUS_OXFMT — the markdown parity assertions would be \
+             silently skipped."
+        );
         eprintln!("[fmt-corpus-md] svelte.dev submodule not checked out; skipping.");
         return;
     };
     let fixtures = root.join("fixtures/fmt-corpus").join(&short_sha);
     let markdown_root = fixtures.join("markdown");
     if !markdown_root.is_dir() {
+        assert!(
+            !in_corpus_job(),
+            "no markdown fixtures at fixtures/fmt-corpus/{short_sha} in the job \
+             that sets FMT_CORPUS_OXFMT — `generate-fmt-corpus` produced nothing, \
+             or wrote a different svelte.dev SHA."
+        );
         eprintln!(
             "[fmt-corpus-md] no markdown fixtures at fixtures/fmt-corpus/{short_sha}; skipping. \
              Run: pnpm run generate-fmt-corpus"
@@ -126,6 +145,12 @@ fn svelte_dev_markdown_cli_parity() {
 
     let oxfmt = oxfmt_bin();
     if !oxfmt_runnable(&oxfmt) {
+        assert!(
+            !in_corpus_job(),
+            "FMT_CORPUS_OXFMT is set but oxfmt is not runnable at {} — the \
+             oracle is broken, not absent.",
+            oxfmt.display()
+        );
         eprintln!(
             "[fmt-corpus-md] oxfmt not runnable at {} (set FMT_CORPUS_OXFMT); skipping.",
             oxfmt.display()
--- a/crates/rsvelte_fmt/tests/svelte_dev_markdown.rs
+++ b/crates/rsvelte_fmt/tests/svelte_dev_markdown.rs
@@ -29,9 +29,15 @@ fn repo_root() -> PathBuf {
 }
 
 fn svelte_dev_short_sha(root: &Path) -> Option<String> {
+    let submodule = root.join("submodules/svelte.dev");
+    // An uninitialised submodule is an empty directory, and `git -C` there walks
+    // up to the superproject and reports *its* HEAD — a valid-looking wrong SHA.
+    if !submodule.join(".git").exists() {
+        return None;
+    }
     let out = Command::new("git")
         .arg("-C")
-        .arg(root.join("submodules/svelte.dev"))
+        .arg(&submodule)
         .args(["rev-parse", "HEAD"])
         .output()
         .ok()?;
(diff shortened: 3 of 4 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. I applied the #2367 test to every guard here and found no workflow that sets either signal without supplying the prerequisite. Two design notes below, one of which is the same coupling defect I raised on #2367.

Read-only review at `564f718` — no build, no test run.

---

### The check that mattered on #2367, run here

For each guard: does the job that sets the signal actually supply the prerequisite, and does any other workflow set it without doing so?

- `FMT_CORPUS_OXFMT` is set at exactly one place in the repo, `ci.yml:603`, in the `Test fmt corpus` job. That job checks out `svelte.dev` at `ci.yml:521` (`git submodule update --init --depth 1 submodules/svelte submodules/svelte.dev`), installs the oxfmt oracle **unconditionally** (`ci.yml:530-549` — no `if:`, and it ends in `./node_modules/.bin/oxfmt --version`, so the binary the env var points at is proven runnable before the tests run), and runs `pnpm run generate-fmt-corpus` unconditionally. All three guards in `svelte_dev_corpus.rs` / `svelte_dev_markdown.rs` — submodule, fixtures, oxfmt-runnable — are supplied by that job. ✓
- `RSVELTE_FMT_TW_TEST_DIR` is set at exactly one place, `ci.yml:613`, in the same job, immediately after an unconditional `npm i tailwindcss prettier-plugin-tailwindcss` into `$RUNNER_TEMP/tw-oracle` (`ci.yml:551-566`). `tw_test_dir()` returns `Some` only when both `node_modules/<pkg>/package.json` files exist, which is exactly what that step creates — so the first assertion's message ("set but does not contain both …") is an accurate description of the only way it can fire. ✓
- Nothing else in `.github/` or `scripts/` sets either variable. ✓
- The two corpus binaries run **only** in that job: `scripts/ci/run-test-shard.sh:56` excludes `svelte_dev_corpus` and `svelte_dev_markdown` from the shard partition. `--test cli` (which carries `tailwind.rs`) *does* run in the shards, where `RSVELTE_FMT_TW_TEST_DIR` is unset, so `tw_oracle_declared()` is false and both new assertions are inert there. ✓

That is the failure mode #2367 had, and this PR does not have it. Job-scoping on a variable the job itself sets is strictly better than gating on `CI`.

### 1. `FMT_CORPUS_OXFMT` is also a documented user-facing knob, and the skip messages advertise it

This is the `TSGO_BIN` coupling from #2367 §3 in a new place. `oxfmt_bin()` reads `FMT_CORPUS_OXFMT` as "path to the oxfmt binary", and the existing skip message tells a developer to set it:

```
[fmt-corpus] oxfmt not runnable at {} (set FMT_CORPUS_OXFMT); skipping.
```

A contributor who follows that advice — points `FMT_CORPUS_OXFMT` at their own oxfmt build to work on the formatter — and has not run `git submodule update --init submodules/svelte.dev` now gets a **panic** from the *first* guard (`svelte_dev_corpus.rs:210`, "submodules/svelte.dev is not checked out in the job that sets FMT_CORPUS_OXFMT") instead of the skip they used to get. Following one skip's printed remedy makes a different skip fatal, and the message names a "job" that does not exist on their machine.

The variable is carrying two meanings: *where the oracle is* (payload, user-settable) and *this run promised the corpus* (signal, CI-only). Splitting them fixes it and costs one line in `ci.yml`. Given #2367 is reportedly moving to a dedicated `RSVELTE_REQUIRE_PREREQS`, that is the natural signal to reuse here — see §3.

`RSVELTE_FMT_TW_TEST_DIR` has the same dual role but is much less exposed: it is not advertised in any skip message, and it is documented as pointing at a directory a consumer project would have.

### 2. The `.git` existence check is load-bearing for guard 2, not cosmetic

`svelte_dev_short_sha` (`svelte_dev_corpus.rs:42-47`, `svelte_dev_markdown.rs:32-37`) now returns `None` when `submodules/svelte.dev/.git` is absent. The comment explains the walk-up correctly, and `Path::exists()` is right for a submodule where `.git` is a *file* (a `gitdir:` pointer), not a directory.

Worth stating why it belongs in *this* PR rather than as a drive-by: without it, an uninitialised submodule yields the **superproject's** SHA, `fixtures/fmt-corpus/<wrong-sha>` does not exist, and control reaches the *second* new assertion — which would then fire with the message "`generate-fmt-corpus` produced nothing, or wrote a different svelte.dev SHA". That is a confidently wrong diagnosis of a missing submodule. The ordering fix is what keeps guard 2's message true.

### 3. Two conventions, if #2367 lands as planned

If #2367 converts to a repo-specific `RSVELTE_REQUIRE_PREREQS` set only by jobs that supply prerequisites, the repo ends up with two idioms for the same idea: a dedicated signal variable in nine test files, and payload-variable-as-signal in these three. Both are defensible in isolation; together they mean the next person adding a guard has to guess.

Recommendation, in order of preference: (a) set `RSVELTE_REQUIRE_PREREQS: 1` on the `Test fmt corpus` job and use it here too, which also resolves §1 for free; or (b) keep the job-scoped form and add one line to the epic saying fmt-corpus is deliberately the exception because its prerequisite is generated rather than checked out. What should not happen is landing both silently.

### 4. Nits

- The three `in_corpus_job()` / `tw_oracle_declared()` helpers are duplicated verbatim across three files. Fine for now; if a fourth appears, it belongs next to whatever #2367 settles on.
- `tw_fixture` is `#[cfg(unix)]`, so on a hypothetical Windows runner the guards vanish along with the tests. Not a problem today (the job is `ubuntu-latest`) and arguably correct, but it means "this suite ran" is not something the guard can assert on every platform.
- The second Tailwind assertion covers oxfmt / node / sidecar as one condition, so its message has to name all three. Splitting it would say which one is missing; low value at two call sites.

baseballyama:
Adopted the proposal. Pushed as `758e1814`. **I am reversing my earlier "keep both idioms" position**, and it is worth saying exactly why, because the reversal is not a concession — the new fact defeats the argument I made.

## Why I was wrong

My argument was that a payload variable catches "declared but unusable", which a boolean cannot. That is still true — but I was comparing against the wrong alternative. The proposal **keeps** that check: inside a gated job the payload check still runs, so a `FMT_CORPUS_OXFMT` pointing at a broken oxfmt, or a TW dir missing one package, still fails. Nothing is lost.

What I had missed is that both variables are **user-facing knobs**, one of them named in a skip message as the remedy. Using them as the fail-gate means a contributor's unrelated local export silently converts every skip into a panic. That is the same payload/signal conflation I refused to fix in `TSGO_BIN` two commits ago — reappearing in a variable I introduced. So the proposal strictly dominates: same coverage, one idiom, hazard gone.

## One correction to the proposal, found while implementing

Gating on `RSVELTE_REQUIRE_PREREQS` **alone** would have broken the sharded `test` job. `run-test-shard.sh:56` excludes only `compile_module_thread_safety`, `svelte_dev_corpus` and `svelte_dev_markdown` — so **`--test cli` does run in the shards**, which declare prerequisites but install no Tailwind oracle. A gate-only guard fails there.

So the condition is **gate AND applicability**:

```rust
std::env::var_os("RSVELTE_REQUIRE_PREREQS").is_some()   // this job promised prerequisites
    && std::env::var_os("FMT_CORPUS_OXFMT").is_some()   // ...and this one is in scope for it
```

The payload variable stays, but demoted from "may I fail?" to "is this prerequisite mine?", which is the question it can actually answer. `RSVELTE_REQUIRE_PREREQS` is added to the `Run svelte.dev formatter-parity corpus tests` step; the `sort_tailwindcss` step gets it from #2367.

## Verification — four regimes, the first being the hazard

| regime | `RSVELTE_REQUIRE_PREREQS` | payload var | required | corpus | markdown | tailwind |
|---|---|---|---|---|---|---|
| **contributor exports the knob locally** | – | set | **skip** | ok | ok | 11 passed |
| sharded `test` job | 1 | – | skip | ok | — | 11 passed |
| `Test fmt corpus` job, prereq missing | 1 | set | **fail** | FAILED | FAILED | 4 failed / 7 passed |
| plain local run | – | – | skip | ok | ok | — |

Row 1 is the bug this fixes. Row 2 is why the gate alone was not enough. Row 3 keeps everything #2371 was for — and the Tailwind cell failing exactly its 4 gated tests while the other 7 pass confirms the guard is still attached to the right sites.

`rustfmt --check` clean, `ci.yml` parses.

## On the skip message

No longer misleading, and fixed at the root rather than by rewording: following "(set `FMT_CORPUS_OXFMT`)" now yields a skip again, because the panic needs `RSVELTE_REQUIRE_PREREQS` too, and no contributor sets that. I left the message alone deliberately — it gives correct advice for the thing it is about.

## Merge-order note

If #2371 lands before #2367, the Tailwind guard is temporarily **inert** (its step gets `RSVELTE_REQUIRE_PREREQS` from #2367), i.e. it skips as it does today. Fail-open in the safe direction, self-resolving once both land. The corpus guards are fully active either way, since their step's variable is added here.

baseballyama:
Re-reviewed at `758e1814`. **LGTM — 0 blocking.** Your correction to the proposal was the load-bearing part and it is right; I re-derived it rather than taking it on trust.

**The correction checks out.** `run-test-shard.sh:56` excludes only `compile_module_thread_safety`, `svelte_dev_corpus` and `svelte_dev_markdown`. `--test cli` — which carries the Tailwind tests — is not on that list, so it *does* run in the sharded `test` job, which sets `RSVELTE_REQUIRE_PREREQS` and installs no Tailwind oracle. Gating on the gate alone would have turned that job red. The two corpus binaries are excluded, so there the second clause is belt-and-braces; in `tailwind.rs` it is the whole fix.

**All sites, not the ones the repro exercises.** Eight assertion sites behind three helpers: `tw_fixture` ×2 (`tailwind.rs:168,178`), `svelte_dev_markdown` ×3 (`:132,144,159`), `svelte_dev_corpus` ×3 (`:215,229,244`). Every one routes through `tw_oracle_declared()` / `in_corpus_job()`, so converting the three helpers converts all eight — there is no site sitting on the old single-clause form. Confirmed negatively too: at head the only surviving reads of `RSVELTE_FMT_TW_TEST_DIR` and `FMT_CORPUS_OXFMT` are the payload lookups (`tw_test_dir`, `oxfmt_bin`) and doc text, not guards.

Routing every guard through one helper per suite is why this is checkable at all — it is the same property #2366's second round lacked, where the state was threaded to seven call sites and read at two.

The only unguarded early returns in the three files are `collect_samples` / `collect_markdown_samples`'s `read_dir` walk bail. An unreadable tree yields an empty sample set, which the *second* assert ("no fixtures at `fixtures/fmt-corpus/{sha}`") then catches. Guarded downstream, not a gap.

**The asymmetry inside `tailwind.rs` is the correct one, and worth naming.** After both PRs land, `node_runnable`'s guard (#2367) is single-clause while `tw_oracle_declared` is two-clause — in the same file. That looks inconsistent and is not: every job that sets the gate supplies node, and not every job that sets the gate supplies the oracle. The rule both files now follow is *one clause per prerequisite the setter does not supply unconditionally*. That sentence is worth putting in one of the two doc comments, because the next reader's instinct will be to "unify" them.

**Merge order — direction confirmed, and both orders are safe.** At #2367's head the gate var sits on the `sort_tailwindcss` step (`ci.yml:621`); this PR adds it to the corpus step (`:601-609`) only.

- **This PR first:** `tw_oracle_declared()` is false in the fmt-corpus Tailwind step, so the two `tw_fixture` asserts skip instead of firing. Fail-open in the safe direction — a guard that does not bite, never a spurious failure — scoped to Tailwind only, since the corpus guards get their var from this PR's own hunk. Self-resolves when #2367 lands. Your claim holds.
- **#2367 first:** it sets the var on a step whose helpers are still single-clause, so behaviour is unchanged; and in the shard job `RSVELTE_FMT_TW_TEST_DIR` is unset, so the old single-clause `tw_oracle_declared` still skips. No window where anything fails.

No job sets `RSVELTE_REQUIRE_PREREQS` together with `FMT_CORPUS_OXFMT` except the corpus step, and none sets it with `RSVELTE_FMT_TW_TEST_DIR` except the Tailwind step, so there is no combination that arms a guard whose prerequisite is absent.

**Textual merge, since both PRs touch the same two files.** The `ci.yml` hunks are disjoint (#2367 at old 291-296 and 611-618; this PR at 601-606). In `tailwind.rs` the hunk *contexts* overlap (old 147-155) but the changed lines do not: #2367 rewrites `node_runnable`'s body, this PR rewrites `tw_oracle_declared`'s doc + body and `tw_fixture`'s head, which #2367 leaves as `let tw_dir = tw_test_dir()?;`. It should merge clean. If it ever needs a hand merge, the result must keep **both** — the `let Some(tw_dir) = … else { assert!(!tw_oracle_declared(), …) }` head from here *and* the in-`node_runnable` assert from #2367. Dropping either silently restores a vacuous skip, which is precisely the failure neither PR would catch.

Issue #2090, opened 2026-07-31
compiler (client, dev): `$.track_reactivity_loss` residuals — legacy scripts, `<script module>` / `.svelte.js`, and comment-position svelte-ignore forms


baseballyama:
## Summary

Follow-ups deliberately left out of PR #2084 (which closed #2026 for the runes instance-script path):

1. **legacy (non-runes) instance scripts are not instrumented** — the official compiler instruments them; the rsvelte AST pass rides a runes-only branch, so covering legacy means relocating the pass, not extending the hook.
2. **`<script module>` / `.svelte.(js|ts)` are not instrumented** — cheap slice: add an await collector to `module_dev_tail_ast.rs`, same shape as the equality collector.
3. **svelte-ignore recognition is line-scan based** — ancestor-node comments, multi-line statements (await on a later line), and same-line block comments are not honored (official reads the analysis-phase ignore stack; proper fix is threading `ignore_stack` from 2_analyze into phase 3).
4. Cosmetic verbatim-copy differences (`await (load())` keeps parens; inner comments dropped) shared with the equality rewrite.

Also worth one extra unit assertion: the comma-form `// svelte-ignore await_reactivity_loss, other` (pins the `extract_svelte_ignore` delegation).

See PR #2084's "Known residuals" section and its review thread for the measured official outputs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FKkTpzZGaGWWzJvw4BnHJs

baseballyama:
Recount from PR #2175 (issue #2116 fix): instance scripts are now clean in **both** modes (legacy `track_reactivity_loss` under-emits 146 → 0, runes 97 → 0). All **215** remaining `track_reactivity_loss` under-emits are module-side — 197 in `<script module>` and 18 in `.svelte.(js|ts)` — the original table attributed files by their *instance* script mode, so the module half is 215, not 18. Concretely: `module_dev_tail_ast`'s batched rewrite has no `AwaitExpression` collector; adding the collector to that existing batch (mirroring the new `instance_dev_tail_ast`) should close this.

baseballyama:
Scope update: sub-problems 1 and 2 (plus the comma-form \`// svelte-ignore await_reactivity_loss, other\` assertion) are already resolved on \`main\`:

[...]

PR #2372, opened 2026-08-06
fix(compiler): copy await trivia into the `track_reactivity_loss` wrap (#2090)

baseballyama:
Partial fix for #2090 — sub-problem 4, `.svelte.(js|ts)` module path.

## What actually reproduces

Sub-problems 1–3 are confirmed fixed on `main` (`88b8d2b7`); only 4 remains, and it is
**smaller and differently located** than the issue's last comment says.

The issue lists four shapes under sub-problem 4. Each was run through the real corpus
normalizer (`flattenTemplateHoles` → oxfmt → `stripBlankLines`, with
`compatibility/.oxfmtrc.json`, the config `one.mjs` uses). **Three are not gate-visible
under oxfmt 0.62.0**:

| shape | raw | after normalization (oxfmt **0.62.0**) | oxfmt rewrote both sides |
|---|---|---|---|
| `await (load())` keeps parens | differs | MATCH | yes |
| one-lined `for await` / method bodies | differs | MATCH | yes |
| blank line before trailing `export { … }` | **already matched raw** | MATCH | yes |
| **inner comment dropped** | differs | **DIFFERS** — the gate-visible one | yes |
| control: identical inputs | — | MATCH | yes |
| control: `= 1` vs `= 2` | — | DIFFERS | yes |

Three notes on how to read that table, because each row rests on different evidence:

- **The version is load-bearing and was initially wrong.** The first run was made with the
  oxfmt actually installed here — **0.61.0** — while `package.json` pins `^0.62.0` (#2249);
  no install had run since the bump. The table above is a re-run against a freshly
  installed 0.62.0, not an inference from the 0.61 numbers. A MATCH under one formatter
  version does **not** imply a MATCH under another: normalization is applied to each side
  independently, and a more-collapsing normalizer *manufactures* MATCHes rather than
  preserving them. The paren row is the one least entitled to be carried by implication,
  since paren handling is exactly what moves between versions — so it was measured, not
  inherited.
- **"Not gate-visible" is not "equivalent output."** The compiler still emits the divergent
  bytes; normalization is comparison-side only. These return the moment oxfmt's paren or
  layout handling shifts.
- **The controls are what make the MATCH rows mean anything.** oxfmt 0.62 has been observed
  to walk zero files and exit 0, which yields MATCH for any input including one it never
  touched. Each row therefore reports whether oxfmt actually rewrote the text, and the two
  controls show the harness moving in both directions.

The `export { … }` row is the only version-independent one — it already matched **raw**.

So sub-problem 4 reduces to exactly one thing: a comment between `await` and its argument
is dropped.

## Root cause

`await_reactivity_loss_ast.rs` sliced the **argument's own span** for the replacement text:

```rust
let arg_span = expr.argument.span();
let arg_text = self.source[arg_span.start as usize..arg_span.end as usize].trim();
```

`/* hi */` in `await /* hi */ load()` sits between the `await` keyword and the argument, so
it fell outside that slice and was dropped. Upstream rebuilds the node as
`b.call('$.track_reactivity_loss', argument)` and esrap flushes the comment positionally
just before the argument, landing it inside the call — so copying from just past the `await`
keyword is the faithful mirror, not a workaround.

## Not a `.svelte.js` verbatim-copy problem

The issue attributes this to the `.svelte.(js|ts)` path's verbatim source copying and defers
it to the AST-ification thread (#2308). Measured: the comment is dropped in **all four**
script kinds — `.svelte.js`, `<script module>`, runes instance, legacy instance — and
official keeps it in all four. It is the shared await rewrite, not the module path.

## Scope of this PR

Fixes the `.svelte.(js|ts)` module path, verified end to end. The three component script
kinds still drop it **downstream of this pass** — instrumented, the collector now hands back
`(await $.track_reactivity_loss(/* hi */ load()))()` correctly for `<script module>` and the
legacy instance script, and the comment is lost after that; the runes instance script never
reaches this collector at all (it goes through `visitors/expression_converter.rs`, which
builds IR nodes and carries no comments). That is a separate, larger cluster and I am
reporting it rather than half-landing it.

Tests are the official compiler's output: block comment, a run of two comments, a line
comment, and the `svelte-ignore` path (which already passed — the wrap is skipped, so the
source is untouched). Each of the five was classified by mutation rather than by reading, because "guard" is
otherwise a label that absorbs any test that happens to pass on both revisions:

| test | mutant that kills it |
|---|---|
| `svelte_js_module_keeps_the_comment` | the copy's start bound |
| `every_comment_in_the_run_is_kept` | the copy's start bound |
| `a_line_comment_survives` | the copy's start bound |
| `an_ignored_await_is_left_alone` | **the `svelte-ignore` gate** — a different property |
| `a_parenthesized_operand_stays_balanced` | **none currently reachable** |

So it is three discriminating the copy fix, one discriminating the ignore gate, and **one
genuinely inert**. Grouping the last two as "guards" would have hidden that only one of
them is inert: disabling `self.ignored.contains(...)` fails the ignore test, while
reverting the end bound to `expr.argument.span().end` fails nothing at all.

That last row is worth stating plainly: **no test in this suite discriminates the end-bound
change**, because this caller parses with `ParseOptions::default()`, so `preserve_parens`
makes the argument span cover the parens and both bounds slice identical text. The change
is kept on a structural argument, not a tested one — the read range now equals the written
range by construction instead of by coincidence, which is what breaks the day a caller
parses with `preserve_parens: false` (`destructure_transforms.rs:1090`, `:1118` already
do). Unobservable under this caller is not the same as unverifiable: a test constructing
such a caller would discriminate it. The line-comment case pins only the comment's survival, because
upstream breaks the call across lines and a source splice cannot; oxfmt collapses both
layouts to the same bytes, verified directly.

## Ratchet

Not run (disk). This should close `client-dev` entries whose only divergence is a dropped
comment inside an awaited expression in a `.svelte.(js|ts)` module; the two-sided ratchet
will confirm.

---

## This PR changes component output, gate-invisible — added after review

Letting the comment survive makes an unrelated pre-existing defect observable. On the component `<script module>` path:

```js
official: return (await $.track_reactivity_loss(/* hi */ load()))();
this PR : return ((await $.track_reactivity_loss(/* hi */ load())))();
```

The comment lands in the right place; one redundant paren pair comes with it. Filed as #2381.

**It is not caused by this PR.** The defect lives in `rsvelte_esrap`'s comment-bearing `print_split` path, which the component path already takes on `origin/main`; it is unreachable for this shape today only because the comment is dropped before it gets there. Isolated: present with a comment, absent without.

**Not visible to any corpus gate** — redundant parens are MATCH after `oxfmt` 0.62 + `stripBlankLines`, measured with a control confirming oxfmt rewrote both sides and negative controls in both directions. Recording it here rather than letting it ride silently: a byte divergence only the author knows about is what the ratchet justifications exist to prevent.

Trade taken deliberately: blocking this would keep the *comment* bug live in order to avoid surfacing a *paren* bug.

### DIFF
--- a/.changeset/quiet-moons-clap.md
+++ b/.changeset/quiet-moons-clap.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): keep a comment between `await` and its argument in the dev `$.track_reactivity_loss` wrap
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/await_reactivity_loss_ast.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/await_reactivity_loss_ast.rs
@@ -293,11 +293,13 @@ impl<'a, 'src> Visit<'a> for AwaitCollector<'src> {
             return;
         }
 
-        // Copy from just past the `await` keyword, not from the argument: the
-        // trivia between them holds comments that upstream flushes inside the
-        // call, and slicing at the argument would drop them.
+        // Copy the operand region the replacement covers — from just past the
+        // `await` keyword to the expression's own end. Ending at the argument
+        // instead would drop the trivia holding comments upstream flushes
+        // inside the call, and would cut a parenthesized argument short of its
+        // `)` wherever the parse does not preserve parens.
         let arg_start = expr.span.start as usize + "await".len();
-        let arg_text = self.source[arg_start..expr.argument.span().end as usize].trim();
+        let arg_text = self.source[arg_start..expr.span.end as usize].trim();
         let wrap = track_reactivity_loss_wrap(arg_text);
         // The `;` rides inside this edit rather than as an insertion of its own,
         // which `splice`'s innermost-only filter would read as nested.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/await_reactivity_loss_ast.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/await_reactivity_loss_ast.rs
@@ -293,8 +293,11 @@ impl<'a, 'src> Visit<'a> for AwaitCollector<'src> {
             return;
         }
 
-        let arg_span = expr.argument.span();
-        let arg_text = self.source[arg_span.start as usize..arg_span.end as usize].trim();
+        // Copy from just past the `await` keyword, not from the argument: the
+        // trivia between them holds comments that upstream flushes inside the
+        // call, and slicing at the argument would drop them.
+        let arg_start = expr.span.start as usize + "await".len();
+        let arg_text = self.source[arg_start..expr.argument.span().end as usize].trim();
         let wrap = track_reactivity_loss_wrap(arg_text);
         // The `;` rides inside this edit rather than as an insertion of its own,
         // which `splice`'s innermost-only filter would read as nested.
--- a/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
+++ b/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
@@ -49,6 +49,21 @@ fn every_comment_in_the_run_is_kept() {
     );
 }
 
+/// The copied region runs to the `await` expression's own end, so a
+/// parenthesized operand keeps its closing paren however the parse treats
+/// parens — ending the copy at the argument would emit `(load()`.
+#[test]
+fn a_parenthesized_operand_stays_balanced() {
+    let src = "export async function f() {\n\treturn (await (load()))();\n}\n";
+    let out = module(src);
+    assert_contains(&out, "$.track_reactivity_loss((load()))");
+    assert_eq!(
+        out.matches('(').count(),
+        out.matches(')').count(),
+        "unbalanced parens:\n{out}"
+    );
+}
+
 /// Upstream breaks the call across lines for a line comment
 /// (`$.track_reactivity_loss(\n\t// hi\n\tload()\n)`); splicing source text
 /// keeps it on the opening line. oxfmt collapses the two layouts to the same
--- a/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
+++ b/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
@@ -49,9 +49,12 @@ fn every_comment_in_the_run_is_kept() {
     );
 }
 
-/// The copied region runs to the `await` expression's own end, so a
-/// parenthesized operand keeps its closing paren however the parse treats
-/// parens — ending the copy at the argument would emit `(load()`.
+/// Preservation guard, not a regression test: this caller parses with
+/// `ParseOptions::default()`, where `preserve_parens` makes the argument span
+/// cover the parens, so it passes on either end bound. It would start
+/// discriminating only under `preserve_parens: false`, which is why the copy
+/// runs to the expression's own end — the read range then equals the written
+/// range by construction rather than by coincidence.
 #[test]
 fn a_parenthesized_operand_stays_balanced() {
     let src = "export async function f() {\n\treturn (await (load()))();\n}\n";
--- a/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
+++ b/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
@@ -1,0 +1,71 @@
+//! A comment between `await` and its argument survives the dev
+//! `$.track_reactivity_loss` wrap.
+//!
+//! Upstream rebuilds the expression as `b.call('$.track_reactivity_loss',
+//! argument)` and esrap flushes the comment positionally just before the
+//! argument, so it lands inside the call. rsvelte splices the argument's source
+//! text, which starts at the argument and therefore left the trivia between the
+//! `await` keyword and the argument outside the replaced range — dropping it.
+//!
+//! Expected strings are the official compiler's output.
+
+use rsvelte_core::compiler::ModuleCompileOptions;
+use rsvelte_core::{GenerateMode, compile_module};
+
+fn module(src: &str) -> String {
+    compile_module(
+        src,
+        ModuleCompileOptions {
+            filename: Some("m.svelte.js".to_string()),
+            generate: GenerateMode::Client,
+            dev: true,
+            ..Default::default()
+        },
+    )
+    .expect("compile")
+    .js
+    .code
+}
+
+#[track_caller]
+fn assert_contains(out: &str, expected: &str) {
+    assert!(out.contains(expected), "expected `{expected}`. Got:\n{out}");
+}
+
+const WRAPPED: &str = "(await $.track_reactivity_loss(/* hi */ load()))()";
+
+#[test]
+fn svelte_js_module_keeps_the_comment() {
+    let src = "export async function f() {\n\treturn await /* hi */ load();\n}\n";
+    assert_contains(&module(src), WRAPPED);
+}
+
+#[test]
+fn every_comment_in_the_run_is_kept() {
+    let src = "export async function f() {\n\treturn await /* a */ /* b */ load();\n}\n";
+    assert_contains(
+        &module(src),
+        "(await $.track_reactivity_loss(/* a */ /* b */ load()))()",
+    );
+}
+
+/// Upstream breaks the call across lines for a line comment
+/// (`$.track_reactivity_loss(\n\t// hi\n\tload()\n)`); splicing source text
+/// keeps it on the opening line. oxfmt collapses the two layouts to the same
+/// bytes, so only the comment's survival is pinned here.
+#[test]
+fn a_line_comment_survives() {
+    let src = "export async function f() {\n\treturn await // hi\n\t\tload();\n}\n";
+    let out = module(src);
+    assert_contains(&out, "$.track_reactivity_loss(");
+    assert_contains(&out, "// hi");
+}
+
+/// The wrap is skipped entirely when ignored, so the source is untouched.
+#[test]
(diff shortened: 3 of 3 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. The fix is range-consistent (I checked the read range against the replaced range, which is where this class of change usually duplicates or truncates), the upstream-mirror claim holds, and three of the four tests discriminate with the fourth correctly labelled. Two nits, one on the normalizer reasoning.

Read-only review at `83a11efe` — no build, no test run.

---

### Verified: the fix does not duplicate or truncate

This is the failure mode a "read from further left" change invites, so I checked it explicitly. The edit is `(start, expr.span.end, replacement)` (`await_reactivity_loss_ast.rs:313`) and `track_reactivity_loss_wrap` is `format!("(await $.track_reactivity_loss({argument_text}))()")` (`:25-27`) — so the **whole** `AwaitExpression`, `await` keyword included, is replaced, and the template re-emits `await`. Widening the read range to start just past `await` therefore moves the trivia from outside the replaced range to inside the emitted text exactly once. No duplicated comment, no orphaned `await`. ✓

The line-comment case is safe for the reason that is easy to miss: `arg_text` retains the newline that terminates the `//`, so the template's closing `))()` lands on the following line rather than being commented out. A `//` between `await` and its argument always has a newline before the argument, so there is no shape where this breaks.

Both offsets are char boundaries — `span.start + 5` is past the ASCII `await`, and a node span end is a boundary — so no non-ASCII hazard here.

### Verified: the upstream-mirror claim

`b.call('$.track_reactivity_loss', argument)` rebuilds the node and esrap flushes pending comments positionally at the argument, so upstream lands the comment inside the call. Copying source from just past the keyword reproduces that placement rather than compensating for it. That is the right justification and it matches what the tests assert.

### 1. Nit — tie the read end to `expr.span.end`, not to the argument's

`:301` reads `self.source[arg_start..expr.argument.span().end as usize]` while `:313` replaces up to `expr.span.end`. Those are equal today only because the caller parses with `ParseOptions::default()` (`module_dev_tail_ast.rs:89`), i.e. `preserve_parens: true`, which makes `await (load())`'s argument a `ParenthesizedExpression` spanning the parens.

Flip that one option and the two ends diverge: `argument.span().end` falls at the inner `)`, the read text becomes `(load()` — unbalanced — and the emitted wrap is a **syntax error**, not a cosmetic diff. That is not hypothetical in this crate; `destructure_transforms.rs:1176` and `:1204` already pass `preserve_parens: false`, so the two conventions coexist a few files apart.

Under the old code the asymmetry was benign because the read start also came from the argument. The new code makes the start independent of the argument, so the end should be too. `expr.span.end` is a byte-for-byte no-op today and immune tomorrow.

(Incidentally this also explains the issue's "`await (load())` keeps parens" row: with parens preserved, rsvelte copies them verbatim while upstream's builder drops them.)

### 2. The normalizer call needs a "did the normalizer do work?" control, and a different label

Striking three shapes because they normalize to MATCH is a reasonable *gate-scoped* conclusion, but two things about it should change.

**First, the verification.** The relevant recent failure was not a config mismatch — it was that oxfmt 0.62 **walked zero files** because the corpus tree was gitignored, and reported `exit 0` on every entry. A run that formats nothing produces "MATCH after normalization" for any input, including a real divergence. The PR says the config matched `one.mjs`'s, which is the right thing to check but not the thing that failed. Please report the count of files oxfmt actually reformatted, or show a byte diff between the pre- and post-normalizer output proving the normalizer moved something on these specific inputs. Two of the three rows claim oxfmt *changed* the text (dropped parens, re-laid-out bodies), so that evidence is one command away and would settle it.

**Second, the label.** "Not worth fixing" and "not visible to the gate" are different claims, and only the second is established. Normalization is comparison-side only, never a compiler post-pass — so a shape oxfmt hides is still a byte divergence a user's unformatted build sees, and it becomes gate-visible again the moment oxfmt's paren or layout handling shifts (it has shifted twice recently, 0.61 → 0.62). Recommend keeping all three recorded in #2090 marked **not gate-visible**, with the oxfmt version that made them so, rather than struck. That costs nothing and means the next oxfmt bump surfaces them as a known list instead of as three mystery regressions.

The fourth row — the dropped comment — is the one you fixed, and it is correctly identified as the only one the gate can see.

### 3. The attribution correction is well supported and worth propagating

Measuring all four script kinds and finding the comment dropped in each is the right way to falsify "this is the `.svelte.(js|ts)` verbatim-copy path", and the follow-up detail is more useful than the headline: the collector now returns the correct text for `<script module>` and the legacy instance script and the comment is lost **downstream**, while the runes instance script never reaches this collector at all (`visitors/expression_converter.rs` builds IR nodes carrying no comments). Those are two different causes, not one. Please make sure both land in #2090 separately — "still drops it" would otherwise read as one remaining bug when it is two.

### 4. Tests

Three of four discriminate against the pre-fix slice (`svelte_js_module_keeps_the_comment`, `every_comment_in_the_run_is_kept`, `a_line_comment_survives` — all three assert a comment the old range excluded by construction). `an_ignored_await_is_left_alone` is correctly declared as already passing; it is a preservation guard for the ignore path, and labelling it rather than counting it 4/4 is the standard this epic has settled on.

`a_line_comment_survives` deliberately asserts substrings rather than the exact official bytes, with the reason stated (upstream breaks the call across lines; a source splice cannot). That is the honest form — it pins survival, which is the property under test, and does not pretend to pin layout.

baseballyama:
Re-reviewed at `19a81015`. **0 blocking.** The span change is right and I verified it end to end. Two things need correcting, both about labels rather than code — one on the new test, one on the normalizer rows, which are still on the PR in their original form.

Read-only — no build, no test run.

### The span fix is correct

`self.source[arg_start..expr.span.end]` now matches the replaced range `(start, expr.span.end, replacement)` at `:313` exactly, so the read range and the write range are the same range. That is the property that makes this class of edit safe to reason about, and it is now structural instead of coincidental. The updated comment states both reasons (trivia, and the paren case) rather than only the one that motivated the original change.

### The new test is a preservation guard, not a repro — please label it as one

This is the only thing I would insist on. `a_parenthesized_operand_stays_balanced` **passes on `83a11efe` as well**, so it does not discriminate against the state it appears to pin.

The caller parses with `ParseOptions::default()` (`module_dev_tail_ast.rs:89`), and oxc's default is `preserve_parens: true` (`oxc_parser/src/lib.rs:242`). Under that option `await (load())`'s argument **is** a `ParenthesizedExpression` spanning the parens, so `expr.argument.span().end == expr.span.end` and both revisions slice identical text. The docstring — "ending the copy at the argument would emit `(load()`" — describes a counterfactual that is false for this caller today; it becomes true only under `preserve_parens: false`, which is what `destructure_transforms.rs:1090` and `:1118` use a few files away.

That is exactly why the change is worth making, and I am not asking for it to be reverted or for a different test. I am asking the docstring to say what it actually pins:

> No-op under today's `ParseOptions::default()` (`preserve_parens: true`), where the argument span already includes the parens. This pins the invariant against a future caller that parses with `preserve_parens: false`, under which ending the copy at the argument would emit `(load()`.

This epic has already settled the standard — `an_ignored_await_is_left_alone` is labelled correctly in this same file — and the count in the PR should be "three discriminating tests plus two preservation guards", not four plus one. The paren-balance assertion is a good addition either way; a global invariant that would catch the failure by a different route is worth having even when the shape assertion already covers it.

### The normalizer rows: the relabel is not on the PR, and the implication does not carry the weight

The PR body is unchanged. It still reads *"three of them do not survive it, so the gate cannot see them and they are not worth fixing"*, and still credits the run to **oxfmt 0.62** — which is the version the run did not use. **The relabel is on none of the three rows, not all three.** Whatever was established since, the artifact that outlives this thread is #2090 and this body, and neither records it.

On the reasoning offered for carrying the rows over — that a MATCH under weaker normalization implies a MATCH under stronger — **that implication does not hold, and here it points the wrong way.** Normalization is applied to each side independently, so `f(a) == f(b)` constrains `g(a) == g(b)` only if `g` factors through `f`; two formatter versions are not ordered that way. And the concrete failure was that the older binary *rewrote* text toward one shape, i.e. it collapsed distinctions — a more-collapsing normalizer manufactures MATCHes. A 0.61 MATCH is therefore the weaker evidence in the adverse direction: it is exactly the result that can flip to DIFFERS under 0.62, not the one that is safe to carry.

Per row, the exposure is not uniform:

- **`blank line before trailing export { … }`** — recorded as `already matched` **raw**. Version-independent; survives unconditionally, no re-run needed.
- **`one-lined for await / method bodies`** — rests on oxfmt re-laying out bodies. Layout is the most stable part of a formatter across a patch bump, but it still rests on the instrument, so it needs the number from the run that was actually made.
- **`await (load())` keeps parens** — rests specifically on oxfmt **dropping redundant parens**. Paren handling is precisely what shifted in the incident that started this. This is the row least entitled to be carried by implication and the one most worth re-stating with a version.

The two controls that were run are the right ones and they answer my original nit properly: identical inputs → MATCH is the positive control (the instrument reports equality when equality holds), `=1` vs `=2` → DIFFERS is the negative control (the instrument can move). Together they rule out the zero-files-walked failure, which is what "did the normalizer do work?" was asking. They are worth two lines in the body — they are stronger evidence than the implication and they do not depend on an ordering between versions that does not exist.

Concretely, what I would like in the body before merge: the three rows kept and marked **not gate-visible under oxfmt 0.62.0** rather than struck as "not worth fixing", the two controls stated, and the version corrected. Then the next oxfmt bump surfaces them as a known list instead of three mystery regressions — which was the whole point of keeping them.

### Unchanged from my earlier review

The no-duplicate/no-truncate check, the upstream-mirror justification, the two-causes attribution correction (`<script module>` + legacy instance lose it downstream; runes never reaches this collector), and the honest substring assertions in `a_line_comment_survives` all still stand. The attribution split is the most reusable thing in this PR and should land in #2090 as two entries.

Issue #2064, opened 2026-07-31
compiler (client, dev): ~1740 corpus entries diverge for causes no dev-cluster issue covers (JSDoc comment loss, legacy bind get/set form, $.assign, $$css shape)


baseballyama:
## Summary

A full expected/actual sweep of the 4565-entry `known-failures.client-dev.json` baseline (done while triaging #2020-#2030/#2039/#2040) shows that fixing all 14 dev-cluster issues turns only ~2825 entries green. **~1740 entries also (or only) diverge for causes no existing issue tracks**:

- **JSDoc / block comment loss** (98 entries as sole cause, many more combined) — likely related to client codegen falling back to the text printer for comment-bearing programs
- **legacy `bind:` accessors emitted as arrow functions** where official emits `function get() {}` / `function set($$value) {}`
- **`$.assign`**: 23 missing / 2 extra
- **`const $$css = { hash, code }` shape** differences

## Data

Per-entry cause attribution: `cause2.mjs` / `model2.json` (session scratchpad; regenerate by diffing `compatibility/{expected,actual}` for the client-dev target). Also note #2030 (26 entries, UTF-8 byte vs char column counting) and #2039 (6 entries) live inside this OTHER bucket in the tables posted to the cluster issues.

Each bullet above probably deserves its own issue once scoped; this issue tracks the umbrella so the '14 issues = the whole baseline' assumption doesn't mislead planning.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FKkTpzZGaGWWzJvw4BnHJs

baseballyama:
Addition from the #2026 implementation (PR #2084): `// svelte-ignore` line comments inside the instance script are dropped from **SSR** output (reproduces with dev=false, so unrelated to the dev instrumentation work). Same family as the JSDoc/block-comment loss tracked here.

baseballyama:
## `known-failures.client-dev.json` residue — full attribution (203 entries on `902bbb1d`)

Method (deliberately **not** first-diff-line, which is a documented mis-attribution trap):
for every baseline entry, compile both compilers at `generate: 'client', dev: true` and
**count occurrences of every `$.<helper>` / `$$<ident>` token** on each side. Group by the
multiset of count deltas (`rsvelte − official`). Entries whose helper counts all match are
structural (ordering / naming / whitespace / comments) and are labelled `STRUCTURAL`.

### Cluster table

[...]

PR #2374, opened 2026-08-06
fix(compiler): only exempt an event attribute's own arrow from the dev `$.assign` wrap (#2064)

baseballyama:
Closes the last `client-dev` ratchet entry. Part of #2064.

## Re-derived first — the issue is very stale

#2064's title says **~1740 entries**. Its own later comments re-measured that to 133, and
`compatibility/known-failures.client-dev.json` on `main` holds **1**. Of the four causes in
the title, only `$.assign` is still live; and of the eight structural sub-causes in the
2026-08-05 comment, **sub-cause B (17 entries, the largest and the one marked "ready to
implement") was already fixed** by #2316 (`ab2d6361`) — `props_transforms.rs` no longer has
the single `loc_cursor` that comment quotes.

So the one thing actually left is the `$.assign` row.

## The bug

`runed/sites/docs/src/lib/components/demos/scroll-state.svelte`:

```svelte
<form onsubmit={preventDefault(() => (scroll.x = x))}>
```

rsvelte emitted the bare `scroll.x = $.get(x)`; upstream emits
`$.assign(scroll, "x", "=", $.get(x), "…scroll-state.svelte:41:69")`.

`AssignmentExpression.js:183-201` exempts an event handler arrow from the coerced-away-proxy
warning **only when the arrow is the attribute's expression**:

```js
if (path.at(-1) === 'ArrowFunctionExpression' &&
    (path.at(-2) === 'RegularElement' || path.at(-2) === 'SvelteElement')) { … }
```

Here `path.at(-2)` is the `CallExpression`, so the exemption does not apply. rsvelte set
`in_event_attribute_handler` around the **whole** attribute expression, so any arrow anywhere
inside it — including one passed as a call argument — was exempted.

## The fix

One line of condition: set the flag only when the attribute expression is itself an arrow.

Dev-only by construction — the wrap returns early on `!context.state.dev`
(`expression_converter.rs:5115`), so the non-dev `client` target, which is at 0, cannot move.

## Tests

- **Repro** (fails on `main`): the arrow inside a call argument is wrapped.
- **Pin**: `onclick={() => (scroll.x = x)}` — the arrow *is* the attribute expression, stays exempt.
- **Pin**: a nested arrow with no event attribute anywhere is wrapped.

Both pins were green before the fix; they are labelled as pins in the file rather than counted
as repros.

## Ratchet

`known-failures.client-dev.json` goes to `[]` in this PR, as the two-sided ratchet requires. If
the fix does not in fact close that entry, CI will say so — that is the confirmation, not a
local corpus run (disk is tight and the corpus was not run).

## Left over, deliberately

The exemption is carried by a level flag that stays set for the whole body conversion, so an
assignment nested *inside* an exempt arrow's body — `onclick={() => (a.b = f(() => (c.d = e)))}` —
is still exempted, where upstream would wrap the inner one. No corpus entry covers it and I
could not justify shipping an unverifiable second change alongside this one. Recorded in
`known-failures.md`.

### DIFF
--- a/.changeset/dev-assign-nested-event-arrow.md
+++ b/.changeset/dev-assign-nested-event-arrow.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): only exempt an event attribute's own arrow from the dev `$.assign` wrap
--- a/compatibility/known-failures.client-dev.json
+++ b/compatibility/known-failures.client-dev.json
@@ -1,3 +1,1 @@
-[
-	"runed/sites/docs/src/lib/components/demos/scroll-state.svelte"
-]
+[]
--- a/compatibility/known-failures.md
+++ b/compatibility/known-failures.md
@@ -225,11 +225,23 @@ the equality instrumentation, `$.track_reactivity_loss`, ownership mutation
 validation, `$.tag()` / `$.tag_proxy()`, `console.*` wrapping and the signal
 read/write row — was already empty.
 
-Still over-reaching, and not covered by any corpus entry: the exemption is
-carried by a level flag that stays set for the whole body conversion, so an
-assignment nested *inside* an exempt arrow's body
-(`onclick={() => (a.b = f(() => (c.d = e)))}`) is exempted too. Upstream would
-wrap the inner one.
+Two divergences remain here, both deferred only because **no corpus entry
+reaches them** — each has a check that distinguishes fixing from not, so neither
+is unverifiable:
+
+- **Over-reach.** The exemption is carried by a level flag that stays set for the
+  whole body conversion, so an assignment nested *inside* an exempt arrow's body
+  is exempted too. `onclick={() => (a.b = f(() => (c.d = e)))}` must emit
+  `$.assign(c, 'd'` and must not emit `$.assign(a, 'b'` — one input, both signs.
+  A boolean cannot express upstream's third conjunct, which is the *identity*
+  test `expression === context.path.at(-1)`; the exempt arrow has to be carried
+  by identity, not by a level.
+- **Under-reach, the opposite direction.** Upstream's guard names
+  `SvelteElement` alongside `RegularElement`, but `visit_event_attribute` is
+  reached only from `regular_element.rs`, so `<svelte:element this={tag}
+  onclick={() => (o.x = v)}>` is never exempt and emits
+  `$.assign(o, 'x', '=', v, …)` where upstream emits none. Measured, not
+  inferred.
 
 Counting method, for whoever picks this up: attribute an entry by **comparing
 how many times each helper appears** on each side, never by the first differing
--- a/compatibility/known-failures.md
+++ b/compatibility/known-failures.md
@@ -63,7 +63,7 @@ quoted key dropped in a destructured `$derived`) and #2034 (`$.to_array` arity
 with a rest element) — were resolved by #2036, which mirrored #2010's client
 destructuring fixes onto the server target.
 
-## Client dev (`known-failures.client-dev.json`, 1 entry)
+## Client dev (`known-failures.client-dev.json`, 0 entries)
 
 The `client-dev` target is the `client` target with `dev: true`. It is a
 separate ratchet because `dev` gates 18 client codegen files plus the CSS
--- a/compatibility/known-failures.md
+++ b/compatibility/known-failures.md
@@ -212,18 +212,24 @@ that separator still carry source-map segments.
 
 ### What is left
 
-One entry, contributed by the `runed` / `svelte-toolbelt` enrolment:
-`runed/sites/docs/src/lib/components/demos/scroll-state.svelte` writes
-`onsubmit={preventDefault(() => (scroll.x = x))}`, and rsvelte emits the bare
-`scroll.x = $.get(x)` where upstream wraps it as
-`$.assign(scroll, "x", "=", $.get(x), "…scroll-state.svelte:41:69")`. That is
-the `$.assign` row of #2064 — the coerced-away-proxy dev warning
-(`AssignmentExpression.js:170-236`) — reached here through an assignment nested
-inside a call argument in a template attribute expression. Every other
-dev-helper cluster the enrolment-era table tracked — the equality
-instrumentation, `$.track_reactivity_loss`, ownership mutation validation,
-`$.tag()` / `$.tag_proxy()`, `console.*` wrapping and the signal read/write
-row — is empty.
+Nothing. The last entry — `runed/…/demos/scroll-state.svelte`, which writes
+`onsubmit={preventDefault(() => (scroll.x = x))}` and had rsvelte emitting the
+bare `scroll.x = $.get(x)` where upstream wraps it as
+`$.assign(scroll, "x", "=", $.get(x), "…scroll-state.svelte:41:69")` — is fixed.
+The event-attribute exemption from the coerced-away-proxy dev warning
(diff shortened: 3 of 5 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. The upstream claim holds, the ratchet edit is exactly scoped, and both pins are correctly labelled. One correction: the reason given for deferring the nested case is wrong in a way worth fixing before it becomes precedent.

Read-only review at `f27bc24f`. No build, no test run.

---

### Upstream claim: verified

`3-transform/client/visitors/AssignmentExpression.js:183-201`:

```js
if (
    path.at(-1) === 'ArrowFunctionExpression' &&
    (path.at(-2) === 'RegularElement' || path.at(-2) === 'SvelteElement')
) {
    const element = context.path.at(-2);
    const attribute = element.attributes.find((attribute) => {
        …
        return expression === context.path.at(-1);
    });
    if (attribute) should_transform = false;
}
```

Three conjuncts, and the `find` makes the third explicit: the arrow must **be** the attribute's expression, by identity. For `onsubmit={preventDefault(() => (scroll.x = x))}` the assignment's `path.at(-1)` is the arrow but `path.at(-2)` is the `CallExpression`, so the exemption never applies. Narrowing the flag to "the attribute expression is itself an arrow" is the faithful mirror. ✓

### 1. "No check distinguishes fixing from not" is not accurate — here is the check

The deferral of `onclick={() => (a.b = f(() => (c.d = e)))}` rests on two claims, and only one of them is true:

- *No corpus entry reaches it* — true, and a perfectly good reason to defer.
- *No check distinguishes fixing from not* — not true. The distinguishing check is the same shape as this PR's own repro, and it fits in one test:

```rust
#[test]
fn an_arrow_nested_inside_an_exempt_arrow_is_still_wrapped() {
    let out = compile_client_dev(
        r#"<script>
	let a = $state({ b: 0 });
	let c = $state({ d: 0 });
	let e = $state(1);
	const f = (fn) => fn();
</script>

<button onclick={() => (a.b = f(() => (c.d = e)))}></button>
"#,
    );
    assert!(out.contains("$.assign(c, 'd', '='"), "inner arrow is not the attribute expression");
    assert!(!out.contains("$.assign(a, 'b'"), "outer arrow is the attribute expression");
}
```

Upstream wraps the inner assignment (`path.at(-2)` is the `CallExpression` `f(…)`) and exempts the outer one (`path.at(-2)` is the `RegularElement`), so the two assertions pin opposite outcomes on the same input — which is exactly what "distinguishes fixing from not" means. It needs no corpus and no submodule.

This matters beyond this PR because it is the same conflation the epic has been unpicking elsewhere: *not visible to the ratchet* and *not verifiable* are different claims, and only the first is established here. Please restate the note in `known-failures.md` as "no corpus entry reaches it" and drop the second half, or the next person reads it as "cannot be tested" and does not look.

For whoever does pick it up: the faithful shape is not a deeper flag but a different one — carry the identity of the exempt arrow (the attribute expression node) rather than a boolean level, and exempt an assignment only when its nearest enclosing arrow **is** that node. That is a direct translation of upstream's `expression === context.path.at(-1)`, and it collapses the nesting case for free. A boolean cannot express identity, which is why it over-reaches.

### 2. Nit — upstream's condition has a `SvelteElement` half that this path cannot reach

The code comment at `attribute.rs:162-164` renders the guard as `path.at(-2) === 'RegularElement'`, dropping `|| path.at(-2) === 'SvelteElement'`. That is cosmetic in the comment, but it points at something real: `visit_event_attribute` is called from exactly one place, `regular_element.rs:531`, and there is no `svelte_element.rs` visitor in `client/visitors/` (the `mod.rs:26` doc lists one that does not exist, and `attribute.rs:180` separately notes "SvelteElement would need separate handling if we add that visitor" about delegation).

So `<svelte:element this="button" onclick={() => (o.x = v)}>` appears never to reach this exemption at all, where upstream exempts it — an under-reach in the opposite direction from the one this PR fixes. I have not traced where SvelteElement event attributes are handled, so I am flagging the asymmetry rather than asserting a divergence; it is worth one grep before the comment is left saying `RegularElement` only.

### 3. Ratchet edit — correctly scoped, and self-checking

The JSON goes from one entry to `[]`, and that entry (`runed/…/demos/scroll-state.svelte`) is precisely the shape `an_arrow_nested_in_a_call_argument_is_still_wrapped` reproduces — `onsubmit={preventDefault(() => (scroll.x = x))}`. Nothing broader is removed. Deferring confirmation to CI is sound here because the ratchet is two-sided: if the fix does not close it the job fails on a new failure, and if it were left listed after passing the job fails on a stale entry. Either way the emptying cannot pass silently.

The dev-only argument checks out as a reason the `client` target cannot move: the wrap returns early on `!context.state.dev`.

### 4. `f27bc24f` — verified, and sufficient rather than partial

The commit is a single line, `1 entry` → `0 entries`, matching `[]`. I also checked the second rule the consistency gate enforces, since a header sync that leaves the other rule stale would be worse than none: `known-failures-md-check.mjs:72` only fires when the file contains an `"N entries are attributed to a cluster; the remaining M"` sentence, and that sentence is **not present** in `known-failures.md` at this head. So the header count was the only thing to sync. Clean.

### 5. Test classification — both pins correct

- `an_arrow_nested_in_a_call_argument_is_still_wrapped` — discriminates: on `main` the flag was set unconditionally, the nested arrow was exempt, and no `$.assign` was emitted.
- `the_attribute_expression_arrow_itself_stays_exempt` — pin. Green before, since the flag was true for this shape too.
- `a_nested_arrow_that_is_not_under_an_event_attribute_is_wrapped` — pin. Green before: `onclick={go}`'s expression is an identifier, and `go`'s body is converted at script level, never under the flag.

Labelling the two in-file rather than counting 3/3 is the standard this epic settled on, and it is applied correctly.

baseballyama:
Re-reviewed at `0273e913`. **LGTM — 0 blocking.** The recorded reason is now the true one, and the record is stronger than what I asked for. Read-only — no build, no test run.

**The false half is gone and the true half is intact.** "No corpus entry reaches them" stays; "no check distinguishes fixing from not" is gone. Those were doing very different work — the first is a legitimate deferral reason, the second was a claim about verifiability that was not true — and separating them is the whole point.

**Spelling the check out inline is better than removing the false claim.** `onclick={() => (a.b = f(() => (c.d = e)))}` must emit `$.assign(c, 'd'` and must not emit `$.assign(a, 'b'` — one input, both signs — is now in the file, so whoever picks this up starts with the check already written rather than re-deriving it. And keeping the identity-vs-boolean diagnosis next to it (upstream's third conjunct is `expression === context.path.at(-1)`, which a level flag cannot express) records *why* the obvious patch will not work, which is the part that would otherwise be rediscovered by someone half-fixing it.

**The `SvelteElement` under-reach is recorded rather than silently carried**, and recorded in the right terms — as the *opposite* direction, with the mechanism (`visit_event_attribute` reached only from `regular_element.rs`), the shape (`<svelte:element this={tag} onclick={() => (o.x = v)}>`), what each side emits, and "Measured, not inferred". Two divergences in opposite directions from one guard is exactly the thing a burn-down note should not collapse into "still over-reaching", which is what the previous text said.

**Gate check.** The section headers still agree with the JSON at this head — `client` 0/`[]`, `server` 1/1 entry, `client-dev` 0/`[]` — so `known-failures-md-check.mjs` passes on counts, and the client-dev "attributed to a cluster / remaining" reconciliation sentence is absent, which its own "when that sentence is present" clause makes a no-op rather than a failure. The edit is prose-only and does not disturb either.

Nothing further from me. The deferral is now recorded in a form that survives the next person reading it.

Issue #2342, opened 2026-08-06
perf(compiler): residual O(n²) backward scans in prop-read rewriting


baseballyama:
Follow-up carved out of #2322 / PR #2339, which removed the dominant quadratic term (`result.char_indices().nth(i)` resolved per character inside the per-character loop, in `transform_prop_reads_in_expr`, `3_transform/client/props_transforms.rs`).

After that fix the fitted log-log slope is **1.799**, down from 2.027 (n = 6 sizes, 2 KB → 63 KB, in-process hrtime). Still above linear, so at least one quadratic term remains.

Suspects, all on the per-*matched-identifier* path rather than the per-character path — each is an O(n) backward scan fired O(n) times:

- `is_shadowed_by_function_param`
- `is_explicit_property_key`
- `is_arrow_param_binding`

Absolute residual is **17x smaller** than what #2339 removed (240 ms vs 4145 ms at a 63 KB script), so this is a tail-of-a-tail. Before spending time here, weigh it against who is affected:

Only **legacy (non-runes) components with a physically large `$:` block** enter this path at all. Of the six shipped corpus projects only `svelte-ux` triggers it — its `Button.svelte` has a single `$: _class = cls(…)` running from line 44 to the end of the script, so ~16 KB of a 17.2 KB instance script arrives as one reactive statement. Modern large Svelte projects have **zero** `$:` statements, and the Svelte test fixtures over-represent legacy syntax by roughly 3-5x, so a fixture-corpus measurement will overstate the impact. This is also not on the svelte-rs gap population, so closing it closes no competitive gap.

Filing it so the measurement is not lost, not because it is next. Any work here should re-measure with a paired A/B (median of the ratio, back-to-back runs) rather than min-of-N — on a shared machine min-of-N inflates an improvement by roughly 1.7x.

PR #2376, opened 2026-08-06
fix(compiler): match destructuring brackets lexically (#2357)

baseballyama:
Batch (b) of #2357, based on `origin/main` — independent of #2366/#2370, no stacking needed since it touches a different file.

`find_matching_open_bracket` in `client/destructure_transforms.rs` walked backwards from a closing bracket counting every `{`/`[`/`}`/`]` it passed, with **no** string, comment or regex handling at all. A destructuring assignment whose pattern carries a bracket inside a string or comment therefore never finds its opener, `find_and_transform_one_destructure` discards the candidate, and the statement is left untransformed.

Both repros confirmed failing before the fix:

```rust
find_matching_open_bracket(r#"{ a = "}" } = obj"#, 10, '{', '}')   // None, want Some(0)
find_matching_open_bracket("{ a, /* } */ b } = obj", 15, '{', '}') // None, want Some(0)
```

The first is ordinary JavaScript — a destructuring default value containing a brace. Both now return `Some(0)`. A third test pins that genuine nesting (`{ a: { b } } = obj`) still resolves to the outer bracket; it passed before and after and is a guard, not a repro.

**Interface change.** This is the `Vec<char>` cluster the epic flagged as unswept, so the conversion is the whole-function kind rather than a `code_bytes` swap: `close_pos` and the return value are now **byte** offsets, because `code_bytes` yields byte offsets and threading char indices through it would reintroduce exactly the char/byte mixing that has already produced three bugs in this epic. The single caller indexes by char, so it converts on the way in with the `b()` table it already builds, and on the way out with a `binary_search` over the same `byte_offsets` vector — no new mapping state.

The backward walk collects `code_bytes` forward and iterates the result in reverse, the same shape used for the reverse scanners in #2366.

Not addressed here, and worth its own look: the helper rebuilt a `Vec<char>` of the whole statement on **every** call and is called once per candidate bracket, so it was O(n) allocation fired O(n) times. The new version still collects a vector per call. That is a performance question with the same shape as #2342, and it should be measured rather than assumed — I have not measured it and am not claiming an improvement.

`cargo test -p rsvelte_core --lib destructure_transforms::tests` 3/3. `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` clean, no `-A`. `--test ssr` / `--test runtime` cannot run here (`Fixtures not found`).

---

**Update: `find_colon_at_depth0` settled — it is reachable, and it panics.** (`00aa2e15`)

This was the epic's one "suspected but unconverted for want of a reaching input" entry. The input exists:

```rust
check_identifier_in_statement("cond ? component = \"ああa\" : component = other", "component", &re)
// panicked: end byte index 7 is not a char boundary; it is inside 'あ' (bytes 5..8)
```

`find_colon_at_depth0` counted **characters**; its single caller (`state_transforms.rs:748`) slices `rhs` with the result, i.e. by **bytes**. The two agree only while the expression is ASCII. This is a compiler panic, not a mis-slice, so it is a harder failure than the other instances of this class — the ternary branch just has to contain a non-ASCII string literal before the `:`.

Three tests, all confirmed failing first: the byte-offset contract, the panic through the real caller, and a `:` written inside a comment being read as the branch separator.

**The `${…}` carve-out applies here and I did not convert this one to `code_bytes`.** The scanner descends into `${…}` deliberately — it increments `depth` on `${` and decrements on the matching `}` — so an opaque template skip would change behaviour. Instead the loop keeps its hand-rolled structure and only changes its index unit from char to byte, with comment handling added by hand. UTF-8 continuation bytes are all `>= 0x80`, so they never match an ASCII arm and the `_ => {}` case absorbs them.

**On accidental compensation** (asked explicitly): there is none to preserve in either function. `find_matching_open_bracket`'s char index was consumed as a char index throughout, so converting it to bytes required converting the boundary and nothing downstream depended on the old unit. `find_colon_at_depth0`'s char index was consumed as a *byte* index by its only caller — the units never matched, so there was no working compensation, only a latent panic. In neither case did the old indexing make something incidentally correct.

---

**Update: (c) added — `transform_props_destructuring`'s raw `find('{')` / `rfind('}')`.** (`111ab248`)

Re-derived from the code rather than from the epic's text, and the realistic trigger is better than the one the epic implied. It is not an exotic string edge case — it is **JSDoc**:

```js
let /** @type {Props} */ { a, b } = $props();
```

which is idiomatic in JavaScript (non-TypeScript) Svelte 5 components. `find('{')` returns the brace in `{Props}`, so the extracted prop list is `Props} */ { a, b` and every declarator parsed from it is garbage. A `}` in a trailing comment moves the closing brace the same way.

Because `transform_props_destructuring` needs a `ComponentAnalysis` (no `Default`) and so is awkward to call from a unit test, I first extracted the two lines into `props_pattern_span` as a **pure no-op** keeping `find`/`rfind`, confirmed the two repros fail against that, and only then made the body lexical. The third test (a plain `{ a, b }` pattern) passed throughout and is a guard, not a repro.

Three commits on this branch now, each a separate defect in the same file family:
| commit | function | defect |
|---|---|---|
| `6104c03e` | `find_matching_open_bracket` | counted brackets inside strings/comments |
| `00aa2e15` | `find_colon_at_depth0` | returned a char index into a byte slice (**panic**) |
| `111ab248` | `props_pattern_span` | located the pattern with a blind `find`/`rfind` |

`cargo test -p rsvelte_core --lib props_transforms` 14/14, `destructure_transforms::tests` 3/3, `state_transforms` 5/5. Bare clippy clean, `rustfmt --edition 2024 --check` clean.

---

**Correction to the `${…}` justification above, and to the reason I gave for it.** (`0d0c405c`)

The sentence I originally wrote — "an opaque template skip would swallow the descent, so behaviour would change" — is **not accurate**, and since it will be read as precedent by later batches it is worth correcting with the measurement rather than leaving.

What `find_colon_at_depth0` actually does: it never returns a position *inside* an interpolation, because `${` raises the depth and the `:` test is gated on `depth == 0`. The `${`/`}` pair only moves the **outer** depth, and cancels when balanced. Probed on four shapes, all returning the correct colon:

| input | returned | correct |
|---|---|---|
| `` `${ a }` : x `` | 9 | 9 |
| `` `${ {a:1} }` : x `` | 13 | 13 |
| `` cond ? `${a}` : y `` | 14 | 14 |
| `` `a${b}c` : x `` | 9 | 9 |

The nested case is the interesting one and it exposes a second accidental cancellation: the template loop does **not** count `{` inside an interpolation, so `{a:1}`'s closing `}` consumes the depth that `${` added, and the interpolation's own `}` is then ignored because the `depth > 0` clamp is already false. The intermediate bookkeeping is wrong in two places and lands on the right answer.

So the honest reason to keep the hand-rolled loop is **"preserve the depth bookkeeping exactly"**, not "the descent would be lost". A `code_bytes` conversion would very likely agree on balanced input — but establishing *when* the two coincide requires reasoning about a counter that is only accidentally correct, and this epic's own record is that such reasoning is where the errors come from. Keeping the loop and changing only the index unit avoids the question entirely.

**Also addressed: the `.ok()` nit.** `byte_offsets.binary_search(&byte).ok()` degraded a would-be-impossible miss into `None`, which discards the destructure candidate silently — precisely the failure mode this sweep exists to remove, reproduced one line away from the fix. It is now `.expect("open bracket is a char boundary")`: the helper only ever returns ASCII bracket positions, which are always char starts, so a miss is a bug in the helper rather than a property of the input and should fail loudly.

### DIFF
--- a/.changeset/colon-depth0-byte-offset.md
+++ b/.changeset/colon-depth0-byte-offset.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): return a byte offset from `find_colon_at_depth0`
+
+The ternary-branch analysis in `check_identifier_in_statement` sliced its right-hand
+side with the position this returned, but the scan counted characters. A ternary
+whose true branch assigns a non-ASCII string literal — `cond ? x = "ああa" : x = y` —
+panicked with "byte index is not a char boundary". The scan also read a `:` written
+inside a comment as the branch separator.
--- a/.changeset/destructure-bracket-lexical.md
+++ b/.changeset/destructure-bracket-lexical.md
@@ -1,0 +1,10 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): match destructuring brackets lexically
+
+`find_matching_open_bracket` walked backwards counting every `{`/`[` it saw,
+including ones inside string literals and comments. A destructuring assignment
+whose pattern carried a brace in a default value (`{ a = "}" } = obj`) or in a
+comment failed to find its opening bracket and was left untransformed.
--- a/.changeset/props-pattern-span-lexical.md
+++ b/.changeset/props-pattern-span-lexical.md
@@ -1,0 +1,12 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): find the `$props()` pattern braces lexically
+
+`transform_props_destructuring` located the destructuring pattern with a raw
+`find('{')` / `rfind('}')`, so a JSDoc type annotation ahead of it —
+`let /** @type {Props} */ { a, b } = $props()`, idiomatic in JavaScript Svelte
+components — made the scan start at the annotation's brace and parse
+`Props} */ { a, b` as the prop list. A `}` in a trailing comment moved the
+closing brace the same way.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
@@ -218,9 +218,11 @@ pub(super) fn find_and_transform_one_destructure(
                 let close_bracket = c;
                 let open_bracket = if c == ']' { '[' } else { '{' };
 
-                // Walk backwards from position `i` to find the matching open bracket
+                // Walk backwards from position `i` to find the matching open bracket.
+                // The helper works in byte offsets; this loop indexes by char.
                 if let Some(pattern_start) =
-                    find_matching_open_bracket(statement, i, open_bracket, close_bracket)
+                    find_matching_open_bracket(statement, b(i), open_bracket, close_bracket)
+                        .and_then(|byte| byte_offsets.binary_search(&byte).ok())
                 {
                     let pattern_str = &statement[b(pattern_start)..b(i + 1)];
                     let rhs_start = j + 1;
@@ -481,24 +483,24 @@ fn is_inside_enclosing_pattern(statement: &str, pattern_open_byte: usize) -> boo
 }
 
 /// Find the matching opening bracket, respecting nesting and strings.
+/// `close_pos` and the returned position are **byte** offsets.
 pub(super) fn find_matching_open_bracket(
     s: &str,
     close_pos: usize,
     open_bracket: char,
     close_bracket: char,
 ) -> Option<usize> {
-    let chars: Vec<char> = s.chars().collect();
+    let (open, close) = (open_bracket as u8, close_bracket as u8);
+    // Collect forward so opaque runs are skipped, then walk the code bytes back.
+    let code: Vec<(usize, u8)> = code_bytes(s.as_bytes())
+        .take_while(|(i, _)| *i < close_pos)
+        .collect();
+
     let mut depth = 1;
-    let mut i = close_pos;
-
-    // Walk backwards
-    while i > 0 {
-        i -= 1;
-        let c = chars[i];
-
-        if c == close_bracket {
+    for &(i, c) in code.iter().rev() {
+        if c == close {
             depth += 1;
-        } else if c == open_bracket {
+        } else if c == open {
             depth -= 1;
             if depth == 0 {
                 return Some(i);
@@ -1605,5 +1607,30 @@ mod non_ascii_tests {
             extract_destructure_targets("{ a /* , b */, c }"),
             vec!["a".to_string(), "c".to_string()]
         );
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn matching_open_bracket_ignores_string_contents() {
+        // `{ a = "}" } = obj` — the default value's `}` is text, not a closer.
+        let s = r#"{ a = "}" } = obj"#;
+        assert_eq!(find_matching_open_bracket(s, 10, '{', '}'), Some(0));
+    }
+
+    #[test]
+    fn matching_open_bracket_ignores_comment_contents() {
+        let s = "{ a, /* } */ b } = obj";
+        let close = s.rfind('}').unwrap();
+        assert_eq!(find_matching_open_bracket(s, close, '{', '}'), Some(0));
+    }
+
+    #[test]
+    fn matching_open_bracket_still_matches_nested() {
+        let s = "{ a: { b } } = obj";
+        assert_eq!(find_matching_open_bracket(s, 11, '{', '}'), Some(0));
     }
 }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
@@ -6,7 +6,7 @@ use std::fmt::Write as _;
 
 use crate::compiler::phases::phase2_analyze::ComponentAnalysis;
 use crate::compiler::phases::phase2_analyze::scope::BindingKind;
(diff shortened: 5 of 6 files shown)


baseballyama:
**LGTM with nits** — 0 blocking. Both index-unit conversions audit clean, every call site checks out, and all five repro tests discriminate. The second commit also closes the one item I left open on #2369 — with a stronger input than I could construct.

Read-only review of both commits, `6104c03e` and `00aa2e15`. No build, no test run.

---

### Index-unit audit — clean, including the direction the epic has burned on three times

**`find_matching_open_bracket`, in.** `close_pos` now arrives as `b(i)`, and `b` is built from `byte_offsets: Vec<usize> = statement.char_indices().map(|(b, _)| b).collect()` (`destructure_transforms.rs:159`) with an out-of-range clamp to `statement.len()` (`:161-166`). One entry per char, ascending. ✓

**`find_matching_open_bracket`, out.** `binary_search(&byte).ok()` (`:225`) can only miss if the returned byte is not some char's first byte. It always is: the returned index comes from `code_bytes` sitting on an ASCII `{`/`[`, which is necessarily a char start and necessarily present in `byte_offsets`. So `Ok` is guaranteed on every input, non-ASCII included. ✓

**Nit on that line.** `.ok()` turns a would-be-impossible miss into `None`, and `None` from this helper means "no matching bracket — discard the candidate and leave the statement untransformed", which is *precisely* the silent-discard failure this PR exists to remove. If the invariant ever breaks, the symptom is the bug you just fixed, reappearing with no diagnostic. `.expect("code_bytes yields char-boundary offsets")` costs nothing and fails loudly instead.

**The backward walk.** `take_while(|(i, _)| *i < close_pos)` is exact rather than merely sufficient — `code_bytes` yields ascending indices, so it stops at the first index ≥ `close_pos` — and it reproduces the old loop's exclusive bound (`while i > 0 { i -= 1; … }` examined `close_pos-1 … 0`). Reversing the collected vector matches the shape used for the reverse scanners in #2366. ✓

**`find_colon_at_depth0`.** The arms that move `i` by hand are the risk, and both are right:

- block comment: after the inner `while`, `i` sits on the `*` of `*/`; the arm's `i += 1` lands on `/` and the loop's trailing `i += 1` steps past it. On an unterminated `/*` the `while` exits at `bytes.len()` and the two increments overshoot to `len + 2` with no indexing after, so it exits cleanly rather than panicking.
- line comment: `continue` skips the trailing increment, leaving `i` on the `\n`, which falls through `_ => {}` on the next pass. ✓

UTF-8 continuation bytes are all ≥ 0x80 so they never match an ASCII arm, and the returned offset is now the same unit the caller slices with (`state_transforms.rs:748`).

### The `find_colon_at_depth0` close-out is stronger than what I flagged

On #2369 I could only build the panic at function level (`rhs = "'日日' : b = 2"`). `check_identifier_in_statement("cond ? component = \"ああa\" : component = other", …)` reaches it through the real caller, which settles reachability rather than arguing it. That is the right way to close a "suspected but unconverted" entry, and `ternary_branch_split_survives_non_ascii` pins it at the caller rather than at the helper — so it survives a future refactor of the helper's signature.

Please update #2369's "Still open in this file" note (and any follow-up issue) — that item is now closed here, and the two PRs are otherwise independent.

### Nit — the `${…}` carve-out justification is overstated

The comment says the scanner "descends into `${…}` on purpose". It does not: the backtick arm's inner `while` never tests for `:`, so no position inside an interpolation is ever returned. What it actually does is let `${` and the matching `}` move the **outer** `depth` variable. For a balanced interpolation those cancel, so `code_bytes` (which would leave `depth` untouched across the whole template) is equivalent for every balanced input and arguably cleaner for unbalanced ones.

The *decision* is still right — a unit-only conversion is the correct risk posture inside a bug-fix, and rewriting the depth bookkeeping in the same commit would have been a second unverifiable change. Just state the reason as "preserve the existing depth bookkeeping exactly" rather than "it descends into `${…}`", so the next sweep does not treat this as a proven carve-out when it is a deferred one.

### Nit — the caller's closer detection is still string-only

`find_and_transform_one_destructure`'s outer loop tracks string literals (`:186-206`) but not comments, so a `}` written inside `/* … */` and followed by ` =` is still taken as a candidate closer. I checked where that lands: `code_bytes` treats the comment as opaque, so `take_while(|(i,_)| *i < close_pos)` yields **nothing** before it, the depth-1 walk never runs, and the helper returns `None` — the candidate is discarded and the statement is left alone. Conservative direction, same as today, no wrong transform. Worth one line in the changeset so the asymmetry (lexical helper, non-lexical caller) is on the record for batch (c).

### Test classification and discrimination

All five new tests re-derived against the pre-commit bodies:

| test | pre-fix result |
|---|---|
| `matching_open_bracket_ignores_string_contents` | `None` — the `}` inside `"}"` pushes depth to 2, the opener only brings it to 1 |
| `matching_open_bracket_ignores_comment_contents` | `None`, same mechanism through `/* } */` |
| `matching_open_bracket_still_matches_nested` | `Some(0)` — **guard, not repro** ✓ classification correct |
| `colon_position_is_a_byte_offset` | `Some(6)` (char index) vs `Some(10)` (byte) |
| `colon_in_comment_is_not_depth0` | `Some(5)` / `Some(5)` instead of `Some(10)` / `None` |
| `ternary_branch_split_survives_non_ascii` | panics |

`matching_open_bracket_still_matches_nested` is the one declared a guard and it is exactly that — I re-derived `Some(0)` on the old char-indexed walk, since nothing in `{ a: { b } } = obj` is inside a literal or comment.

### Noted, not a finding

The honest perf note is the right call: the helper still collects a vector per call and is called once per candidate bracket, and you explicitly decline to claim an improvement. Measuring that belongs with #2342's shape, not here.

PR #2377, opened 2026-08-06
test: cover carriage returns in the whitespace predicates (#1791)

baseballyama:
Closes the CRLF lead from #1791's axis-B sweep. Test-only; **no compiler change** — the guards are
already correct.

## What the sweep found

Twelve `matches!` whitespace sets in `crates/rsvelte_core/src` list `\r`. Dropping `\r` from all
twelve left the entire 1,403-test unit suite green, because every existing input is LF-only: no
test ever produces a value those predicates could reject.

`\r` is load-bearing by spec, not vestigial — `3_transform/utils.rs:142,149,155` document these as
ports of official's `/[^ \t\r\n]/`, `/^[ \t\r\n]+/` and `/[ \t\r\n]+$/`, and nothing upstream
normalises CRLF away.

## The differential experiment

Compiled one component with LF and with CRLF line endings, through official `svelte@5.56.8` (the
version this repo pins) and through rsvelte, client and server:

| | rsvelte vs official, LF | rsvelte vs official, CRLF |
|---|---|---|
| **current code** | MATCH | **MATCH** |
| **12 guards mutated** | MATCH | ***DIFFERS*** |

So: **no bug in shipped behaviour** — rsvelte handles CRLF correctly today and is byte-identical to
official. But the mutation *is* behaviour-changing, which makes this a real coverage gap rather
than dead code. The LF column staying MATCH in both rows confirms the mutation only moves the
CRLF axis.

The divergence is substantial, not cosmetic. Without `\r` a CRLF text node is no longer
whitespace-only, so inter-element whitespace stops collapsing and leaks into the template verbatim:

```
official:  `<div><span></span> <b>a</b> <i>b</i></div> <pre>
mutated:   `<div><span></span>\r\n\t\r\n <b>a</b>\r\n <i>b</i></div>\r\n\r\n <pre>
```

## The tests

Two, covering the three predicates' territory — leading trim, trailing trim, whitespace-only
detection — plus `<pre>`, where whitespace is significant:

- `crlf_matches_its_lf_twin_apart_from_significant_whitespace` — CRLF and LF outputs must be equal
  once the `<pre>` carriage returns are stripped, for both client and server. It also asserts the
  CRLF output *contains* a `\r`, so the test cannot pass by accidentally never exercising them.
- `crlf_whitespace_only_text_nodes_still_collapse` — the direct assertion on collapsing.

Verified three ways, which is the point of the PR:

1. passes on clean code — `2 passed`
2. **fails with all 12 guards mutated** — `0 passed; 2 failed`, both by name
3. passes again after revert — `2 passed`

No fixtures or submodules needed, so it runs in every shard.

## Scope

`test(...)`, no shipped behaviour change, no changeset. One new file. `rustfmt --check` clean.

Remaining axis-B work stays itemised on #1791: 96 of the 108 enumerable guards fall outside this
operator, and the 347 non-`matches!` enum guards are untouched.

### DIFF
--- a/crates/rsvelte_core/tests/crlf_whitespace.rs
+++ b/crates/rsvelte_core/tests/crlf_whitespace.rs
@@ -1,0 +1,59 @@
+//! CRLF sources must compile identically to their LF twin, except inside
+//! elements where whitespace is significant.
+//!
+//! The whitespace predicates are ports of official's `/[^ \t\r\n]/`,
+//! `/^[ \t\r\n]+/` and `/[ \t\r\n]+$/`, so `\r` is part of their contract. No
+//! test reached that: every existing input is LF-only, and dropping `\r` from
+//! all twelve `matches!` whitespace sets in `crates/rsvelte_core/src` left the
+//! entire unit-test suite green. It stops being green here — without `\r` a
+//! CRLF text node is no longer whitespace-only, so inter-element whitespace
+//! leaks into the template verbatim instead of collapsing to one space.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+const SOURCE: &str = "<script>\n\tlet name = 'world';\n</script>\n\n<div>\n\t<span>{name}</span>\n\t\n\t<b>a</b>\n\t<i>b</i>\n</div>\n\n<pre>\n  keep   these\n  spaces\n</pre>\n";
+
+fn compile_to_js(source: &str, generate: GenerateMode) -> String {
+    let options = CompileOptions {
+        filename: Some("App.svelte".to_string()),
+        generate,
+        ..Default::default()
+    };
+    compile(source, options)
+        .unwrap_or_else(|e| panic!("compile failed: {e:?}"))
+        .js
+        .code
+}
+
+/// `<pre>` keeps its carriage returns, so the two outputs differ only by the
+/// `\r`s inside it — stripping those must make them identical.
+#[test]
+fn crlf_matches_its_lf_twin_apart_from_significant_whitespace() {
+    for generate in [GenerateMode::Client, GenerateMode::Server] {
+        let lf = compile_to_js(SOURCE, generate);
+        let crlf = compile_to_js(&SOURCE.replace('\n', "\r\n"), generate);
+
+        assert!(
+            crlf.contains('\r'),
+            "the CRLF input must still reach the output via <pre>, or this \
+             test is not exercising carriage returns at all"
+        );
+        assert_eq!(
+            crlf.replace('\r', ""),
+            lf,
+            "{generate:?}: CRLF and LF sources must compile to the same code \
+             once the <pre> carriage returns are removed"
+        );
+    }
+}
+
+/// The whitespace-only text nodes between `</span>`, `<b>` and `<i>` collapse
+/// to a single space. With `\r` missing from the predicates they do not.
+#[test]
+fn crlf_whitespace_only_text_nodes_still_collapse() {
+    let crlf = compile_to_js(&SOURCE.replace('\n', "\r\n"), GenerateMode::Client);
+    assert!(
+        crlf.contains("<b>a</b> <i>b</i>"),
+        "CRLF whitespace between elements must collapse to one space; got:\n{crlf}"
+    );
+}


baseballyama:
**LGTM** — 0 blocking. The oracle is the pinned version, the mutation isolates the CRLF axis with a working control, and the non-empty guard is in the shipped test rather than only in the throwaway harness. One correction to the claim about "both tests", and one scope note.

Read-only review at `53884203`. No build, no test run.

---

### Verified

**Oracle is the pinned Svelte.** `git submodule status submodules/svelte` → `44a7813730579b94004e182e5a67aab27aa9d2a6 (svelte@5.56.8)`, and `submodules/svelte/packages/svelte/package.json` reads `"version": "5.56.8"`. That matches the PR's stated `svelte@5.56.8` and AGENTS.md's pin. ✓

**The mutation isolates the axis, and the control works.** The four-cell table has the property that makes it readable: the LF column stays MATCH in *both* rows. A mutation that also moved LF would leave the CRLF result unattributable — as it is, the only cell that moves is the one the axis names. That is the right shape for this kind of claim, and it is the part most such tables omit.

**The non-empty guard is in the shipped test.** `crlf_whitespace.rs:42-46`:

```rust
assert!(
    crlf.contains('\r'),
    "the CRLF input must still reach the output via <pre>, or this \
     test is not exercising carriage returns at all"
);
```

It sits inside `crlf_matches_its_lf_twin_apart_from_significant_whitespace`, ahead of the `assert_eq!`, inside the `for generate in [Client, Server]` loop — so it runs on both targets, not once. Given the first harness run compared empty to empty and reported IDENTICAL, having the guard *ordered before* the equality assertion is what makes the equality meaningful; if both compiles returned `""` the guard fires first with a message naming the actual failure. Correctly built.

### Correction — that is not true of both tests

`crlf_whitespace_only_text_nodes_still_collapse` has **no** `\r` self-check. It asserts `crlf.contains("<b>a</b> <i>b</i>")`.

It does not need one, and the reasoning is worth writing down rather than waving at:

- It cannot pass on empty output — `contains` on `""` is false for a non-empty needle, so the empty-vs-empty failure mode is closed by the assertion's own shape rather than by a guard.
- It cannot pass under the mutation — without `\r` in the predicates the CRLF text node is not whitespace-only, the inter-element whitespace leaks in verbatim, and the collapsed `<b>a</b> <i>b</i>` never appears.
- What it does *not* prove on its own is that the input was CRLF: the same string appears for an LF input. That is fine here because the input is constructed inline by `SOURCE.replace('\n', "\r\n")` on the line above, so the axis is exercised by construction rather than by assertion.

So: both tests die under the mutation, neither can pass on empty output, and one of the two self-verifies the axis. Worth saying that precisely in the PR body instead of "it also asserts the CRLF output contains a `\r`", which reads as a property of the suite.

### Scope note — the shipped tests encode an invariant, not parity

The differential against official lives in the PR description; what lands in the repo is a self-consistency assertion — *CRLF output ≡ LF output, modulo the `\r`s that survive inside `<pre>`*. If rsvelte and official both mishandled CRLF in the same way, these tests would still pass.

That is the right trade for the stated purpose. Mutation coverage of the twelve guards is what the #1791 sweep is after, an internal invariant covers it, and the test needs no fixtures or submodules so it runs in every shard — which is worth more than oracle-coupling a test whose job is to stop a predicate from silently losing a character class.

The one thing to add is provenance: the invariant is derived from official's behaviour (`/[^ \t\r\n]/` et al., and nothing upstream normalising CRLF away), and the module doc states it as fact without saying where it came from or which version was measured. One line — "verified against `svelte@5.56.8`; official does not normalise CRLF" — means a future reader who sees this fail knows whether to suspect rsvelte or the assumption.

Related, and already handled well: `crlf.contains('\r')` depends on `<pre>` preserving carriage returns. If a future change normalises line endings at parse time, that assertion fails loudly instead of the test going quietly vacuous. That is the correct direction for a guard to fail in.

### Nit

The `SOURCE` constant is a single escaped line, which makes the fixture hard to read for a test whose whole subject is whitespace layout. A raw string with the `\r\n` substitution applied at use — as it already is — would let a reader see the `<pre>` block and the inter-element gaps that the two tests are actually about.

PR #2380, opened 2026-08-06
fix(compiler): print the client `.svelte.(js|ts)` module through esrap (#2341)

### LINKED ISSUE #2341: compiler (client): the text re-printer drops esrap's statement-level margins inside block bodies
Follow-up carved out of #2284, which fixed the same rule for class **members**. The remainder is statements **inside a block**, which is a much larger surface.

esrap's `body()` handler — used for both `ClassBody` and `BlockStatement` — emits a blank line between two siblings iff either prints multiline, or their node types differ:

```js
if (child_context.multiline || prev_multiline || child.type !== prev_type) context.margin();
```

Note this is **structural and source-independent**: official *adds* a blank line where the source has none, and *drops* a source blank line between two single-line `const` declarations. So the fix is never "preserve the source blank line" — the margin has to be recomputed from node types and multiline-ness.

#2284 implemented that predicate for class members. Two hunks in its repro still diverge, both inside blocks:

- inside a re-printed method body: `const a = 1;` → `return $.get(this.#x);` (VariableDeclaration vs ReturnStatement ⇒ official emits a margin)
- inside the rebuilt constructor body: `const b = 2;` → `$.set(this.#x, s, true);`

Why it was not done there: rsvelte's client class path is a **text** re-printer, so applying the rule needs statement-type classification for arbitrary JS text, not just for the member kinds a class body can contain. That belongs to the burn-down of the text re-printer rather than to a targeted fix.

Note the corpus ratchets cannot see this at all — `verify.mjs` normalizes with oxfmt **and** `stripBlankLines`, so blank-line divergence is invisible to every corpus gate. Any fix here needs its own test coverage; a green corpus is not evidence.

Related: #2336 (the comment-cursor epic) is a *different* mechanism — no `loc`, no comment cursor involved here.

baseballyama:
Implements **#2336 item 4** — route the client `.svelte.(js|ts)` module path through `rsvelte_esrap`. Closes #2341. Scoped out of the epic because it turned out to be separable; see https://github.com/baseballyama/rsvelte/issues/2336#issuecomment-5204219007.

## The epic's stated blocker was false

#2336 says "anything still emitted as `Raw` cannot go through `to_oxc`", and `to_oxc`'s own module doc listed `JsStatement::Raw` among the forms "the structural esrap printer cannot reconstruct". **Both were wrong**, and the doc is how the claim propagated into the epic — so this PR fixes the doc at the source.

`raw_single_statement` and `expand_stmt` *parse* raw text, and `expand_stmt` flattens a multi-statement chunk inline at statement-list sites. `transform_client_module` emits the whole module body as one `Raw`, and it converts: 3 IR statements → 2 oxc statements. Nothing needed AST-ifying first.

## The change

`transform_client_module` called `js_ast::codegen::generate` directly. It now attempts `program_to_oxc` → `rsvelte_esrap` and falls back to the text generator unchanged on `None`, exactly as the component path at `client/mod.rs:2343` does.

## Why the header moved

The `/* X generated by Svelte vN */` header was pushed as a `JsStatement::Raw`. A comment-only statement parses to **zero** statements, so under esrap the comment has no node to anchor to and flushes at the *body tail* — the file ended with its own header.

Upstream does not model it as a statement either. `phases/3-transform/index.js:104`:

```js
js.code = `/* ${basename} generated by Svelte v${VERSION} */\n${js.code}`;
```

Building the body without it and prepending after printing is a faithful mirror of that, not a workaround.

## Result

#2341's two hunks close with **no margin-specific code**, because `rsvelte_esrap`'s `body_elems` (`printer.rs:859`) already implements esrap's predicate against real oxc `Statement` discriminants. The new test asserts the whole file byte-for-byte against the official compiler, plus that the predicate *removes* a source blank line between two single-line `const`s — it is structural, not source-preserving.

## Known interaction, measured

A comment inside an awaited operand prints one redundant paren pair on the comment-bearing `print_split` path: `((await X))()` where official gives `(await X)()`.

This is **pre-existing in `rsvelte_esrap` and independent of this PR** — I reproduced it on the *component* `<script module>` path, which already routes through esrap on `origin/main`, with this PR's branch not applied. It only becomes reachable once #2372 lets that comment survive at all. Isolated: present with a comment, absent without one. Redundant parens are MATCH after oxfmt 0.62 + `stripBlankLines` (measured, with a reformat control), so no corpus gate sees it — but it is a real byte divergence and wants its own fix in the printer.

I have **not** claimed the epic is decomposed. The shapes exercised here are block-body statements and class bodies, **not #2307's Program-level comments** — that is the case most likely to still need items 1–3, and it should be the acceptance test for whether #2307 can follow this.

Green: new suite 3/3, `module_dev_await_instrumentation` 13/13, `class_member_margins_2284` 3/3, `compile_module_thread_safety`, `--lib` 1403. Clippy clean, no suppressions.

---

## CI failure in `ctor_private_state_set_paren_2253.rs` — resolved, and it was the tests that were wrong

Two assertions failed on the first run. Answering the three questions that matter before touching them:

**Which of this PR's parts causes it: the routing.** Reverting the routing while keeping the header change leaves all 13 green; the header change alone causes nothing.

**The "reflow" is the official compiler's own output.** The failing assertion expected

```js
$.set(this.#x, {
	a: s
	// c
});
```

The official compiler emits — byte for byte, verified by diffing whole files, not by eye:

```js
$.set(this.#x, { a: s

	// c
	 });
```

which is exactly what this PR produces. So the two tests were pinning a **divergence from upstream**, recorded from the old text printer's layout. Correcting them is an argument against upstream's output, not a relaxation of assertions: the assertions stay exact-substring, and the property each test exists for — the comment staying inside the literal — is unchanged. In the second test the remaining differences from upstream (`, true`, `$.get(this.#n)` instead of `this.#n.v`) are pre-existing and untouched by this PR.

**The corpus gate cannot see this shape.** Official's layout and the old rsvelte layout are **MATCH** after `flattenTemplateHoles` → oxfmt 0.62.0 → `stripBlankLines` (measured, with a control confirming oxfmt rewrote both sides). The difference is entirely blank lines and intra-line layout, which is what `stripBlankLines` and the formatter exist to absorb.

That answers why ~14,000 corpus entries across three targets could pass while two unit tests failed: **the gate is blind here, so these unit tests were the only thing watching this shape — and they were watching the wrong side.** It is the same lesson as the three normalization rows above, arriving from the opposite direction: gate-invisible divergences need unit tests, and a unit test written from our own output rather than upstream's will happily pin the bug.


baseballyama:
Reviewed at `45bd290a`. **0 blocking.** The doc's new claim is true — I checked it against the code rather than against the PR body — and the header move is a faithful mirror. One correction to the doc fix itself: it is narrower than the claim it replaces, and one of its sentences over-promises in a way the code next to it disproves. Read-only — no build, no test run.

### The doc's new claim is true, and its symbols exist

All three named symbols resolve, which is the first thing worth checking in a doc fix that names code: `raw_single_statement` (`to_oxc.rs:473`), `parse_raw_statements` (`:1188`), `expand_stmt` (`:1324`). And they do what the doc says: `parse_raw_statements` calls `parse_chunk(code.trim())` — a real parse — and `expand_stmt` returns `Vec<Statement>` so a multi-statement chunk flattens inline at list sites. The old claim ("anything still `Raw` cannot go through `to_oxc`") was false, and fixing it at the source rather than working around it downstream is the right call.

### But the correction is narrower than the claim it replaces

The old text named **four** things as unreconstructable: `JsExpr::Raw`, `JsExpr::Spanned`, `JsStatement::Raw`, `JsStatement::RawMapped`. The new text says "The `Raw` family is handled, not excluded" and then names only the two `JsStatement` variants.

The other two are handled as well — `JsExpr::Spanned` at `:1149`, `JsExpr::Raw` at `:1159` via `parse_raw_expression`. So the doc deletes a false claim about four variants and replaces it with a true claim about two, leaving a reader who asks "does `JsExpr::Raw` convert?" with no answer where they previously had a wrong one. That is an improvement, but this doc has just demonstrated that its errors propagate into epics, so it is worth closing completely: add `JsExpr::Raw` / `JsExpr::Spanned` to the same sentence with `parse_raw_expression` as their counterpart.

### One sentence over-promises, and the fix is two lines below it

> Conversion fails only when that text does not parse (`chunk-parse`).

Taken literally this says a successful parse is a successful conversion. The two functions immediately following `parse_raw_statements` say otherwise: `restore_legacy_pre_effect_deps` and `restore_single_target_destructure_sequences` exist precisely because text that *parses fine* round-trips wrong — the first because esrap prints a single-element `SequenceExpression` with parens that re-parse as a `ParenthesizedExpression` the printer then drops. Each is a bespoke repair for a specific loss.

So the real failure mode is not binary. It is: **fails loudly when the text does not parse, and can differ silently when it does.** Given that the previous version of this sentence is what blocked item 4 for an entire epic, the replacement should not invite the opposite over-correction — someone reading "fails only when it does not parse" will route the next `Raw` through `to_oxc` and trust byte-equality without checking. One clause fixes it: "…does not parse; note that parsing is not the same as round-tripping faithfully — see the `restore_*` helpers below, each of which repairs a specific loss."

### Verified: the header move preserves the fallback byte-for-byte

This is the part that could have regressed silently, because no test covers the fallback. It does not, and the argument is short: upstream is `js.code = \`/* … */\n${js.code}\`` (`3-transform/index.js:104`) — exactly one newline. rsvelte's module output matched upstream on the corpus before this PR (`client` at 0 entries throughout), so the text generator must already have emitted exactly one `\n` after the header statement. `format!("{header}\n{code}")` therefore reproduces it. Both branches now build the header the same way, so the two paths cannot drift apart.

The rest of the routing mirrors the component path exactly — same `PrintOptions::default().with_empty_statements(true)`, same `comment_source`-present/absent split (`client/mod.rs:2352-2383`). The one difference is `map_source: None` where the component passes `options.enable_sourcemap.then_some(source)`, which is correct here because `transform_client_module` returns only a `String` — there is nowhere for a map to go. If `.svelte.(js|ts)` maps are ever wired up, this is the call site that needs the same plumbing.

### The caveat is stated correctly, and I want to reinforce one part of it

"I have **not** claimed the epic is decomposed… not #2307's Program-level comments — that should be the acceptance test for whether #2307 can follow this" is exactly right, and it is the sentence that keeps this PR from being read as more than it is. Two shapes were exercised (block bodies, class bodies); Program-level comments are the case items 1-3 exist for. Worth carrying into #2336 verbatim so the next reader of the epic does not infer from "item 4 landed" that the blocker is gone generally.

The known-interaction section is also the right shape, and I want to name why, because I asked for exactly this on #2372 an hour ago and did not get it: the redundant-paren divergence is reported as **pre-existing, isolated by a control (present with a comment, absent without), reproduced on a path this PR does not touch, and kept as "a real byte divergence [that] wants its own fix"** rather than struck because the gate cannot see it. That is the distinction between *not gate-visible* and *not worth fixing*, applied without being asked.

One caution on it: "MATCH after oxfmt 0.62 + `stripBlankLines` (measured, with a reformat control)" is a **paren**-normalization result, and paren handling is the specific behaviour that shifted between the two recent oxfmt versions. Please record the version alongside it in #2336, as you did in the prose here — a bare "MATCH" ages badly for exactly this row.

### Tests

`block_statement_margins_match_upstream` and `same_kind_single_line_statements_get_no_margin` both discriminate against `main`: the first asserts the blank lines esrap's predicate inserts inside the constructor and method, which the text generator never produced; the second asserts the predicate *removes* a source blank line between two single-line `const`s, which a text path that copies source could not do. Testing the removal direction as well as the insertion direction is what makes it a test of the predicate rather than of one output.

`the_generated_by_header_leads_the_file` is the odd one out and should say so. It passes on `main` — the header led the file there too, as a `Raw` statement. What it actually pins is the hazard introduced *within this PR*: that routing through esrap makes a comment-only statement parse to zero statements and flush at the body tail. That is a real thing to pin and the docstring explains the mechanism, but by this epic's labelling standard it is a preservation guard against a specific regression, not a repro against the pre-fix state. One clause ("passes on `main`; it pins the failure mode the esrap routing introduces") keeps the count honest.

### One nit

`// EXPERIMENT: route through to_oxc -> esrap like the component path.` — this is now the primary path with a silent fallback, not an experiment. The word will make the next reader look for a feature flag that does not exist. `AGENTS.md` also asks comments to carry the WHY rather than the status; the WHY here is the sentence already in the PR body ("falls back to the text generator unchanged on `None`, exactly as the component path does").

baseballyama:
Re-reviewed at `21e2fb97`, specifically against the two failures in `Test (ubuntu-latest, 1)`. **1 blocking, pending evidence.** The `EXPERIMENT` nit is applied (`21e2fb97`) — thank you.

Correcting my own earlier review first: I wrote "0 blocking" on `45bd290a` without reading the checks, while `Test (ubuntu-latest, 1)` was already red on these two tests. The static argument I made about the header was right and reached nothing, because the reflow is esrap reprinting a node the old path emitted verbatim. That was my miss.

### The fix changes the two assertions, and the provenance is asserted rather than shown

`55a1ad26` resolves both failures by rewriting the expected strings:

```
- "\t\t$.set(this.#x, {\n\t\t\ta: s\n\t\t\t// c\n\t\t});"
+ "\t\t$.set(this.#x, { a: s\n\n\t\t// c\n\t\t });"

- "?? {\na: s,\n// c\nb: s\n}, true);"
+ "?? {\na: s,\n// c\nb: s\n},\ntrue\n);"
```

with the justification "The official compiler's own bytes". If that is true, the old expectations were pinning rsvelte's text-generator layout rather than upstream's, they were wrong, and updating them is not merely acceptable but required. If it is not true, this PR has shipped a comment-layout regression underneath a test that now certifies it — and a test updated to match new output is the single most effective way to make a regression permanent.

So the claim needs its evidence attached, not because I doubt the intent, but because **this is the one class of change where a reviewer cannot tell the two cases apart from the diff**. One command against `submodules/svelte` compiling the same two sources, with the bytes pasted, settles it. The file header does not help here: it says the tests "pin the synthesized indentation", i.e. these strings were rsvelte's layout, so they carry no upstream provenance of their own.

### A concrete reason to doubt at least one of them

The two new expectations encode **contradictory printer behaviour for the same construct**:

- the second breaks every argument of a multi-line `$.set(...)` onto its own line — `},\ntrue\n);`
- the first does not — `$.set(this.#x, { a: s` keeps the receiver and the literal on the opening line, in a call that also spans lines

Both are `$.set(...)` calls whose argument list spans multiple lines. esrap's argument breaking is not supposed to be a coin flip between them. Either there is a rule that explains both — in which case say what it is, because it is not obvious — or one of the two strings is not upstream's output.

The first is the one I would check hardest. `{ a: s\n\n\t\t// c\n\t\t }` has a **blank line** between the only property and the trailing comment, and a space before the closing brace (`\t\t });`). That is not a shape upstream's printer produces for a one-property object literal, and it is very much the shape a comment-flush artifact produces. It also sits in the printer path your own PR body already flags as having a live defect ("a comment inside an awaited operand prints one redundant paren pair on the comment-bearing `print_split` path"). Two independent comment-placement artifacts on the same path is a more economical explanation than upstream emitting a blank line there.

The second is plausible on its face — breaking arguments one per line when a call spans lines is standard esrap behaviour, and it is consistent with the byte-for-byte class-body expectations in this PR's own new test.

### What would settle it, in order of cost

1. Compile both sources with the official compiler and paste the output. If they match the new strings, this is a correction of two wrong tests and I withdraw the finding entirely — please put the command in the commit message, since the next reader will have the same doubt.
2. If they do not match, the divergence belongs in the same place as the redundant-paren one: recorded against `rsvelte_esrap`'s comment path, with the module routing either held until it is fixed or landed with the divergence named. Not absorbed into the expectations.

Worth noting what does **not** need re-litigating: `assert_structurally_valid` still runs in both tests, so the parse invariant that #2253 exists to protect is untouched by this change. The question is purely whether the new layout is upstream's or rsvelte's — which is exactly the question the project's byte-identical goal makes load-bearing.

### Everything else from my previous review stands

The two doc corrections (the `Raw` family answer covering two of four variants; "fails only when that text does not parse" over-promising, where `restore_legacy_pre_effect_deps` / `restore_single_target_destructure_sequences` are the counterexamples) are unchanged at this head and still worth fixing before merge.

baseballyama:
Re-reviewed at `21e2fb97` (test change `55a1ad26`). **I withdraw the blocking finding — 0 blocking.** The diagnosis inverts cleanly, and I can now see from the test inputs alone why my doubt was wrong.

### Why I was wrong, from the fixture rather than from the measurement

I argued that `{ a: s\n\n\t\t// c\n\t\t }` — blank line, space before the closing brace — is not a shape upstream produces. That reasoning assumed a prettier-style printer. esrap is not one; it is a minimal structural printer, and both artefacts are its ordinary output: a short object literal prints inline as `{ a: s }` **with padding spaces**, and a trailing comment forces the break while the closing `` }`` keeps its padding — which is exactly the leading space in `\t\t });`. The blank line is the margin predicate, the same one this PR's own module test pins byte-for-byte.

The fixture makes it plainer still. The *source* is

```
this.#x = {
	a: s
	// c
};
```

and the old expectation was that layout with a `$.set(` wrapper around it — i.e. **source layout preserved**, which is what a text/verbatim transform emits and what no structural printer would produce. So the old string recorded rsvelte's text path, not upstream. Once the module routes through esrap, reflow is the *expected* outcome, and the assertion had to move. My "internal inconsistency" point (one `$.set` breaking its arguments, the other not) has the same answer: the first call's object starts inline, so its arguments never leave the opening line, while the second's object is multi-line from the start.

So: the old tests were pinning a divergence *from* upstream, and correcting them is a fix, not a relaxation.

### The third measurement is what makes it legitimate, and it should stay in the record

Relaxing a test toward upstream and relaxing it toward our own output are indistinguishable in a diff; the only thing separating them here is the normalization result — that official's layout and old-rsvelte's are **MATCH** under `flattenTemplateHoles` → oxfmt 0.62.0 → `stripBlankLines`. That is also the explanation for how a wrong expectation survived this long: the gate was blind to the entire difference, so nothing but this hand-written test could ever have contradicted it, and it was pinning the wrong side.

Please keep that in the commit message or the PR body with the oxfmt version attached, not just in the thread. It is the same class as the paren row already recorded here, and a bare "MATCH" ages badly for exactly the normalizations that recently shifted.

The isolation control — revert the routing, keep the header change, 13 still green — is the right shape for the question it answers: one variable moved, and it clears the part of the change I had verified statically. Good that the two were separated rather than reasoned about together.

### Confirmed: the property the tests exist for is still pinned

The file's purpose is that the emitted module **parses** and that comments survive inside the construct they were written in. Both hold at this head:

- `// c` still sits between `{` and `}` in the new expectation, so "the comment stays inside the literal" is still what the assertion tests — the message on the assert is still accurate.
- `assert_structurally_valid` is still called in both tests, unchanged. The parse invariant #2253 exists to protect is untouched by this edit.

### Confirmed: the two pre-existing divergences were not absorbed

This is checkable from the diff and it comes out clean. The second test's assertion changed **only in its layout portion** — `"}, true);"` → `"},\ntrue\n);"`. The `, true` proxy flag and `$.get(this.#n) + […].length` are byte-identical to what they were before this PR: the `true` is still asserted, and the second assertion (`$.set(this.#n, $.get(this.#n) + [\n1,\n// c\n2\n].length);`) is not touched at all. So neither divergence was quietly folded into the new expectation; both remain exactly as they were, still asserted, still available to be fixed separately.

For the `, true` specifically: it is current deliberate behaviour rather than an oversight — the `binary-expression-never-proxies` changeset suppresses the flag for a `BinaryExpression` and states that `LogicalExpression` and `ConditionalExpression` keep proxying, which is the `??=` case here. Whether upstream agrees is a separate question and belongs in its own issue, not in this file.

### Still open from my earlier review

Both `to_oxc` doc points are unchanged at this head and still worth fixing before merge: the `Raw`-family sentence answers two of the four variants the old text named (`JsExpr::Spanned` at `:1149` and `JsExpr::Raw` at `:1159` are handled too), and "conversion fails only when that text does not parse" over-promises in the opposite direction from the sentence it replaces — `restore_legacy_pre_effect_deps` and `restore_single_target_destructure_sequences` exist because text that parses can round-trip wrong. *Fails loudly when it does not parse, can differ silently when it does.*

baseballyama:
Bytes pasted, not asserted. Command:

```
node -e "const {compileModule}=await import('submodules/svelte/packages/svelte/src/compiler/index.js'); …"
compileModule(src, { filename: 'A.svelte.js', generate: 'client', dev: false })
```

The outcome is **inverted** relative to the prediction: the *first* expectation
is byte-for-byte upstream; the *second* is the one carrying divergences.

## Expectation 1 — `constructor_literal_with_trailing_line_comment_in_last_position`

Official's output:

```js
	constructor(s) {
		$.set(this.#x, { a: s

		// c
		 });
	}
```

Escaped, so the blank line and the space before `}` are unambiguous:

```
"…\n\tconstructor(s) {\n\t\t$.set(this.#x, { a: s\n\n\t\t// c\n\t\t });\n\t}\n}"
```

The new assertion is `"\t\t$.set(this.#x, { a: s\n\n\t\t// c\n\t\t });"` — an exact
substring. **Upstream really does emit the blank line and the ` });`.** The
"two comment-placement artifacts is the more economical explanation" reasoning
was a good prior and it is simply false here: this is what a trailing comment in
last position looks like out of official esrap. Outcome 1 for this one — the old
string was wrong and the routing fixed it.

## Expectation 2 — `compound_and_logical_assignments_keep_nested_comments`

Official:

```js
	constructor(s) {
		$.set(this.#x, this.#x.v ?? {
			a: s,
			// c
			b: s
		});

		$.set(this.#n, this.#n.v + [
			1,
			// c
			2
		].length);
	}
```

rsvelte on this branch:

```js
	constructor(s) {
		$.set(
			this.#x,
			this.#x.v ?? {
				a: s,
				// c
				b: s
			},
			true
		);

		$.set(this.#n, $.get(this.#n) + [
			1,
			// c
			2
		].length);
	}
```

Two divergences, **neither introduced by this PR**:

1. **A spurious third argument `true`.** Official emits a two-argument `$.set`
   for a logical-assignment RHS. The diff for this expectation is
   `}, true);` → `},\ntrue\n);` — the `true` was already there on `origin/main`;
   only the line breaking changed.
2. **`$.get(this.#n)` where official reads `this.#n.v`.** rsvelte is internally
   inconsistent about this — it uses `.v` in the first statement and `$.get` in
   the second — which is why it reads as a real defect rather than a style
   choice. `.v` is the untracked read, so this may be semantic and not cosmetic.

**This resolves the internal contradiction you spotted, and it was pointing at
something real.** The two `$.set` calls do not break their arguments the same way
because they do not have the same number of arguments: the spurious `true` makes
the first call wide enough that esrap breaks every argument onto its own line.
The observation was right; the cause is the extra argument, not the printer, and
not expectation 1.

## What I changed in response

- The file header no longer says the tests "pin the synthesized indentation" —
  it now states the expectations are upstream's bytes, obtained by compiling with
  `submodules/svelte`, and that residual divergences are named at their site.
- Expectation 2 now carries a comment naming both divergences and official's
  actual bytes, so the assertion cannot be read as upstream's layout. I did not
  absorb them silently and did not relax the assertions; they are pre-existing
  and out of scope for a printing-route change, and I am filing them separately.

`assert_structurally_valid` still runs in both tests; 13/13 in the file pass.

PR #2383, opened 2026-08-06
fix(compiler): fail loudly on an impossible bracket-offset miss (#2357)

baseballyama:
Follow-up to #2376, which merged while this commit was in flight. Same file family, two small corrections that came out of its review.

**`.ok()` on a `Result` whose `Err` is unreachable is a silent-discard generator.** `byte_offsets.binary_search(&byte).ok()` degraded a would-be-impossible miss into `None`, which discards the destructure candidate without a trace — the exact failure mode #2376 exists to remove, reproduced one line away from the fix that removes it. The helper only ever returns ASCII bracket positions, which are always char starts, so a miss is a bug in the helper rather than a property of the input: `.expect("open bracket is a char boundary")`.

The tell for this shape, worth carrying: **you cannot name an input that produces the `Err`.** It reads as defensive and behaves as suppression.

**The `${…}` carve-out comment stated a reason that is false.** It said an opaque template skip "would swallow the descent, so behaviour would change". Probed on four shapes, the scanner returns the correct colon in every one, because `${` raises the depth and the `:` test is gated on `depth == 0` — it never returns a position inside an interpolation at all.

| input | returned | correct |
|---|---|---|
| `` `${ a }` : x `` | 9 | 9 |
| `` `${ {a:1} }` : x `` | 13 | 13 |
| `` cond ? `${a}` : y `` | 14 | 14 |
| `` `a${b}c` : x `` | 9 | 9 |

The correct reason is **"preserve the depth bookkeeping exactly"**, and the nested case shows why that is not pedantry: the template loop does not count `{` inside an interpolation, so in `` `${ {a:1} }` `` the closing `}` of `{a:1}` consumes the depth that `${` added, and the interpolation's own `}` is then ignored because the `depth > 0` clamp is already false. **The intermediate bookkeeping is wrong in two places and lands on the right answer** — a third accidental cancellation in this epic. Converting the index unit only avoids having to reason about when a counter that is merely accidentally correct coincides with a correct one, which is where this epic's errors have come from.

Tests: 6/6 destructure, 5/5 state, 14/14 props. Bare clippy clean with no `-A`, rustfmt clean.

### DIFF
--- a/.changeset/bracket-offset-expect.md
+++ b/.changeset/bracket-offset-expect.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): fail loudly on an impossible bracket-offset miss
--- a/.changeset/bracket-offset-expect.md
+++ b/.changeset/bracket-offset-expect.md
@@ -2,4 +2,8 @@
 "@rsvelte/compiler": patch
 ---
 
-fix(compiler): fail loudly on an impossible bracket-offset miss
+chore(compiler): fail loudly on an impossible bracket-offset miss
+
+No behavioural change: the discarded branch is unreachable for any `&str`
+input, so this only replaces a silent `.ok()` discard with a panic that
+names the offset.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
@@ -220,17 +220,22 @@ pub(super) fn find_and_transform_one_destructure(
 
                 // Walk backwards from position `i` to find the matching open bracket.
                 // The helper works in byte offsets; this loop indexes by char.
-                if let Some(pattern_start) =
-                    find_matching_open_bracket(statement, b(i), open_bracket, close_bracket).map(
-                        |byte| {
-                            // The helper only returns ASCII bracket positions, which
-                            // are always char starts; a miss would be a bug, not input.
-                            byte_offsets
-                                .binary_search(&byte)
-                                .expect("open bracket is a char boundary")
-                        },
-                    )
-                {
+                if let Some(pattern_start) = find_matching_open_bracket(
+                    statement,
+                    b(i),
+                    open_bracket,
+                    close_bracket,
+                )
+                .map(|byte| {
+                    // The helper only returns ASCII bracket positions, which
+                    // are always char starts; a miss would be a bug, not input.
+                    byte_offsets.binary_search(&byte).unwrap_or_else(|_| {
+                        panic!(
+                            "bracket byte offset {byte} is not a char start in a {}-byte statement",
+                            statement.len()
+                        )
+                    })
+                }) {
                     let pattern_str = &statement[b(pattern_start)..b(i + 1)];
                     let rhs_start = j + 1;
 
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/destructure_transforms.rs
@@ -221,8 +221,15 @@ pub(super) fn find_and_transform_one_destructure(
                 // Walk backwards from position `i` to find the matching open bracket.
                 // The helper works in byte offsets; this loop indexes by char.
                 if let Some(pattern_start) =
-                    find_matching_open_bracket(statement, b(i), open_bracket, close_bracket)
-                        .and_then(|byte| byte_offsets.binary_search(&byte).ok())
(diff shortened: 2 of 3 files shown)


baseballyama:
Reviewed at `8b1803ea`. **LGTM — 0 blocking.** The `.expect` is genuinely unreachable, and I checked it by construction rather than by accepting the comment. Read-only — no build, no test run.

### The panic cannot fire

The claim is that the helper only ever returns ASCII bracket positions. That is true, and the chain that makes it true is short enough to state completely:

1. `find_matching_open_bracket` (`destructure_transforms.rs:494`) returns `Some(i)` from exactly one place — the `else if c == open` arm, after `depth -= 1` reaches zero. So the returned `i` always indexes a byte whose value is `open_bracket as u8`.
2. The only production call site is `:224`, and `open_bracket` is `if c == ']' { '[' } else { '{' }` — never anything else, so the byte value is `0x5B` or `0x7B`. (The three other references are unit tests, which also pass `'{'`.)
3. In valid UTF-8, `0x5B` / `0x7B` cannot occur inside a multi-byte sequence: lead bytes are `0xC2..=0xF4` and continuations `0x80..=0xBF`. `s` comes from `&str`, so it is valid UTF-8 by construction. Therefore `i` is always a char start.
4. `byte_offsets` is `statement.char_indices().map(|(b, _)| b).collect()` — the *complete* ascending list of char starts of the same string the helper scanned. `binary_search` over it therefore hits for any char start.

The indices are also in the right coordinate space: `code_bytes(s.as_bytes())` yields absolute offsets into `s` (the `take_while(|(i, _)| *i < close_pos)` comparison against a byte offset into `s` depends on that too), so there is no slice-relative offset to go wrong.

And the `.map` placement preserves the short-circuit: `find_matching_open_bracket` returning `None` still yields `None` and skips the block without evaluating the `expect`. Only a `Some` is searched.

So this converts a discard that could never trigger into a panic that can never fire — which is the right direction, since the previous form left a reader unable to tell whether the `.ok()` was load-bearing. That question is now answered in the code.

### The comment correction is a real fix, not a rewording

Worth recording, because the old comment was confidently wrong. It said the scanner "descends into `${…}` on purpose". It does not: the backtick arm at `state_transforms.rs:953` runs its own inner loop to the closing backtick and never tests for `:`, so a colon inside an interpolation is never returned regardless of depth. The new comment states what the arm actually does — `${` does `depth += 1`, `}` does `depth -= 1` — and that is verifiable in the four lines below it.

"the bookkeeping is preserved exactly rather than re-derived" is also the honest framing for why the sweep skipped this site, because the bookkeeping is *approximate*: a literal `}` in template text also decrements the outer depth when `depth > 0`, so `f(\`a } b\`) : x` loses its colon. That is pre-existing and out of scope here — but it is the reason "preserved exactly" is the right claim to make rather than "correct".

### Two nits, neither worth blocking on

**The changeset asserts a fix where the PR asserts nothing changes.** `fix(compiler): fail loudly on an impossible bracket-offset miss` will land in the public changelog, and a user reading it will look for the behaviour that changed. By this PR's own argument there is none. Compare its sibling `colon-depth0-byte-offset.md`, which states the observable symptom (`cond ? x = "ああa" : x = y` panicked with "byte index is not a char boundary") — that is a changelog entry; this one is a commit subject. Either add a body saying "no behavioural change: an internal invariant now panics instead of silently skipping", or drop the changeset.

**The panic message states the invariant, not the violation.** `"open bracket is a char boundary"` reads as true even in the moment it fires. If it ever does, the person reading the backtrace wants the offset and the statement. `expect` cannot format, but `unwrap_or_else(|_| panic!("bracket offset {byte} is not a char boundary in {statement:?}"))` costs nothing on the success path and turns an impossible panic into a diagnosable one — which is the whole point of preferring it to a silent skip.

baseballyama:
Collision notice: #2398 edits the same expression at `destructure_transforms.rs:221-228`, and its pre-image is `main`'s, so it does not contain `72dcfa29`. **Whichever lands second must produce both behaviours** — resolving toward #2398 alone would restore the silent `.ok()` discard, and by this PR's own unreachability argument the panic never fires, so nothing would report the regression.

`f708e71a` on #2398 now carries the combined form, keeping your `#[cold] bracket_offset_miss` verbatim and routing the conversion through a new `CharToByte::char_of`:

```rust
.map(|byte| {
    table
        .char_of(byte)
        .unwrap_or_else(|| bracket_offset_miss(byte.get(), statement.len()))
})
```

That also deletes the `byte_offsets` vector, whose only remaining use was the `binary_search` this replaces.

**If #2383 lands first**, I rebase #2398 onto it and keep this expression. **If #2398 lands first**, this hunk is already satisfied and should be dropped rather than re-applied — the `#[cold]` helper travels with it, so no authorship is lost either way.

PR #2384, opened 2026-08-06
fix(analyze): give attribute-scoped a11y warnings the attribute's span

baseballyama:
Part of #2281 (warning parity). Fixes the larger of the **two** systemic causes behind the `warning-position-known-failures.*` ratchet.

## The cause

`2_analyze/visitors/regular_element.rs:414-421` stamps `element.start`/`element.end` on any a11y warning that arrives without a span, and `shared/a11y/mod.rs` pushed **every** warning without one — so the element fallback always won, including for the warnings upstream attaches to an *attribute*. Of the **17** warn sites inside `a11y/index.js`'s first attribute loop (`:108`-`:287`), **14** pass `attribute` and exactly three pass `node`:

- `a11y_interactive_supports_focus`
- `a11y_no_interactive_element_to_noninteractive_role`
- `a11y_no_noninteractive_element_to_interactive_role`

This PR raises each attribute-scoped warning with its attribute's span, leaving the element fallback covering exactly those three.

## The measurement

Over the 625 entries listed in `warning-position-known-failures.client.json`, which carry **967** mismatching `(code, line, column)` tuples:

| cause | tuples | share |
|---|---|---|
| rsvelte emits no span at all | 649 | 67.1% |
| span present but element-wide instead of attribute-wide | **318** | **32.9%** |

The second bucket is **100% a11y codes** and **100% column disagreements** — 315 column-only, 3 line-and-column, **0 line-only**. Zero line-only is the signature: the attribute and its element are on the same line, so only the column can move. That is what separates this cause from the missing-span one and is why "attach the span already available at the emission site" was the wrong description for a third of the backlog.

After the fix the 318 becomes **1**: a single `a11y_figcaption_index`, and it is out of reach **structurally**, not merely unobserved. Upstream raises it at `:532`, outside both attribute loops, on `children[index]` — so none of the four stamp sites can reach it, and `stamp_attribute` skips anything that already carries a span. That argument holds regardless of what any run showed, which is a better reason than "it appears in the pre-fix run too".

Codes cleared, by tuple count — **summing to all 317** (`120 + 59 + 41 + 24 + 24 + 20 + 10 + 4 + 3 + 3 + 2 + 7x1`):

`a11y_role_supports_aria_props` 120, `a11y_role_supports_aria_props_implicit` 59, `a11y_no_redundant_roles` 41, `a11y_no_abstract_role` 24, the `a11y_incorrect_aria_attribute_type*` family **24** (boolean 6, tokenlist 5, integer 4, token 4, bare 3, tristate 2), `a11y_invalid_attribute` 20, `a11y_autofocus` 10, `a11y_role_has_required_aria_props` 4, `a11y_autocomplete_valid` 3, `a11y_misplaced_scope` 3, `a11y_unknown_role` 2, and one tuple each of `a11y_aria_activedescendant_has_tabindex`, `a11y_unknown_aria_attribute`, `a11y_aria_attributes`, `a11y_misplaced_role`, `a11y_hidden`, `a11y_accesskey`, `a11y_positive_tabindex` (seven codes).

An earlier revision of this list stopped at the counts of 3 and reported the type family as 22, so it summed to **306** against a measured 317 while reading as exhaustive. That is the same defect as a cause inferred from a code name, and it is now gated: `known-failures-md-check.mjs` — which already guarded this file's header counts — also checks that a stated total equals the addends printed beside it.

**Measuring this needed no corpus staging.** Both compilers load in memory — official from `submodules/svelte/packages/svelte/src/compiler/index.js`, rsvelte from `.corpus-cache/rsvelte.node` — and the ratchet ids resolve straight to files already present under `submodules/`. No `expected/` or `actual/` tree is written, so the ~0.57 GiB a full corpus run costs is avoided entirely. Anyone reproducing the table above can do so on a checkout with no corpus trees at all.

## Tests

`crates/rsvelte_core/tests/a11y_attribute_warning_span.rs` — 5 repros (all FAILED before the fix, all pass after) plus 2 guards:

| test | kind |
|---|---|
| `incorrect_aria_attribute_type_points_at_the_attribute` | repro |
| `no_abstract_role_points_at_the_role_attribute` | repro |
| `role_supports_aria_props_points_at_the_aria_attribute` | repro |
| `invalid_attribute_points_at_the_href` | repro |
| `autofocus_points_at_the_attribute` | repro |
| `codes_upstream_scopes_to_the_element_inside_the_attribute_loop` | **carve-out control** — the only test that can see `ELEMENT_SCOPED_CODES` |
| `element_scoped_warning_still_points_at_the_element` | pin — passes pre-fix; guards the fallback, not the carve-out |

**Correcting an earlier claim in this description:** I said the pin was the negative control and would fail if `ELEMENT_SCOPED_CODES` were emptied. It does not. `a11y_no_static_element_interactions` is raised at `:372`, *outside* the attribute loop, so it never passes through `stamp_attribute` — I emptied the list and all six tests stayed green. The exclusion list is the one judgment call in this change and nothing pinned it; only the corpus ratchet would have objected, later and less legibly.

`codes_upstream_scopes_to_the_element_inside_the_attribute_loop` fixes that, with one case per carved-out code, each raised *inside* the loop:

| input | code | upstream site |
|---|---|---|
| `<h1 class="x" role="button">y</h1>` | `a11y_no_noninteractive_element_to_interactive_role` | `:252` |
| `<button class="x" role="presentation">y</button>` | `a11y_no_interactive_element_to_noninteractive_role` | `:240` |
| `<div class="x" role="button" on:click={f}>y</div>` | `a11y_interactive_supports_focus` | `:230` |

Checked in both directions rather than asserted: with the carve-out it passes; with `ELEMENT_SCOPED_CODES` neutered it fails on the first case with `got column 14 -> "role=\"button\">y</h1>"`. A test that passes with *and* without the mechanism it is named for is a guard, not a control — and here it was standing in for the only part of the change a reviewer cannot verify by reading.

`cargo test -p rsvelte_core --test validator` stays at 333/333.

## Docs

`compatibility/warning-known-failures.md` carries three corrections, all of them things the previous single-cause reading of this bucket got wrong:

1. "Almost all of these are one systemic cause: missing span" — there are two, 67.1% / 32.9%.
2. `a11y_role_supports_aria_props` was listed as a *missing-position* code. It never was; rsvelte always emitted a span for it. It is the single largest member of the wrong-span bucket.
3. "Burning this down is mostly mechanical — attach the span already available at each emission site" — true of the 67.1%, false of the 32.9%, where the span was attached, by the wrong owner.

## Ratchet

The position ratchets shrink. They are re-baselined in this PR (see the follow-up commit) — a listed entry that now passes fails CI as loudly as a regression.


baseballyama:
Reviewed at `edbe5295`. **0 blocking.** The exclusion list is complete — I verified it by reading every warn call in upstream's attribute loop rather than sampling — and the surviving `a11y_figcaption_index` is structurally pre-existing. Three findings, none blocking: the negative control does not exercise the carve-out it exists to protect, one tuple count does not reconcile, and one figure needs its population stated. Read-only — no build, no test run. CI: 4 pass, 27 pending, no failures.

### The exclusion list is complete

Upstream's attribute loop is `for (const attribute of node.attributes)` at `a11y/index.js:108`, closing at `:287` (the next statement is `const role = attribute_map.get('role')` at `:289`). Every `w.a11y_*` call inside that range:

`:116, :122, :126, :145, :154, :165, :168, :179, :188, :204, :259, :265, :272, :281` pass `attribute` — including `:204`, which is multi-line and whose first argument I checked separately because a wrapped call is where a sampled read goes wrong.

`:230, :240, :252` pass `node`: `a11y_interactive_supports_focus`, `a11y_no_interactive_element_to_noninteractive_role`, `a11y_no_noninteractive_element_to_interactive_role`.

That is exactly the three in `ELEMENT_SCOPED_CODES`, and there is no fourth. The second stamped loop is consistent too — `:331` and `:333` both pass `attr` — as are the two individually stamped sites, `:408` (`href`) and `:436` (autocomplete).

The `warning.start.is_some()` guard in `stamp_attribute` is the right second half: a warning that already carries its own span is left alone rather than overwritten, so the stamp only fills the gap the element fallback used to fill.

### `a11y_figcaption_index` is pre-existing, and the reason is structural

Upstream raises it at `:532` as `w.a11y_figcaption_index(children[index])` — the target is a **child node**, not the element and not an attribute, and it is raised outside both attribute loops. None of this PR's four stamp sites can reach it, and `stamp_attribute` skips anything that already has a span. So it cannot have been introduced here, independently of what the ratchet measured. The md's claim that it is "a third cause, not a remnant of this one" is right, and the fact that it disagrees on **both** line and column — where the whole a11y bucket was column-only bar three — is the corroborating signal.

### The negative control does not exercise the exclusion list

This is the one I would fix before merge. `element_scoped_warning_still_points_at_the_element` uses `a11y_no_static_element_interactions`, which upstream raises at `:372` — **outside** the attribute loop. A warning raised outside the loop is never passed to `stamp_attribute` at all, so that test passes whether `ELEMENT_SCOPED_CODES` contains three entries or none.

The carve-out is the only judgment call in this change, and nothing currently pins it: delete the constant and its `contains` check, and all six tests stay green — only the corpus ratchet would object, on a later run, as three codes' worth of new position failures.

One source fixes it, in the same shape as the others (leading `class` so the element and attribute columns cannot coincide):

```rust
let src = "<h1 class=\"x\" role=\"button\">y</h1>\n";
// a11y_no_noninteractive_element_to_interactive_role must point at `<h1`, not `role=`
```

`h1` is non-interactive and `button` is an interactive role, so this reaches `:252` — inside the loop, element-scoped — which is precisely the path the constant governs.

The five positive tests are well built, and the leading `class="…"` in each is doing real work: without it the element and the attribute would be a few columns apart and an off-by-a-little would still pass. Asserting on the *text* at the reported column rather than on a number is also the right form.

### The tuple arithmetic does not close

The md states the a11y bucket as **318 tuples**, and that what remains is **650** = 649 missing-span + 1 `a11y_figcaption_index`. From 967 total that makes **317 tuples cleared**. The itemised list sums to **306**:

120 + 59 + 41 + 24 + 20 + 10 + 22 + 4 + 3 + 3 = 306

so 11 cleared tuples are unaccounted for. Either the list is not exhaustive — in which case say so, since it currently reads as a full accounting — or one of the per-code figures is off. This file's counts are prose that CI does not check (`known-failures-md-check.mjs` gates `known-failures.md`, not this one), which is exactly the condition under which the numbers drifted last time; a total that does not reconcile with its own parts is the cheapest place to catch it.

The two counts that *are* checkable against the JSON both reconcile: all three position files hold **529** entries at this head and all three code files hold **70**, matching "the same 529" and "the same 70 entries", and 625 − 96 = 529.

### One figure needs its population

"`a11y/index.js` passes `attribute`, not `node`, at 16 of its warn sites" is true of the two attribute loops (14 + 2) but not of the file: `:408`, `:436` and the eight `a11y_incorrect_aria_attribute_type*` calls at `:876-:933` also pass an attribute, which would make it 26. The cleared-codes list right below it includes that `incorrect_aria_attribute_type*` family, so a reader reconciling the two will land on the larger number. Saying "at 16 of the warn sites in its two attribute loops" makes it checkable.

### The retraction is the most valuable line in the file

> Note that `a11y_role_supports_aria_props` was previously listed above as a missing-position code. It never was: rsvelte always emitted a span for it.

Correcting a prior mis-attribution in the record, and naming the mechanism that produced it — a single-cause reading of a bucket, inferred from code names rather than measured per tuple — is worth more than the burn-down itself. The two-cause split with per-tuple figures (649 / 318, and 315 column-only / 3 line-and-column / **0 line-only**) is what makes the next slice of this backlog estimable instead of guessed, and the zero is the load-bearing number: it is what rules out a line-attribution bug hiding inside the column one.

baseballyama:
Re-reviewed at `c4495ae9` (delta from `edbe5295`). **0 blocking.** All three items are applied, and the third one came back stronger than what I asked for. Read-only — no build, no test run. CI: 20 pass / 2 skipped / 12 pending, no failures.

### 1. The carve-out control is real — it can move

`codes_upstream_scopes_to_the_element_inside_the_attribute_loop` covers all three `ELEMENT_SCOPED_CODES` with `assert_eq!(col, 0)`, and each fixture places the triggering `role` attribute at a non-zero column, so emptying the constant flips every one of the three from 0 to the attribute's column. That is the property the previous six tests did not have: I could delete `ELEMENT_SCOPED_CODES` and they all stayed green.

The relabel on the old test is the other half and I want to be explicit about why it matters more than a doc tweak. `element_scoped_warning_still_points_at_the_element` uses `a11y_no_static_element_interactions`, which is raised **outside** the attribute loop and never reaches `stamp_attribute` — so it could never have detected the carve-out no matter what the constant contained. Naming it "Guards the fallback, not the carve-out" stops the next reader counting it as coverage it does not provide. Two tests that look like one test is how the hole survived.

### 2. The arithmetic, and the gate that now holds it

I ran the new gate's own regex over the file rather than adding the numbers by hand. Exactly **one** `summing to` phrase, addends `120 + 59 + 41 + 24 + 24 + 20 + 10 + 4 + 3 + 3 + 2 + 7x1` = **317** = stated. The type family is itemised (boolean 6, tokenlist 5, integer 4, token 4, bare 3, tristate 2 = 24), the seven single-tuple codes are named, and the regex tolerates the newline inside the backticks that this file happens to have.

The `a11y_role_supports_aria_props` self-correction — "it never was a missing-position code; rsvelte always emitted a span for it" — is the part I would keep. A wrong entry retracted in the same document that carried it is worth more than the 317 being right.

And `known-failures-md-check.mjs` is genuinely wired, not just written: `corpus-compat.yml` runs it at two steps plus the `corpus:known-failures-check` script. Turning an unchecked prose number into a checked one is the correct generalisation of the finding, and it is the durable half of this PR.

### 3. The 17 warn sites

"Of the 17 warn sites inside `a11y/index.js`'s first attribute loop (`:108`-`:287`), 14 pass `attribute` and exactly three pass `node`" matches my own independent count of the upstream file exactly, including which three (`:230` / `:240` / `:252`). This is the sentence that justifies the carve-out being exactly three codes long, so it being derived from the loop rather than from the code names is what makes the fix bounded.

### Two residuals on the new gate, neither blocking

**(a) It is opt-in by phrase, so its own "measured nothing" state is green.** The loop runs over `matchAll`; a file with zero matches passes. If a future entry writes "totalling 317" or drops the parenthetical, that list is silently unchecked and nothing says so — the exact shape #1791 catalogues, in the gate written to prevent it. One line closes it: fail when the match count is zero, with a comment that removing the convention must be deliberate. That is this repo's own `assert!(checked > 0, "oracle ran no fixtures")` idiom from #1791 item 3, applied to the check rather than the corpus.

**(b) It checks self-consistency, not the measurement.** 317 is not derivable from anything in the tree — the three ratchets carry file ids, and their movement is 625 → **529** entries per target, i.e. 96 removed. The gate can prove the addends agree with the stated total; only a verify run against the oracle can prove either number is the right one. That is the correct scope for a gate that must run without a corpus, and it is worth stating in the file so nobody reads a green check as confirmation of 317.

That second point has a one-word consequence in the prose: the section states 96 (entries) at one paragraph and 317 (tuples) three paragraphs later, and the 317 sentence says only "Codes cleared, summing to all 317". Both units are present in the section and the reader has to infer which is which — writing "317 **tuples**" removes it. Unit confusion between entries and tuples is the same class of drift this section exists to document.

### One thing to carry forward

The heuristic added here — "count line-only mismatches; a bucket with *zero* of them is a wrong-**owner** bucket, not a missing-span one" — is the reusable output of this PR and it generalises past a11y. It is the same move as grouping by *warn target* instead of by symptom in #2402: the symptom bucket (`start === undefined`) cannot separate mechanisms, the structural bucket can. Two independent instances in a week is enough to treat it as a rule rather than a trick.

### Merge ordering, restated because it has not changed

#2384, #2391 and #2402 all rewrite `warning-known-failures.md` from the same base, and #2384 and #2402 replace the **same paragraph** with independently written prose. This one goes first; the other two rebase and **regenerate** their ratchets rather than hand-merging, because the post-merge totals are not derivable arithmetically — the removal sets need not be disjoint at entry granularity, so any hand-written sum would be a number no run has ever produced.

Issue #2378, opened 2026-08-06
compiler: char indices escaping into byte slices — `find_colon_at_depth0` and the wider class


baseballyama:
## Summary

`client/state_transforms.rs::find_colon_at_depth0` returns a **char** index; its single
caller slices the same string by **bytes**. One non-ASCII character before the `:` is enough
for a silent mis-slice, and two are enough to panic.

This is the same defect class as the panic fixed in #2369
(`transform_legacy_destructure_declarations`, `let { ああ } = obj;`) and as
`extract_store_set_targets` (a `memmem` byte offset indexing a `Vec<char>`). Three instances
in one day, so it is a bug class in its own right, riding along with #2357 rather than being
part of it — note that **the `depth ±= 1` proxy does not find any of them**.

## The code

`find_colon_at_depth0` builds `let chars: Vec<char> = expr.chars().collect()` and returns `i`
into that vector. The caller (`check_identifier_in_statement`) does:

```rust
if let Some(colon_pos) = find_colon_at_depth0(rhs) {
    let true_rhs = rhs[..colon_pos].trim();          // byte slice
    let false_branch = rhs[colon_pos + 1..].trim();  // byte slice
```

## Confirmed reproduction (function level)

`rhs = "'日日' : b = 2"` — 16 bytes, 12 chars. The scan returns char index **5**; byte 5 is
inside the second `日` (bytes 4..7):

```
end byte index 5 is not a char boundary; it is inside '日' (bytes 4..7) of `'日日' : b = 2`
```

With a single non-ASCII character the slice usually stays on a boundary and silently cuts the
ternary in the wrong place instead — wrong dependency tracking rather than a crash.

## Not yet established

**End-to-end reachability.** Two of us independently failed to construct a `.svelte` source
that reaches this call with such an `rhs`; it sits behind `lhs.contains('?')` inside
`check_identifier_in_statement`, so it needs a legacy `$:` reactive statement whose assignment
LHS contains a `?` and whose RHS carries a non-ASCII string before the ternary colon. That is
why #2369 left it unconverted rather than converting it blind — but the function-level
reproduction above is real, so whoever picks this up starts from a failing input, not from zero.

## Also in scope, same function

`find_colon_at_depth0` deliberately descends into `${…}` interpolations (it increments `depth`
on `${` and decrements on the matching `}`), which is why it cannot simply take
`shared/js_scan.rs::code_bytes` — that treats a template literal as fully opaque. It is
[...]

PR #2385, opened 2026-08-06
fix(compiler): return byte offsets from two more position scanners (#2378)

baseballyama:
Sweep for #2378, run as the mechanical check that issue asks for: **does an index escape the function into a caller that uses a different unit?** — not *does this function build a `Vec<char>`*.

## Method

Enumerated every function in `crates/*/src/*.rs` that returns a `usize`-shaped position **and** touches `chars`/`char_indices` — 50 candidates. Classified each producer by index unit (`Vec<char>` / `.chars().enumerate()` → char; `char_indices()` / `as_bytes()` → byte), discarded the ones returning distances or line numbers rather than offsets (`levenshtein`, `levenshtein_distance`, `full_line_comment_lines`), then read **each caller** to see which unit it consumes.

## Found: two more mismatches

| function | returns | caller does | result |
|---|---|---|---|
| `client/props_transforms.rs::find_destructuring_pattern_end` | char index into `Vec<char>` | `decl[..pattern_end]`, `rest[..pattern_end]` | byte slice |
| `server/transform_script.rs::find_simple_assignment` | char index from `chars.iter().enumerate()` | `rest_trimmed[..eq_pos]`, `rest_trimmed[eq_pos + 1..]` | byte slice |

Both now return byte offsets. Repros confirmed failing first — for `find_simple_assignment` by grafting the new tests onto the **pre-fix** function and running them (`Some(5)` vs `None`, and the byte-offset assertion), rather than reasoning that they would have failed.

`let { café } = obj` loses its closing brace (`"{ café "`), so the `= rhs` split fails and the declaration is silently left untransformed. `let { ああ } = obj` straddles the offset and panics. Same two-mode signature as the rest of this class.

`find_destructuring_pattern_end` also returned an offset into a **trimmed** copy while callers sliced the original. Both current callers pass already-trimmed input so the offset is 0 today; the function now adds the trim base so the contract is honest rather than accidentally satisfied.

Both conversions route through `code_bytes`, which also drops the hand-rolled string tracking and gains comment handling — `find_simple_assignment("x /* = */ 1")` previously reported an assignment at the commented `=`.

## Audited and clean, so nobody re-checks them

- `find_destructure_rhs_end` — returns a char index. **Corrected reason:** it is not that all three
  callers wrap it in the `b()` table — `:355` never converts it at all. It compares `rhs_ends`
  against `other.close_pos`, itself a char index (documented as such at `:179`/`:181`), so that
  site is consistent by *staying in char space* rather than by converting. The function is
  char-space end to end: it builds its own `Vec<char>` and indexes `start` into it.
  Its `start` argument is a char index at all three sites, so it is clean on both sides.
- `scan_assignment_rhs_end` — byte-based throughout (`as_bytes()`), caller slices by byte.
- `skip_string_literal` — takes and returns byte offsets over a `bytes` slice.
- `dollar_param_body_range` — takes `chars` and its caller indexes `chars`. Consistent.

## Scope note

This is not the whole class. Per the visibility asymmetry recorded on #2378: a byte position landing *on* a char boundary mis-slices **silently**, and only positions landing mid-character panic — so the reported count measures **reporting**, not prevalence. The sweep above covers producers that return a position *and* mention chars; it does **not** cover positions passed as arguments between functions, nor arithmetic on an offset after it is returned. #2378 should stay open.

`cargo test -p rsvelte_core --lib props_transforms` 15/15, `transform_script` unit tests pass, bare clippy clean with no `-A`, `rustfmt --edition 2024 --check` clean. `--test ssr` / `--test runtime` cannot run here (`Fixtures not found`).

## Review round

**`base != 0` fixture added** (`30ca832c`). All four original patterns start at the bracket, so
the trim `base` — the only arithmetic this conversion introduced — was zero in every case and
went unexercised. `"  { café }"` joins the same array under the identical assertion. Confirmed
discriminating by dropping the `base` term and re-running: the mutant fails on that pattern and
on no other.

**The corrected clean-reason above changes the sweep's question set.** "Does the caller wrap it?"
is the wrong question for a caller that never slices — and asking only that would wave through a
char-vs-byte **comparison**, which mis-selects silently and never panics. The visibility asymmetry
that made "five instances" a statement about reporting applies here at full strength: panics get
reported, comparisons do not. Recorded on #2378.

### DIFF
--- a/.changeset/index-unit-sweep.md
+++ b/.changeset/index-unit-sweep.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): return byte offsets from two more position scanners
+
+`find_destructuring_pattern_end` and `find_simple_assignment` counted characters
+while their callers sliced the same string by bytes, so a non-ASCII identifier or
+string literal in a destructuring pattern or a `let` initialiser sliced short —
+`let { café } = obj` lost its closing brace — and a multi-byte character straddling
+the offset panicked outright.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
@@ -4384,8 +4384,9 @@ mod pattern_end_unit_tests {
 
     #[test]
     fn pattern_end_is_a_byte_offset() {
-        // Callers slice `&str` with this, so it must be a byte offset.
-        for pattern in ["{ a }", "{ café }", "{ ああ }", "[ あ, い ]"] {
+        // Callers slice `&str` with this, so it must be a byte offset. The
+        // leading-space case is the only one where the trim `base` is non-zero.
+        for pattern in ["{ a }", "{ café }", "{ ああ }", "[ あ, い ]", "  { café }"] {
             let end = find_destructuring_pattern_end(pattern).unwrap();
             assert_eq!(&pattern[..end], pattern, "pattern {pattern:?}");
         }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
@@ -1308,49 +1308,26 @@ pub(super) fn transform_destructured_export_let(
 
 /// Find the end position of a destructuring pattern in `{ ... } = RHS` or `[ ... ] = RHS`.
 /// Returns the position after the closing `}` or `]`.
+/// Byte offset just past the pattern's closing bracket, relative to `s` as passed.
 pub(super) fn find_destructuring_pattern_end(s: &str) -> Option<usize> {
-    let s = s.trim();
-    let first = s.chars().next()?;
-    if first != '{' && first != '[' {
+    let trimmed = s.trim_start();
+    let base = s.len() - trimmed.len();
+    if !matches!(trimmed.as_bytes().first(), Some(b'{' | b'[')) {
         return None;
     }
 
-    let chars: Vec<char> = s.chars().collect();
     let mut depth = 0;
-    let mut i = 0;
-    let mut in_string = false;
-    let mut string_char = ' ';
-
-    while i < chars.len() {
-        if in_string {
-            if chars[i] == '\\' {
-                i += 2;
-                continue;
+    for (i, c) in code_bytes(trimmed.as_bytes()) {
+        match c {
+            b'{' | b'[' => depth += 1,
+            b'}' | b']' => {
+                depth -= 1;
+                if depth == 0 {
+                    return Some(base + i + 1);
+                }
             }
-            if chars[i] == string_char {
-                in_string = false;
-            }
-            i += 1;
-            continue;
+            _ => {}
         }
-
-        if chars[i] == '\'' || chars[i] == '"' || chars[i] == '`' {
-            in_string = true;
-            string_char = chars[i];
-            i += 1;
-            continue;
-        }
-
-        if chars[i] == '{' || chars[i] == '[' {
-            depth += 1;
-        } else if chars[i] == '}' || chars[i] == ']' {
-            depth -= 1;
-            if depth == 0 {
-                return Some(i + 1);
-            }
-        }
-
-        i += 1;
     }
     None
 }
@@ -4398,5 +4375,19 @@ mod props_pattern_span_tests {
         let line = "let { a, b } = $props();";
         let (open, close) = props_pattern_span(line).unwrap();
         assert_eq!(&line[open..=close], "{ a, b }");
+    }
+}
+
+#[cfg(test)]
+mod pattern_end_unit_tests {
+    use super::find_destructuring_pattern_end;
+
+    #[test]
+    fn pattern_end_is_a_byte_offset() {
+        // Callers slice `&str` with this, so it must be a byte offset.
+        for pattern in ["{ a }", "{ café }", "{ ああ }", "[ あ, い ]"] {
+            let end = find_destructuring_pattern_end(pattern).unwrap();
+            assert_eq!(&pattern[..end], pattern, "pattern {pattern:?}");
+        }
     }
 }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_script.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_script.rs
@@ -6153,39 +6153,24 @@ fn transform_reexported_prop_declarations(
 }
 
 /// Find assignment `=` in a simple declarator (not inside parentheses, brackets, etc.)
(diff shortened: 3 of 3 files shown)


baseballyama:
Reviewed at `9163e872`. **No blocking findings in the code.** Two corrections, one to the audit table and one gap in the new test. Read-only — no build, no test run.

**CI position first, since it is not settled:** 13 checks pass, including `Test (ubuntu-latest, 1)` and `Compiler parity`. `Test unit` — the job that actually runs the two new `#[cfg(test)]` modules, since both live in `src/` — is **still pending**, as are shards 2/3, `Test runtime` and `Formatter parity`. Everything below is about the code; it is not a green-CI sign-off.

### Byte-consistency with every caller — checked, all three

The interesting direction here is not whether the functions changed unit, but whether anything downstream still consumes the old one. Three production call sites, and each slices the same string it passed:

- `props_transforms.rs:512` — `decl` (already `.trim()`ed at `:505`), slices `decl[..pattern_end]` and `decl[pattern_end..]`.
- `props_transforms.rs:1267` — `rest` (`.strip_prefix("export let ")?.trim()`), slices `rest[..pattern_end]` and `rest[pattern_end..]`.
- `transform_script.rs:6067` — `rest_trimmed`, slices `rest_trimmed[..eq_pos]` and `rest_trimmed[eq_pos + 1..]`. The `+ 1` is safe for a reason worth stating: `=` is one byte, so `eq_pos + 1` is a char boundary whenever `eq_pos` is. Under the old char index it was neither.

No caller carries the offset onto a *different* string, stores it, or does arithmetic on it beyond that `+ 1`. So the conversion is complete on the consumer side, not just at the two definitions.

### The new `base` term is correct — and untested

`find_destructuring_pattern_end` now returns `base + i + 1` where `base` is the trim offset, which makes the documented contract ("relative to `s` as passed") true independently of caller discipline. That is the right call, and the PR body is honest that both current callers pass trimmed input so `base` is 0 today.

But that is also why nothing exercises it: all four fixtures in `pattern_end_is_a_byte_offset` — `"{ a }"`, `"{ café }"`, `"{ ああ }"`, `"[ あ, い ]"` — start at the bracket, so `base == 0` in every case. The one genuinely new line of arithmetic in this function has no coverage. Adding `"  { café }"` to the same array exercises it under the same assertion (`base` 2 + `}` at trimmed byte 8 + 1 = 11 = the full length, so `&pattern[..end] == pattern` still holds). One array element.

### The fixtures do discriminate, and they cover both failure modes

Derived independently rather than taken from the body, because the two modes fail differently and a fixture set that only hits one of them would look sufficient:

- `"{ café }"` is 8 chars / 9 bytes. The pre-fix function returns char 8, so `&pattern[..8]` is `"{ café "` — the closing brace is gone and the assertion fails **without panicking**. That is the silent mis-slice mode.
- `"{ ああ }"` is 6 chars / 10 bytes, and byte 6 falls inside the second `あ` (bytes 5-7), so the pre-fix slice **panics**. That is the loud mode.
- `"ああ = 1"` puts `=` at char 3 but byte 7, so `assert_eq!(pos, src.find('=').unwrap())` fails on the value while `src[..3]` remains a valid boundary — silent again, and caught only because the test asserts the position rather than just slicing.

Grafting the tests onto the restored pre-fix functions rather than reasoning that they would fail is the right standard, and `commented_equals_is_not_an_assignment` (`Some(5)` → `None`) pins the behaviour change that came along with `code_bytes` rather than leaving it as an unremarked side effect.

### Spot-check of the audit table: the conclusion is right, one stated reason is not

I checked `find_destructure_rhs_end`, since it is the row whose claim is most falsifiable — it is the one that returns a **char** index, so an unwrapped caller would be a live bug.

It is clean. But not for the reason the table gives:

> `find_destructure_rhs_end` — returns a char index, and **all three** callers wrap it in the `b()` char→byte table.

Two of them do: `:297` and `:388` both slice `statement[b(rhs_start)..b(rhs_end)]` (and `:393` re-wraps it for `actual_end`). The third, `:355`, **never converts it** — it collects `rhs_ends` and compares each against `c.close_pos`, which is itself a char index (`close_pos: usize, // char index of ] or }`). It is consistent by staying in char space, not by crossing into byte space.

That distinction is worth fixing in the table because the corrected reason is the stronger one: "does the caller wrap it in `b()`?" is the wrong question for a caller that never slices, and asking only that would have waved through a char-vs-**byte** *comparison*, which mis-selects a candidate silently and never panics. The audit found the right answer; the row records a check that would not always produce it.

While there I also checked the argument direction, which the scope note correctly excludes from the sweep: `start` is a char index at all three sites (`rhs_start = j + 1` from a char-indexed loop, and `c.eq_pos + 1` where `eq_pos` is documented as a char index). Consistent, so this function is clean on both sides.

### `code_bytes` side effects, checked

Neither conversion picks up a regex/division hazard. `find_simple_assignment` returns at the first top-level `=`, which precedes any division in a declarator; and inside a destructuring pattern a `/` follows an alphanumeric (`{ a = b / c }`), so `slash_starts_regex` is false. Template interpolations stay opaque exactly as the hand-rolled tracker treated them, with the nesting handled properly rather than by "scan to the next backtick".

### Two nits

**The scope note is the most valuable paragraph in the PR and should outlive it.** "The reported count measures reporting, not prevalence" plus the two uncovered directions (positions passed *as arguments*, and arithmetic on a returned offset) is exactly the shape of what is left. It is on #2378 already — good — and the argument-direction check I did above is one instance of it coming out clean, which is worth adding there as data rather than leaving the category wholly unmeasured.

**Style:** `find_simple_assignment` calls `crate::compiler::phases::phase3_transform::shared::js_scan::code_bytes(bytes)` fully qualified inline, while `props_transforms.rs` imports `code_bytes`. Matching the sibling keeps the call site readable.

The changeset is the right shape, for the record — it names the observable symptom (`let { café } = obj` lost its closing brace) rather than restating the commit subject.

baseballyama:
Re-checked at `30ca832c`. **0 blocking.** Both items from my previous review are applied, and the coverage one is applied in the form that actually exercises the new arithmetic.

`"  { café }"` is the right fixture: `base` is 2, the closing brace sits at trimmed byte 8, so `end` is 11 and `&pattern[..11] == pattern` holds — while the pre-fix function returns 8 (a char index into the *trimmed* copy) and yields `"  { caf"`, so the case still discriminates as well as covering `base`. Naming in the comment why that one entry is there keeps it from being tidied away later as a duplicate of `"{ café }"`.

The audit-table correction is the more valuable of the two, and it now records the stronger check: `:355` never converts, it compares against `other.close_pos` in char space. The reason that matters beyond this one row is that "does the caller wrap it in `b()`?" cannot see a char-vs-byte **comparison** — which mis-selects a candidate silently and never panics, so it belongs to the invisible half of the class #2378 describes.

CI note, since my last comment was explicitly conditional on it: the new push reset the run. At `9163e872` shards 1 and 2 had gone green with no failures anywhere; `Test unit` — the job that runs both new `#[cfg(test)]` modules — had not resolved before the push. It is the one to watch on this run, since it is the only job that executes the two new tests.

PR #2388, opened 2026-08-06
perf(compiler): stop rescanning the statement body once per reactive variable

baseballyama:
`extract_reactive_statement_deps` asks two questions — "does this statement body
reference this reactive variable" and "does it assign to it" — once per (`$:`
statement x reactive variable) pair. Each answer copied and rescanned the whole
body three times, formatted and searched twenty patterns, or compiled and ran a
boundary regex. Almost every pair is a miss.

Four commits, each a no-op by construction:

1. **One scan rules a variable out.** Every strip in `body_references_identifier`
   only blanks or deletes bytes, so a name absent from the raw body is absent from
   all of them; every pattern in `is_assigned_anywhere_in_body` contains the name.
   A single `memmem` scan settles both. The three strips also borrow instead of
   copying when they have nothing to strip.
2. **No per-name rebuild on a miss.** `transform_prop_reads_in_expr` and
   `transform_store_reads_client` walk and re-emit the whole text once per name;
   a name that does not occur rebuilds it unchanged, at the cost of a `Vec<char>`,
   an offset table, a scan index and a `String`.
3. **No regex.** The pattern only asked whether the byte after the name is an
   identifier byte, whether the byte before is, and whether a leading `.` is a
   member access or the tail of a spread. `IdentifierMatcher` asks those directly;
   the scan advances by one byte so overlapping occurrences stay reachable. A test
   reconstructs the regex it replaced and asserts they agree on 20 cases,
   including `obj.count`, `f(...count)`, `$count`, and non-ASCII names.

## Measurement

carbon-components-svelte v0.110.2, 287 components, client target. Interleaved
paired runs, branch point vs tip, 8 pairs:

| arm | median |
| --- | ---: |
| branch point | 1128.2 ms |
| tip | 752.0 ms |

**-33.3%**, and the tip wins all 8 pairs. The machine was under load ~116, which
inflates both arms; on a quiet machine the same binaries measure 363.7-382.0 ms
and 247.0 ms, the same -33%.

Attribution: `extract_reactive_statement_deps` was 48.7% of total compile time
here and is now 6.0%.

## Scope

This is the legacy `$:` path only, and it is client-only — the server transform
does not call these functions. It moves nothing on runes code: the corpus this
compiler had been profiled against (flowbite, 1296 files) contains **zero** files
with a `$:` line, which is why a hotspot worth half of compile time stayed
invisible. carbon carries one in 173 of 287.

### DIFF
--- a/.changeset/identifier-boundary-without-regex.md
+++ b/.changeset/identifier-boundary-without-regex.md
@@ -1,0 +1,17 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Answer the identifier-boundary question without a regex
+
+`body_references_identifier` compiled-and-cached one boundary regex per reactive
+variable and then ran it over the stripped statement body, once per (`$:`
+statement × variable) pair. The pattern only ever asked three things — is the
+byte after the name an identifier byte, is the byte before one, and is a leading
+`.` a member access or the tail of a spread — so a matcher that asks them
+directly replaces it. Overlapping occurrences stay reachable because the scan
+advances by one byte, not by the match length.
+
+On carbon-components-svelte this regex was 70% of the remaining time in
+`extract_reactive_statement_deps`; its share of total compile time drops from
+19.6% to 6.0%.
--- a/.changeset/reactive-deps-substring-gate.md
+++ b/.changeset/reactive-deps-substring-gate.md
@@ -1,0 +1,19 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Rule a reactive variable out with one scan before rewriting the statement body
+
+`extract_reactive_statement_deps` asks `body_references_identifier` and
+`is_assigned_anywhere_in_body` once per (`$:` statement × reactive variable)
+pair, and each answer copied and rescanned the whole statement body three
+times — or formatted and searched twenty patterns. Almost every pair is a miss,
+and a name absent from the raw body is absent from every stripped derivative of
+it, because the strips only blank or delete bytes. One substring scan now
+settles those. The three strips also borrow instead of copying when they have
+nothing to strip.
+
+On carbon-components-svelte, whose components are legacy (173 of 287 files
+carry a `$:` line), this was 48.7% of total compile time; the corpus this had
+been profiled against carries no `$:` at all. Compiling the 287 components
+drops from 366-380 ms to 268-279 ms.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -5354,12 +5354,14 @@ fn transform_instance_script_for_visitors(
             let _reactive_guard = super::profile::ReactiveStmtGuard(_reactive_start);
             // Extract assignment targets and dependencies from the raw statement
             // for topological sorting (matching official compiler's order_reactive_statements)
+            let _rs_deps_start = super::profile::timer_start();
             let (assigned_vars, dep_vars) = extract_reactive_statement_deps(
                 &statement,
                 state_vars,
                 prop_assignment_transform_vars,
                 store_sub_vars,
             );
+            super::profile::record_rs_deps(super::profile::timer_elapsed(_rs_deps_start));
 
             let var_state_vars: Vec<String> = legacy_state_vars
                 .iter()
@@ -5375,6 +5377,7 @@ fn transform_instance_script_for_visitors(
                 .map(|v| v.as_slice())
                 .unwrap_or(&[]);
             *reactive_ordinal += 1;
+            let _rs_body_start = super::profile::timer_start();
             let transformed = transform_reactive_statement(
                 &statement,
                 state_vars,
@@ -5388,6 +5391,8 @@ fn transform_instance_script_for_visitors(
                 analysis,
                 &prop_invalidate_bodies,
             );
+            super::profile::record_rs_body(super::profile::timer_elapsed(_rs_body_start));
+            let _rs_assign_start = super::profile::timer_start();
             // Also apply state assignment transformations to the reactive statement body
             // This handles cases like: `$: selected ? component = Sub : component = banana`
             // where state variables are assigned inside conditional expressions.
@@ -5401,6 +5406,7 @@ fn transform_instance_script_for_visitors(
                 &non_proxy_vars,
             )
             .unwrap_or(transformed);
+            super::profile::record_rs_assigns(super::profile::timer_elapsed(_rs_assign_start));
             // Collect reactive statements to append at end (matching official compiler behavior
             // which appends all reactive statements after the rest of instance body code)
             let mut reactive_code = transformed;
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
@@ -182,6 +182,13 @@ pub(super) fn transform_prop_reads_in_expr(expr: &str, prop_vars: &[String]) -> 
     let mut result = expr.to_string();
 
     for prop_name in prop_vars {
+        // The walk below pushes every character it reads, so a name that does
+        // not occur rebuilds the expression unchanged -- at the cost of a
+        // `Vec<char>`, an offset table, a scan index and a `String` per name.
+        if memmem::find(result.as_bytes(), prop_name.as_bytes()).is_none() {
+            continue;
+        }
+
         // Use word boundary matching to replace identifier references
         // But avoid replacing function calls that already have ()
         // Note: Rust's regex crate doesn't support lookahead, so we use a different approach:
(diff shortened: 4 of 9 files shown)


baseballyama:
## Review — correctness is solid, output parity is proven, the perf evidence is one tier short

Recommendation: **merge after (a) tier (c) is measured and (b) the cancelled check is re-run.**
Nothing in the code is wrong; the gap is entirely in what was measured and how it was reported.

### CI

`gh pr checks 2388`: 33 pass, 1 fail — **VPS native VERSION drift check**. It is not a real
failure: the job's only step is `Set up job / cancelled`, and run `31111838072` has
`conclusion: cancelled`. Per repo precedent (cancelled / exit-143 ⇒ re-run, not investigate) it just
needs re-running before merge; `mergeable_state` is `unstable` only because of it.

The gates that matter all completed green **at the head SHA `daa3193e`**: run `31111839499`
(Corpus Compat, success) containing **Compiler parity** (job `92651325794`, 8m8s) and **Shape matrix
parity** (job `92651325932`, 2m59s), plus Test unit / Test 1-3 / Test runtime / Clippy / Format.

Ignore the "Compatibility Report" bot comment (validator 192/334): it is byte-identical on #2389 and
#2394 with the same `Current commit: 44a781373057`. That bot is broken and is worth its own issue —
it is not a signal about this PR.

### Output parity — proven, by the right gate

Named evidence, not the claim: Compiler parity + Shape matrix parity at `daa3193e`. And the corpus
is not runes-only — `scripts/compat-corpus/corpus-sources.json` enrols sveltejs/svelte's
legacy-heavy fixtures plus `sveltestrap`, `smelte`, `attractions`, `svelte-table`, `powertable`,
`svelte-pivottable`, `svelte-calendar`, so the legacy `$:` path this PR touches is genuinely
exercised; Runtime Legacy (1207) is green via Test runtime.

### Correctness

No memoization is introduced, so the "insufficient cache key" risk does not apply: the change is
three per-call early-outs plus a per-call `IdentifierMatcher`. It *removes* shared state — the
unbounded, never-cleared thread-local `FxHashMap<String, Arc<Regex>>` in `get_or_compile_regex`
(`client/mod.rs:194`) loses its highest-cardinality producer.

Each early-out was checked for equivalence rather than read for plausibility:

- The raw-body pre-scan in `body_references_identifier` (`state_transforms.rs:34`) is sound only if
  the three strip passes never *create* an occurrence. All three are length-preserving — blanking
  bytes in place (`result[i] = b' '`) or `replace_range(pos..end, " ".repeat(end - pos))`. Nothing
  deletes or joins, so no two fragments can fuse into the identifier. ✔
- The `memchr(':')` and `"function"`/`"=>"` guards sit on the sole mutation sites, and since
  mutations only blank, neither substring can appear later. ✔
- The 20 `is_assigned_anywhere_in_body` patterns are all `format!`-built around `var_name`. ✔
- `IdentifierMatcher` ≡ the deleted regex: beyond the committed 20-case test, a **402,730-case
  differential fuzz** (exhaustive short strings over `. $ a ␠` × 10 identifiers including
  `$$restProps`, `$0`, `あn`, plus 400k random strings with non-ASCII) found **0 mismatches** —
  overlapping occurrences, `obj.count` vs `f(...count)`, `$count` vs `count`, UTF-8 continuation
  bytes immediately before the name.
- The props/store per-name skips are exact identity when the name does not occur (both walks end in
  copy-through and every replacement is guarded by `starts_with(name)`), and ordering is preserved
  because the `is_inside_getter_context` prefix check reads a buffer that is unchanged when skipped.

Only behavioural edge: an empty identifier now returns `false` where the regex could match.
Unreachable from declaration names.

### Perf evidence — the blocking gap

| tier | reported? |
|---|---|
| (a) synthetic scaling fixture | **no** — only CodSpeed's `transform_client[05-legacy-reactive]` 7.8 → 6.7 ms (+16.48 %): CI-generated, a single size rather than a scaling curve, and CodSpeed itself flags *"Different runtime environments detected"* and compared against `b152751` because no successful run existed on the base |
| (b) real trigger | **yes** — carbon-components-svelte v0.110.2, 287 components, 8 interleaved pairs: 1128.2 → 752.0 ms, tip wins 8/8 |
| (c) whole shipped population | **no** — replaced by the argument "flowbite, 1296 files, contains zero `$:`" |

1. **Tier (c) is missing, and the substitute argument does not cover the PR.** The zero-`$:`
   argument retires commits 1 and 3 only. **Commit 2 touches paths reachable with no `$:` at all**:
   `transform_store_reads_client` runs for `$store` auto-subscriptions and
   `transform_prop_reads_in_expr` for `export let` props. On the *hit* path both now pay an extra
   `memmem` scan before the walk. So tier (c) is not structurally 1.000 — it is exactly where a
   small regression would hide. You already touched `compile_profile.rs`, whose `--shipped` mode
   runs the six vendored projects; this costs minutes.
2. **The tier-(b) statistic is not the required one.** The bar is *median of the per-pair ratio*;
   the PR reports median-of-each-arm and derives −33.3 % from those. "Wins 8/8" carries the sign so
   the conclusion survives, but please restate it from the 8 raw pairs you already have. The primary
   run was under load ~116 (your own note) and the quiet-machine follow-up is given asymmetrically
   as a range vs a point (`363.7–382.0` vs `247.0`), which reads as min/range rather than paired
   medians — two machines should not be read as one series.
3. **Layout effect.** Before/after as separate binaries mixes ~5 % of code-layout effect into every
   ratio. Irrelevant at −33 %; decisive for tier (c), where the honest claim will be **"no
   regression"**, not a number.

I have deliberately not computed the missing tier for you — a measurement produced by the reviewer
is not evidence the change was measured.

### Smaller points

- Three unconditional `Instant::now()` / `elapsed()` pairs per `$:` statement now ship in release
  builds (`rs_deps` / `rs_body` / `rs_assigns` in `profile.rs`). ~300 ns per reactive statement,
  consistent with the module's existing practice and its documented budget — but it is permanent tax
  on the very path being optimised.
- `CLAUDE.md` "Code Comments" (no change history / provenance, single line): the `IdentifierMatcher`
  doc comment opens with *"This used to be a regex … was 70% of that scan's cost"*. Keep the second
  half (the rule it encodes), drop the provenance and the figure — the figure will go stale.
- Commit 2 has no changeset.
- `rsvelte_devtools/src/bin/prop_reads_work_count.rs`'s `record_pass` counts become non-comparable
  across this commit (it now counts only passes actually performed — arguably more correct). Worth a
  line so nobody compares across the boundary.

### Prior review

None. Zero review comments and zero threads; the only review is
`copilot-pull-request-reviewer[bot]` reporting a quota limit. Nothing was outstanding.

### What is still owed

1. Tier (c): `compile_profile --shipped`, branch point vs tip, interleaved, **median of the per-pair
   ratio**, reported as its own row. Expected ≈1.00 — the point is to prove commit 2's added
   `memmem` on the hit path costs nothing.
2. Tier (b) restated as median-of-per-pair-ratio over the 8 pairs; relabel or drop the load-116 run.
3. Tier (a): either build a scaling fixture, or state plainly that none exists and that CodSpeed's
   `05-legacy-reactive` is the only microbenchmark, with its mixed-environment and wrong-base
   caveats. The CodSpeed +16.48 % should not stand in as an author-produced tier (a).
4. Re-run the cancelled **VPS native VERSION drift check**.

baseballyama:
### Correction to my previous comment, §"Perf evidence", point 1

I claimed commit 2 reaches non-`$:` code through **two** functions. Half of that is wrong, and the
half that is wrong is the half I did not check before writing it. Re-derived by enumerating call
sites (`grep -rn 'transform_store_reads_client\|transform_prop_reads_in_expr' crates/rsvelte_core/src/`):

- **`transform_prop_reads_in_expr` — `$:`-only.** Every call site is in
  `3_transform/client/reactive_transforms.rs` (lines 459, 502, 528, 543, 587, 622, 640, 642, 656,
  703). The rest of the hits are `tests.rs`, a doc comment in `store_transforms.rs:353`, and the
  `measure_prop_reads.rs` devtool. My "for `export let` props" was wrong — retract it.
- **`transform_store_reads_client` — reachable with no `$:` in the file.** Confirmed:
  `3_transform/client/mod.rs:5774` sits in the general legacy instance-script statement loop under
  `if !store_sub_vars.is_empty() && !analysis.runes && …`, i.e. it is keyed on `$store`
  auto-subscriptions, not on reactive statements. (The other call site, `reactive_transforms.rs:745`,
  is the `$:` path.)

The request is unchanged but narrower and more precise: **tier (c) is still owed, and
`transform_store_reads_client` is the single reason.** Note that the `mod.rs:5774` site already
carries a `store_sub_vars.iter().any(|store| transformed.contains(store))` guard, so commit 2's
per-name skip adds a *second* containment scan per name on the path that does hit. That is the
specific thing a shipped-population measurement has to show costs nothing — a legacy component with
`$store` and no `$:` is the shape to look at, and it is not what carbon was chosen to exercise.

Sorry for the noise. The rest of the comment stands; the two CI claims in it I have since verified
directly (`runs/31111838072` → `conclusion: cancelled`, sole step "Set up job"; `runs/31111839499`
Corpus Compat → success at `daa3193e`, with **Compiler parity** job `92651325794` and **Shape matrix
parity** job `92651325932` both success).

Issue #2386, opened 2026-08-06
compiler: distinguish byte and char offsets by type, so #2378 stops needing a sweep


baseballyama:
## Summary

Every silent mis-slice in #2378 has the same shape: a function computes an offset in one unit
and a caller consumes it in another. Nothing in the type system objects, because both units are
`usize`. Fixing instances one at a time keeps working, and keeps needing a sweep to find the
next one.

Give the two units distinct types — `ByteOffset` and `CharOffset` — so the mismatch is a compile
error at the boundary where it happens, and the sweep stops being the detection mechanism.

## What this does **not** catch

Stating this up front, because the failure mode of a fix like this is that the next sweep gets
skipped on false confidence:

- **Right unit, wrong string.** A `ByteOffset` computed against `trimmed` and used to slice
  `original` type-checks perfectly and mis-slices exactly as badly. The newtype carries no
  provenance. This is a real shape here — `props_pattern_span` and
  `find_destructuring_pattern_end` both take a trimmed base and their callers re-add the trim
  delta by hand.
- **Arithmetic that leaves the unit valid but the position wrong** (`+ 1` past a multi-byte
  character). Still a byte offset, still type-correct, still mid-character.
- **Anything crossing FFI or a serialized boundary**, where the type is erased anyway.

So this closes the *unit-confusion* half of #2378 and leaves the *provenance* half open. #2378
stays open after this lands.

## Correction to the scoping premise

The plan was to start with "the module holding the five known instances". **There is no such
module** — the five are spread across four:

| instance | module | fixed in |
|---|---|---|
| `transform_legacy_destructure_declarations` | `server/transform_legacy.rs` | #2369 |
| `extract_store_set_targets` | `server/transform_legacy.rs` | #2370 |
| `find_colon_at_depth0` | `client/state_transforms.rs` | #2376 |
| `find_destructuring_pattern_end` | `client/props_transforms.rs` | #2385 |
| `find_simple_assignment` | `server/transform_script.rs` | #2385 |

Density of *known* instances is the wrong slicing criterion anyway: known instances are the ones
already fixed, so a module scores high precisely where the remaining risk is lowest.

## Proposed slice 1: `client/scan_index.rs`

261 lines, three caller modules (`expression_utils.rs`, `props_transforms.rs`,
`store_transforms.rs`), and — the reason it is the right first slice — it is **uniformly
char-indexed by design**. Its own doc comment opens with "One-pass bracket/token index over the
[...]

PR #2389, opened 2026-08-06
fix(compiler): keep store reads intact when the store name is not ASCII

baseballyama:
Found while scoping slice 1 of #2386. It is a **two-layer** defect where the upper layer hides the lower one, so the two fixes cannot be separated.

## Layer 1 — the identifier pre-filter is ASCII-only

`client/utils.rs::text_contains_any_identifier` extracts words with
`is_ident_start_byte` / `is_ident_continue_byte`, both of which accept only
`[A-Za-z0-9_$]`. A store subscription named `$名前` is therefore extracted as the
word `$`, which is not in the variable set, so `transform_store_reads_client`
returns the line untouched at its first guard.

The function's own doc comment already states the stakes: *"false positives just mean we do a
bit more work in the downstream transform, while false negatives would cause correctness bugs."*
This is that false negative.

## Layer 2 — the read rewriter mixes char index and byte length

`transform_store_reads_client` walks `chars: Vec<char>` with an index `i`, and at
**nine sites** advanced or offset it by `store_sub.len()` — the name's length in
**bytes**:

```rust
let after_idx = i + store_sub.len();   // :400, :427
i += store_sub.len();                  // ×6
is_shorthand_object_property(&index, &chars, i, store_sub.len())
```

For `$名前` that is 7 instead of 3, so every match skipped four characters of source.

## The masking, demonstrated rather than argued

The pin `store_read_survives_a_non_ascii_store_name` was written first and confirmed failing.
It failed **for the wrong reason** — output was the input verbatim, not the mis-sliced form I
predicted — which is what exposed layer 1. Re-running it with layer 1 patched and layer 2 still
broken produced the predicted corruption:

| state | `x = $名前 + 1;` |
|---|---|
| `main` | `x = $名前 + 1;` (no transform — layer 1 masks) |
| layer 1 fixed only | `x = $名前();` — **` + 1` silently dropped** |
| both fixed | `x = $名前() + 1;` |

So **fixing the pre-filter alone turns a missing transform into lost output.** That is the
reason this is one commit and not two.

The second pin (`…is_not_already_a_call…`) passes on `main` — it is non-discriminating there,
because both the correct and the masked path leave the input alone. It only becomes
discriminating in the middle column, where it fails. Recording that so nobody reads it as
coverage of `main`.

## Layer 1 also stops a *wrong* transform, not only a missing one

Under-claimed in the first version of this body. `名$count` is a single JavaScript identifier,
but the ASCII-only pre-filter extracted `$count` from it as a standalone word, and
`transform_store_sub_calls:238` then checks the preceding byte:

```rust
let prev_byte = result.as_bytes()[abs_pos - 1];
let prev_char = prev_byte as char;
!prev_char.is_alphanumeric() && prev_char != '_' && prev_char != '$'
```

That byte is `0x8D`, the last continuation byte of `名`. `0x8D as char` is U+008D, a control
character, so `is_alphanumeric()` is false and `before_ok` came out **true**. Measured on
`origin/main`, with the ASCII case as the control that distinguishes the mechanism from a
general non-ASCII bail:

| input | `main` | this PR |
|---|---|---|
| `x = $count(1);` | `x = $count()(1);` | same |
| `x = a$count(1);` | unchanged — correctly rejected | same |
| `x = 名$count(1);` | **`x = 名$count()(1);`** — wrongly transformed | **unchanged** |

So layer 1 is a correctness fix in **both** directions: it stops a missed transform *and* a
wrong one. The new predicate makes the whole run one word, so the gate bails before `:238` is
reached.

**What it does not fix, stated because the table above could be read as more than it is**: the
`byte as char` defect at `:238` survives. On a mixed line the gate stays open via the standalone
occurrence and the wrong transform still happens — measured on this branch:

```
x = $count(1) + 名$count(1);   →   x = $count()(1) + 名$count()(1);
```

That is a distinct shape (a byte *decoded* as a char, not an index in the wrong unit) and is
filed separately rather than fixed here.

## Risk of the layer-1 change

The old and new predicates differ by exactly one disjunct, `b >= 0x80`, which is false for every
byte below `0x80`. On all-ASCII input the new functions **are** the old functions, extensionally
— this is algebraic, not a sampling result, so there is no input class left to check.

The one shape where the new predicate reads *differently* is an ASCII name immediately followed
by a non-ASCII character (`abc日`): previously two words, now one. In JavaScript source that is
genuinely a single identifier, so the new reading is the correct one; inside a string literal it
converts a harmless false positive into a false negative, which the downstream `is_inside_string`
check would have discarded anyway. **The corpus job on this PR is the oracle for that claim** — I
did not run it locally.

## Verification

- 2 pins, written first, confirmed failing (see table).
- `cargo test -p rsvelte_core --lib` — **1420 passed, 0 failed**.
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — clean, no `-A`.
- `rustfmt --edition 2024 --check` on both files — clean.

## Relation to #2386

This is slice 0, not slice 1: it is the *bug* that justified re-scoping the newtype away from
`client/scan_index.rs`. `store_transforms.rs` holds exactly one `Vec<char>` region meeting 17
byte-slice sites, which is the coexistence the newtype exists to police —
`scan_index.rs` is the only module in the population where the two units **cannot** meet, so the
newtype could never have caught anything there. Details in #2386.

Related: #2378, #2386.


baseballyama:
Reviewed at `86892bb3`. **0 blocking.** I reproduced the middle column and checked the safety argument directly; both hold. The two-layer framing is right, and the PR already records the labelling point I would otherwise have raised. Read-only — no build, no test run. CI: 1 pass, 30 pending at time of writing, no failures.

### The middle column reproduces

This is the cell that decides one commit versus two, so I derived it rather than accepting it. With layer 1 fixed and layer 2 unfixed, for `x = $名前 + 1;` and `store_sub = "$名前"`:

`store_sub.len()` is **7** bytes (`$` + 3 + 3) while the name is **3** chars. `result` is 12 chars, and the match begins at char index 4. The branch pushes `$名前` + `()` into `new_result`, then `i += 7` lands on index 11 — the `;`. Chars 7..11 (` `, `+`, ` `, `1`) are never pushed, because every arm `continue`s after the advance rather than copying the skipped region. Output: `x = $名前();`. **` + 1` is silently dropped, exactly as the table says.**

That is a strictly worse failure mode than the one it replaces — a missing transform leaves valid code with wrong runtime semantics, whereas this deletes source text — so bundling the two layers into one commit is the correct call and the reasoning should stay in the changeset, which it does.

### The ASCII-identity claim is stronger than "no existing input can move"

It is not a sample, it is extensional equality. `is_ident_start_byte` and `is_ident_continue_byte` each gain exactly one disjunct, `b >= 0x80`, which is **false for every byte below 0x80**. So on all-ASCII input the new predicates return the same value as the old ones for every byte, `extract_identifiers` produces the identical word set, and `text_contains_any_identifier` reaches the identical decision. There is no input class to sample — the two functions are the same function on ASCII. Worth stating that way rather than as a risk assessment.

### An unclaimed second fix, worth putting in the PR body

The `abc日` analysis in the body covers the *less permissive* direction and argues it is more correct. It is, and there is a stronger version: the change also closes a live mis-transform in the **other** direction.

Take `名$count` with a store named `$count`. Under the old predicate, `名`'s bytes were not identifier bytes, so extraction produced the word `$count` — a match — and `transform_store_sub_calls` proceeded to its boundary check at `:238`, which reads `result.as_bytes()[abs_pos - 1] as char`. That byte is `0x8D`, the last byte of `名`, and `0x8D as char` is U+008D, a control character: `is_alphanumeric()` is false, so `before_ok` came out **true** and `$count` inside the identifier `名$count` was treated as a standalone store reference. Under the new predicate the whole run extracts as one word `名$count`, which does not match, and the gate at `:361`/`:215` bails before any of that.

So the pre-filter fix does not only stop missing transforms; it stops a wrong one. That is worth a line, because it changes how the layer-1 change should be described — it is a correctness fix in both directions, not a permissiveness trade.

### Residual of the same class, pre-existing and out of scope

The `byte as char` boundary checks survive at `:238`, `:246` and `:256`. They are now unreachable through the gate for a *pure* `名$count` line, but not for a mixed one: `x = $count + 名$count;` keeps the gate open via the standalone occurrence, and `transform_store_sub_calls` then walks byte offsets and hits the same U+008D reading at the second occurrence. `transform_store_reads_client` is safe here because it walks `chars` and its before-check sees the real `名`, which `is_alphanumeric()` correctly reports as true — the two functions disagree only because one is byte-indexed and the other char-indexed.

Rare, pre-existing, and squarely #2378's population rather than this PR's. Worth an issue so the `byte as char` pattern gets counted, since it is a third shape of the class: not a char index used as a byte, but a byte *decoded* as a char.

### Checked rather than assumed

- Nine `sub_chars` sites (`:402, :429, :481, :486, :491, :498, :500, :508, :514`), matching the claim.
- The four surviving `store_sub.len()` uses (`:119, :244, :245, :249`) are in byte space throughout — `statement.len()`, `statement[pos..].find()`, `result.as_bytes()`, and slices of `result` — so they are consistent, not missed.
- `is_shorthand_object_property`'s fourth parameter is a **char** count: it computes `var_end = var_start + var_len` and then indexes `chars[k]` (`expression_utils.rs:1372-1385`). Passing `sub_chars` is right, and this was the site most likely to be wrong, since the length crosses a function boundary where the unit is not in the name.
- All three `transform_*` entry points sit behind `text_contains_any_identifier` (`:30`, `:215`, `:361`), so the pre-filter really is the single gate the two-layer story depends on.

### On the tests

The body already states that the second pin passes on `main` and is non-discriminating there, discriminating only in the middle column. That is the right call and the right place to say it — recording a guard as a guard is exactly what keeps a "2 pins" line from reading as two repros. The first pin covers both broken states with one input, which is the economical shape: it fails on `main` (no transform) *and* on layer-1-only (`x = $名前();`), so it cannot be satisfied by a partial fix.

Naming the corpus job as the oracle for the `abc日` claim, rather than asserting the corpus was clean, is also correctly scoped — the claim is pending, not made.

PR #2390, opened 2026-08-06
test(core): make the CSS scoping tests reach the path they are named for (#1791)

baseballyama:
Part of #1791.

## Items 2 and 5 were already done

Before writing anything I checked the tree rather than the issue text. Both items the issue
still lists as open are complete:

- **Item 5 (16 empty `#[test]` functions)** — `expression_walker_pin.rs` no longer exists and
  there are **zero** empty `#[test]` bodies anywhere under `crates/`. Dropped by `e2337bfc`
  ("make the skip audit assert and drop 16 assertion-less pin tests (#1791)", PR #1811).
- **Item 2, `audit_skipped.rs`** — same commit. It is now a two-sided ratchet: `unverifiable`
  (a skip whose fixture is gone), `new_stale` (a skipped fixture that now passes), and
  `dead_ratchet` (a `KNOWN_STALE_SKIPS` entry that no longer applies).
- **Item 2, `svelte2tsx_fixtures.rs`** — lives at `crates/rsvelte_projection/`, not
  `crates/rsvelte_core/`, and already asserts: `total_tested > 0` plus a regression assert
  against `compatibility/svelte2tsx-fixtures-known-failures.json`. The asserts predate the file
  move in `b311eecc`.

Both are wired, not merely written: `scripts/ci/run-test-shard.sh` enumerates every
integration-test target from `cargo metadata` and excludes only three by name, neither of them
these.

## What was actually still vacuous

Two `#[cfg(test)]` tests in `3_transform/css.rs` — item 2's exact failure mode, in a file the
issue does not name:

```rust
if let Some((css_content, css_start)) = extract_css_content(input) {
    ...
    let output = transform_css(&children, selector, hash, &css_content, css_start, &ctx);
    println!("CSS Output:\n{}", output.text);
}
```

Vacuous twice over. They assert nothing, and the whole body is inside an `if let` with no
`else`, so a `None` from `extract_css_content` passes silently.

**Adding an assertion would have been the wrong fix**, which is why they are replaced rather
than patched. Both built a `CssContext` by hand with `used_elements: FxHashSet::default()` —
empty. In the real pipeline that set comes from `analysis.css.used_elements` (phase 2). With it
empty, *every* selector is pruned, and the run prints:

```
/* (unused) main button { ... }*/
/* (unused) main div > button { ... }*/
```

So `test_combinator_handling` never reached combinator handling at all. An assertion on that
output would have pinned the pruning of everything — passing for a reason nobody stated.

## Replacement

`crates/rsvelte_core/tests/css_scoping_combinators.rs` drives the real `compile()` pipeline, so
`used_elements` is populated for real, and pins three things:

- a matched element selector is kept and scoped, not pruned;
- descendant and child combinators are both scoped —
  `main.svelte-H div:where(.svelte-H) > button:where(.svelte-H)`;
- a selector no element matches **is** pruned (negative control — without it, an implementation
  that kept and scoped everything unconditionally would satisfy the first two).

The scoping hash is content-derived, so it is normalised to `svelte-H` and the exact selector
shape is asserted, rather than "a hash appears somewhere".

## Proof that the assertions discriminate

Each mutation was applied to `main`, run, and reverted.

| mutation | result |
|---|---|
| **A** — invert the used-element test: `return !ctx.used_elements.contains(&decoded)` → `return ctx...` (`css.rs:4468`) | **kills all 3** |
| **D** — drop every `:where({})` wrapper from the scoping emitters | **kills only `descendant_and_child_combinators_are_both_scoped`** |

A proves the prune/retain axis is covered. D proves the scoped-shape assertions discriminate on
exactly the axis they claim, and that the other two tests are correctly insensitive to it.

Two further mutations were **inert and are reported as neither kills nor survivals**: dropping
the `:where()` wrapper at `css.rs:933` alone, and at `css.rs:7230` alone. Under each, the
emitted CSS is byte-identical to the unmutated output — those lines never run for these inputs,
so they are not valid negative controls and say nothing about the tests either way. Only the
combined mutation D moves the output.

## Changeset

None. Test-only: one integration test added, two assertion-free `#[cfg(test)]` tests removed.
No shipped crate behaviour changes — `css.rs`'s non-test code is untouched.


baseballyama:
Reviewed at `70ef629b`. **0 blocking.** The diagnosis of why patching was wrong is the valuable part, the replacement reaches the path it names, and the two inert mutations really are inert — I confirmed both by reading, and each has a *reason* worth recording alongside the empirical result. Read-only — no build, no test run. CI: 1 pass, 30 pending, no failures.

### Why patching would have been wrong is the finding

Both deleted tests built a `CssContext` with an empty `used_elements`, so every selector was pruned and the run printed `/* (unused) … */` for all of them. `test_combinator_handling` never reached combinator handling. The consequence worth stating exactly as the body does: **an added assertion would have pinned the pruning of everything** — the test would then pass, permanently, for a reason nobody stated, and it would have looked like coverage of the combinator path forever after.

That is a sharper case than "the test had no assertions". A test with no assertion is visibly worthless; a test whose fixture puts the code in the wrong state and *then* gets an assertion is worse, because it now claims something. Deleting rather than patching is the right call and the reason generalises.

### The replacement reaches the path, and proves it inline

Driving the real `compile()` populates `used_elements` from phase 2, and the `!out.contains("(unused)")` assertion in both positive tests is not decoration — it is the proof that pruning did not happen, i.e. that execution got past the gate the old fixtures never cleared. That is the right way to assert "we reached the code": make the precondition an assertion rather than a comment.

The scoped-shape assertions then pin the combinator output itself — `main.svelte-H button:where(.svelte-H)` and `main.svelte-H div:where(.svelte-H) > button:where(.svelte-H)` — with the hash normalised so the test pins the selector *shape* rather than "a hash appears somewhere". Normalising and asserting the exact string is the difference between a test that survives a codegen change and one that catches it.

`a_selector_no_element_matches_is_pruned` is the necessary third leg: without it, "keep and scope everything unconditionally" satisfies the first two. Mutation **A** confirms that axis empirically, and **D** confirms the scoped-shape assertions discriminate on exactly the axis they claim while the other two tests are correctly insensitive to it. A kills 3, D kills 1 — different kill sets from different mutations is what makes the three tests non-redundant.

### The two inert mutations are genuinely inert, and here is why

Reporting them as **neither kills nor survivals** is the correct handling, and it is the part most likely to have been fudged, so I checked both:

- **`css.rs:933`** — `let where_modifier = format!(":where({modifier})")` sits in the segment-alignment routine that walks `emitted` against `css_source` to build source-map runs (`:928-940`). It exists to *recognise* the inserted text while mapping, not to produce it. These tests read `.css.code` only, so nothing they assert on can move. Inert for two independent reasons.
- **`css.rs:7230`** — the `write!(selector_parts, ":where({})", selector)` is inside the branch guarded by the `skip` computation over `n == "root" || n == "host" || ((n == "is" || n == "where") && selectors.len() == 1)` (`:7222-7231`), i.e. the pseudo-class path. The four test inputs are `div`, `main button`, `main div > button` and `section` — **no pseudo-class selector anywhere** — so the branch is never entered.

Both check out. One suggestion, since the table is going into the record: state *what input would make each live* rather than only that the output did not move. For `:7230` that is a selector like `:is(main) button`; for `:933` it is an assertion on the CSS **source map** rather than on `.code`. That converts "inert" from a dead end into a note about which coverage is genuinely absent versus which is not applicable — and it is the difference between "we could not kill it" and "nothing here should".

### One observation, not a request

`css()` passes `runes: Some(true)` while the fixtures contain no runes. It is harmless and probably deliberate (forcing runes mode keeps the component off the legacy analysis path), but since every other option in that struct is load-bearing for what is asserted, a reader will wonder. One clause in the helper's doc would settle it — or drop it, if the CSS pipeline is mode-independent here, which would make the test's dependencies exactly the ones it needs.

No changeset is the right call: `css.rs`'s non-test code is untouched, so there is no shipped behaviour to describe.

PR #2391, opened 2026-08-06
fix(analyze): gate attribute_quoted on runes mode

baseballyama:
Part of #2281 (warning parity). Burns `attribute_quoted` out of the code ratchet: **70 → 51 entries**, 19 removed, 0 added, on all three targets.

## One predicate

Upstream reaches the quoted-attribute check only through `validate_attribute`, and **both** of its callers guard it with `analysis.runes`:

- `2-analyze/visitors/shared/component.js:92` — `if (context.state.analysis.runes) { validate_attribute(attribute, node); … }`
- `2-analyze/visitors/shared/element.js:42` — same guard

So in legacy mode upstream never emits `attribute_quoted` at all. rsvelte ran the check unconditionally at all four of its emission sites (`shared/component.rs`, `svelte_component.rs`, `svelte_self.rs`, `regular_element.rs` for custom elements), giving legacy users a warning they cannot act on — `<Component id="{item.id}" />` in a file with no runes.

The check itself was correct; only the gate was missing. Verified against both compilers rather than read off the source:

| input | official | rsvelte (before) |
|---|---|---|
| `<script>let x=1;</script><Component id="{x}" />` | — | `attribute_quoted` |
| `<script>let x=1;</script><my-el id="{x}"></my-el>` | — | `attribute_quoted` |
| `<svelte:options runes /><Component id="{x}" />` | `attribute_quoted` | `attribute_quoted` |
| `<svelte:options runes /><div id="{x}"></div>` | — | — |

## Two things found on the way

**The span was missing.** Upstream's `w.attribute_quoted(attribute)` takes the attribute as its warn node; rsvelte emitted the warning with no span, so even where the code matched, the position did not. Fixed here — the upstream `validator/samples/attribute-quoted` fixture now matches official on all four `attribute_quoted` positions (`13:11`, `15:29`, `19:14`, `22:16`). It stays in `validator-known-failures.json` for an unrelated reason: `svelte_component_deprecated` in the same file still has no span.

**The message was wrong.** rsvelte said "Quoted attribute **values** will be stringified"; upstream says "Quoted attributes on **components and custom elements** will be stringified". rsvelte's wording implies the warning applies to plain elements, which it never does. This one is invisible to the corpus gate — that gate compares `(code, line, column)` and never the message — and the validator fixture that would catch it is in `validator-known-failures.json` for a different reason, so the wrong text sat behind an entry justified by something else.

## Cleanup

The predicate existed in four copies: two identical private `is_quoted_single_expression` functions and two inline `matches!` chains. All four now call one helper in `shared/attribute.rs`, which is where upstream keeps it — and, being one function, is where the runes gate could be applied once instead of four times.

## Tests

`crates/rsvelte_core/tests/attribute_quoted_runes_only.rs` — 6 repros (all FAILED before, all pass now) and 2 pins:

| test | kind |
|---|---|
| `legacy_component_attribute_is_not_flagged` | repro |
| `legacy_svelte_component_attribute_is_not_flagged` | repro |
| `legacy_svelte_self_attribute_is_not_flagged` | repro |
| `legacy_custom_element_attribute_is_not_flagged` | repro |
| `warning_points_at_the_attribute` | repro (span) |
| `message_matches_upstream` | repro (text) |
| `runes_component_attribute_is_still_flagged` | **pin** — fails if the fix suppresses instead of gating |
| `runes_custom_element_warns_but_plain_element_does_not` | **pin** — both halves of the element rule |

One repro per emission site, deliberately: a single gate now feeds all four, but four call sites can drift apart again.

`--test validator` stays 333/333, as do `validator_message_wording`, `attribute_shorthand_validation`, `svelte_component_validation` and `warning_filter`.

## Severity of what is left

Worth recording, because the ratchet count does not distinguish it: of the 51 entries that still diverge on codes, **6 are under-warnings** — rsvelte silent where upstream warns (`a11y_no_static_element_interactions` ×3, `state_referenced_locally` ×2, `options_missing_custom_element` ×1). The other 45 are noise the user cannot suppress, 116 tuples over five codes, and no entry diverges in both directions at once. A missing diagnostic and an extra one are both defects but fail differently, and the ratchet reports them as one number.

## Docs

`compatibility/warning-known-failures.md` had this cluster recorded as "`attribute_quoted` over-warning on **namespaced SVG child components**". That was never the cause — it was inferred from where the entries clustered in the corpus rather than from upstream's control flow. The file now says what the cause was, and flags that the remaining cluster descriptions were written the same way.

## Ratchet

`warning-known-failures.*.json` 70 → 51 (19 removed, 0 added). `warning-position-known-failures.*.json` unchanged at 529, set for set. Re-baselined in this PR with a release build of the branch.

**Rebased onto post-#2384 main and regenerated from a run** (not hand-merged, not derived):

```
node scripts/compat-corpus/compile.mjs                              # 14,130 entries × 3 targets
```

`git status` was **clean** afterwards — the run reproduced this branch's committed ratchets byte-for-byte, so no hand-edit exists anywhere in the history. Entry floor met (14,130 ≥ the 12,000 `--update-warning-baseline` refuses below).

The `--no-fmt` seeding was then checked against the real CI configuration by re-running **with** formatting and **without** any update flag, so the committed baseline was verified two-sided:

```
node scripts/compat-corpus/verify.mjs
  warning-code-mismatch      51    (--no-fmt run: 51)
  warning-position-mismatch  529   (--no-fmt run: 529)
  js-mismatch                0     (--no-fmt run: 11)
  ✅ no warning regressions (580 known warning failures remain)
  exit 0
```

That pair is its own positive control: the flag demonstrably moved the output family (11 → 0, exactly the formatting false positives `--no-fmt` is documented to produce) and left the warning family bit-identical. Warning tuples are formatter-independent by construction — they are read from `warnings.json`, written by `compile.mjs` before verify runs and parsed with `JSON.parse`.

## CI status at time of writing

**No check has run on this rebased head.** GitHub Actions was in `major_outage`; the repo queue stood at 42 queued / 1 in progress and none of this PR's 31 checks had dequeued. The pre-rebase head `5344902d` was fully green including `CodSpeed Performance Analysis` (`Performance Gate Passed`), and this PR carries no known performance concern — but that is the *previous* head, and it should not be read as this one having passed.

The branch is also one commit behind `main` (`73aef74f`, legacy SSR bracket scanners). That commit touches only `3_transform/server/`, so it cannot move warning tuples, which are produced in Phases 1-2 — but the ratchets here were measured against `a8a653d5`.


baseballyama:
Reviewed at `5e08bdc1`. **0 blocking, 1 nit that matters more than usual.** Both upstream guards check out, the four sites really do collapse to one gate, and the message text now matches upstream byte for byte — but the only thing in the repo that can observe the wording asserts a prefix of it. Read-only — no build, no test run. CI: 1 pass, 30 pending, no failures.

### Both upstream callers are guarded, and there are exactly two

`validate_attribute` has two call sites in the whole upstream visitor tree, and both are inside `if (context.state.analysis.runes)`:

- `shared/component.js:92-93`
- `shared/element.js:42-43`

So legacy mode never reaches the check, and the gate belongs on the emission rather than at the call sites. Verified by enumerating callers, not by reading the two the PR names — a third caller outside the guard would have inverted the conclusion.

### Four sites, one gate — confirmed

At this head there is exactly **one** direct emitter of `warnings::attribute_quoted()` left, at `shared/attribute.rs:69` inside `warn_attribute_quoted`. `regular_element.rs`, `shared/component.rs`, `svelte_component.rs` and `svelte_self.rs` all route through it, and the duplicated `is_quoted_single_expression` in `component.rs` is deleted rather than left as a second definition. That is the shape that makes the gate hold: not four copies of the same condition, but one predicate with no way around it.

Folding the span assignment into the same helper is right too — the warning now carries the attribute's own start/end at the single point it is raised, which is the same principle #2384 applies to the a11y codes.

### The nit: the sole observer of the wording observes 40% of it

The message is corrected here — `warnings.rs` goes from "Quoted attribute **values** will be stringified…" to "Quoted attributes **on components and custom elements** will be stringified…", which is byte-identical to upstream `warnings.js:732`, `\nhttps://svelte.dev/e/attribute_quoted` suffix included. Good catch, and it is not a cosmetic one: the old text described a rule that does not exist.

But the corpus gate compares `(code, line, column)` **only** — by construction it cannot see a message at all — so `attribute_quoted_runes_only.rs` is the single thing in the repo that observes this string. And it asserts:

```rust
.starts_with("Quoted attributes on components and custom elements will be stringified")
```

That pins the first nine words. The clause that carries the actual advice — "in a future version of Svelte. If this isn't what you want, remove the quotes" — and the docs URL are unobserved by anything: change either and every gate in the repo stays green. Given that a wrong message is exactly the defect this PR is fixing, the one test that can see it should see all of it:

```rust
assert_eq!(
    w.message,
    "Quoted attributes on components and custom elements will be stringified in a future version of Svelte. If this isn't what you want, remove the quotes\nhttps://svelte.dev/e/attribute_quoted"
);
```

(If the public `Warning.message` strips the URL at the API boundary, assert the full sentence and drop the suffix — the point is that the assertion should end where the observable string ends, not nine words in.)

This is the general shape worth keeping: when a gate is blind to a field **by construction**, the unit test is not a convenience, it is the only oracle, and a partial assertion there leaves the field genuinely unchecked. Same reasoning as the a11y carve-out on #2384 — the one judgment call needs the one test that can see it.

### Tests otherwise

Four legacy negatives across all four node kinds (`Component`, `svelte:component`, `svelte:self`, custom element) plus two runes positives is the right matrix: the negatives fail before the fix and the positives fail if the fix suppresses the warning outright instead of gating it. The `runes_component_attribute_is_still_flagged` docstring says exactly that — "the half of the check that stays" — which is the distinction that keeps a gating change from silently becoming a deletion.

### Ratchet, and the merge order

Counts reconcile at this head: all three code files hold **51** entries, matching "the same 51 entries" and "51 entries each" in the md.

One thing to watch, and it is the same sentence in both PRs. `warning-known-failures.md:38` reads "the same 51 entries; the three position files hold the same **625**". #2384 rewrites the *other* number on that same line (625 → 529) while leaving 70. So the two PRs edit one line from opposite ends, and the merged truth is **51 and 529** — a pair neither PR's run produced. Landing #2384 first and then rebasing this one and regenerating is the only sequence that ends with a number a run actually measured; resolving the conflict by hand would produce a line that looks right and was never measured, which is precisely the failure mode the four corrections shipped today were cleaning up.

Issue #2381, opened 2026-08-06
esrap: the comment-bearing print path emits a redundant paren pair around a parenthesized `await`


baseballyama:
`rsvelte_esrap`'s comment-bearing print path emits one redundant paren pair around an `await` whose printed form needs parens. Pre-existing; **#2372 is what makes it observable**, because until that PR the comment was dropped and the comment-bearing path was never taken for this shape.

## Repro

```svelte
<script module>
	export async function f() {
		return await /* hi */ load();
	}
</script>
<p>hi</p>
```

`generate: 'client'`, `dev: true`.

```js
official: return (await $.track_reactivity_loss(/* hi */ load()))();
rsvelte : return ((await $.track_reactivity_loss(/* hi */ load())))();
```

## It is the comment, not the shape

Same input with the comment removed is correct:

```js
return (await $.track_reactivity_loss(load()))();   // both compilers
```

So the extra pair appears exactly when `to_oxc` produces a `comment_source` and the program prints through `rsvelte_esrap::print_split` rather than `print_with`. Statement position (`await load();` with no `return`) is also correct.

## It is pre-existing, not caused by the module routing

Reproduced on the **component** `<script module>` path, which already routes through `to_oxc` → `rsvelte_esrap` on `origin/main` — with #2380 (the `.svelte.(js|ts)` routing change) **not** applied. #2380 does not introduce it; it will make the same defect reachable from `.svelte.(js|ts)` modules once both land.

## Where to look

`parse_chunk` parses with default `ParseOptions` (`preserve_parens: true`), so the re-parsed text carries an explicit `ParenthesizedExpression`. `unparen` (`crates/rsvelte_esrap/src/printer.rs:197-205`) is supposed to strip those before printing, since acorn elides source parens and official esrap never sees them. The printer then re-adds parens because an `AwaitExpression` in callee position requires them. One of those two steps is running twice on the `print_split` path.

Note `restore_legacy_pre_effect_deps` (`to_oxc.rs:1190-1195`) documents the stripping as load-bearing in the other direction — it rebuilds a sequence *because* the printer drops the parens — so the two paths disagree about whether a re-parsed `ParenthesizedExpression` survives. That disagreement is the likely seam.

## Gate visibility

[...]

PR #2392, opened 2026-08-06
fix(compiler): keep a comment between `await` and its operand in a runes instance script (#2375)

### LINKED ISSUE #2375: compiler (client, dev): component scripts drop a comment inside the `track_reactivity_loss` wrap, by two separate mechanisms
> [!WARNING]
> **Mechanism A below is RETRACTED by its author — do not build on it.** On #2372's
> branch a component `<script module>` now *keeps* the comment (with a redundant paren
> pair, tracked separately as #2381), which contradicts the "lost downstream" claim
> recorded here. I could not reconcile that against the earlier failing-test observation
> the section was written from, so **mechanism A is unverified**: re-measure it on
> current `main` before treating any of it as a starting point. Mechanism B (below)
> stands — it was established by instrumentation that produced no output at all, which
> is not subject to the same ambiguity.

Split out of #2090. #2372 fixed the `.svelte.(js|ts)` module path; the three **component** script kinds still drop the comment, through **two different mechanisms**, neither of which is the pass #2372 changed.

## Repro

```svelte
<script>
	async function f() {
		return await /* hi */ load();
	}
	let x = $state(1);
</script>
<p>{x}</p>
```

`generate: 'client'`, `dev: true`. Official:

```js
return (await $.track_reactivity_loss(/* hi */ load()))();
```

rsvelte:

```js
return (await $.track_reactivity_loss(load()))();
```

Same for a legacy instance script (`export let a = 1` instead of the rune) and for `<script module>`. Official keeps the comment in all four script kinds; measured, not inferred.

The divergence **survives the corpus normalizer** (`flattenTemplateHoles` → oxfmt 0.62 → `stripBlankLines`) — checked against `compatibility/.oxfmtrc.json`, so `client-dev` can see it. The other three shapes #2090 listed alongside it do not survive; see the measurement table in https://github.com/baseballyama/rsvelte/issues/2090#issuecomment-5203976965.

## Mechanism A — `<script module>` and the legacy instance script: lost downstream ⚠️ RETRACTED, see warning above

These do reach `await_reactivity_loss_ast.rs`. Instrumented, the collector now hands back the **correct** text post-#2372:

```
[collector output] return (await $.track_reactivity_loss(/* hi */ load()))();
```

and the comment is gone from the final output. So the loss is strictly downstream of that pass, somewhere in the component script pipeline — a pass that reaches component scripts but not `.svelte.(js|ts)` modules, since the identical collector output survives intact there.

Note comments in general survive a component instance script (`let y = /* hi */ 1;` is kept), so this is not blanket comment stripping.

## Mechanism B — the runes instance script: never reaches that pass

Instrumenting `collect_await_reactivity_loss_edits` produces **no output at all** for a runes instance script. It goes through `visitors/expression_converter.rs:477`, which builds the wrap as IR nodes (`JsExpr::Call` / `JsExpr::Await` / `JsExpr::Member`) rather than splicing source text. IR nodes carry no comments, so there is nothing to lose — the comment never enters the representation.

The module doc-comment on `await_reactivity_loss_ast.rs` says it is "shared by every script kind that reaches it through a batched source rewrite"; the runes instance script is the kind that does not.

## Relationship to #2336

Mechanism B is the same territory as the comment-cursor epic: a synthesized IR subtree with no source position, where upstream's node keeps its `loc` and esrap flushes the comment positionally. Whoever picks this up should first decide whether B is a **consumer** of #2336 (fixed for free once IR nodes carry positions / the cursor discipline is ported) or independent of it. Mechanism A's relationship cannot be assessed until it is re-measured (see the warning at the top); the "text pass eating an already-correct string" reading was based on the retracted observation.

Do not attempt B in isolation without reading #2336's "why the one-line fix does not work" section.

## Scope

Depends on nothing; #2372 is orthogonal and already landed the module half. Parent: #2090 (which has nothing else left in it). Related: #2336, #2308.

Measured at `origin/main` = `88b8d2b7` plus #2372, against the pinned `submodules/svelte`.

baseballyama:
Fixes #2375 (partially — see "What this does not fix").

## What was actually wrong

The issue records **two mechanisms**, and I retracted mechanism A myself before
starting. Re-measured at `origin/main` = `cf911651` by instrumenting every
`$.track_reactivity_loss` emission site, the real picture is:

| script kind | path taken | status |
|---|---|---|
| legacy instance | `instance_dev_tail_ast` → `await_reactivity_loss_ast` collector | fixed by **#2372** (already open) |
| `<script module>` | `module_dev_tail_ast` → same collector | fixed by **#2372** |
| **runes instance** | **`ast_state_transform::try_rewrite_await_reactivity_loss`** | **fixed here** |

Two corrections to the issue:

- **Mechanism A was not "lost downstream".** The collector's own edit already
  lacked the comment; the loss was *inside* the pass #2372 changed, not after
  it. Applying #2372's bound to `origin/main` makes both those kinds keep the
  comment, confirmed by direct measurement.
- **Mechanism B named the wrong site.** `visitors/expression_converter.rs:477`
  never executes for any of the three kinds — an `eprintln!` there produced no
  output. B's *observation* (the runes script never reaches the collector) is
  correct; the wrap is built by a **second copy** of the same logic at
  `ast_state_transform.rs:2028`, which the issue does not mention.

So it is one defect duplicated across two files, not two mechanisms.

## The fix

`try_rewrite_await_reactivity_loss` sliced the operand from
`expr.argument.span()`, which begins *after* any comment separating the operand
from `await`. It now starts just past the keyword — the same bound #2372 uses
in the collector. Widening only the start is safe because
`apply_and_drain_inner_replacements` re-bases the drained inner replacements on
whatever start it is given.

Upstream reference:
`submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/AwaitExpression.js:11-21`
— upstream hands `b.call('$.track_reactivity_loss', argument)` the *visited
node*, so comments ride along with its `loc`; rsvelte copies source text, which
is why the bound is load-bearing.

## Tests

`crates/rsvelte_core/tests/track_loss_comment_runes_instance_2375.rs`.
Run against unfixed code, three of the four fail with the real divergence:

```
return (await $.track_reactivity_loss(load()))();     // rsvelte
return (await $.track_reactivity_loss(/* hi */ load()))();   // official
```

`an_ignored_await_is_left_alone` passes on both sides — it is a guard for a
**different** property (the `svelte-ignore await_reactivity_loss` gate), which
is what kills it; it does not discriminate the copy bound.

## What this does not fix

Once the comment is kept, this shape reaches `rsvelte_esrap`'s comment-bearing
print path and picks up redundant parens — **#2381**:

```
official: return (await $.track_reactivity_loss(/* hi */ load()))();
rsvelte : return (((await $.track_reactivity_loss(/* hi */ load()))))();
```

Note #2381's repro shows **two** paren pairs on the `<script module>` path; the
runes path shows **three**. The fix for that lives in
`crates/rsvelte_esrap/src/printer.rs`, which another agent holds, so this PR
touches only `client/ast_state_transform.rs`. The assertions here pin the
comment's placement inside the call and deliberately stop short of the paren
count, with the upstream bytes written out in the test file so the gap is
explicit and greppable.

## Gate visibility

Measured through the real normalizer (`flattenTemplateHoles` → oxfmt 0.62.0 →
`stripBlankLines`, `compatibility/.oxfmtrc.json`, with a control confirming
oxfmt rewrote both sides):

| | vs official |
|---|---|
| before this PR | **DIFFER** — the gate sees it |
| after this PR | **MATCH** |

So `client-dev` can see this change. Worth stating plainly: the residual #2381
parens are *absorbed* by oxfmt, so merging this before #2381 converts a
gate-visible divergence into a gate-invisible one. `client-dev` is at 0 known
failures, which means no corpus entry carries this shape at all — consistent
with it being an interaction shape the collected corpus does not sample.

### DIFF
--- a/.changeset/runes-await-comment.md
+++ b/.changeset/runes-await-comment.md
@@ -1,0 +1,10 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): keep a comment between `await` and its operand in a runes instance script
+
+Dev-mode client output wrapped `await X` as `(await $.track_reactivity_loss(X))()`
+by copying the operand from the argument's own span, which begins past any
+comment separating it from the `await` keyword. The copy now starts just past
+the keyword, matching what upstream preserves by passing the visited node.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
@@ -2022,8 +2022,12 @@ impl<'a, 's> StateVarCollector<'a, 's> {
         // Walk the argument so inner state-var refs (and nested awaits)
         // register their replacements, then drain them into the argument text.
         self.visit_expression(&expr.argument);
-        let arg_span = expr.argument.span();
-        let arg_text = self.apply_and_drain_inner_replacements(arg_span.start, arg_span.end);
+        // Copy from just past the `await` keyword, not from the argument's own
+        // start: the trivia between them holds comments upstream keeps inside
+        // the call. Widening only the start is safe because the drained inner
+        // replacements are re-based on whatever start is passed here.
+        let arg_start = expr.span.start + "await".len() as u32;
+        let arg_text = self.apply_and_drain_inner_replacements(arg_start, expr.span.end);
 
         let wrap = format!("(await $.track_reactivity_loss({}))()", arg_text.trim());
         // The `;` rides inside this replacement rather than being appended to
--- a/crates/rsvelte_core/tests/track_loss_comment_runes_instance_2375.rs
+++ b/crates/rsvelte_core/tests/track_loss_comment_runes_instance_2375.rs
@@ -1,0 +1,77 @@
+//! A comment between `await` and its operand must survive the dev-mode
+//! `(await $.track_reactivity_loss(X))()` wrap in a **runes** instance script.
+//!
+//! The runes script does not reach `await_reactivity_loss_ast`'s collector; it
+//! is rewritten by `ast_state_transform::try_rewrite_await_reactivity_loss`,
+//! which copied the operand from the argument's own span and so began past any
+//! trivia the `await` keyword was separated from it by. Expectations are the
+//! official compiler's bytes.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+fn dev_client(src: &str) -> String {
+    compile(
+        src,
+        CompileOptions {
+            filename: Some("T.svelte".into()),
+            generate: GenerateMode::Client,
+            dev: true,
+            ..Default::default()
+        },
+    )
+    .expect("compile")
+    .js
+    .code
+}
+
+fn runes(body: &str) -> String {
+    dev_client(&format!(
+        "<script>\n{body}\tlet x = $state(1);\n</script>\n<p>{{x}}</p>"
+    ))
+}
+
+// The assertions below pin the comment's placement inside the call — the
+// property this pass owns — and stop short of the paren pairs around the
+// `await`. Official emits `return (await $.track_reactivity_loss(/* hi */
+// load()))();` with one pair; the comment-bearing esrap print path emits three
+// here, which is #2381 in `rsvelte_esrap`, reachable only once a comment is
+// kept at all.
+#[test]
+fn a_block_comment_before_the_operand_is_kept() {
+    let out = runes("\tasync function f() {\n\t\treturn await /* hi */ load();\n\t}\n");
+    assert!(
+        out.contains("$.track_reactivity_loss(/* hi */ load())"),
+        "got: {out}"
+    );
+}
+
+#[test]
+fn every_comment_in_the_run_is_kept() {
+    let out = runes("\tasync function f() {\n\t\treturn await /* a */ /* b */ load();\n\t}\n");
+    assert!(
+        out.contains("$.track_reactivity_loss(/* a */ /* b */ load())"),
+        "got: {out}"
+    );
+}
+
+#[test]
+fn a_line_comment_survives() {
+    let out = runes("\tasync function f() {\n\t\treturn await // hi\n\t\t\tload();\n\t}\n");
+    assert!(
+        out.contains("// hi"),
+        "the line comment was dropped; got: {out}"
+    );
+}
+
+/// Discriminates the `svelte-ignore` gate, not the copy bound: with the gate
+/// removed this shape gains a wrap it must not have.
+#[test]
+fn an_ignored_await_is_left_alone() {
+    let out = runes(
+        "\tasync function f() {\n\t\t// svelte-ignore await_reactivity_loss\n\t\treturn await /* hi */ load();\n\t}\n",
+    );
+    assert!(
+        !out.contains("track_reactivity_loss"),
+        "the ignored await was wrapped; got: {out}"
+    );
+}


baseballyama:
Reviewed at `b71ade8a`. **0 blocking.** Both retractions hold, and the duplication claim — the one that decides whether #2375 can close — is verifiable without running anything. One correction to how Mechanism B is described. Read-only — no build, no test run. CI: 1 pass, 30 pending, no failures.

### Mechanism A: the retraction is right, and it is structural

The issue said the collector returns correct text for `<script module>` and the legacy instance script and the comment is lost downstream. On `origin/main` the collector slices `expr.argument.span()` — which by definition starts *after* any trivia between the `await` keyword and its operand. The comment is therefore absent from the collector's own output; there is nothing downstream to lose. No instrumentation needed: the bound is the mechanism, and it visibly excludes the region the comment occupies.

And the routing confirms which script kinds that covers. `instance_dev_tail_ast.rs:98` and `module_dev_tail_ast.rs:72` both call `await_reactivity_loss_ast::collect_await_reactivity_loss_edits` — the collector #2372 changes. So the legacy instance script and the module path share that one bound, and #2372's fix reaches both. The three-row table is accurate on this axis.

### Mechanism B: the site is real, but "never executes" is the wrong reason

The second copy is real, and it is the one that matters: `ast_state_transform.rs:2028`, in `try_rewrite_await_reactivity_loss`, reached for the **runes** instance script via `:2406`. That is the path this PR fixes, and the issue never names it.

The two copies are genuinely duplicated logic, not superficially similar. Side by side:

| | `await_reactivity_loss_ast.rs` (#2372) | `ast_state_transform.rs` (this PR) |
|---|---|---|
| start | `expr.span.start + "await".len()` | `expr.span.start + "await".len()` |
| end | `expr.span.end` | `expr.span.end` |
| text | direct source slice | `apply_and_drain_inner_replacements` |
| wrap | `(await $.track_reactivity_loss({}))()` | `(await $.track_reactivity_loss({}))()` |
| trim | yes | yes |

Same bound arithmetic, same template, same trim — the only difference is how the text is obtained, and both had the same pre-fix bug (starting at the argument). **One defect duplicated across two files** is the correct reading, and the two now agree byte for byte on the bound, including the `expr.span.end` end that #2372 only reached after review. Fixing the second copy to match the first's *corrected* form rather than its original is the right call and is worth a line in the body, because it is what stops the two drifting again.

**Where I would change the wording**: `expression_converter.rs:477` is not a site that never executes — it is a site that cannot exhibit this defect. It builds the wrap as **IR nodes** (`JsExpr::Call` / `JsExpr::Await` / …) inside `convert_js_node`'s `JsNode::AwaitExpression` arm, i.e. the typed template-expression path. It is not a text splice, so there is no copy bound to get wrong, and it carries no comments by construction.

That distinction matters for two reasons:

1. **An `eprintln!` reporting zero output is bounded by the inputs exercised**, exactly like the probe on #2394 — and the fixtures here are instance scripts. A template-level `await` (`{await /* c */ p}` in a runes component) would plausibly reach that arm. So "produced no output" supports "not reached by these tests", not "dead".
2. If it *is* reachable, it drops the comment for a **different** reason — comments never attach to IR nodes — which is a third instance of the class, not covered by #2372 or by this PR. That bears directly on whether #2375 can close: two of three paths are fixed, and the third is unmeasured rather than absent.

One input settles it. If `{await /* c */ p}` in a runes template keeps the comment, the arm is either unreachable or harmless and #2375 closes; if it drops it, that is the remaining row. Cheaper than the instrumentation already done, and it converts the weakest cell in the table into a measured one.

### Tests

Three positives plus a correctly-labelled gate control, and the label on `an_ignored_await_is_left_alone` says exactly what it discriminates ("the `svelte-ignore` gate, not the copy bound") — which is the distinction that stops it being counted as a fourth repro.

The comment above the assertions is the part I would keep verbatim: pinning the comment's placement inside the call while deliberately stopping short of the paren pairs, with the reason (#2381, reachable only once a comment is kept at all) and the official byte sequence both stated. That is an assertion that knows its own boundary — it pins the property this pass owns and does not silently absorb a printer defect that belongs to another crate. The alternative, asserting the full official string, would have forced either a wrong expectation or a blocked PR.

PR #2394, opened 2026-08-06
fix(compiler): guard member properties in the legacy `$:` compound-assignment scan (#2373)

### LINKED ISSUE #2373: Legacy SSR: `extract_simple_assignments` skips the member-property guard for compound assignment
Split out of the #2357 scan sweep (batch 4, #2370), where it was found but deliberately not fixed.

`extract_simple_assignments` in `crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs` records the identifiers a legacy `$:` statement assigns to. It guards against member properties in two of its three branches but not the third:

| form | `is_member_prop` checked? | `foo.x` records… |
|---|---|---|
| `foo.x = 1` | yes | nothing |
| `foo.x++` / `foo.x--` | yes | nothing |
| `foo.x += 1` | **no** | `x` |

The guard exists because a member assignment mutates the *base object*, not a declared variable — recording the property invents a reactive dependency for any statement that reads an identifier of that name, which reorders otherwise-independent `$:` statements away from source order. That reasoning applies identically to compound assignment, so the omission looks like an oversight rather than a decision.

```rust
extract_simple_assignments("foo.x = 1")  // []
extract_simple_assignments("foo.x += 1") // ["x"]  <-- inconsistent
```

**Why it was not fixed in #2370.** The function feeds `extract_reactive_lhs_vars` → `sort_reactive_statements_topologically`, so removing `x` from the set here deletes a dependency edge and changes which `$:` statements get reordered. That is a semantic change to emitted output, not the lexical cleanup #2357 is about, and it needs corpus evidence rather than a unit test: the right gate is whether any corpus entry's `client`/`server` output moves when the guard is added to the compound branch.

Note the denominator problem when measuring: the corpus is overwhelmingly runes-mode and legacy `$:` density in modern repos is near zero, so an empty diff is weak evidence and should be reported alongside the count of entries that actually contain a `$:` statement with a compound member assignment.

baseballyama:
Fixes #2373.

## What

`extract_simple_assignments` in `crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs`
applied its `is_member_prop` guard to the `=` and `++`/`--` branches but not to
the compound-assignment branch, so `$: obj.x += 1` claimed to assign `x`:

```rust
extract_simple_assignments("obj.x = 1")  // []
extract_simple_assignments("obj.x += 1") // ["x"]  <-- before
extract_simple_assignments("obj.x += 1") // []     <-- after
```

The recorded name becomes a node in `sort_reactive_statements_topologically`'s
declared-variable map, so any `$:` statement reading a plain `x` gained a
dependency edge on the member statement and was pushed below it.

## Upstream reference

`submodules/svelte/packages/svelte/src/compiler/phases/2-analyze/visitors/AssignmentExpression.js`
lines 10-24. The visitor does **not** branch on `node.operator` — every
assignment operator takes the same path — and the target goes through
`extract_identifiers(node.left)`, whose `unwrap_pattern`
(`utils/ast.js:89-138`) pushes a `MemberExpression` and then filters it out by
`node.type === 'Identifier'`. So upstream records nothing into
`reactive_statement.assignments` for a member target under any operator, and
`order_reactive_statements` (`phases/2-analyze/index.js:1273-1334`) therefore
never builds an edge from it.

Verified against the pinned official compiler (5.56.8): for
`$: a = x * 2; $: obj.x += 1;` official SSR keeps source order.

## Tests

**Discriminating** (unit tests in `transform_legacy.rs`; both fail on the
pre-fix function, output below):

- `member_assignment_targets_are_not_recorded` — kills the mutant that drops
  `!is_member_prop` from the compound branch.
  ```
  assertion failed: extract_simple_assignments("obj.x += 1;").is_empty()
  ```
- `compound_member_assignment_does_not_reorder` — same mutant, at the level of
  the observable consequence.
  ```
    left: "\tlet a = 1;\n\t$: obj.x += 1;\n\t$: a = x * 2;"
   right: "\tlet a = 1;\n\t$: a = x * 2;\n\t$: obj.x += 1;"
  ```

**Discriminating in the opposite direction** (they fail if the compound branch
is disabled outright rather than guarded — verified by running that mutant):

- `plain_identifier_assignment_targets_are_recorded` (`left: []`, `right: ["x"]`)
- `compound_identifier_assignment_still_reorders`

**Guards, explicitly labelled as such** in
`crates/rsvelte_core/tests/legacy_reactive_member_compound_2373.rs`: they pass
before and after. The mechanism that makes them inert is that SSR reactive
ordering goes through the AST port of `order_reactive_statements` in
`server/ast/script.rs`, not through the text scanner this PR changes. They pin
that the live path keeps agreeing with upstream for `obj.x = …`, `obj.x += …`,
`obj.x++` and `x += …`.

## Found but not fixed

1. **The changed text path is dead code in the shipping pipeline.**
   `extract_simple_assignments` is reachable only via
   `extract_reactive_lhs_vars` → `sort_reactive_statements_topologically` →
   `sort_reactive_in_place` → `reorder_reactive_statements_after_functions`,
   whose only caller is `transform_script.rs:266` inside
   `transform_script_content_inner`; that function's only entry
   (`transform_script_content_with_imports_and_derived`) has exactly one caller,
   `server/ast/visitors/declaration_tag.rs:346`, which passes a single
   declaration-tag body — where a top-level `$:` cannot appear. Evidence: an
   instrumented build printing on `has_reactive == true` produced **0** hits
   across the 18 `rsvelte_core` integration suites that contain `$:` sources (42
   tests), with a positive control confirming the probe fires when the function
   is called directly. That is why this PR claims no output change; it also
   means #2373's stated blocker (needing corpus evidence that `client`/`server`
   output moves) no longer applies. Deleting the ~450 lines of now-unreachable
   legacy text reorder machinery is a separate change and is left out
   deliberately.
2. **`$: obj.x++` diverges from upstream in the same function, in the opposite
   direction.** Upstream's `UpdateExpression` visitor uses
   `object(node.argument)` and records the *base object* (`obj`), whereas the
   `=`/`+=` path records nothing. rsvelte records nothing in all three cases.
   Left alone: it is a different defect, adds edges rather than removing them,
   and is unobservable for the same reachability reason.

## Checks

- `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings` clean.
- `cargo test -p rsvelte_core --lib`, `--test ssr`, `--test compiler_fixtures`,
  `--test compiler_errors`, `--test runtime` (release), plus the 18 `$:`-bearing
  integration suites.

### DIFF
--- a/.changeset/wild-pugs-brush.md
+++ b/.changeset/wild-pugs-brush.md
@@ -1,0 +1,15 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Apply the member-property guard to compound assignment in the legacy server
+`$:` reorder scanner. `extract_simple_assignments` recorded `x` for
+`$: obj.x += 1` while recording nothing for `$: obj.x = 1` and `$: obj.x++`,
+which invented a reactive dependency and hoisted the statement above any `$:`
+that reads a plain `x`. Upstream's `AssignmentExpression` visitor takes the same
+branch for every operator and records no target for a member expression.
+
+No change to emitted output: the text scanner is reachable only from the
+declaration-tag script path, where a `$:` statement cannot occur, and SSR
+reactive ordering runs through the AST port of `order_reactive_statements`,
+which was already correct for these shapes.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -1559,8 +1559,8 @@ fn extract_simple_assignments(code: &str) -> Vec<String> {
             }
             let ident: String = chars[start..i].iter().collect();
 
-            // A member property (`foo.x = …` / `foo.x++`) is not a declared
-            // variable: the assignment mutates the *base object*, not the
+            // A member property (`foo.x = …` / `foo.x += …` / `foo.x++`) is not
+            // a declared variable: the assignment mutates the *base object*, not the
             // property. Recording the property would create a false reactive
             // dependency for any statement that reads an identifier of that
             // name (e.g. `$: { if (x) … }` spuriously depending on
@@ -1618,7 +1618,11 @@ fn extract_simple_assignments(code: &str) -> Vec<String> {
                 if matches!(op, '+' | '-' | '*' | '/' | '%' | '&' | '|' | '^') {
                     // Check it's not `==` following
                     let after_eq = chars.get(j + 2).copied().unwrap_or('\0');
-                    if after_eq != '=' && !is_reactive_keyword(&ident) && !vars.contains(&ident) {
+                    if after_eq != '='
+                        && !is_member_prop
+                        && !is_reactive_keyword(&ident)
+                        && !vars.contains(&ident)
+                    {
                         vars.push(ident.clone());
                     }
                 }
@@ -2254,4 +2258,43 @@ fn split_destructuring_properties_ssr(s: &str) -> Vec<&str> {
     }
     result.push(&s[start..]);
     result
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn member_assignment_targets_are_not_recorded() {
+        // Upstream's `AssignmentExpression` visitor takes the same branch for
+        // every operator, and `extract_identifiers` yields nothing for a
+        // member-expression target, so no operator records the property.
+        assert!(extract_simple_assignments("obj.x = 1;").is_empty());
+        assert!(extract_simple_assignments("obj.x += 1;").is_empty());
+        assert!(extract_simple_assignments("a.b.c *= 2;").is_empty());
+        assert!(extract_simple_assignments("obj.x++;").is_empty());
+    }
+
+    #[test]
+    fn compound_member_assignment_does_not_reorder() {
+        let out = reorder_reactive_statements_after_functions(
+            "\tlet a = 1;\n\t$: a = x * 2;\n\t$: obj.x += 1;\n",
+        );
+        assert_eq!(out, "\tlet a = 1;\n\t$: a = x * 2;\n\t$: obj.x += 1;");
+    }
+
+    #[test]
+    fn compound_identifier_assignment_still_reorders() {
+        let out = reorder_reactive_statements_after_functions(
+            "\tlet a = 1;\n\t$: a = x * 2;\n\t$: x += 1;\n",
+        );
+        assert_eq!(out, "\tlet a = 1;\n\t$: x += 1;\n\t$: a = x * 2;");
+    }
+
+    #[test]
+    fn plain_identifier_assignment_targets_are_recorded() {
+        assert_eq!(extract_simple_assignments("x = 1;"), vec!["x"]);
+        assert_eq!(extract_simple_assignments("x += 1;"), vec!["x"]);
+        assert_eq!(extract_simple_assignments("x++;"), vec!["x"]);
+    }
 }
--- a/crates/rsvelte_core/tests/legacy_reactive_member_compound_2373.rs
+++ b/crates/rsvelte_core/tests/legacy_reactive_member_compound_2373.rs
@@ -1,0 +1,101 @@
+//! Upstream parity pins for `$:` statements whose only assignment target is a
+//! member expression. A member assignment mutates the *base object*, so
+//! upstream records no assignment target at all for it and leaves the statement
+//! in source order — regardless of the operator.
+//!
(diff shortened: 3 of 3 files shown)


baseballyama:
Reviewed at `8834a393`. **0 blocking.** I verified the unreachability claim link by link rather than accepting the chain, and it holds — with a stronger corroboration than the instrumented build. Read-only — no build, no test run. CI: 31 pending, no failures.

### The call chain, verified

Every link checked by enumerating callers at this head, not by following the PR's write-up:

- `extract_simple_assignments` — one production caller, `transform_legacy.rs:1456`, inside `extract_reactive_lhs_vars` (`:1447`).
- `extract_reactive_lhs_vars` — one caller, `:1887`, inside `sort_reactive_statements_topologically` (`:1875`).
- that function — two callers, `:1250` and `:1412`, both inside `reorder_reactive_statements_after_functions` (`:899`) and `sort_reactive_in_place` (`:1280`).
- `reorder_reactive_statements_after_functions` — one production caller, `transform_script.rs:266`, inside `transform_script_content_inner` (`:37`), guarded by `!is_module`.
- that function — one caller, `transform_script_content_with_imports_and_derived` (`:20`) at `:26`.
- and that — **one caller in the entire workspace**: `declaration_tag.rs:346`.

No branch of that tree has a second production entry point. Everything else the grep returns is a `#[cfg(test)]` call.

### The terminus really is `$:`-free

`declaration_tag.rs:346` sits in `try_async_declaration_tag`, whose input is a `DeclarationTag` — a `{@const …}` — with the tag braces stripped and a `;` appended (`:317-332`). The argument is the declaration's own source text, and it is passed with `is_module = false`, which is what keeps the `!is_module` branch live. A top-level labelled statement cannot appear there: `{@const}` accepts a declaration, and a `$:` label is not one, so the parser never produces this shape.

**Independent corroboration, which I find more convincing than the probe.** `AGENTS.md` states that server SSR is pure-AST and *the legacy text generator is deleted*. That is the same fact from the other end: what survives in `transform_legacy.rs` is whatever the `{@const}` lowering still calls, and the `$:` reorder machinery is residue of the deleted path. The instrumented 0-of-42 result and the positive control are good practice, but they measure the tests; the caller graph and the architecture note together establish it for all inputs.

### The upstream derivation goes through the mechanism

Checked in `2-analyze/visitors/AssignmentExpression.js`:

```js
const id = node.left.type === 'MemberExpression' ? object(node.left) : node.left;
if (id !== null) {
	for (const id of extract_identifiers(node.left)) { … }
}
```

The function body **never reads `node.operator`** — there is no branch on it anywhere, so `=`, `+=` and `*=` are literally the same code path. And `extract_identifiers` is `unwrap_pattern(pattern, []).filter((node) => node.type === 'Identifier')` (`utils/ast.js:136-138`), so for a `MemberExpression` left the single unwrapped node is filtered out and the loop body never executes. The outer `id` (via `object()`) is non-null for `obj.x`, so it gates entry and then the inner binding shadows it — nothing is added to `assignments`.

That is the right derivation: it holds for **every** operator because the operator is never consulted, not because three operators were sampled. Worth keeping in the changeset in that form.

### What this changes beyond the diff

If the chain is right — and I believe it is — then the `$:` reorder machinery in `transform_legacy.rs` is unreachable in the shipping pipeline, and that reframes work already done today. My blocking finding on #2366 (the leading-guard exits truncating a `$:` continuation into an unterminated template literal) is **correct about the code and wrong about the impact**: it cannot occur in a shipping compile, because the only entry is a `{@const}` body. I would not withdraw the fix — it removes a trap in code that is still `pub(crate)` and still compiled — but it should be re-labelled from "emits non-parsing output" to "would, if the path were ever revived", and it is not the priority I gave it.

The durable question is the one this PR raises rather than answers: whether the `$:` half of `transform_legacy.rs` should be deleted rather than maintained. Three PRs today have carried review and CI cost on it. If it is deleted, both this guard and #2366's exits go with it; if it is kept, it wants a comment at `reorder_reactive_statements_after_functions` saying what still reaches it, because the next person will re-derive this chain from scratch. Either way the finding is worth an issue of its own — it is larger than a one-condition diff.

### Tests

`compound_member_assignment_does_not_reorder` and `compound_identifier_assignment_still_reorders` are a proper pair — same shape, one operand differing, opposite expected outcomes — and they call `reorder_reactive_statements_after_functions` directly, which is the only way to observe this code at all now.

The integration file is labelled correctly and unusually carefully: "**These are guards, not discriminating tests**", with the reason (the live AST path already agreed) stated in the module doc rather than left for a reader to discover. Given that four of its five cases pass before and after, that header is what stops the file from reading as four repros. That is the standard this epic has been converging on, applied without being asked.

PR #2398, opened 2026-08-06
chore(compiler): give byte and char offsets distinct types (#2386)

baseballyama:
Slice 1 of #2386. **No behavioural change** — 1420 unit tests unchanged, and the diff is two signatures plus their call sites.

## What lands

`shared/offsets.rs`: `ByteOffset`, `CharOffset`, and `CharToByte` (an O(1) char→byte table for one string). Deliberately **no `From`/`Into` between the two** — an implicit conversion restores exactly the silence being removed, so every crossing is a named method and stays greppable.

Converted in `client/destructure_transforms.rs`:

| function | unit | was |
|---|---|---|
| `find_matching_open_bracket` | bytes in and out | `usize`, unit stated in a doc comment |
| `find_destructure_rhs_end` | chars in and out | `usize`, unit stated in a field comment |

That file was chosen because it holds **both** in one place: a byte-offset scanner and a char-offset scanner, with `Candidate` carrying char indices between them. The crossing at `:224` — where a char index is converted before being handed to the byte-based helper — is exactly the shape that produced #2369, #2376 and #2385.

## The control moved

A newtype whose rejection is never demonstrated is an assumption. I applied the mis-use in both directions and rebuilt:

```
error[E0308]: expected `ByteOffset`, found `CharOffset`
error[E0308]: expected `CharOffset`, found `ByteOffset`
```

Both directions rejected, then reverted. The three existing `find_matching_open_bracket` tests also failed to compile until updated, which is the same signal from a second source.

No `trybuild` compile-fail test is checked in: it would add a proc-macro dev-dependency to a crate whose CI cost is already dominated by test-binary compilation, for one assertion. The demonstration above is recorded here instead.

## What this does not catch — unchanged from #2386

- **Right unit, wrong string.** `ByteOffset` carries no provenance; one measured against a trimmed copy still slices the original. The cheap fix (carry the `&str` in the type) is written up on #2386 and is **not** in this slice.
- **Arithmetic that keeps the unit valid but the position wrong** (`+ 1` past a multi-byte character).
- **`Add<usize>` accepts any `usize`, so `char_offset + name.len()` still compiles.** That is
  #2389's exact shape — nine sites of `i += store_sub.len()` — and **even with `i: CharOffset`
  all nine would compile and still drop source text**. Verified, not assumed:
  `CharOffset::new(4) + "$名前".len()` yields 11, the wrong value #2389 had. This is the largest
  cluster the epic has found and the newtype does **not** catch it; closing it needs a length
  type, not an offset type.
- The remaining **16 `b()` calls across 13 slice expressions** in this file still go through the
  untyped closure. They are correct by construction today — every one reads the same table — and
  converting them is slice 1b, not this PR.

Stating that plainly because the failure mode of a change like this is that the next sweep gets skipped on false confidence.

## Scope note: slice 1 is **not** the module named in #2386's original plan

`client/scan_index.rs` was dropped. Measuring `Vec<char>` sites against variable `&str` slice sites across the 13 `Vec<char>`-holding modules in `3_transform`, it is the **only** one with zero string slices — the one module where the two units cannot meet, so the newtype could not have rejected anything there. Table and reasoning in #2386.

## Collision with #2383, resolved here

Both PRs edit `destructure_transforms.rs:221-228`. #2383 replaces the `.and_then(…ok())` silent
discard with a `#[cold]` panic; this PR's pre-image is `main`'s and therefore does not contain it.
**Resolving the conflict toward this PR alone would revert #2383 with nothing to notice** — by
#2383's own unreachability argument the panic never fires, so no test and no gate would report it.

`f708e71a` produces both behaviours in one expression:

```rust
.map(|byte| {
    table
        .char_of(byte)
        .unwrap_or_else(|| bracket_offset_miss(byte.get(), statement.len()))
})
```

It also closes a direction the review found and I had missed: the old `binary_search` result
dropped to a bare `usize` that is a **char** index, so the typed value re-entered untyped space
immediately. `char_of` keeps the conversion inside the table and **deletes the duplicate
`byte_offsets` vector** — that raw vector had exactly one remaining use.

`Candidate`'s three fields are now `CharOffset`, which retires their `// char index of …`
comments; a comment asserting a unit is the thing this type replaces.

**Sequencing**: this branch already contains the combined form, so if #2398 lands first, #2383's
hunk is satisfied and should be **dropped rather than re-applied**. If #2383 lands first I rebase
onto it and keep this expression.

Related: #2386, #2378, #2383.


baseballyama:
Reviewed at `63e10792`. **0 blocking.** The newtype is load-bearing at the crossing it targets, the no-`trybuild` call is right, and the limits section is the reason this reads as a slice rather than a conversion. Two things to fix: one stated figure is ~2.5× high, and this PR collides with #2383 on the exact line both change. Read-only — no build, no test run. CI: 1 pass, 30 pending, no failures.

### The crossing at `:224` is the right target and the conversion is load-bearing

That call is the shape behind #2369, #2376 and #2385: `find_matching_open_bracket` produces a **byte** offset, the surrounding loop indexes **chars**, and `Candidate` carries char indices between them. Before, both sides were `usize` and the untyped `b()` closure was the only thing keeping them apart. Now `close_pos: ByteOffset` forces `table.byte(CharOffset::new(i))` at the call, so the conversion is a compile-checked step rather than a convention.

`CharToByte::byte` also preserves the old closure's semantics exactly — past-the-end char offsets map to `byte_len`, matching an exclusive slice end — and says so in a doc comment. That is what makes "no behavioural change" true rather than hoped, and it is the detail a newtype conversion most often gets wrong.

Refusing `From`/`Into` between the two, so every crossing is a named greppable method, is the right design and the module doc gives the reason.

### One direction is typed; the other is not

At the same call, the return trip is still untyped:

```rust
.and_then(|byte| byte_offsets.binary_search(&byte.get()).ok())
```

`byte.get()` drops back to `usize` and the `binary_search` result — a **char** index — re-enters bare-`usize` space immediately, feeding `pattern_start` and then `b(pattern_start)`. So the byte→char half of the very crossing this PR is about remains exactly as unpoliced as before. A `CharToByte::char_of(&self, at: ByteOffset) -> Option<CharOffset>` closes it, removes the raw `byte_offsets` vector from this call site entirely, and would make `pattern_start` a `CharOffset` — which is what makes `b(pattern_start)` on the next line checked rather than conventional.

### A limit worth adding to the list, because a reader will assume otherwise

`Add<usize>` / `Sub<usize>` on both newtypes means `CharOffset + <a byte length>` compiles. That is precisely #2389's defect: `i += store_sub.len()` at nine sites, where `i` walked a `Vec<char>` and `.len()` was bytes. Even with `i: CharOffset`, every one of those nine sites would still compile and still drop source text.

The existing bullet — "arithmetic that keeps the unit valid but the position wrong (`+ 1` past a multi-byte character)" — is a different and much smaller case. The one to state is: **a delta in the wrong unit is unchecked**, because deltas are bare `usize` by construction. It matters because #2389 is the largest single cluster this epic has found, and a reader looking at these types will reasonably assume they would have caught it. They would not have. Saying so protects the next slice from being scoped on false confidence, which is the stated purpose of that section.

### The ~40 figure does not reconcile

> The remaining ~40 byte-slice sites in this file still go through the untyped `b()` closure.

At this head `destructure_transforms.rs` has **16** occurrences of `b(` and **13** `statement[…]` slices. Under either reading the remainder is ~16, not ~40. Since that number is what sizes slice 1b, it is worth either correcting or saying which population produces 40 — this is the same unchecked-prose-number shape as the 317-vs-306 tuple count on #2384, and prose figures in this epic are not gated by anything.

The qualitative claim it supports is correct and is the important half: those sites are correct today because they all read the same table, and they are unpoliced. The PR does not read as having converted the file, which is the right posture.

### Sequencing: this collides with #2383 on the same line

Both PRs edit `:221-228`. #2383 (open, `af6ef76b`) turns `.and_then(|byte| byte_offsets.binary_search(&byte).ok())` into `.map(|byte| …binary_search(&byte).expect("open bracket is a char boundary"))`; this PR keeps `.and_then(…ok())` and changes the argument to `byte.get()`. Its pre-image is main's, so this branch does not contain #2383.

Whichever lands second must produce **both** properties — the typed conversion *and* the `.expect` — and the failure mode is silent: resolving in this PR's favour reverts #2383 and quietly restores the discard that PR exists to remove, with no test to notice, since the panic is unreachable by #2383's own argument.

Worth deciding the order now rather than at merge time. If #2383 lands first, the rebase here is a one-line adaptation to `byte.get()`. And the `char_of` suggestion above is the shape that reconciles them properly: `table.char_of(byte).expect("open bracket is a char boundary")` carries the type *and* the assertion in one expression.

### The `trybuild` call is right

Agreed, and for a reason stronger than proportionality: this repo's CI cost is dominated by **test-binary compilation**, not test execution — that is why `run-test-shard.sh` partitions binaries rather than tests. Adding a proc-macro dev-dependency to `rsvelte_core` for one compile-fail assertion pays that cost on every shard, forever, to check something the two captured `E0308`s already establish. Recording the demonstration in the PR body is the proportionate form.

The second signal is the better evidence anyway and deserves a sentence of its own: the three existing `find_matching_open_bracket` tests **stopped compiling** until updated. That is the type doing its job on code nobody deliberately mutated — an unplanned positive control, and worth more than a constructed one.

baseballyama:
The `Add<usize>` limit noted in the body is now tracked as #2404 (`ByteLen`/`CharLen`), with the `+ 11` demonstration. Nothing in this PR changes; the pointer just stops the limit living only in a PR body.

PR #2402, opened 2026-08-06
fix(analyze): attach spans to five spanless warnings

baseballyama:
Part of #2281 (warning parity). Takes the position ratchet from **625 to 175** entries — 450 removed, 0 added, all three targets.

## Read this first: the shrink is semantically inert

This bucket is *position-only*. In every entry here rsvelte already emits the right diagnostic, with the right message; it reports it at a wrong or absent location. A 450-entry shrink reads like a large correctness win to anyone skimming the number, and it is not one. The code ratchet — which does measure semantics — is **unmoved at 70** by this PR, and that is the intended result. The two ratchets are kept apart precisely so the larger position backlog cannot mask a semantic regression.

What it does buy: `warning.start` was `undefined`, so an editor or CLI had nothing to underline. That is a real defect for tooling, just not a compilation one.

## "One systemic cause" was a hypothesis, and it was false

Five codes dominated the missing-span half — 525 of 649 tuples:

| code | tuples |
|---|---|
| `event_directive_deprecated` | 142 |
| `element_invalid_self_closing_tag` | 118 |
| `export_let_unused` | 110 |
| `non_reactive_update` | 102 |
| `options_missing_custom_element` | 53 |

They share a symptom — `start === undefined` — and nothing else. Checking each one's upstream warn target before writing a fix turned up **three** mechanisms:

**A — the visitor already holds the warn target.** `element_invalid_self_closing_tag` (`RegularElement.js:223` warns on `node`) and `event_directive_deprecated` (`OnDirective.js:16` warns on `node`, the directive, *not* the element) had `element` / `on` in scope and simply did not pass them. This is the only mechanism for which "attach the span already available at the emission site" — the description this file previously gave the whole bucket — is accurate.

**B — the visitor holds the wrong node.** Upstream warns `options_missing_custom_element` on the `customElement` **attribute** (`index.js:692`), while rsvelte's analysis holds `<svelte:options>`. Attaching what was in hand would have produced a confident span pointing at the wrong thing. That is exactly the failure that put the a11y bucket in this list under the wrong label, so it seemed worth not repeating one PR later.

**C — the target is not a node in the tree.** `non_reactive_update` (`index.js:765`) and `export_let_unused` (`index.js:815`) warn on `binding.node`, the declaration identifier. The binding records only `declaration_start`; there is no end. It is reconstructed from the name's byte length. This is a data-availability problem, not a plumbing one, and a single shared "attach the node" fix could not have covered it.

## Tests

`crates/rsvelte_core/tests/warning_span_attach.rs` — 6 tests, all FAILED before with `has no start`, grouped by mechanism so a future reader sees the three rather than one:

| test | mechanism |
|---|---|
| `self_closing_tag_spans_the_element` | A |
| `event_directive_spans_the_directive_not_the_element` | A — pins directive, not element |
| `options_missing_custom_element_spans_the_attribute` | B — pins attribute, not `<svelte:options>` |
| `export_let_unused_spans_the_identifier` | C |
| `non_reactive_update_spans_the_identifier` | C |
| `non_ascii_identifier_span_is_byte_correct` | C — unit pin |

Two of these assert the *narrower* of two plausible spans (directive not element; attribute not options element). Those are the cases where a wrong fix still produces a span and still looks right in aggregate, so an equality assertion on the exact source text is the only thing that separates them.

The non-ASCII test earned its place immediately: it caught a genuine unit confusion, though in my test harness rather than in the compiler. `end` is derived from the name's **byte** length while the reported `column` counts **characters**; slicing a line by `column` as if it were a byte index panics on `export let プロップ`. rsvelte and official both report columns 19-23 there, so the compiler was right and the check was wrong. Kept as a pin on the byte/char boundary surviving the conversion.

`--test validator` stays 333/333; `warning_filter`, `validator_message_wording` and `svelte_options_validation` unchanged.

## What is left

Measured on the **rebased** tree (post-#2384), so this supersedes the pre-rebase figures this section used to carry: 75 entries, one mismatching tuple each. 74 are missing-span over 25 codes with none above 9 (`svelte_component_deprecated` 9, `slot_element_deprecated` 6, `block_empty` 6, …). There is no next large cluster in this bucket. The 75th is a single `a11y_figcaption_index` disagreeing on both line and column — the residual #2384 predicted would be structurally out of reach of its four stamp sites, and it survived exactly as predicted.

## Ratchet

Now that #2384 has merged, this is **rebased onto it and regenerated from a run**: `warning-position-known-failures.*.json` **529 → 75**. `warning-known-failures.*.json` unchanged at 70.

```
node scripts/compat-corpus/compile.mjs                              # 14,130 entries × 3 targets
```

Entry floor met (14,130 ≥ the 12,000 below which `--update-warning-baseline` refuses, because it deletes every id it did not measure).

**The 75 is not reachable arithmetically, and that is worth reading before trusting any future re-baseline.** From 625 (before either fix), #2384 alone removed 96 and this change alone removed 450 — and those two removal sets are **disjoint**. So both set arithmetic (625 − 546) and the subtraction a rebase invites (529 − 450) give **79**. The measured answer is **75**, because four entries clear only when *both* fixes are present and by neither alone:

```
svelte/packages/svelte/messages/compile-warnings/a11y.md/16.svelte
svelte/documentation/docs/98-reference/.generated/compile-warnings.md/16.svelte
svelte.dev/…/98-reference/.generated/compile-warnings.md/16.svelte
svelte.dev/…/98-reference/30-compiler-warnings.md/18.svelte
```

Each was verified by membership: present in the 625, in #2384's 529 **and** in this PR's 175, absent from the 75. The ratchet is keyed per *entry* while the comparison is per *tuple*, so an entry with any surviving tuple stays listed — meaning the interaction lives **inside** entries, not between the removal sets. Disjointness is therefore not sufficient for the counts to add. Written up in `compatibility/warning-known-failures.md`.

## CI status at time of writing

The `CodSpeed Performance Analysis` failure on the pre-rebase head `4e2b60ab` (`Performance Regression: -11.52%`) was diagnosed as a **measurement artifact, not a regression**. The benchmark job itself succeeded (11 steps, all green, 25 min — not `exit 143`), so this was a threshold trip on the comparison. Evidence:

- The diff is entirely in Phase 2. `analyze[synthetic-state-heavy]` — which contains every instruction it adds — was **measured** in that same run and is in the untouched set, as are `parse`, `transform_client`, `transform_server` and `client` for the same fixture. The log carries exactly 127 `Measured:` lines = 3 regressed + 124 untouched, so every measured benchmark is accounted for.
- The three flagged rows are one measurement counted three times (`two_compiles` and `compile_both` both contain the server compile).
- The deltas scale wrongly for added work: `two_compiles` pays analyze twice and `compile_both` once, so a per-analyze cost would make their absolute deltas differ 2:1 — observed +5.1 ms vs +5.0 ms, i.e. 1:1. Meanwhile the three ratios are 1.137 / 1.137 / 1.114 across benchmarks of 18 / 37 / 45 ms. A fixed amount of added work cannot produce a constant *ratio* across different absolute sizes.
- CodSpeed's own report flags two comparability defects: cross-runtime-environment comparison scoped to the changed benchmarks, and a base fallback to `250be54` because no successful run existed on `main`.

**The empirical retest did not run.** Actions was in `major_outage` (repo queue 42 queued / 1 in progress), so the re-run against a correct base never dequeued. The verdict above is analytic. It should not be read as having been confirmed by a green CodSpeed check, because there has not been one on a rebased head.


baseballyama:
Reviewed at `3263a8d8`. **0 blocking.** I checked the three-way split by reading the upstream warn target for all five codes, and it is right in every case. The two mechanism-B-shaped tests discriminate, mechanism C's arithmetic is correct, and the inertness statement is where it needs to be. One residual worth naming and one merge hazard that is now three-way. Read-only — no build, no test run. CI: 31 pending, no failures.

### The three-way split, verified by reading all five

Not sampled — each upstream call site, with its first argument:

| code | upstream site | warn target | mechanism |
|---|---|---|---|
| `element_invalid_self_closing_tag` | `RegularElement.js:223` | `node` — the element | A |
| `event_directive_deprecated` | `OnDirective.js:16` | `node` — the **directive** | A |
| `options_missing_custom_element` | `index.js:692` | `attribute` — not `root.options` | B |
| `non_reactive_update` | `index.js:765`, `:773` | `binding.node` | C |
| `export_let_unused` | `index.js:815` | `binding.node` | C |

Every row matches what the PR claims and what the code now stamps. The two `non_reactive_update` sites both pass `binding.node`, so there is no fourth mechanism hiding in the second one.

This is the second bucket in two PRs whose recorded cause turned out to be inferred from where entries clustered rather than from control flow, and it is worth saying what actually distinguishes the two rounds: the earlier reading grouped by *symptom* (`start === undefined`), which is downstream of all three mechanisms and therefore cannot separate them. Checking the warn **target** — the input to the diagnostic rather than its output — is what splits them. That generalises to the rest of #2281's backlog.

### Mechanism B: the two narrow-span tests do discriminate

This was the place a plausible wrong fix could hide, and it cannot here, for a reason worth stating precisely: `spanned()` slices `start.column..end.column` and the assertions are `assert_eq!` on the resulting source text, so **both** ends are pinned. A span is not merely "present and confident" — it is checked against an exact substring.

- `event_directive_spans_the_directive_not_the_element` asserts `on:click={() => {}}`. Attaching the element in hand yields `<div on:click={() => {}}></div>` — a different string, so the test fails. The two plausible fixes are separated.
- `options_missing_custom_element_spans_the_attribute` asserts `customElement="my-el"`. Attaching `<svelte:options>` yields `<svelte:options customElement="my-el" />` — likewise separated.

That is the property the aggregate count cannot supply: 450 entries clearing tells you the position moved, not that it moved to the right node. These two tests are what make the B fix checkable, and pinning both ends rather than a prefix is what makes them airtight.

### Mechanism C: the arithmetic is right, and the sixth test pins it

`stamp_declaration` computes `end = start + binding.name.len()` — a byte length added to a **byte** start, so the result is a byte offset, which is the only internally consistent choice. The doc comment's justification is exactly right in the direction that matters: a `chars().count()` there would produce `start_byte + char_count`, a byte offset landing mid-character for any non-ASCII name.

I re-derived the sixth test's numbers rather than taking them. For `<script>export let プロップ;</script>`: `<script>` is 8 chars, `export let ` is 11, so the identifier starts at 19 — byte and char alike, since everything before it is ASCII. `プロップ` is 4 chars and 12 bytes, so the character end is 23 and the byte end is 31. The doc's "19-23 … would show as 31" is correct on all three figures, and `assert_eq!(spanned(…), "プロップ")` fails if the byte end leaks into the column. The conclusion that the harness was wrong and the compiler right holds up.

Also worth noting because it is structural: the ratchet compares one `(code, line, column)` per entry, so `end` is **not gate-observable at all**. These unit tests are its only oracle — the same situation as the message text on #2391 — and here they observe it properly, at both ends, including the non-ASCII case. That is the right handling of a field a gate cannot see.

**The one residual**, and it is a convention question rather than a defect in this PR: rsvelte's `column` counts **characters** (the harness comment says so, and the byte→column conversion is what makes the プロップ case come out at 23), while upstream's columns come from a locator over a JavaScript string and are therefore UTF-16 code units. Those agree for every BMP character, which is why プロップ matches, and diverge for an **astral** one — an identifier containing e.g. `𝕏` (U+1D54F, a valid `ID_Start`) is 1 char, 2 UTF-16 units, 4 UTF-8 bytes, so a char-counted column would report one less than upstream per astral character on the line.

That affects every code, not these five, and the corpus is the oracle for it — the 450 clearing entries show the conventions agree on everything the corpus contains. But this file is where the two units are already being reasoned about explicitly, so it is the natural place to record the boundary: `プロップ` proves byte-length end vs character column, and does **not** probe char-column vs UTF-16-column. One sentence in the sixth test's doc, or an issue, so the next person does not read that test as covering the whole class.

### The inertness statement is where it needs to be

"Read this first: the shrink is semantically inert" leads the PR body, and the md now carries the same claim in the section header rather than only in the commit. Both say the code ratchet is unmoved at 70 and why the two are kept apart. A 450-entry shrink is exactly the number that gets quoted without its caveat, so putting the caveat above the number rather than below it is the right call — and the counts reconcile at this head: all three position files hold **175**, all three code files **70**.

### The merge hazard is now three-way, and it is not one number

`warning-known-failures.md` is being rewritten from the same base by three open PRs:

- #2384 → position count 625 → **529**, and it rewrites the whole "Warning positions" section into a two-cause split (649 missing-span / 318 a11y tuples).
- #2391 → code count 70 → **51**, same line `:38`.
- this PR → position count 625 → **175**, and it *also* rewrites the same section, into the three-mechanism account.

So the conflict is not the shared `:38` line alone — **#2384 and this PR replace the same paragraph with independently written prose.** Worse, the post-merge position count cannot be derived arithmetically: 625 − 96 − 450 assumes the two removal sets are disjoint at **entry** granularity, and an entry may carry both an a11y tuple and a missing-span tuple, in which case it clears only after both land. Any hand-written total is a number no run produced.

The only safe sequence is the one already planned — #2384 first, then this and #2391 rebased and **regenerated**, with the two prose accounts reconciled by hand rather than by git. Concretely, the merged section wants #2384's per-tuple split as the frame and this PR's three mechanisms as the sub-structure of its missing-span half; they are complementary analyses of the same bucket, not competing ones.

baseballyama:
CodSpeed's `-11.54%` here is real, and it is not caused by anything in this diff.

Giving these five warnings a span moves them from "no position" to "has a position", and `generate_frame` only runs for a warning that has one. That function reached the five lines it quotes with `source.split('\n').collect::<Vec<_>>()` — a full scan of the source **per warning**. `synthetic-state-heavy` emits `non_reactive_update` many times, so the three regressed benchmarks are all paying O(source × warnings) that this PR newly switched on for them.

Fixed in #2587: the frame now slices its five lines out of the `Utf8ToUtf16` line index the caller already builds to convert the offsets. Output is byte-identical (eight shapes diffed against the pre-change implementation, plus a poisoned-control run to prove the harness sees the function). On 3200 warnings that path goes from 126 ms to 13 ms.

Rebase this branch once #2587 lands and CodSpeed should go back to green. Two caveats worth noting on the report itself, neither of which changes the conclusion: it compared against `5f6de88` rather than the real merge base (no successful run on `main` at `723df2f`), and it flagged "different runtime environments detected".

baseballyama:
Merged `main` and re-baselined the validator ratchet, which this PR made stale in seven places:

```
custom-element-props-identifier-props-option   invalid-self-closing-tag
runes-referenced-nonstate                      runes-referenced-nonstate-bind-this
tag-custom-element-options-missing             unreferenced-variables
unreferenced-variables-each
```

179 → 172. Each is a fixture whose only remaining divergence was one of the five spans this PR attaches, so they are the ratchet's own definition of a burn-down.

The cluster counts in `validator-known-failures.md` were re-measured from the suite's failure report rather than subtracted, per that doc's own rule, and the measurement turned up two entries whose recorded cause is no longer true:

| cluster | was | now |
|---|---:|---:|
| error span not populated | 141 | 141 |
| warning span-only | 35 | 30 |
| warning content | 3 | **1** |
| | 179 | 172 |

`unknown-code` was documented as an emission-**order** bug (the `svelte-ignore` comment-code warnings emitted as a batch ahead of the a11y ones). The report now shows all six interleaved in upstream's source order, so only the `None` spans remain. `attribute-quoted` was documented as a message-wording difference; its messages now match verbatim. Both moved to span-only, and the doc records the correction instead of quietly renumbering — a ratchet entry whose stated cause is gone is an entry that would have been "fixed" by work that had already happened.

Only `svelte-self-deprecated` is left in the content cluster: the format string interpolates the component's own name into its example import and rsvelte lowercases it — `import Input from './input.svelte'` against upstream's `import Self from './Self.svelte'`.

Separately, the earlier CodSpeed `-11.54%` here was real and is fixed by #2587 (now merged), which stops `generate_frame` from splitting the whole source once per spanned warning.

baseballyama:

```
```

PR #2406, opened 2026-08-06
ci: run the blocking gates on pull requests to any base (#2405)

### LINKED ISSUE #2405: ci: a stacked PR silently gets a reduced check set (8 vs 35) because ci.yml is gated on base=main
## What happens

`.github/workflows/ci.yml` is gated on `pull_request: branches: [main]`. A PR whose **base is
another PR's branch** therefore never fires it — silently, with a green check mark.

Verified on PR #2370 (base `fix/2357-scan-sweep-batch2`) against PR #2389 (base `main`) at the
same point in time:

| | #2370 (stacked) | #2389 (base `main`) |
|---|---|---|
| total checks | **8** | **35** |
| `Format` / `Clippy` | never ran | ran |
| `Test unit` / `Test core` / `Test runtime` / `Test server` | never ran | ran |
| `C ABI`, `Code coverage`, lockfile gates | never ran | ran |

#2370's eight checks are all parity workflows that carry no branch filter (`Compiler parity`,
`Formatter parity`, `Lint parity`, `Shape matrix parity`, `svelte-check parity` x3,
`MSRV, package, and docs`).

## Why it is worse than "fewer checks"

1. **It reads as green.** Nothing on the PR page says a check set was reduced. The only tell is
   counting checks against a base-`main` PR — 8 versus 35.
2. **It removes exactly the jobs a stack needs most.** `Test unit` is the job that runs
   `#[cfg(test)]` modules living in `src/`, which is where stacked compiler work tends to put its
   new tests.
3. **It defeats the reason for re-pointing a stacked PR.** Re-pointing #2370 onto its parent's
   current head was done to get a genuine combined verification of both branches. The combined
   run is precisely the one that excludes the test suite, so the only evidence the combination
   compiles and passes is a local run.
4. **Retargeting does not fix it.** When the parent merges, GitHub auto-retargets the child to
   `main` and `ci.yml` becomes eligible — but a retarget alone does not re-run checks. Without a
   push or a manual re-run, the child can be merged having never run them.

## Suggested fixes (pick one)

- Drop the `branches:` filter from `pull_request` in `ci.yml` (keep it on `push`), so every PR
  runs CI regardless of base. Simplest; costs runner time on stacked PRs, which is the point.
- Or add a required job that fails when `github.base_ref != 'main'` and `ci.yml` has not run,
  making the reduction visible instead of silent.

Either way, a stacked PR should not be mergeable on a check set nobody chose.

## Interim rule

Do not merge a stacked PR until `ci.yml` has run once against a `main` base — after retargeting,
push or manually re-run first.

baseballyama:
Closes #2405.

A PR whose base is not `main` matched **none** of the blocking workflows, because each declared
`pull_request: branches: [main]`. It therefore displayed **green having run no tests, no clippy
and no fmt.** The only tell was counting its checks against a base-`main` PR: #2370 had **8**,
#2389 had **35**, at the same moment.

## Which option, and what failure mode it optimises against

Two candidates were on the issue. I took **dropping the filter** rather than **a required job that
detects the reduction**, on one piece of evidence:

**Stacked PRs are 0.29% of the population** — 5 of 1741 merged PRs, full history. So the cost
argument that would favour the cheaper option does not exist at this volume; the extra runner time
is five PRs' worth over nine months.

That leaves only the failure mode. The detector makes the reduction **visible**; dropping the
filter makes it **impossible**. The defect being fixed is precisely that a reduction *was* visible
— in the check count — and nobody looked, for nine months. Optimising for "impossible" over
"visible" is the whole lesson of the bug.

## Scope: the three blocking gates, not everything

| workflow | change | why |
|---|---|---|
| `ci.yml` | filter dropped | `Format`, `Clippy`, `Test unit/core/runtime/server`, `Embeddable core`, lockfile gates |
| `capi.yml` | filter dropped | `C ABI`, a real gate; keeps its `paths:` filter |
| `changeset.yml` | filter dropped | `Changeset`, a real gate |
| `coverage.yml`, `benchmark.yml`, `codspeed.yml` | **unchanged** | expensive and informational, and already label-gated behind job-level `if:` — running them on stacked PRs adds queue load without adding a gate |

`push: branches: [main]` is **kept everywhere.** Removing that too would run CI on every branch
push and multiply a queue that is already saturated.

## Concurrency: checked, not assumed

The group is `${{ github.workflow }}-${{ github.head_ref || github.ref }}` in all three.
`github.head_ref` is the PR's **head** branch, and a parent and its child necessarily have
different heads (`…-batch2` vs `…-batch4`), so they land in different groups and cannot cancel
each other. Pushing the parent's branch cancels only the parent's own in-flight run, which is the
intended supersede. GitHub does not re-run a PR when its *base* branch is pushed, so the child is
not triggered by parent activity either.

This mattered: trading a silent gap for a silent cancellation would have been a worse bug, because
`CANCELLED` reads as a failure and gets re-run rather than investigated.

## Verification is a check count, not this diff

The thing being fixed is *a check that does not run*, so a workflow diff cannot demonstrate it —
the symptom of failure is indistinguishable from success by construction. A throwaway PR stacked
on **this branch** is the control; its count is reported in a comment below. GitHub evaluates
`on.pull_request.branches` against the workflow file in the PR's merge ref, so a PR based here
sees the fixed triggers.

## Historical exposure: 5 of 1741, none of them untested

| PR | base | merged |
|---|---|---|
| #10, #11, #12 | `review/tier-1-5-fixes` → `-1-6-` → `-1-7-` | 2026-05-03 |
| #1267 | `feat/corpus-burndown-2` | 2026-06-27 |
| #1913 | `perf/compile-wave2` | 2026-07-30 |

None put untested code on `main`. `ci.yml`'s `pull_request:` declares no `types:`, so it defaults
to `[opened, synchronize, reopened]` — merging a child into its parent branch pushes that branch,
synchronizes the **parent** PR, and runs full CI on the combination. All five had a parent
targeting `main`. What was lost is review signal at the child PR, not coverage.

**0.29% is scope evidence for how to fix this, not evidence that it is small.** Coverage arriving
later at a different PR is not the same as the signal being right where the decision is made — a
child PR showing green having never run `Format`, `Clippy` or any `Test` job is a defect whatever
the parent's run covers, because the reviewer is looking at the child.

Two merge orders remain available while this lands, and they are safe for different reasons:

- **Child into parent first** — the parent PR synchronizes and runs full CI on the combination,
  with nothing to remember. *Assumes you are not squash-merging*: under squash this produces one
  commit on `main` carrying both changes under one title.
- **Parent to `main` first** — the child auto-retargets and, after a push or manual re-run, gets
  the full check set on top of the merged parent. Stronger verification and one commit per change.
  *Assumes someone re-runs*, since retargeting is not a `synchronize` event; the mitigation is that
  the miss is visible in the check count rather than silent.

This repo squash-merges, so the second is being used, with the count as the gate.

## What this does **not** change

This alters what runs on every future PR, so the boundaries are worth stating rather than
inferring.

**1. `coverage.yml`, `benchmark.yml`, `codspeed.yml` remain base-filtered — by choice, not by
oversight.** They are expensive, informational, and already gated behind job-level `if:`/label
conditions, so running them on stacked PRs would add queue load without adding a gate. If a
stacked PR needs coverage numbers, the label mechanism still works once it targets `main`.

**2. "CI now runs" is not "CI now blocks".** Verified against the repo, not assumed:

```
GET /repos/baseballyama/rsvelte/branches/main/protection  →  404 Branch not protected
GET /repos/baseballyama/rsvelte/rulesets/16341833         →  rules: deletion, non_fast_forward
```

The `main` ruleset carries **no `required_status_checks` rule at all**. So nothing here is
enforced on any PR, stacked or not — a red `Test unit` is still merge-able by anyone who clicks
through it. What this PR fixes is that the checks now **exist and report** on a stacked PR; it
does not make them binding, because they are not binding anywhere.

That is an owner-level setting rather than something a workflow file can reach, so it is out of
scope here — but a reader who takes "the gates now run on stacked PRs" as "stacked PRs are now
gated" would be wrong in a way that matters.

### DIFF
--- a/.github/workflows/capi.yml
+++ b/.github/workflows/capi.yml
@@ -23,8 +23,8 @@ on:
       - 'Cargo.toml'
       - 'Cargo.lock'
       - '.github/workflows/capi.yml'
+  # Unfiltered: see ci.yml.
   pull_request:
-    branches: [main]
     paths:
       - 'crates/rsvelte_capi/**'
       - 'crates/rsvelte_core/**'
--- a/.github/workflows/changeset.yml
+++ b/.github/workflows/changeset.yml
@@ -9,8 +9,8 @@ name: Changeset Gate
 # re-trigger the gate, since the default `pull_request` trigger types
 # (`opened`/`synchronize`/`reopened`) don't include `labeled`.
 on:
+  # Unfiltered: see ci.yml.
   pull_request:
-    branches: [main]
     types: [opened, synchronize, reopened, labeled, unlabeled]
 
 concurrency:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,9 @@ name: CI
 on:
   push:
     branches: [main]
+  # Unfiltered: a PR stacked on another branch must still run these gates,
+  # otherwise it reports green having run none of them.
   pull_request:
-    branches: [main]
 
 permissions:
   contents: read


baseballyama:
## Verified by check count on a throwaway stacked PR, not by this diff

#2407 was opened with base `fix/2405-stacked-pr-ci` — i.e. stacked on this branch, so GitHub
evaluates the triggers from the fixed workflow files in the merge ref.

**Before (a stacked PR on unfixed triggers, #2370): 8 checks, none from `ci.yml`.**

**After (#2407, first push — a markdown-only change):** 20 checks, including every job that was
missing:

```
Format · Clippy · Embeddable core · Playground lockfile · Lint-types lockfile
Test unit · Test runtime · Test thread-safety · Test (ubuntu-latest, 1/2/3)
Changeset · Documentation · Compatibility Report · Language server
Test fmt corpus · Corpus verify baseline-flag contract · Auto-update branch guard
VPS native shim smoke test · VPS native VERSION drift check
```

**Second push** touched `crates/rsvelte_capi/src/lib.rs` so the probe would also reach the
`paths:`-filtered gates, which a markdown diff legitimately skips. That brought in the two
remaining ones:

```
C ABI (ubuntu-latest) · C ABI (macos-latest) · Compiler parity · Formatter parity
Lint parity · Shape matrix parity · svelte-check parity ×3
```

So all three edited workflows are confirmed firing on a non-`main` base: `ci.yml`, `capi.yml`
(`C ABI`) and `changeset.yml` (`Changeset`).

The run list is the cleaner form of the evidence, because it shows the trigger rather than the
rendering:

```
pull_request  pending  CI            9404eec7
pull_request  queued   C ABI         9404eec7
pull_request  queued   Changeset     9404eec7
pull_request  queued   Corpus Compat 9404eec7
```

`event = pull_request` on a PR whose base is a topic branch. Before this change `ci.yml` produced
**zero** runs for such a PR.

## Concurrency: verified empirically, not only by reading the expression

The concern was trading a silent gap for a silent cancellation. Two observations:

1. **Parent and child coexist.** While #2407's runs were queued, this PR (#2406, head
   `fix/2405-stacked-pr-ci`) independently held 25 checks with its own `CI`, `C ABI` and
   `Changeset Gate` runs queued — neither cancelled the other. The group keys on
   `github.head_ref`, and a parent and child necessarily have different heads.
2. **Supersede still works within one PR.** #2407's `Changeset Gate` on `52dc76af` shows
   `cancelled` when `9404eec7` arrived — the intended behaviour, and the control showing the
   concurrency group is actually active rather than inert.

## Cleanup

#2407 and its branch are throwaway and are being closed/deleted now that the counts are recorded.
Nothing from it is merged.

baseballyama:
Reviewed at `98c93886`. **0 blocking.** The scope call is right, and the one interaction that could have made unfiltering *wrong* holds for a reason worth recording. Read-only — no build, no workflow run. CI: 19 pass / 2 skipped / 4 pending, no failures.

### The partition, enumerated rather than sampled

At this head, across all 18 workflows, the ones with a `pull_request:` trigger split exactly two ways:

| still `branches: [main]` | unfiltered |
|---|---|
| `benchmark.yml`, `codspeed.yml`, `coverage.yml` | `ci.yml`, `capi.yml`, `changeset.yml` (this PR) + `corpus-compat.yml`, `crates-io-packages.yml`, `type-aware-lint.yml` (already) |

The axis that separates them is the right one: everything that *gates correctness* now runs on every PR, and what stays filtered is the three that measure **against a main baseline** — a benchmark or coverage delta computed against a parent branch is not a number anyone would act on. So the filter that remains is not leftover, it is the case where the filter is the correct behaviour.

`push: branches: [main]` is kept, so this does not double post-merge cost either. And ci.yml has **no** base-relative logic at all — zero hits for `origin/main` / `base.sha` / `merge-base` / `github.base_ref` — so nothing else inside it silently assumed the base was main.

### Unfiltering `changeset.yml` is safe because of #1799, and that link is invisible

This is the part I checked hardest, because a changeset gate on a stacked PR is exactly where unfiltering could have produced a wall of false failures. The gate diffs against

```
base="$(git merge-base HEAD origin/main)"
git diff --name-only --diff-filter=A "$base"...HEAD -- '.changeset/*.md'
```

— the merge-base with **main**, not `github.event.pull_request.base.sha` and not the PR's base branch. That was #1799's fix for a different problem (a stale `base.sha` misreading other PRs' changesets as this PR's). Its side effect is what makes today's change safe: for a PR stacked on a parent branch, the merge-base with main sits *below* the parent's commits, so the parent's changeset is inside the diff range and the child does not fail for lacking one of its own. Had the gate stayed base-branch-relative, unfiltering it would have failed every stacked `fix:`/`feat:` PR.

The flip side is real and I think acceptable: a stacked child now passes on its **parent's** changeset. That is fail-open, but in the correct direction — one changeset per logical stack is what Changesets wants anyway, and the alternative (a changeset per stacked PR) would produce duplicate changelog entries.

The concurrency groups key on `${{ github.workflow }}-${{ github.head_ref || github.ref }}` in both files, so parent and child have distinct head refs and cannot cancel each other's runs. That was the other way this could have gone wrong.

### The residual: the defect was "green with checks absent", and only the instance is fixed

Worth separating what was wrong in #2405. The filter was the *cause*; the thing that made it cost a day is that a PR reported **green having run 8 of 35 checks** — an absent check is not a red check, and nothing looks at the difference. Removing three filters fixes the instances that exist today. It does not stop the next workflow from being added with `branches: [main]`, and after this merges the intent behind the three that *keep* the filter is recorded only by their absence from this diff — a reader six months out sees three filtered workflows and no statement of which are deliberate.

Both halves close with the same cheap artifact: a script over `.github/workflows/*.yml` that fails when a `pull_request:` trigger carries a `branches:` key unless the file is on an allow-list with a one-line reason. That turns the intent into something checked, and makes the next filtered addition a deliberate act rather than a default someone copied. Same shape as the sum gate #2384 just added to `known-failures-md-check.mjs`, and it is the CI-layer instance of #1791's whole thesis — "reports success without measuring". Not blocking this PR; worth filing.

One smaller note in the same direction: the comment on `capi.yml` and `changeset.yml` is `# Unfiltered: see ci.yml.` That is the right amount of comment, but it points at a file whose comment explains the *general* rule and not why these two specifically must run on stacked PRs. For `changeset.yml` the reason is the merge-base property above, and it is not obvious — one clause there would save the next reader the trip I just made.

Issue #2409, opened 2026-08-06
compiler: `{#await p　then v}` with an ideographic space emits invalid JS — byte-decoded whitespace, 22 sites


baseballyama:
Split out of the question-4 sweep on #2378. 23 sites, one fix, but **one of them is a user-visible
parse failure on the shipping path and is the reason to do this before the rest.**

## ⚠ `tag.rs:1407` — `{#await …}` emits invalid JavaScript when the separator is an ideographic space

`U+3000` is JavaScript whitespace. `{#await p　then v}` is valid Svelte. Compiled on
`origin/main` (`b152751b`), server mode:

```js
// separator is U+0020 — correct
$.await($$renderer, p, () => {}, (v) => {
    $$renderer.push(`<p>${$.escape(v)}</p>`);
});

// separator is U+3000 — emitted verbatim below
$.await(
    $$renderer,
    ,                       // ← empty argument: this does not parse
    () => {
        $$renderer.push(`<p>${$.escape(v)}</p>`);
    },
    () => {}
);
```

Two failures, not one: the awaited expression is **lost**, and the pending/then branches are
**swapped**. `compile()` returns `Ok`, so nothing reports it — the output is simply not JavaScript.

Cause: `tag.rs:1407` decides `preceded_by_ws` from `self.source.as_bytes()[self.index - 1] as char`.
`U+3000` is `E3 80 80`; the byte before `then` is `0x80`, which casts to `U+0080`, a control
character. `is_whitespace()` is false, so `then` is not recognised as a keyword and is swallowed
into the expression.

`:1415` and `:1428` are the same check on the trailing side, for `then` and `catch`.

**This is phase-1 parse, not the legacy text pipeline.** It is the only item in the whole #2378 map
that an ordinary user can hit today by writing CJK-formatted markup.

Worth settling early: does official Svelte accept `U+3000` here? If it does, rsvelte must too. If
it rejects it, rsvelte must *error* — emitting `,` as an argument is wrong under either answer.

## The full site list

**`is_whitespace()` on a byte — 10 sites.** Unicode whitespace at or below `U+00FF` includes
`U+0085` (NEL) and `U+00A0` (NBSP), and `0x85`/`0xA0` are valid UTF-8 **continuation** bytes. Any
multi-byte character containing one reads as whitespace one byte at a time — `々` is `U+3005` =
`E3 80 85`, so a skip-whitespace loop walks into the middle of it.

```
[...]

PR #2416, opened 2026-08-06
fix(compiler): recognise `then`/`catch` after a non-ASCII space in `{#await}` (#2409)

baseballyama:
Fixes the `tag.rs` half of #2409. **`tag.rs` only** — the other 22 sites in that issue are a
different severity and stay there.

## This is a silent miscompile, not a parse failure

`U+3000` is JavaScript whitespace and `{#await p　then v}` is ordinary CJK-formatted markup.
On `b152751b`, server mode:

```js
// separator U+0020 — correct
$.await($$renderer, p, () => {}, (v) => {
    $$renderer.push(`<p>${$.escape(v)}</p>`);
});

// separator U+3000 — before this PR
$.await(
    $$renderer,
    ,                    // the awaited expression is gone: this does not parse
    () => {
        $$renderer.push(`<p>${$.escape(v)}</p>`);
    },
    () => {}             // and pending / then are transposed
);
```

`compile()` returns `Ok`. No error, no warning.

## Cause

The keyword scan decided word boundaries from a raw byte:

```rust
let prev = self.source.as_bytes()[self.index - 1] as char;
prev.is_whitespace() || prev == ')' || prev == ']'
```

`u8 as char` is a Latin-1 decode. `U+3000` is `E3 80 80`; the byte before `then` is `0x80`, which
becomes `U+0080`, a control character. `is_whitespace()` is false, so `then` is not treated as a
keyword and is absorbed into the awaited expression — which then fails to produce an argument.

Three sites, all in the same scan: the leading check and the trailing checks for `then` and
`catch`. Each now reads a **character**:

```rust
self.source[..self.index].chars().next_back()
self.source[after_idx.min(self.source.len())..].chars().next()
```

### The safety claim, with the enumeration behind it

Slicing `self.source[..self.index]` panics if `self.index` is not a char boundary, where the old
byte indexing could not. "This is safe because `advance` decodes multi-byte sequences" is a claim
about every path that moves `self.index`, so here is the check rather than the assertion:

- `Parser::advance` decodes UTF-8 and steps whole characters.
- `advance_by(n)` sets the index arithmetically and **could** land mid-character. In this scan it
  is called only as `advance_by(4)` / `advance_by(5)` immediately after `match_str("then")` /
  `match_str("catch")` — consuming ASCII keywords that were just matched at the current position,
  so the result is a boundary.
- **The invariant is already load-bearing**: the parser slices `self.source` by an `index`-derived
  bound in **105 places**, including `self.source[self.index..].chars().next()` on its hottest
  path. This change relies on an invariant the parser already depends on everywhere; it does not
  introduce a new class of risk.

`then`/`catch` are ASCII, so `index + 4` / `index + 5` are boundaries too. Both reads are O(1).

## Pins — three, written first and confirmed failing

All three fail on `b152751b`; the third prints the malformed output above.

1. **`the_awaited_expression_is_not_dropped`** — no empty argument slot, and
   `$.await($$renderer, p,` is present.
2. **`the_pending_and_then_branches_are_not_transposed`** — the empty pending arm precedes the
   `then` body. **Deliberately separate from (1)**: a fix that restores the argument while leaving
   the order swapped must still fail.
3. **`a_full_width_space_compiles_like_an_ascii_space`** — asserts the **exact** emitted call for
   the ASCII form, then byte-equality between the two. A structural-validity check would have
   passed on the transposition, so the assertion is on the text.

The ASCII control is what isolates this to the decode rather than to non-ASCII input generally.

## Shippable regardless of the upstream question

#2409 asks whether official Svelte accepts `U+3000` there. That question is still open and this
PR does not settle it — **but emitting `,` as an argument is wrong under either answer**: if
official accepts it, rsvelte must too; if official rejects it, rsvelte must error. Fixing the
malformed output does not foreclose either outcome.

## Verification

- 3 new pins, written before the fix, confirmed failing.
- `cargo test -p rsvelte_core --lib` — 1429 passed, 0 failed.
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — clean, no `-A`.
- `rustfmt --edition 2024 --check` — clean.
- Fixture suites (`parser_fixtures`) do not run locally: `pnpm run generate-fixtures` has not been
  run in this checkout. CI covers them.

## This does **not** close the class — #2409 stays open

Two things a merged #2416 must not be read as fixing:

1. **The other 19 byte-decode sites in #2409** are untouched, by design — different severity, and
   folding them in would gate this behind a larger review. (19, not 22: #2409 scopes 22 and this PR
   fixes three of them. The reconciliation of the four totals circulating for this class, each with
   the command that produced it, is on #2409.)
2. **The same `{#await …　then}` shape exists one layer down in phase 3.**
   `client/state_transforms.rs:378` and `:457` test the character after an identifier against an
   **ASCII-only delimiter whitelist**:

   ```rust
   let next_char = result.as_bytes()[after_ident] as char;
   if next_char != ',' && next_char != ')' && next_char != ' ' && next_char != ':' {
   ```

   Note the cause is different: those comparisons are against ASCII literals, so the *decode* is
   harmless — a non-ASCII byte equals none of them either way. The defect is that the **whitelist
   is incomplete**: a full-width space is a real delimiter and is not listed. Fixing the parse
   layer leaves phase 3 wrong for the same input, and the fix there is not the same fix.

3. **The 83 sites the sweep excluded are unexamined for question 5.** They were cleared on
   "compared only against ASCII literals, so the cast cannot change the answer" — which is true,
   and which answers question 4 only. An ASCII-only *delimiter whitelist* is a different defect and
   is invisible to that query, because there is nothing wrong with the cast to find. Read
   "83 excluded" as **83 cleared against question 4**, not as 83 cleared.

Related: #2409, #2378.

### DIFF
--- a/.changeset/await-ideographic-space.md
+++ b/.changeset/await-ideographic-space.md
@@ -1,0 +1,11 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): recognise `then`/`catch` after a non-ASCII space in `{#await}`
+
+The keyword scan decided word boundaries by casting a raw byte to `char`, which
+decodes UTF-8 as Latin-1. A full-width space before `then` presented its last
+byte as a control character, so the keyword was swallowed into the awaited
+expression and the compiler emitted a call with an empty argument — output that
+does not parse — with the pending and `then` branches transposed.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
@@ -1401,20 +1401,19 @@ impl<'a> Parser<'a> {
             // Only honor `then`/`catch` at the top level of the expression
             if paren_depth == 0 && bracket_depth == 0 && brace_depth == 0 {
                 // Require preceding character to be a word-boundary (whitespace or start)
-                let preceded_by_ws = if self.index == expr_start {
-                    true
-                } else {
-                    let prev = self.source.as_bytes()[self.index - 1] as char;
-                    prev.is_whitespace() || prev == ')' || prev == ']'
-                };
+                // A byte here decodes as Latin-1, so `U+3000`'s last byte read as a
+                // control character and the keyword was swallowed into the expression.
+                let preceded_by_ws = self.index == expr_start
+                    || self.source[..self.index]
+                        .chars()
+                        .next_back()
+                        .is_some_and(|c| c.is_whitespace() || c == ')' || c == ']');
                 if preceded_by_ws && self.match_str("then") {
                     let after_idx = self.index + 4;
-                    let is_word_boundary = if after_idx >= self.source.len() {
-                        true
-                    } else {
-                        let next_char = self.source.as_bytes()[after_idx] as char;
-                        next_char.is_whitespace() || next_char == '}'
-                    };
+                    let is_word_boundary = self.source[after_idx.min(self.source.len())..]
+                        .chars()
+                        .next()
+                        .is_none_or(|c| c.is_whitespace() || c == '}');
                     if is_word_boundary {
                         has_then = true;
                         break;
@@ -1422,12 +1421,10 @@ impl<'a> Parser<'a> {
                 }
                 if preceded_by_ws && self.match_str("catch") {
                     let after_idx = self.index + 5;
-                    let is_word_boundary = if after_idx >= self.source.len() {
-                        true
-                    } else {
-                        let next_char = self.source.as_bytes()[after_idx] as char;
-                        next_char.is_whitespace() || next_char == '}'
-                    };
+                    let is_word_boundary = self.source[after_idx.min(self.source.len())..]
+                        .chars()
+                        .next()
+                        .is_none_or(|c| c.is_whitespace() || c == '}');
                     if is_word_boundary {
                         has_catch = true;
                         break;
--- a/crates/rsvelte_core/tests/await_ideographic_space_2409.rs
+++ b/crates/rsvelte_core/tests/await_ideographic_space_2409.rs
@@ -1,0 +1,75 @@
+//! `{#await p　then v}` — a full-width space before `then` is ordinary in
+//! CJK-formatted markup, and `U+3000` is JavaScript whitespace. The keyword scan
+//! decided word boundaries from a raw byte, so the last byte of `U+3000` read as
+//! `U+0080` and `then` was swallowed into the awaited expression.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+fn compile_server(src: &str) -> String {
+    compile(
+        src,
+        CompileOptions {
+            generate: GenerateMode::Server,
+            ..Default::default()
+        },
+    )
+    .expect("compile")
+    .js
+    .code
+}
+
+fn await_component(separator: &str) -> String {
+    format!(
+        "<script>\nlet p = Promise.resolve(1);\n</script>\n\n{{#await p{separator}then v}}\n  <p>{{v}}</p>\n{{/await}}"
+    )
+}
+
+/// The awaited expression must survive. It used to be dropped entirely, leaving
+/// an empty argument slot — output that does not parse.
+#[test]
+fn the_awaited_expression_is_not_dropped() {
+    let out = compile_server(&await_component("\u{3000}"));
+
+    assert!(
+        !out.contains(",\n\t\t,") && !out.contains("$.await($$renderer, ,"),
+        "empty argument slot in the emitted call: {out}"
+    );
+    assert!(
+        out.contains("$.await($$renderer, p,"),
+        "awaited expression missing: {out}"
+    );
+}
+
+/// Separate from the argument itself: a fix that restores the expression while
+/// leaving the branches in the wrong order must still fail.
+#[test]
+fn the_pending_and_then_branches_are_not_transposed() {
+    let out = compile_server(&await_component("\u{3000}"));
+
+    let then_body = out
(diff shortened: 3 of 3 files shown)


baseballyama:
Reviewed at `b2616fa2`. **0 blocking.** The fix cannot panic — I derived the boundary property rather than taking the `advance` check on trust — the pins are genuinely separate, and the ASCII control isolates what it claims. Read-only — no build, no test run. CI: 31 pending, no failures.

### The char-boundary assumption holds, and the proof is short enough to state

This was the thing that could have turned a miscompile fix into a panic, so I checked it end to end rather than checking `advance` alone:

1. **`advance()` preserves boundaries** (`parser.rs:517-538`): it reads the leading byte and adds the full UTF-8 length (`b < 0xE0 → 2`, `< 0xF0 → 3`, else 4). Given a boundary it lands on a boundary.
2. **Every arm of this loop advances only via `advance()`** — the quote/paren/bracket/brace `continue` arms and the tail at `:1434`. Nothing in the scan does raw byte arithmetic on `self.index`, and the loop is entered at `expr_start`, itself a boundary. So `self.source[..self.index]` is always a valid split.
3. **`after_idx` is a boundary too, for a different reason**: it is only computed after `match_str("then")` / `match_str("catch")` succeeds, so bytes `[index..index+4]` are literally `then` — ASCII — and `index + 4` lands on the byte after them. Same for `+ 5`.
4. `.min(self.source.len())` covers the past-the-end case, so the slice cannot be out of range either.

All four are local to this function plus one method. No panic is reachable.

**A forward-looking caveat for the other 22 sites**, because the property is not global: `advance_by` (`parser.rs:540`) is `self.index = (self.index + n).min(self.bytes.len())` with **no** boundary check. Any site reached through a cursor moved by `advance_by` with a non-boundary `n` would make this same `chars().next_back()` pattern panic rather than mis-read. So the correct generalisation for #2409 is not "replace `byte as char` with `chars()`" — it is "replace it, **after** establishing that the cursor at that site is boundary-preserving". Here that holds; it will not hold everywhere by construction. Worth stating in #2409 explicitly, because the mechanical form of this fix is exactly the shape that would introduce the panic the sibling work just hit.

### The two pins are separate

Under a partial fix that restores the argument but leaves the branches transposed, pin 1 (`$.await($$renderer, p,` present, no empty slot) passes and pin 2 fails — the transposed emission puts the `then` body before the empty pending arrow, so `empty_pending < then_body` is false. The separation is real, not decorative.

One fragility worth naming: pin 2 locates `() => {}` with `str::find`, i.e. the **first** occurrence anywhere in the file. It discriminates today because nothing earlier in this output emits that exact text, but a future codegen change that introduces an earlier empty arrow would silently turn the assertion vacuous, and it would still pass. Scoping the search to the `$.await(` call — `find("$.await(")` and then searching from that offset — costs one line and removes the dependency.

Related observation: pin 3 already pins the order exactly for the ASCII form (`$.await($$renderer, p, () => {}, (v) => {`) and transfers it to the full-width form by byte-equality, so pin 3 **subsumes** pin 2's guarantee. That is not an argument for deleting pin 2 — it fails with a clearer message and it is the one a reader of the issue will look for — but it does mean the order is doubly pinned, and pin 3 is the stronger of the two.

### Pin 3's ASCII control does isolate the decode

The two inputs differ in exactly one character. The control asserts the ASCII form emits the exact expected call, so "non-ASCII input breaks compilation generally" is ruled out by construction: if that were the defect, the control would still pass and the equality would fail, and the equality's only variable is the separator. One-variable difference plus a passing control on the other value of that variable is the right structure for an isolation claim.

Asserting text rather than structure is also correct for the reason given — `assert_structurally_valid` passes on the transposition, so a structural assertion would have been worthless here. That is the same distinction that made #2390's deleted tests worse than useless: an assertion that holds in the broken state.

Worth recording one dependency of the byte-equality assertion: it holds only because a non-dev **server** compile carries no source positions. The two sources differ in length (1 byte vs 3), so any offset leaking into the generated code would break the equality. That makes it a bonus invariant — it also proves no position leaks — but it means the test must stay `GenerateMode::Server` with `dev: false`; switching to dev would fail it legitimately and look like a regression. One clause in the test doc would save that.

### Scope is honest, and I verified the boundary of it

`tag.rs` has **zero** remaining `as char` casts of any kind at this head, so the three sites fixed are all of them in that file and there is no fourth site sharing the shape. Confirmed by exhaustive grep of the file, not by sampling.

One number to check: searching `rsvelte_core/src` for `bytes[…] as char` / `as_bytes()[…] as char` returns **29** occurrences at this head, three of them elsewhere in `1_parse`. If #2409's population is 22, it is worth saying which files that covers — the residual count is what sizes that issue, and this epic has already shipped two corrections to prose figures nothing gates. My pattern may simply be broader than the issue's; the qualitative claim (`tag.rs` complete, the rest deferred) is what I verified and it holds.

### On the two disclosures

**The skipped suite is disclosed and it is the relevant one, which is the right way round.** `parser_fixtures` is precisely the suite that would exercise a phase-1 keyword scan, so saying "these do not run in this checkout, CI covers them" rather than letting a green `--lib` run stand in for the whole is the honest form. The alternative — reporting 1429 passing without the caveat — would have read as broader coverage than was obtained. CI carries it, and the three new pins are integration tests that do run.

**I agree with not settling the upstream question, and the argument is stronger than "it can ship anyway."** Emitting `,` in an argument position is not a candidate behaviour under either answer: if official accepts `U+3000` there, rsvelte must produce the same parse; if official rejects it, rsvelte must produce an error. Malformed output is wrong in both branches, so fixing it cannot foreclose either. That is a genuine independence argument rather than a deferral, and it is what makes this reviewable now. The upstream comparison still needs to happen — it decides whether a *third* pin (error vs accept) is owed — but it is not a precondition for this.

Issue #2413, opened 2026-08-06
compiler: `a11y_invalid_attribute` names `href` where the SVG attribute is `xlink:href`


baseballyama:
## The defect

Inside SVG, `a11y_invalid_attribute` names the wrong attribute. The element carries `xlink:href`, and rsvelte reports it as `href`.

Fixture: `validator/samples/a11y-anchor-in-svg-is-valid`.

**rsvelte:**

> `''` is not a valid href attribute

**official:**

> `''` is not a valid xlink:href attribute

## Why it matters

The message names the attribute the user must fix. In an SVG `<a>` the offending attribute is literally spelled `xlink:href`; telling the reader it is `href` points them at an attribute that is not there. If the element has both, it points them at the wrong one.

The warning is otherwise correct — the code and the fact of the warning are right, only the interpolated attribute name is wrong. That is the failure mode that survives longest, because everything except the text agrees.

## Why nothing caught it

Currently invisible to every gate:

- The corpus warning gate compares `(code, line, column)` only — message text is never compared (#2403).
- `DIAGNOSTICS_DIGEST` in `2_analyze/diagnostics_test.rs` pins code + message, but with **fixed placeholder arguments**, so an interpolation bug is structurally invisible to it; and its oracle is a human confirming the wording, so it detects change rather than incorrectness.
- `validator/samples/a11y-anchor-in-svg-is-valid` compares messages for real, but it is listed in `compatibility/validator-known-failures.json`, which is per-fixture and all-or-nothing — once listed for any reason, the entry is unmonitored for every reason.

## Reproducing

Compare against **official compiled on the identical input** (`fixtures/*/validator/a11y-anchor-in-svg-is-valid/warnings.json`), not the sample's checked-in `warnings.json` — see the oracle note in the sibling issue for why that distinction bites message comparison specifically.

Found while measuring #2403.

Issue #2411, opened 2026-08-06
compiler: `svelte_self_deprecated` suggests a self-import path with the wrong case (`./Input.svelte`)


baseballyama:
## The defect

`svelte_self_deprecated`'s message suggests a self-import, and rsvelte capitalises the **file path** as well as the identifier. The path it prints does not exist.

Fixture: `validator/samples/svelte-self-deprecated` (`filename` = `svelte-self-deprecated/input.svelte`).

**rsvelte:**

> `<svelte:self>` is deprecated — use self-imports (e.g. `import Input from './Input.svelte'`) instead

**official:**

> `<svelte:self>` is deprecated — use self-imports (e.g. `import Input from './input.svelte'`) instead

The identifier `Input` is correct in both — it is the basename capitalised, which is what a component identifier should be. The **path** must stay the real filename: `'./input.svelte'`, not `'./Input.svelte'`.

## Why it matters

The message is a copy-pasteable suggestion. On a case-sensitive filesystem (Linux CI, most containers) `import Input from './Input.svelte'` fails to resolve. A user who follows the deprecation advice gets a broken build, and the compiler told them to write it.

## Why nothing caught it

Currently invisible to every gate:

- The corpus warning gate compares `(code, line, column)` only — message text is never compared (#2403).
- `DIAGNOSTICS_DIGEST` in `2_analyze/diagnostics_test.rs` pins code + message, but with **fixed placeholder arguments** and with a human as the oracle ("update the constant after confirming the new wording matches official Svelte"). It detects *change*, not *incorrectness*, and it cannot see an interpolation bug at all because the interpolated value is a placeholder.
- `validator/samples/svelte-self-deprecated` compares messages for real, but it is listed in `compatibility/validator-known-failures.json`, which is per-fixture and all-or-nothing — once listed for any reason, the entry is unmonitored for every reason.

## Oracle note (matters for anyone reproducing this)

Do **not** diff against the sample's checked-in `submodules/svelte/.../svelte-self-deprecated/warnings.json`. Upstream committed that file under a different `filename` than the test harness passes, so it reads `import Self from './Self.svelte'` and any message that interpolates the filename disagrees spuriously. The valid oracle is **official compiled on the identical input** — the generated `fixtures/*/validator/svelte-self-deprecated/warnings.json`. Using the wrong one turns this into a completely different (and non-existent) bug.

Found while measuring #2403.

Issue #2403, opened 2026-08-06
corpus: the warning gate never compares message text


baseballyama:
## The gap

`scripts/compat-corpus/compile.mjs` records each warning as `(code, line, column)`; `verify.mjs` ratchets the codes and the positions separately. **The message text is never compared.** A warning can carry completely wrong wording on all 14,131 corpus entries and every gate stays green.

This is not hypothetical. #2391 found `attribute_quoted` reading:

> Quoted attribute **values** will be stringified in a future version of Svelte.

where upstream says:

> Quoted attributes on **components and custom elements** will be stringified in a future version of Svelte.

rsvelte's wording asserts the warning applies to plain elements, which it never does. It was wrong for as long as the code existed and nothing in the repo noticed.

## Why it survived: two stacked blind spots

1. **The gate is blind by construction.** Message is a third dimension alongside code and position, and the corpus never looks at it. No amount of corpus growth finds this class — it is the same shape as #2281's original defect, where the pipeline discarded `result.warnings` outright and a warning divergence scored `MATCH` on every entry.

2. **The one test that would have caught it was suppressed for an unrelated reason.** `validator/samples/attribute-quoted` compares codes, messages and positions, and it is listed in `compatibility/validator-known-failures.json` — justified there by `svelte_component_deprecated` having no span in the same file. So the entry was silencing the message defect while its justification named something else entirely.

The generalisation is worth stating in the ratchet docs: **a known-failure entry suppresses everything about that entry, not the thing its justification names.** Once listed for any reason, an entry is unmonitored for every reason. That argues for justifications that say what the entry *stops covering*, not only why it fails.

## The axis this sits on

The lesson from #2281 was that the question to ask is **"what does the oracle not look at?"**, not "how big is the corpus". Corpus size is saturated as a growth axis. The two axes that were not: what we compare (warning parity, #2281), and how inputs are constructed (the generated shape matrix). Message text is the next unexamined column in the first of those, and it is free — both compilers already run on every entry and both already return the string.

## Proposal: a third ratchet

`warning-message-known-failures.{client,server,client-dev}.json`, shrink-only, justified per entry in a paired `.md`, on the same argument that split codes from positions:

- A wrong message is **semantic** — the user reads it and acts on it. It belongs with the codes in severity, not with the positions.
- Folding it into the position backlog (currently the larger by an order of magnitude) would hide it, which is precisely why codes and positions were split in the first place.

Comparison notes:

[...]

PR #2418, opened 2026-08-06
test(validator): ratchet warning message text against the same-input oracle (#2403)

baseballyama:
Part of #2403. Surfaces #2391, #2411, #2413.

This is the **first half** of #2403 — the fixture-side gate. It deliberately does not use a
closing keyword: the corpus-side message ratchet the issue proposes is still to come.

## Why here first, and why this is not the ratchet the issue proposed

#2403 proposes a third **corpus** ratchet for warning message text. Before sizing it I ran the
sweep the issue asks for as a follow-up — classifying all **207** `validator-known-failures.json`
entries by which dimension actually diverges — and it changed the order of work:

| bucket | count |
|---|---|
| **message-only** (position fine) | **0** |
| **message AND position** | **3** |
| position-only | 59 |
| code-level | 1 |
| other (mostly fixtures that legitimately error) | 144 |

The validator suite **already compares message text** against official (`validator.rs:202`).
The comparison exists and is being thrown away, because `validator-known-failures.json` is
per-fixture and all-or-nothing. So the cheapest gate with proven yield needs no corpus run, no
artifact regeneration and no new oracle — just stop discarding a comparison we already do.

**"0 message-only" is not a refutation of #2403 — it is its generalisation confirmed exactly.**
Every one of the three also has a position divergence, so all three are listed in the ratchet
for a *position* reason and are still unmonitored for *message*. An entry suppresses everything
about itself, not the thing its justification names.

## The three, with both-sides text

- **`attribute_quoted`** — rsvelte "Quoted attribute **values**" vs official "Quoted attributes
  on **components and custom elements**". Claims the warning applies to plain elements, which it
  never does. (#2391)
- **`svelte_self_deprecated`** — rsvelte `import Input from './Input.svelte'` vs official
  `'./input.svelte'`. The identifier is right; the **path** is capitalised and does not exist, so
  following the suggestion breaks the build on a case-sensitive filesystem. (#2411, filed from
  this work)
- **`a11y_invalid_attribute`** — rsvelte "`''` is not a valid **href** attribute" vs official
  "**xlink:href**". Names an attribute that is not on the element. (#2413, filed from this work)

## Why `DIAGNOSTICS_DIGEST` does not already cover this

`2_analyze/diagnostics_test.rs` pins every diagnostic's code and message — so message text is
not entirely unwatched, and the case for this gate has to be made on what the digest cannot do,
not on "nothing looks at it". It calls each constructor with **fixed placeholder arguments**,
and its failure message asks a human to confirm the new wording matches official Svelte. So
**the oracle is a person and the unit is the template**: a change-detector, not a correctness
comparator. `attribute_quoted`'s wrong wording was baked into the digest on day one and stayed
green forever, and an interpolation bug like #2411/#2413 is structurally invisible to it because
the interpolated value is a placeholder.

## The oracle: "official run on this input", never "the official expectation"

The gate reads the **generated** fixture (`fixtures/*/validator/<name>/warnings.json`, official
compiled on the identical input), not the sample's checked-in
`submodules/svelte/.../warnings.json`.

This is load-bearing, and I got it wrong first. Upstream committed those files under a different
`filename` than the harness passes. Measured against the checked-in file,
`svelte-self-deprecated` reads as `Self`/`./Self.svelte` vs `Input`/`./Input.svelte` — a bug that
does not exist — instead of the real path-capitalisation bug. I was one step from reporting a
real defect as a *different* defect, which is worse than a false positive: the wrong fix would
have earned a green tick and closed the issue.

It bites message comparison specifically, because codes and positions do not depend on the
input's *name*. The corpus gate is immune by construction — it runs both compilers on the same
source in the same process — so **a fixture-side gate has to reproduce that property
deliberately**. That reasoning is recorded in the paired `.md` for whoever adds the next one.

## Proof that it discriminates

Each mutation applied, run, reverted.

| # | mutation | result |
|---|---|---|
| **N4** | perturb a real, currently-matching message in `warnings.rs` (`a11y_missing_content`: "should contain text" → "should contain content") | **kills**: names `a11y-heading-has-content` with both-sides text |
| **N4-control** | the **existing** `test_validator` suite under that same mutation | **PASSES** — `a11y-heading-has-content` is listed in `validator-known-failures.json`, so the old gate is structurally blind to it |
| **N1** | drop `attribute-quoted` from the new ratchet | **kills**: "diverge … and are not in validator-message-known-failures.json" |
| **N2** | add a passing fixture to the ratchet | **kills**: "now match — remove them" (two-sided) |
| **N3** | delete the generated `warnings.json` oracles | **kills**: "only 0 fixture(s) reached message comparison, floor is 158" |

**N4 with its control is the whole argument**: one input, two opposite results. A real message
regression is invisible to the gate that exists and caught by the gate this PR adds. N1/N2 prove
the ratchet is two-sided; N3 proves it cannot pass by measuring nothing.

The floor is the measured value (158 of 334 samples emit warnings whose codes already line up),
grow-only, matching `MIN_VALIDATOR_FIXTURES` / `MIN_SOURCEMAP_FIXTURES` convention.

## Wiring

`validator` is an existing integration-test target and `scripts/ci/run-test-shard.sh` excludes
only `compile_module_thread_safety`, `svelte_dev_corpus` and `svelte_dev_markdown`, so the new
`#[test]` is picked up by the `test` shard automatically. Written **and** wired.

## Deliberately not claimed

Three entries (`a11y-alt-text`, `a11y-no-noninteractive-element-to-interactive-role`,
`inline-new-class-2`) now match on **warnings**, but I have not verified they pass the full
`test_validator`, so I am not touching `validator-known-failures.json`. A two-sided ratchet fails
on a listed entry that passes, so a wrong claim there would break CI for everyone.

## Changeset

None. Test-only: one `#[test]` plus two `compatibility/` ratchet files. No shipped crate
behaviour changes — the three defects are recorded, not fixed, and each has its own issue.


baseballyama:
Reviewed at `52335e13`. **0 blocking.** The N4 control is stronger than "we ran it once" — it is blind *by construction*, and I verified the two facts that make it a one-variable experiment. One real finding on the floor: it is coherent and tight, but it counts the wrong unit, and I can put numbers on how much padding is in it. Read-only — no build, no test run.

**CI position**: every `Test (ubuntu-latest, N)` shard is still pending, and the shard that carries `validator` is the *only* evidence that `compared >= 158` is achievable outside the author's machine. Content signoff with that outstanding.

### The N4 control: verified structurally, not taken on report

Two independent facts, both checked against the pinned submodule (`44a78137`, identical in this PR and on main):

1. **`a11y-heading-has-content` is in `validator-known-failures.json` (line 17).** The existing ratchet is per-fixture and all-or-nothing, so `test_validator` is blind to *anything* about that fixture. The control does not pass by luck under this particular mutation — it would pass under every message mutation to that fixture, which is the property the argument needs.
2. **Exactly one of the 334 samples expects `a11y_missing_content`.** I enumerated the sample tree rather than assuming. So the mutation moves exactly one fixture and nothing else in the suite could have been the reason either gate did or did not fire. That is what makes "one input, two opposite results" a fact about the *gates* rather than about the blast radius of the mutation.

I checked the same property for the three ratchet entries: `a11y-anchor-in-svg-is-valid`, `attribute-quoted` and `svelte-self-deprecated` are at lines 3, 46 and 189 of `validator-known-failures.json`. So "all three were suppressed behind a position justification" is verified per entry, not asserted — and it is the concrete instance of the generalisation the `.md` states.

### One row missing from the N4 table, and it sharpens rather than weakens the claim

`DIAGNOSTICS_DIGEST` (`diagnostics_test.rs:37`) is an FNV-1a over `"<code>\t<message>\n"` for **every** constructor, so the N4 mutation would also turn `diagnostic_codes_and_messages_are_pinned` red — in `test-unit`, not in the shards. The table's `test_validator` row is true as written, but a reader takes away "a real message regression is invisible today", and that is not quite the shape. The honest form is three-valued, and this PR's own prose already argues it:

| gate | under N4 | why |
|---|---|---|
| `DIAGNOSTICS_DIGEST` | **red** | detects the *change*; failure text asks a human to confirm the wording |
| `test_validator` | **green** | fixture is listed — blind to everything about it |
| new message ratchet | **red** | compares against official run on the same input |

The middle column is the interesting one: the digest's red is a prompt to re-baseline, and `attribute_quoted`'s wrong wording sitting green in that digest since day one is the empirical proof that the human step blesses incorrect text. Adding the digest row makes the table say what §"Why `DIAGNOSTICS_DIGEST` does not already cover this" already says, and forestalls the weaker reading.

### The floor: coherent, tight, and counting fixtures rather than messages

I cannot recompute 158 without the generated tree, so I bounded it from the **sample population**, which is independent of both compilers. The 334 samples partition cleanly:

| population | count | reaches `compared`? |
|---|---|---|
| non-empty `errors.json` | 167 | **no** — `compile` returns `Err`, and `let Ok(Ok(result)) … else { continue }` skips |
| ≥1 expected warning | 95 | yes |
| no expectation at all (43) + **empty** `errors.json` (29) | 72 | yes, with an empty warning list |

167 + 95 + 72 = 334, and 0 samples carry both. So the **ceiling on `compared` is 167**, and the floor sits 9 below it. Two things follow:

- **158 is coherent with the mechanism the PR claims.** It exceeds 334 − 207 = 127, which is only possible if fixtures *listed* in `validator-known-failures.json` are reaching comparison — exactly the thesis. Had the floor come in at or below 127 it would have quietly contradicted the argument. And it is a genuinely tight floor, not a comfortable one.
- **But its unit is "fixtures reached", not "messages compared".** 72 of the ≥158 are zero-warning fixtures: they take `compared += 1` and then iterate an empty `zip`. The quantity the assert names can therefore sit up to 72 above the quantity the gate exists to measure, and the teeth come from the accident that 158 happens to be within 9 of the ceiling. A second counter costs one line and floors the real thing:

```rust
messages += result.warnings.len();   // floor it too
```

Upstream expects **549** warnings across those 95 samples, so that floor has two orders of magnitude more resolution than the fixture count and cannot be padded by empty fixtures at all. Non-blocking — but it is the same distinction this PR draws about `DIAGNOSTICS_DIGEST`: measuring the proxy and measuring the thing are different, and here the proxy is 46% padding.

For what it is worth, N3 (delete the oracles → `only 0 fixture(s)`) proves the floor fires at zero, which is the important half. The suggestion above is about the range between 72 and 158, not about zero.

### Wiring: verified, and "automatically" is the right word

`run-test-shard.sh` builds `ALL_TARGETS` from `cargo metadata` and drops exactly three binaries by name (`compile_module_thread_safety`, `svelte_dev_corpus`, `svelte_dev_markdown`); the separate `RUN_FILTER` excludes three *test names*, none of which matches `validator_warning_messages_match_official`. `validator` is an existing integration target, so a new `#[test]` inside it is picked up with no CI change. Written **and** wired, as claimed. (Which shard runs it is a modulo over the sorted target list, so it will move as targets are added — deterministic, but not something to hardcode anywhere.)

### The oracle choice is the most valuable thing here

Reading the **generated** fixture rather than the sample's checked-in `warnings.json` is right, and the reason it is right generalises past this gate: a committed expectation file is "what upstream's CI produced on upstream's input", not "official run on this input". Any message that interpolates the filename diverges spuriously against it, which is why this bites *message* comparison and no earlier gate hit it — codes and positions do not depend on the input's name.

I verified the trap's shape independently: upstream's committed file for `a11y-anchor-in-svg-is-valid` carries `'' is not a valid xlink:href attribute`, while the harness passes `<name>/input.svelte` as `filename`. Reporting a real defect as a *different* defect is the failure mode, and "the wrong fix earns a green tick" is exactly why it is worse than a false positive. Recording that in the paired `.md` for the next gate author is the part I would keep.

### Two nits

**Naming.** The PR body heads its three bullets with warning **codes** (`a11y_invalid_attribute`) while the JSON lists fixture **names** (`a11y-anchor-in-svg-is-valid`), and there is no sample called `a11y-invalid-attribute`. A reader cross-checking the body against the ratchet finds two of three. The `.md` gets this right — it writes `fixture — code` pairs — so it is only the PR body.

**The new `.md` is not gated.** `known-failures-md-check.mjs` covers `known-failures.md`, `warning-known-failures.md` and `matrix-known-failures.md`; nothing checks that `validator-message-known-failures.md` has an entry for each id in the `.json`. So the correspondence this file establishes is by convention only, and the first entry added without a justification will not be noticed. Same class as the residual on #2384, and the cheap fix is the same shape — extend the existing script rather than adding a new one.

PR #2426, opened 2026-08-06
fix(compiler): attribute every validation diagnostic to its source range

baseballyama:
Every PR's Compatibility Report showed `validator` at 192/334 and `svelte2tsx` at 249/254. Neither was a new regression: both were pre-existing divergences parked in shrink-only ratchets, and the report does not consult the ratchets, so it printed the raw numbers. This closes both.

| | before | after |
|---|---|---|
| validator (Compatibility Report) | 192/334 | **334/334** |
| svelte2tsx | 249/254 | **254/254** |
| validator gate (code/message/span compared strictly) | 128/332 | **332/332** |
| `compatibility/validator-known-failures.json` | 207 | **0** |
| `compatibility/svelte2tsx-fixtures-known-failures.json` | 5 | **0** |

Note the two views disagree by construction: the report compares only the *number* of warnings a successful compile emits, so 59 warning position divergences and 4 wording divergences were invisible to it. The gate is the only thing that sees them. Both are fixed here.

## Approach

Error and warning constructors still build a span-less diagnostic; each raising site attaches the range with `AnalysisError::at(start, end)` / `AnalysisWarning::at(start, end)`, taking the node upstream passes to its `e.*` / `w.*` constructor. That node is frequently *not* the one the enclosing visitor is looking at — `attribute_invalid_type` points at the sibling `type` attribute rather than the `bind:` directive, `constant_assignment` at the assignment rather than its target.

For accessibility the cause was concrete: `a11y::check_element` returned span-less warnings and `regular_element.rs` back-filled every one of them with the whole element's range, so each a11y diagnostic pointed at `<div …>` instead of the offending attribute. The back-fill stays, but only as the fallback for warnings upstream really does attribute to the element.

## Structural fixes that fell out

Five things could not be fixed by attaching a span at the call site:

1. **`svelte_element_missing_this` moves to the parser**, where upstream raises it. By Phase 2 the `this` attribute has been folded into `tag`, so a missing `this` is indistinguishable from a valueless one and the attribute's span is gone.
2. **`ParseWarning` carries a span**, so `element_implicitly_closed` keeps its position across the hop from the parser into analysis.
3. **`unknown_code` / `legacy_code` emission order.** They were emitted up front; upstream raises them from the node that collects the preceding comment. The gate zips actual against expected, so order is asserted, not incidental.
4. **`compile_module` marks its input as a module directly** instead of inferring it from a `.svelte.(js|ts)` filename. Upstream knows this from the `analyze_module` entry point; the filename is optional. This adds one non-upstream `#[doc(hidden)]` field to `CompileOptions` — threading it internally would have touched every `ComponentAnalysis::new` caller.
5. **The `context` attribute stays on the `Script` node**, as upstream's `read_script` leaves it, so `script_context_deprecated` can point at it.

The validator harness now passes no `filename`, mirroring upstream's `tests/validator/test.ts`. That is what `svelte_self_deprecated` needs to reach the unset-filename sentinel (`Self` / `Self.svelte`), and it is what exposed the module-detection dependency in (4).

## svelte2tsx

- `preserveAttributeCase` was never honoured, so attributes on foreign-namespace elements were lower-cased. The fixture harness also never derived `namespace` from the `-foreign-ns` suffix, which upstream's `test/helpers.ts` does.
- `type $$ComponentProps` and module-level snippets were emitted in the wrong order relative to `function $$render()`.
- Angle-bracket type assertions in an instance script were rewritten to `as`. Upstream gates that on `mode !== 'ts'`; the module script still rewrites unconditionally. One unit test pinned the pre-fix behaviour and was corrected.

## Verification

`validator` `compiler_errors` `compiler_fixtures` `parser_fixtures` `css` `ssr` `print` green; `runtime` all 5 (legacy 1207, runes 1007, hydration, browser) green; `--lib` 1403 green; `rsvelte_projection` and `rsvelte_lint` green. `cargo fmt --check` clean, `cargo clippy --all-targets` clean.

Locally the Compatibility Report itself reported `validator 333/333 (100.0%)` but then died after `runtime-runes` — no panic, exit 144 — on this machine with the disk at 98%. The same runtime suite passes standalone, and main's CI artifact completes through `runtime-legacy`, so this looks environmental rather than a regression. The `Compatibility Report` job here is the real check.

https://claude.ai/code/session_01WhzBu6eT1ai7ZRaq8DrJhN


baseballyama:
## Review — the span work is faithful; the PR cannot be approved until it rebases, re-baselines and runs CI once

Recommendation: **do not merge as-is.** Merge after (1) the rebase, resolved as described below,
(2) the corpus **warning** ratchets re-baselined *in this PR*, and (3) one green CI run — which this
branch has never had. The content review below is so the rebase lands in a known-good state.

### 1. The conflict — one file, one competing commit, and the resolution is not "take main"

Exactly **one** file conflicts, in **two** hunks:
`crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/shared/a11y/mod.rs`. Everything else in
the 76-file diff merges clean.

The competing commit is **`122af14b` "fix(analyze): give attribute-scoped a11y warnings the
attribute's span (#2384)"** — the suspicion is confirmed, and it is the only cause. This PR was
opened 103 seconds after #2384 merged, so it has been conflicting since creation. The two are
competing implementations of the same fix: #2384 stamps spans after the fact
(`stamp_attribute(&mut warnings[mark..], attr)` plus an `ELEMENT_SCOPED_CODES` opt-out), this PR
attaches them at the emit site (`.at(a.start, a.end)`). The per-site form is the closer port —
upstream passes a node to every `w.*` call.

**Hunk 1 is not a mechanism choice, and resolving it the easy way would reintroduce a bug.**

```rust
// main (122af14b), a11y/mod.rs:490
warnings.push(w::a11y_invalid_attribute(href_value, "href"));
// this PR
warnings.push(w::a11y_invalid_attribute(href_value, &a.name).at(a.start, a.end));
```

The lookup above it is `attribute_map.get("href").or_else(|| attribute_map.get("xlink:href"))`, and
upstream is
`w.a11y_invalid_attribute(href, href_value, href.name)`
(`packages/svelte/src/compiler/phases/2-analyze/visitors/shared/a11y/index.js:408`) — the
attribute's **own** name. Main's literal `"href"` therefore emits
`'#' is not a valid href attribute` where official emits `… valid xlink:href attribute`.

That is not hypothetical: upstream fixture `a11y-anchor-in-svg-is-valid` asserts the `xlink:href`
wording, and `a11y-anchor-in-svg-is-valid` **is** currently listed in
`compatibility/validator-known-failures.json` (207 entries on main). This PR empties that file to
`[]`, i.e. it fixes it. So the rebase must keep **this PR's `&a.name`**, whichever span mechanism
survives — and must keep `crates/rsvelte_core/tests/a11y_attribute_warning_span.rs`, added by #2384
and absent from this PR's file set.

Hunk 2 (`a11y_autocomplete_valid`) is a genuine mechanism-only overlap; either side is fine.

### 2. There is no CI, and that is a consequence of the conflict

`gh pr checks 2426` → *"no checks reported"*. Not a flake:
`actions/runs?head_sha=d427578e` → `total_count: 0`, and the same for `05f3d155`. `ci.yml` and
`corpus-compat.yml` are `pull_request`-triggered, GitHub cannot build `refs/pull/2426/merge` for a
conflicting PR, so nothing was ever scheduled.

A **76-file, +1434/−817** change touching the parser, the analyzer, the CSS reader and svelte2tsx
currently has **zero** CI evidence. The rebase is a precondition for review, not a formality.

### 3. Correctness against upstream — checked, and it holds

~15 changed diagnostics were compared against the pinned reference (`submodules/svelte` @
`44a78137`, v5.56.8) with no mismatches, including:

| diagnostic | this PR | upstream |
|---|---|---|
| `svelte_element_missing_this` (absent / valueless) | `(start, start)` / attribute span | `element.js:289` / `:295` |
| `script_duplicate` | `(start, start)` | `element.js:353/356` |
| `directive_invalid_value` | `self.index` (past the quote) | `element.js:678` |
| `css_empty_declaration` | `start … first ':' + 1` | `read/style.js:475` |
| `css_selector_invalid` | `parser.index` / combinator node | `read/style.js:363`, `css-analyze.js:151` |
| `block_unexpected_character` | `(start, start + 5)` | `shared/utils.js:135` |
| `constant_assignment` / `constant_binding` | outer assignment node threaded through the recursion | `shared/utils.js:84/119` |
| `a11y_invalid_attribute` | attribute node + `a.name` | `a11y/index.js:408` |

The two structural moves check out as well: upstream `read_script` (`1-parse/read/script.js:88`)
returns `attributes` unfiltered, so keeping the `context` attribute matches upstream and retires a
lenient/strict fork; and upstream raises `svelte_element_missing_this` in the parser and splices the
`this` attribute out, so both the move and the `tag` folding are faithful.

### 4. The spans are load-bearing — proven by removing them

`tests/validator.rs` compares `code` + `message` + `start` + `end` in an ordered zip against
upstream's `warnings.json` / `errors.json`, and this PR empties the validator ratchet to `[]`, so no
entry can absorb a span regression. Baseline on the PR tree:
`cargo test -p rsvelte_core --test validator` → **2 passed (332/332)**. Reverting three spans:

| reverted | tests that fail |
|---|---|
| `.at(on.start, on.end)` on `event_directive_deprecated` | `runes-legacy-syntax-warnings` |
| `binding_node_span` → `None` | `unreferenced-variables`, `unreferenced-variables-each`, `runes-referenced-nonstate`, `runes-referenced-nonstate-bind-this` |
| `.at(a.start, a.end)` on `a11y_autocomplete_valid` | `a11y-autocomplete-valid` |

`6 validator regressions`, assertion fired, files restored byte-exactly. The failure output also
confirms `binding_node_span`'s arithmetic (`28:6..28:8` for `d2`, `32:13..32:15` for `g2` =
`declaration_start + name.len()`).

Worth knowing about that gate: the validator ratchet is **one-sided** — `fixed` is computed and
printed but never asserted (`validator.rs:429-441`), unlike the corpus ratchets. Harmless while the
list is `[]`; it will not catch a stale baseline later.

### 5. Blocking: the corpus warning ratchets are not re-baselined

`verify.mjs:526-534` is explicitly two-sided (*"N warning baseline entries already PASS — the
ratchet is stale"*), and `corpus-compat.yml` runs on `pull_request` for `crates/**` and
`compatibility/**` — both touched here. This PR modifies **no** corpus ratchet.

Per `compatibility/warning-known-failures.md` on main, the 529 remaining position entries are
dominated by five missing-span codes: `event_directive_deprecated` (142),
`element_invalid_self_closing_tag` (118), `export_let_unused` (110), `non_reactive_update` (102),
`options_missing_custom_element` (53) — **525 of the 649 missing-span tuples. This PR attaches a
span to every one of those five.** The warning-*code* ratchet (70 entries) is affected too: it lists
a `svelte_self_deprecated` over-warning cluster, and this PR gates that warning on
`analysis.runes`.

So after the rebase, `warning-position-known-failures.{client,server,client-dev}.json`,
`warning-known-failures.*.json` and the header counts in `warning-known-failures.md` (enforced by
`known-failures-md-check.mjs:96-125`) will all be stale and Corpus Compat will fail. Per the repo
rule, the re-baseline lands **in this PR**:
`node scripts/compat-corpus/verify.mjs --update-warning-baseline` — and **never** in the same
invocation as `--update-baseline`, which is a silent no-op (#2350). The output ratchets may move
too: `node_invalid_placement` is an error, and the hyphen change below can flip `error-mismatch`.

### 6. Scope — this is not only positions

Each of these is correct against upstream, but each belongs to a *different* ratchet family than
"positions", and none has a changeset:

1. **`svelte_self_deprecated` gated on `analysis.runes`** (`visitors/svelte_self.rs`) — a warning-**set**
   change. Matches `SvelteSelf.js:27`.
2. **`reset_by` ancestor walk gains `|| a.contains('-')`** (`visitors/regular_element.rs:~941`) — an
   error-set change; `node_invalid_placement` stops firing for some inputs. Matches
   `html-tree-validation.js:154`, and the `.any()` reformulation is sound because both upstream
   branches `return null`. **No test covers the new hyphen case.**
3. **`attribute_quoted` reworded**, forcing a `DIAGNOSTICS_DIGEST` bump — wording is its own class.
4. **CSS selector splitting rewritten** (`1_parse/read/style.rs:2394-2414`): backslash escapes,
   quoted strings and `[`/`]` now affect depth. A **CSS parser behaviour change**, unrelated to
   spans, in no changeset, with **no test**.
5. **The svelte2tsx half (17 files)** is orthogonal to validation spans. Its two riskiest claims do
   hold against pinned upstream — `processInstanceScriptContent.ts:341` really is
   `if (mode !== 'ts' && ts.isTypeAssertionExpression?.(node))` while `processModuleScriptTag.ts:114`
   is unconditional, so the changed unit test was corrected *toward* upstream rather than bent to the
   new code; and `test/helpers.ts:383` really is
   `namespace: sampleName.endsWith('-foreign-ns') ? 'foreign' : null`, so the harness change makes
   the suite stricter.

Splitting svelte2tsx into its own PR would make both halves reviewable and let the compiler half
land without waiting on the other. Given there is no CI history at all, a smaller first PR is also
the cheaper way to get a green run.

### 7. Non-blocking

- `CompileOptions::is_module_source` (`compiler/mod.rs:260`) is a non-upstream `pub` field on the
  drop-in options struct, documented as "leave it `false`". `#[doc(hidden)]` hides it from docs, not
  from the API — a `compile_module`-internal parameter avoids the public-surface leak.
- `binding_node_span` (`2_analyze/mod.rs:1048`) reconstructs the identifier end as
  `declaration_start + name.len()`. Correct for every fixture, but `declare_binding` now *receives*
  the real `(start, end)` and discards it; storing it on `Binding` removes both the heuristic and its
  only failure mode (unicode-escaped identifiers).
- `find_rest_binding_span` / `find_rest_identifier` (`visitors/bind_directive.rs:280-340`)
  re-implement `ScopeBuilder`'s pattern traversal to recover a span `Binding` does not keep. Two
  copies of one traversal will drift; the same `Binding`-span fix collapses them.

### 8. Prior review

Only `copilot-pull-request-reviewer[bot]` reporting a quota limit. No inline comments, no human
reviews — nothing outstanding.

---

### Rebase checklist

1. Rebase `fix/validator-error-spans` onto `origin/main` (currently `a8a653d5`).
2. In `a11y/mod.rs`: keep one span mechanism, keep **this PR's `&a.name`** for
   `a11y_invalid_attribute`, keep `tests/a11y_attribute_warning_span.rs`.
3. `node scripts/compat-corpus/verify.mjs --update-warning-baseline` (alone), then update the header
   counts in `compatibility/warning-known-failures.md`.
4. Add a test for the `reset_by` hyphen case and for the CSS selector-splitting change, or split both
   out.
5. Changesets for the set/wording changes in §6.

baseballyama:
### Supplement — four items the first comment missed

A second pass over the same diff turned up these. The recommendation is unchanged (rebase →
re-baseline → first CI run); items 1 and 2 add to the rebase checklist.

**1. This PR falsifies a checked-in ratchet justification, not just its counts.**
`compatibility/warning-known-failures.md` justifies `a11y_figcaption_index` as *"structurally out of
reach rather than merely unobserved"* — because upstream raises it outside the attribute loop, which
is exactly where main's `stamp_attribute` mechanism (#2384) cannot reach. This PR's per-site `.at()`
does reach it. So the re-baseline is not a re-count: that justification and the
`svelte_self_deprecated` cluster prose both have to be **rewritten**, and the "structurally out of
reach" phrasing should not survive. This also settles the a11y conflict on grounds beyond the
`xlink:href` message: #2426's file is a strict superset of #2384's, not an alternative to it.

**2. `svelte_element_missing_this` becoming a *parse* error is a behaviour change for parse-only
consumers.** It is correct parity — upstream raises it in `1-parse/state/element.js:289` — but the
blast radius is wider than the validator: `parse()`, `rsvelte-fmt`, `rsvelte_lint` and svelte2tsx
now hard-fail on `<svelte:element>` without `this`, where before they got an AST. Those suites
(Parser Modern/Legacy, Print, the fmt-parity corpus, the lint ratchet) have **never run on this
branch**, per the zero-CI finding. Worth calling out in the changeset.

**3. Two more span omissions in the same family as `binding_node_span`.** Both are fixed by the same
"store the declaring identifier's real `(start, end)` on `Binding`" change already recommended:
- `LetDirective` passes `None` to `declare_binding` where upstream `scope.js:1103-1109` has a real
  span, so `<Comp let:x let:x />` yields a span-less `declaration_duplicate`.
- `mixed_event_handler_syntaxes` (`2_analyze/visitors/mod.rs:683-690`) *tests*
  `context.event_directive_node` but *reads the span from* `context.analysis.event_directive_node` —
  two independently-set fields. When they diverge the span is silently dropped. A single source for
  both is the fix.

**4. `binding_node_span`'s `None` paths are reachable, not just theoretical.** Beyond the
unicode-escape arithmetic issue, `declaration_start` is never set on four declare paths
(`scope_builder.rs:2278`, `:2311`, `:2683`, `:3326`, `:3384` — `export function` / `export class` via
oxc, and pattern-declared bindings). `non_reactive_update` on `export function foo(){}; foo = bar`
lands there and silently loses its span — which matters because `non_reactive_update` is 102 of the
529 position-ratchet tuples this PR is meant to retire. Worth a fixture before the re-baseline, so
the ratchet does not record a partial fix as a complete one.

One process note on the first comment: my §3 upstream table and §4 revert experiment were written up
from a parallel review pass rather than re-run by me at the time of posting. I have since confirmed
the load-bearing pieces independently — the sole conflicting file and its two hunks (`git merge-tree`
against merge-base `88b8d2b7`), the competing commit `122af14b`, upstream
`a11y/index.js:408` passing `href.name`, the upstream fixture `a11y-anchor-in-svg-is-valid` asserting
the `xlink:href` wording, that sample's presence in main's 207-entry
`compatibility/validator-known-failures.json`, and `gh api …/check-runs` + `actions/runs?head_sha=`
both returning 0 for `d427578e`.

baseballyama:
Merged current `main`. The `foreign` namespace work in this branch collided with #2528, which landed the same feature independently — worth recording how it was resolved, because the two implementations are not interchangeable line-for-line.

**Resolved to `main`'s implementation** for every conflicting file: `interfaces.rs`, `template/attributes/attribute.rs`, `template/nodes/{element,dynamic_element,special_element}.rs` and `tests/common/svelte2tsx.rs`. `main` reads the flag straight off `options.namespace.preserves_attribute_case()` at each attribute site; this branch threaded it through `Counter`. Keeping both left `Counter::preserve_attribute_case` written and never read — a `dead_code` warning, which is `-D warnings` in CI — so the redundant ctx field and its `Counter::new` parameter are removed here.

**The branch's other four svelte2tsx fixes are untouched**: `snippet_hoisting.rs`, `script/type_assertion.rs`, `nodes/generics.rs` and `template/mod.rs` did not conflict, so `module-snippet-component-instance-reference.v5`, `ts-runes-hoistable-props-false-6.v5`, `ts-await-generics.v5` and `ts-type-assertion` still pass. `svelte2tsx-fixtures-known-failures.json` stays at 0 entries; `attributes-foreign-ns` is the one `main` had already cleared, which is why the set merge of the two ratchets nets to empty rather than losing anything.

`a11y/mod.rs` had the same duplicate shape at a smaller scale: `main`'s version re-matched `href_attr` inside a block this branch had already guarded with `if let AttributeNode::Attribute(a)`. Folded into the outer binding.

`validator-message-known-failures` is now 0 entries here and on `main` (#2565), so this PR no longer moves it.

Verified locally: `cargo test -p rsvelte_projection` (all suites, 0 failed), `cargo test -p rsvelte_core --test validator` 3/3, `known-failures-md-check` 19/16, and `cargo clippy -p rsvelte_core -p rsvelte_projection --all-targets --all-features -- -D warnings` clean.

baseballyama:
I reduced all four of the new svelte2tsx corpus failures. **Three are fixed on this branch** (pushed as `fix(svelte2tsx): keep a dropped whitespace node's entity out of the TSX`); the fourth needs #2589.

I checked each verdict against the gate's own pipeline — `svelte2tsx-compile.mjs`'s exact options, `oxfmtTree` + `stripBlankLines`, and the `oracle-invalid` rule — with `main` as a control, and `main` reproduces its CI verdict on every one.

## 1–3. `&nbsp;` in a dropped whitespace node

`layercake`, `svar-core/RichSelect`, `svelte-ux/Field` and the `non-empty-block-dev` fixture.

Routing `{#if}` / `{#each}` / `{#key}` through `process_fragment_trimmed` brought them under `remove_surrounding_whitespace_nodes`, which **drops** a whitespace-only first/last `Text` child — and a dropped node is never visited, so its source range survives verbatim. `&nbsp;` decodes to U+00A0, which `char::is_whitespace` accepts, so the node was classified from its `data` while its six raw characters stayed in the TSX. The output then does not parse, oxfmt skips it, and the whole file is compared as raw text — which is why the reported first diff was `;function $$render() {` on line 2 rather than anything near the entity.

The classification now also requires the **raw range** to be whitespace. `Text.ts` already documents this exact hazard for the blanking path; the drop path is the same hazard with the node removed instead. `<svelte:boundary>` has it too and fails on `main` — that test in the commit is a pre-existing bug, not one this branch introduced.

## 4. `powertable/PowerTable.svelte` — an exposure, not a regression

This one is **not** this branch's defect. Gating `rewrite_type_assertions` on `is_dts_mode` is correct — `processInstanceScriptContent.ts:342` really does gate on `mode !== 'ts'` while `processModuleScriptTag.ts:115` rewrites unconditionally. The consequence is that rsvelte now leaves `<SortString>e` alone, exactly as official does, so **both** sides become unparseable; the entry stops qualifying for `oracle-invalid` and falls back to a raw-text comparison.

That raw comparison then surfaces three lines of pre-existing divergence, which reduce to two shapes:

```
{@html (a ?? '')}   official ` a ?? '';`     rsvelte `a ?? '';`
{(b ?? '')}         official `(b ?? '');`    rsvelte `b ?? '';`
```

Both reproduce on `main`. #2589 fixes them by rewriting the brace positions the way upstream does.

## Combined

Building the NAPI binding from this branch + #2589 and running the gate:

| entry | before | after |
|---|---|---|
| `svelte/…/non-empty-block-dev/input.svelte` | ts-mismatch | **MATCH** |
| `svelte-ux/…/Field.svelte` | ts-mismatch | **MATCH** |
| `svar-core/…/RichSelect.svelte` | ts-mismatch | **MATCH** |
| `powertable/…/PowerTable.svelte` | ts-mismatch | **MATCH** |

So: merge #2589, rebase this branch on it, and Compiler parity should go green.

baseballyama:
Merged current `main`. The only conflict was `validator-known-failures.{json,md}`, resolved to this branch's side and then **verified by running the suite rather than by reasoning about the merge**: against `main` as of now, `test_validator` passes with the ratchet at **0 entries**.

That is the two-sided check doing its job in both directions — a stale entry would have failed just as loudly as a new failure, so 0 here means all 332 runnable fixtures match upstream on code, message and position, not that the list was emptied.

Also green locally on the merged tree: `--lib` 1523, `compiler_fixtures`, `css`, `ssr`, `cargo test -p rsvelte_projection`, `known-failures-md-check`, and `cargo clippy -p rsvelte_core -p rsvelte_projection --all-targets --all-features -- -D warnings`.

Still waiting on #2589 for the last svelte2tsx corpus entry (`powertable`); rebase once that lands.

baseballyama:
Merged `main` and re-baselined the **error-position** ratchet, which this PR made stale in 531 places (177 per target):

```
error-position-known-failures.client.json      403 -> 226
error-position-known-failures.server.json      403 -> 226
error-position-known-failures.client-dev.json  403 -> 226
error-message-known-failures.*                 unchanged (121 / 120 / 121)
```

The `warning-position-*` ratchets were already re-baselined here; `error-position-*` was the half that was missed, which is what `Compiler parity` was failing on.

The cluster table in `error-known-failures.md` is **re-measured from the run's own `report.json`**, not subtracted, and the measurement is worth reading:

| shape | was | now |
|---|---:|---:|
| rsvelte reports no span at all | 349 | **174** |
| same line, different column | 35 | 33 |
| different line entirely | 19 | **19** |
| | 403 | 226 |

The shrink is **entirely inside the no-span cluster**, and the different-line entries are the same 19. That is the shape a span-attachment change should have, and it rules out the failure mode `validator-known-failures.md` explicitly names — *"the same fallback lands a plausible wrong span rather than none, which is the worse symptom of the same defect"*. A fix that did that would have moved entries from no-span into different-line, **shrinking the ratchet while making the diagnostics worse**, and the count alone would not have shown it.

(I nearly reported the opposite: my first pass classified `?:?` by splitting on `:` and comparing halves, which files every no-span entry under "different line" and would have claimed 174 diagnostics had regressed from *absent* to *wrong*. The corrected classifier keys on `actual === "?:?"` first.)

Local verification on the merged tree, freshly built binding, 14,131 entries × 3 targets:

```
[verify] ✅ all corpus outputs identical after normalization
[verify] ✅ no warning regressions (32 known warning failures remain)
[verify] ✅ no error regressions (236 known error failures remain)
  → 25 declared ratchets across 19 docs match their JSON
```

`Shape matrix parity` is still the inherited `main` failure (#2593); nothing to do here for it.

PR #2433, opened 2026-08-06
fix(esrap): keep redundant parens only for a leading comment

baseballyama:
## What broke

`main` is red. `Coverage` (run 31119636222, `a8a653d5`) fails with three real test failures in `crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs`:

```
expected `(await $.track_reactivity_loss(/* hi */ load()))()`. Got:
	return ((await $.track_reactivity_loss(/* hi */ load())))();
```

(The other red runs on `main` right now — `CI`, `Deploy Docs`, `C ABI`, `crates.io packages` — are the ongoing GitHub Actions outage, `Failed to resolve action download info. Error: Service Unavailable`, and are unrelated.)

## Why

Two PRs merged within a minute of each other and interacted:

- #2372 emits the dev-mode `await X` wrap as spliced **text**: `(await $.track_reactivity_loss(X))()`.
- #2380 routed the client `.svelte.(js|ts)` module through **esrap**, which re-parses each generated chunk with oxc.

oxc preserves explicit parens as a `ParenthesizedExpression`; acorn (esrap's own parser) elides them, so the printer unwraps paren nodes and lets precedence re-add whatever the grammar needs. It kept a node's literal parens whenever **any** comment fell inside its span — and the spliced wrap's span contains the user's comment. So the pair was emitted twice: once by the paren node, once by the callee's precedence rule.

## Fix

Narrow the window to `[paren.start, inner.start)`. A comment *leading* the inner expression is the only one that dangles if the parens go (`return (/*c*/ x)`); a comment deeper inside is already bracketed by that expression's own syntax.

## Verification

**Unit level.** 15 paren/comment shapes through the official compiler (`compileModule`, dev, client) vs rsvelte:

| | diverging from official |
|---|---|
| before | 9 / 15 |
| after | **3 / 15** |

The 3 that remain — `(/* c */ a + b) * 2`, `(/* c */ a, b)`, and an arrow body whose comment official drops — **diverged before this change too**, confirmed by re-running the same probe against `origin/main`'s printer as a control. They are the same doubling family (a paren kept for a genuine leading comment, plus a pair the parent adds) and want a separate fix; not widening this PR.

**Corpus level.** Built the NAPI binding from each printer and compiled the whole corpus twice. **106 outputs change.** Every one moves *toward* official:

| | of the 106 changed outputs |
|---|---|
| byte-identical to official, before | **0** |
| byte-identical to official, after | **68** |
| moved closer to official | **106** |
| moved further from official | **0** |

Real shipping code hits this. `bits-ui`'s `checkbox.svelte.ts`, for example:

```diff
-	#props = $.derived(() => (({     // main
+	#props = $.derived(() => ({      // this PR, == official
```

**The corpus gate cannot see any of it.** `pnpm run corpus:verify` passes with byte-identical numbers on both printers (`match 13183`, `js-mismatch 0`, same 599 known warning failures) — the comparison-side oxfmt normalization collapses `(({…}))` and `({…})` to the same bytes. The corpus also contains zero instances of `track_reactivity_loss(` followed by a comment, so it could not have caught the break either. `await_reactivity_loss_comment_2090` is the only gate that sees this, which is exactly why it is the one that went red.

**Suite.** Full workspace: 410 suites green, 0 failures. `cargo fmt` + `clippy --all-targets --all-features -D warnings` clean.

Newly pinned: `crates/rsvelte_esrap/tests/compat.rs::redundant_parens_survive_only_for_a_leading_comment` — leading comment keeps the parens, deeper comment drops them, and the `(await g(/* c */ x))()` callee shape stays single.

## One corrected expectation

`a_parenthesized_operand_stays_balanced` asserted `$.track_reactivity_loss((load()))`, under a docstring claiming the expected strings are the official compiler's output. That one was not: official recomputes parenthesisation from the AST and prints `load()`. It passed only because the pre-#2380 text splice copied the source bytes verbatim. Repointed at official's actual output, `return (await $.track_reactivity_loss(load()))()();`, with the docstring corrected to say what the test does and does not catch.

https://claude.ai/code/session_01Q94UFSt4NWQ7JpZMaeAoMT

### DIFF
--- a/.changeset/paren-comment-doubling.md
+++ b/.changeset/paren-comment-doubling.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(esrap): keep a redundant paren pair only for a comment that leads the parenthesized expression, so `(await $.track_reactivity_loss(/* c */ load()))()` no longer prints a doubled pair
--- a/.changeset/paren-comment-doubling.md
+++ b/.changeset/paren-comment-doubling.md
@@ -2,4 +2,10 @@
 '@rsvelte/compiler': patch
 ---
 
-fix(esrap): keep a redundant paren pair only for a comment that leads the parenthesized expression, so `(await $.track_reactivity_loss(/* c */ load()))()` no longer prints a doubled pair
+fix(esrap): keep a redundant paren pair only for a comment that *leads* the
+parenthesized expression. A comment deeper inside is already bracketed by that
+expression's own syntax, and keeping the parens for it doubled the pair a parent
+adds from precedence — `(await $.track_reactivity_loss(/* c */ load()))()` printed
+as `((await $.track_reactivity_loss(/* c */ load())))()`, and an object-literal
+arrow body as `() => (({ … }))`. `rsvelte_esrap` is released as 0.10.2 and
+`rsvelte_core` pins the new exact requirement.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2508,7 +2508,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_esrap"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "compact_str 0.10.0",
  "memchr",
--- a/crates/rsvelte_core/Cargo.toml
+++ b/crates/rsvelte_core/Cargo.toml
@@ -73,7 +73,7 @@ oxc_ast = { version = "0.143", features = ["serialize"] }
 oxc_span = "0.143"
 oxc_diagnostics = "0.143"
 oxc_codegen = "0.143"
-rsvelte_esrap = { version = "=0.10.1", path = "../rsvelte_esrap" }
+rsvelte_esrap = { version = "=0.10.2", path = "../rsvelte_esrap" }
 oxc_syntax = "0.143"
 oxc_semantic = "0.143"
 
--- a/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
+++ b/crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs
@@ -49,17 +49,17 @@ fn every_comment_in_the_run_is_kept() {
     );
 }
 
-/// Preservation guard, not a regression test: this caller parses with
-/// `ParseOptions::default()`, where `preserve_parens` makes the argument span
-/// cover the parens, so it passes on either end bound. It would start
-/// discriminating only under `preserve_parens: false`, which is why the copy
-/// runs to the expression's own end — the read range then equals the written
-/// range by construction rather than by coincidence.
+/// The operand's own parens do NOT survive: esrap prints from the AST and
+/// recomputes parenthesisation from precedence, so a redundant `(load())`
+/// collapses to `load()` — official does the same. What this pins is the
+/// surrounding shape, where the source's trailing call and the wrap's own
+/// `()` stack into `…()()`; the balance check then catches a splice that
+/// reads and writes different ranges.
 #[test]
 fn a_parenthesized_operand_stays_balanced() {
     let src = "export async function f() {\n\treturn (await (load()))();\n}\n";
     let out = module(src);
-    assert_contains(&out, "$.track_reactivity_loss((load()))");
+    assert_contains(&out, "return (await $.track_reactivity_loss(load()))()();");
     assert_eq!(
         out.matches('(').count(),
         out.matches(')').count(),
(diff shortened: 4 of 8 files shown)


baseballyama:
Independent verification from the #2375 side. I arrived at the same predicate from a different direction (measuring component script kinds rather than the `.svelte.(js|ts)` regression), and then checked your branch against a case your PR was **not** written against. Short version: **it fixes strictly more than it claims**, and I found one thing it does not fix that I want on the record so nobody assumes it did.

All measurements against official `svelte@5.56.8` (the version `submodules/svelte` pins), `generate: 'client'`, `dev: true`, on your branch at `8e254369`. GitHub Actions is in a repo-wide outage, so none of this is CI.

## Your PR was validated on +1; there is also a +2 case, and you fix it

`origin/main` over-parenthesizes at **two different degrees**, which #2381 does not record — it documents only the `+1`:

| script kind | official | `origin/main` |
|---|---|---|
| `.svelte.js` module | `(await …)()` | `((await …))()` — **+1** |
| component `<script module>` | `(await …)()` | `((await …))()` — **+1** |
| component runes instance | `(await …)()` | `(((await …)))()` — **+2** |
| component legacy instance | `(await …)()` | `(((await …)))()` — **+2** |

The component instance scripts get an extra pair on top, consistent with the round-trip applying the same defect twice — the pair added by one pass becomes a literal `ParenthesizedExpression` for the next, and the over-broad `comment_in_span` window preserves it again.

**On your branch all four collapse to exactly official's one pair.** Eight shapes, all matching:

| shape | official opens | #2433 opens |
|---|---|---|
| component runes instance | 1 | 1 ✅ |
| component legacy instance | 1 | 1 ✅ |
| component `<script module>` | 1 | 1 ✅ |
| `.svelte.js` module | 1 | 1 ✅ |
| runes instance, line comment | 1 | 1 ✅ |
| runes instance, two comments | 1 | 1 ✅ |
| legacy instance, line comment | 1 | 1 ✅ |
| plain comment, no wrap (control) | n/a | n/a ✅ |

So the single predicate change handles both degrees. That is the outcome I was checking for — a fix validated only on `+1` that left `+2` alive would have closed #2381 and buried the remainder, and since #2381 records only `+1`, nothing in the issue would have prompted anyone to look.

## Independent shape sweep: 9 of 10 match

Ten paren/comment shapes through the module path, comparing your branch to official:

| shape | source | result |
|---|---|---|
| G1 | `return (/* c */ x)` | ✅ parens kept |
| G2 | `return (\n// c\n x)` | ✅ parens kept |
| G3 | `return (x /* c */)` | ✅ → `return x;` (official drops both parens *and* the trailing comment) |
| G4 | `return (g(/* c */ x))` | ✅ → `g(/* c */ x)` — was `(g(/* c */ x))` on main |
| G5 | `(await load())()`, no comment | ✅ |
| G6 | `(await load(/* c */ 1))()` | ✅ — was `(((…)))()` on main |
| G7 | `(/* c */ await load())()` | ❌ see below |
| G8 | `await /* hi */ load()` | ✅ — the PR's own case |
| G9 | `(a + /* c */ b) * 2` | ✅ — was `((a + /* c */ b)) * 2` on main |
| G10 | `return ((/* c */ x))` | ✅ → collapses to one pair, matching official |

G3 and G10 are worth calling out as confirmations of your rule rather than mine: official drops a *trailing*-gap comment entirely, and collapses explicit double parens to one. Both fall out of the leading-gap window correctly.

## The one that still differs — pre-existing, not yours

```js
// source: return (/* c */ await load())();
official : return (await $.track_reactivity_loss(/* c */ load()))()();
#2433    : return (/* c */ (await $.track_reactivity_loss(load()))())();
```

**This is not a regression** — `origin/main` produces the identical string, so your PR neither fixes nor worsens it. It is also not a paren bug: your predicate is *right* to keep those parens, because the comment genuinely is in the leading gap. The divergence is that official **relocates** the comment to just before the wrapped argument, while rsvelte leaves it outside the wrap. That is comment placement in the `track_reactivity_loss` rewrite, a different pass from the printer.

I mention it only so that "the paren doubling is fixed" is not read as "this shape now matches official". If it would help I can file it separately; say the word and I will, otherwise I will leave it to avoid duplicating something you may already be tracking.

## The three tests that are red on `main` are green here

`crates/rsvelte_core/tests/await_reactivity_loss_comment_2090.rs` fails **3 of 5 on `origin/main`** right now (I A/B'd it against a clean checkout — pre-existing, unrelated to any in-flight work). On your branch: **5/5 pass**. Worth noting in the PR body, since with Actions down that is otherwise invisible and it makes the change look narrower than it is.

## One housekeeping note

Your branch is based on `a8a653d5`; `origin/main` is now `73aef74f`. Not a conflict — I applied no rebase and everything above builds and passes as-is — but the gap includes several compiler changes, so a rebase before merge would be worth it.

Nice fix. The predicate is exactly right, and `p.expression.span().start` as the window bound is the part that generalizes.

PR #2451, opened 2026-08-06
fix(analyze): name the offending attribute in `a11y_invalid_attribute` (#2413)

### LINKED ISSUE #2413: compiler: `a11y_invalid_attribute` names `href` where the SVG attribute is `xlink:href`

baseballyama:
Fixes #2413.

## The bug

`a11y_invalid_attribute` was constructed with the string literal `"href"`, but the
attribute it is reporting on may be `xlink:href` — the lookup one line above is
`attribute_map.get("href").or_else(|| attribute_map.get("xlink:href"))`. Upstream
(`submodules/svelte/packages/svelte/src/compiler/phases/2-analyze/visitors/shared/a11y/index.js:408`)
passes `href.name`, i.e. the name of the attribute that was actually found.

```
<svg><text><a xlink:href=''>not actually a link</a></text></svg>

official   '' is not a valid xlink:href attribute
rsvelte    '' is not a valid href attribute
```

The fix reads the name off the attribute node. `attribute_map` is populated only
from the `AttributeNode::Attribute` arm (`a11y/mod.rs:85-95`), so the `_ => "href"`
arm is a total-match fallback that cannot fire today and preserves the previous
behaviour if the map ever gains another variant.

## Test

No new test. The official fixture `a11y-anchor-in-svg-is-valid` already covers
exactly this — it asserts `'' is not a valid xlink:href attribute` and
`'#' is not a valid xlink:href attribute` — and was suppressed in
`compatibility/validator-known-failures.json`. The entry is removed, so the
fixture is now a live assertion: 207 -> 206.

Three-value check (`cargo test -p rsvelte_core --test validator -- --nocapture`):

| fix | entry in ratchet | result |
|---|---|---|
| applied | listed (207) | passes; prints `a11y-anchor-in-svg-is-valid` under "ratchet entries now pass" |
| applied | removed (206) | passes |
| reverted | removed (206) | **fails**, naming the fixture as a regression |

The reverted run fails for the predicted reason and nothing else — codes and
spans already matched, only the one word differed:

```
actual:   [a11y_invalid_attribute] '' is not a valid href attribute       @ Some((2, 14))..Some((2, 27))
expected: [a11y_invalid_attribute] '' is not a valid xlink:href attribute @ 2:14..2:27
```

Attribution is clean: the same command on `origin/main` reports **26** entries as
now-passing, and with the fix it reports **27**. The delta is exactly
`a11y-anchor-in-svg-is-valid` — this change fixes one fixture, no more and no less.

Also run: `a11y_validator_fixes` (3 passed) and `a11y_attribute_warning_span`
(7 passed). Both are green, but neither **can observe** this change — they assert
on warning codes and spans only, never on message text. That is the same blind
spot this fix walked out of.

`cargo +1.97.1 fmt --all -- --check` and `cargo clippy -p rsvelte_core
--all-targets --all-features -- -D warnings` are clean.

## Two findings, reported not fixed here

1. **The validator ratchet is one-sided.** `crates/rsvelte_core/tests/validator.rs:454`
   asserts only on `regressions`; the `fixed` set (line 431) is *printed* and then
   dropped. So a fixture that starts passing stays listed forever. The measurement
   above shows **26 entries on `origin/main` already pass** — the file's own claim
   that "every listed fixture is a real divergence from the last confirmed test
   run" is false for 12.6% of it, and the 207 in its header was fiction before
   this PR touched it. I shrank only my own entry; the other 26 need per-entry
   verification and belong in their own change.

2. **`compatibility/lint-known-failures.json`** was baselined while this message
   read `href`. I could not run that gate here (it needs the corpus). Since the
   change moves rsvelte *towards* upstream wording, the expected effect is neutral
   or an improvement, but it is unverified from this branch.

## Relationship to other open PRs — please do not close either as a duplicate

- **#2426 contains this same fix** inside a +1434/−817 diff that is blocked on a worktree lock and
  already conflicting. This PR was written independently against `origin/main`, not lifted from
  that branch, so that it can land on its own. Reconciling #2426 is owned separately.

- **#2418 will conflict with this on merge order.** It adds
  `compatibility/validator-message-known-failures.json` containing
  `["a11y-anchor-in-svg-is-valid", "attribute-quoted", "svelte-self-deprecated"]`, and its gate is
  two-sided (`now_passing.is_empty()`). This fix makes the first entry match, so whichever of the
  two merges second must drop it — 3 -> 2, plus its justification in the paired `.md`. Their
  `MIN_MESSAGE_COMPARISONS` floor is unaffected: that fixture already reached message comparison
  (counts and codes matched), so `compared` does not move. #2418's body already cites #2413 as one
  of the three divergences it surfaced; the two PRs are the same finding approached from opposite
  ends.

## Why the `_ => "href"` arm cannot fire

Checked rather than assumed. `attribute_map` is built at `a11y/mod.rs:83-114`, and the only arm
that inserts into it is `AttributeNode::Attribute(attr)` (`:85-95`). `SpreadAttribute`,
`BindDirective`, `OnDirective` and the `_` catch-all all fall through without inserting. So every
value reachable at the warning site is an `Attribute` and the fallback is dead today — it exists so
the match stays total and preserves the old behaviour if the map ever gains another variant, not
because a variant is expected.


baseballyama:
I rediscovered this bug independently while auditing the validator ratchet for #2457, and landed on the same root cause and the same fix site — `a11y/mod.rs:490` passing the literal `"href"` where upstream `a11y/index.js:408` passes `href.name`. **This PR is correct and I am not opening anything competing.** Two things to add, one of which needs action here.

## Needs action: this PR's `206` is only right against the old ratchet

#2496 makes the validator ratchet two-sided and re-baselines it from a measured run: **207 → 181 entries.** That changes what this PR has to do, in a way that depends on merge order.

- **If #2496 merges first**, rebase and the JSON removal still applies — `a11y-anchor-in-svg-is-valid` is still listed among the 181. But the doc line becomes **180**, not 206, and the doc has been rewritten, so the one-line `207 → 206` edit here will conflict. #2496's doc also describes this fixture as one of "the four content divergences"; that becomes three.
- **If this PR merges first**, #2496 goes red, because with the fix in place the fixture passes and #2496's new `assert!(fixed.is_empty())` will report a stale entry. That is my problem, not yours — I re-run and re-baseline. Do not hand-edit anything on my behalf.

Either way the number must come from a run, not from subtraction. I'd suggest **#2496 first, then a rebase here**, since that direction costs one conflict resolution instead of a re-baseline, but I'm happy either way.

Worth flagging generally: once the ratchet is two-sided, **any** PR that fixes a listed fixture must remove its entry in the same PR or CI fails. This PR already does that, which is exactly right — I'm noting it because it becomes a hard requirement rather than a courtesy.

## Context: this entry had no justification at all

`compatibility/validator-known-failures.md` organises its 207 entries into three clusters with counts. `a11y-anchor-in-svg-is-valid` appeared in **none** of them — not in the error-span list, not in the span-only list, not among the content bullets.

That's a different failure from a mislabelled entry, and worth separating. A mislabelled entry has a claim attached that is wrong, so it can be checked and found wrong. An **unlisted** entry has no claim attached at all — there is nothing to falsify and nothing to expire. If the doc's clusters are the unit of accountability, an entry outside every cluster is accountable to nobody, and this bug sat behind one.

For contrast, the audit also turned up `unknown-code`, which is the *mislabelled* case: filed under span-only, but actually a warning-emission-order bug that the promised span fix would not have cleared. Filed as #2498. And the bookkeeping defect that let clusters double-cite entries while the counts still summed correctly is #2500.

PR #2464, opened 2026-08-06
fix(compiler): mirror upstream's constructor read form and $state proxy gate for private-field $.set (#2395)

### LINKED ISSUE #2395: codegen (client): `$.set` for a private field emits a spurious proxy flag on a logical-assignment RHS and reads the compound operand with `$.get` instead of `.v`
Surfaced while pasting official's bytes for #2380's expectations
(https://github.com/baseballyama/rsvelte/pull/2380#issuecomment-5204892735).
**Pre-existing on `origin/main`; #2380 does not cause either.**

## Repro

`A.svelte.js`, `generate: 'client'`, `dev: false`:

```js
export class R {
	#x = $state.raw({});
	#n = $state(0);

	constructor(s) {
		this.#x ??= { a: s, b: s };
		this.#n += 1;
	}
}
```

Official:

```js
$.set(this.#x, this.#x.v ?? { a: s, b: s });
$.set(this.#n, this.#n.v + 1);
```

rsvelte:

```js
$.set(this.#x, this.#x.v ?? { a: s, b: s }, true);
$.set(this.#n, $.get(this.#n) + 1);
```

## Two divergences

**1. Spurious third argument on a logical-assignment RHS.** Official emits a
two-argument `$.set`; rsvelte adds the `true` proxy flag. This is the same shape
as #2343 (closed), which fixed it for a `BinaryExpression` value — the
`LogicalExpression` produced by `??=` / `||=` / `&&=` was not covered by that fix.

**2. `$.get(this.#n)` where official reads `this.#n.v`.** rsvelte is *internally
inconsistent* here: the same output uses `this.#x.v` in the logical-assignment
statement and `$.get(this.#n)` in the compound one. `.v` is the untracked read
and `$.get` registers a reactive dependency, so unlike (1) this may be a
**semantic** difference rather than a layout one — worth confirming against
`internal/client/reactivity/sources.js` before assuming it is cosmetic.

## Gate visibility

Both are ordinary token differences, not comments or parens, so oxfmt +
`stripBlankLines` cannot absorb them: the corpus gate **can** see both. That
`client` is at 0 known failures therefore means no corpus entry has a private
`$state` field under a compound or logical assignment in a `.svelte.(js|ts)`
class — an interaction shape the collected corpus does not sample, and a
candidate for the generated matrix rather than for more corpus repos.

## Where it is pinned today

`crates/rsvelte_core/tests/ctor_private_state_set_paren_2253.rs`,
`compound_and_logical_assignments_keep_nested_comments`. That test's subject is
comment survival, so on #2380's branch the two divergences are named in a
comment at the assertion site rather than absorbed silently. Whoever fixes this
should tighten those assertions to official's bytes in the same PR.

baseballyama:
Fixes #2395.

## What was wrong

Both divergences are in the **constructor** rewrite for a `#private` `$state`
field in a `.svelte.(js|ts)` class. Reproduced against `submodules/svelte`
(v5.56.8) before touching anything; the issue's description held up.

```js
export class R {
	#x = $state.raw({});
	#n = $state(0);

	constructor(s) {
		this.#x ??= { a: s, b: s };
		this.#n += 1;
	}
}
```

| | official | `origin/main` |
|---|---|---|
| `??=` on `$state.raw` | `$.set(this.#x, this.#x.v ?? { a: s, b: s });` | `…, true);` |
| `+=` on `$state` | `$.set(this.#n, this.#n.v + 1);` | `$.set(this.#n, $.get(this.#n) + 1);` |

## Upstream mirrored

* `phases/3-transform/client/visitors/AssignmentExpression.js:86-91` —
  `needs_proxy = field.type === '$state' && is_non_coercive_operator(operator) &&
  should_proxy(value, scope)`. The logical compounds build a `LogicalExpression`,
  which `should_proxy` never rejects, so for those the flag reduces to
  `field.type === '$state'` — `$state.raw` and `$derived` must not carry it.
* `phases/3-transform/client/visitors/MemberExpression.js:15-18` — the compound
  operand is the *visited* LHS: `in_constructor && (field.type === '$state.raw' ||
  field.type === '$state') ? b.member(node, 'v') : b.call('$.get', node)`.
* `phases/3-transform/client/visitors/shared/function.js:9-13` — `in_constructor`
  is cleared on entry to any nested function, so `.v` must not leak into a
  callback body in the constructor. A plain `if` / `for` block is *not* a
  function and keeps `.v`.

## Reachability

Both sites are reachable from the public `compile_module()` / `compile()` entry
points: `transform_class_fields_client` → the constructor line loop →
`transform_constructor_assignment` (the `??=` / `+=` branches) and →
`private_class_assign_ast::transform_private_class_assign_ast` (the shift /
bitwise compounds, and anything nested inside a grouped statement). Every output
above is real `compile_module(generate: 'client')` output, not a unit-test
fixture.

## Tests

`crates/rsvelte_core/tests/ctor_private_state_read_and_proxy_2395.rs` (6 tests),
all expectations taken from official's bytes:

* constructor `??=` proxies only for a plain `$state` field;
* constructor compound reads `.v` (including `<<=`, which takes the AST path);
* method bodies still read through `$.get` — the control that fails if the fix
  is applied too widely, and that **passes on `origin/main`**;
* a function nested in the constructor reads through `$.get`;
* a plain `if` / `for` block in the constructor still reads `.v`;
* the reported repro, with negative assertions on the two wrong strings.

Checked against `origin/main`: the three constructor tests fail there, and they
fail *with the predicted wrong output* (`…{ a: s }, true` and
`$.get(this.#n) + 1`), not merely differently. The method-body control passes on
`origin/main`, so it discriminates.

`crates/rsvelte_core/tests/ctor_private_state_set_paren_2253.rs::compound_and_logical_assignments_keep_nested_comments`
carried both divergences as a named exception at its assertion site; it is now
tightened to official's bytes and the exception comment removed.

## What the existing suites can and cannot see

Worth stating plainly, because "the suites are green" would be misleading here.
Grepping the whole official fixture corpus for `this.#x <op>= …` returns four
files, and **none of them reaches a branch this PR changes**:

* `class-private-fields-assignment-shorthand` — `||= &&= ??=` in a constructor,
  but all three fields are plain `$state`, which proxies before and after. A
  negative control, not coverage.
* `inspect-trace-class` — `this.#count += 1`, but inside an
  `increment = () => {…}` class property, which takes the method path.
* `class-private-state-increment`, `class-state-public-private` — method bodies.

So `runtime-runes` et al. are a **regression control** for this change, not
evidence of it. The new test file is the coverage. This is also why the corpus
gate reports `client` at 0 failures despite main shipping wrong bytes: nothing in
either corpus samples the interaction.

## Ratchets

No delta. `compatibility/known-failures.{client,server,client-dev}.json` are all
`[]`, and every shape this PR changes was *wrong* on `origin/main` — so any
corpus entry containing one would have been listed. An empty list therefore
proves no entry exercises these shapes, in either direction. (This matches the
issue's own gate-visibility analysis: both divergences are ordinary token
differences that oxfmt + `stripBlankLines` cannot absorb.)

## Still diverging, deliberately out of scope

A **non-`this` prefix** in a constructor (`const inst = this; inst.#n += 1`) is a
separate, larger defect: `transform_class_methods_non_this` /
`private_field_assign_ast` model neither the proxy flag nor the logical
operators there, and today emit the invalid `$.get(inst.#n) ??= s`. Untouched by
this PR.


baseballyama:

```
```

PR #2470, opened 2026-08-06
fix(corpus): refuse baseline rewrites from a narrowed run in one place

baseballyama:
Answers the pattern question raised in review of #2466: *which other scripts
accept both a subset selector and a baseline-writing flag?*

The answer was **more than three**, so this is the shared helper rather than a
fourth hand-written copy.

## The live bug

```
$ node scripts/compat-corpus/matrix/run.mjs --targets client --update-baseline
[matrix] baseline: 50 known -> compatibility/matrix-known-failures.json
EXIT=0
```

**350 entries to 50 — 300 deleted (86%), silently, exit 0.** The ratchet is
`server: 250, client: 50, client-dev: 50`, so a client-only run erases every
server and client-dev entry. It would read as an enormous clean improvement.

Observed with the write redirected to a temp file, so nothing checked in was
touched (`what it WOULD have written: 50 entries`).

What makes it sharp: `run.mjs` already knows the rule at both ends and misses it
only in the middle. Baseline ids embed the target (`:180`), the staleness check
scopes by `measuredTargets` (`:202`), and `--families` is refused by name for
this exact reason (`:188`) — but the rewrite itself was unscoped.

## The survey

| script | ratchet | narrowing axes | refused before this PR |
|---|---|---|---|
| `mutate-corpus.mjs` | single | sampling, `--targets`, `--no-fmt`, orphaned seeds, missing sources | **all** — the exemplar |
| `matrix/run.mjs` | single | `--no-fmt`, `--families`, `--targets` | 2 of 3 |
| `css-prune-sweep.mjs` | single | `--filter` | none (fixed in #2466) |
| `verify.mjs` | **per-target files** | `--no-fmt` | none |
| `fmt-verify.mjs`, `svelte2tsx-verify.mjs` | single | none (`--baseline` picks the file, not a subset) | n/a |

So the rule is **not under-known**. `mutate-corpus.mjs` gets all four axes right,
and `matrix/run.mjs` refuses `--families` *by name for this exact reason* twenty
lines above the site that ignores it. An author who understood the invariant well
enough to write it twice still left the third site open.

That is the argument for a helper rather than a fourth refusal. These are
**incomplete copies, each missing the axis its author happened not to be thinking
about** — a failure of recall at the moment of writing, not of understanding.
Education cannot fix that and a data structure can: the axes are now a list at
the call site, so adding one is an array element and forgetting one is visible in
review as a missing line rather than invisible as an unwritten `if`.

## Also fixed

- **`verify.mjs` gains `--no-fmt`.** Its `--targets` stays allowed *on purpose* —
  every target ratchets to its own file (`baselinePath(t)`), so a narrowed run
  rewrites only what it measured. That distinction is the reason the helper takes
  a list of reasons rather than trying to infer narrowing itself.
- **Matrix refuses at parse time.** The rejected run was previously paid for in
  full before its result was thrown away: **1s instead of a full run**.

## Every arm observed

| Arm | Result |
|---|---|
| matrix `--targets client --update-baseline` | `EXIT=2`, `measured 1 of 3 (client)`, 1s |
| matrix `--families comment-slot --update-baseline` | `EXIT=2`, `measured 1 of 2 families` |
| matrix `--no-fmt --update-baseline` | `EXIT=2` |
| all three narrowings at once | `EXIT=2`, **all three reasons listed** |
| matrix full `--update-baseline` (control) | `EXIT=0`, writes normally — guard does not over-refuse |
| verify `--no-fmt --update-baseline` | `EXIT=2` at parse time |
| verify `--targets client --update-baseline` (control) | **not refused**, proceeds |

The two controls matter as much as the refusals. A guard that fired everywhere
would pass every refusal test above while making `--targets` unusable — a
correct-looking gate that breaks a legitimate workflow, which no amount of
testing the positive cases would catch.

## Unrelated finding

While running matrix in a `feat/corpus-mutation-fuzz` (#2353) worktree I hit 6
`comment-slot/legacy-reactive__*` `js-mismatch` failures on server, where rsvelte
drops the `$:` label when a delimiter-bearing comment precedes the block. **Not a
regression** — that branch is 20 commits behind and predates #2366, which fixed
exactly this. Noting it because baselining the mutation ratchet from that stale
branch would bake in 6 failures main has already fixed, which reinforces the
existing hold on that ratchet.

## CI

Actions is in a major outage; verified locally only. Destructive forms were run
with the baseline write redirected to a temp path — no checked-in ratchet was
modified, and each was confirmed unchanged afterwards.


baseballyama:
**Correcting a count of mine that was being used against this PR. I scored `matrix/run.mjs` as guarded; it is guarded on 2 of 3 axes, and your reproduction is right.**

I ran a repo-wide scan (**1087 tracked `.mjs`/`.rs` files**; 5 accept both a subset selector and a baseline-writing flag) and reported **exactly 1 unguarded instance**, concluding a shared helper was not warranted. The correct figure is **2 of 5**. I verified yours in code before posting this.

## Where I went wrong

`matrix/run.mjs:182-196` refuses two things — `--no-fmt` (`:183`) and the `--families` subset (`:188`, naming FALSE-SHRINK outright). I read "refuses, with the right reason, by name" and scored the file guarded. But `--targets` narrows the same population and is not refused: `ids` (`:180`) is built only from the selected `TARGETS` (`:52`), and `:193` then writes the **whole** baseline from it. So `--targets client --update-baseline` deletes every `server` and `client-dev` entry — your **350 → 50, 300 deleted (86%), exit 0**, which is what a 3-target ratchet collapsing to one target should look like.

## This file is the best argument in the repo for the helper, and it isn't the count

Two of five is weak evidence on frequency. The mechanism is the evidence, and `matrix/run.mjs` demonstrates it **within a single file, fifteen lines apart**:

- the **write** path (`:182-196`) narrows on families only;
- the **compare** path (`:203-208`) narrows on `measuredFamilies.has(family) && measuredTargets.has(target)` — **both** axes;
- the comment introducing it (`:200`) reads *"Only entries in the families this run measured can be judged stale"* — naming one axis while the code directly beneath it handles two.

So the author was **not** unaware that `--targets` narrows the measured set. They wrote `measuredTargets` at `:202` and used it at `:207`. The knowledge was present in the file and did not travel fifteen lines up into the guard. Even their own comment is an incomplete copy of their own code.

Across the population: `mutate-corpus.mjs` refuses all four of its axes, `matrix/run.mjs` 2 of 3, `verify.mjs` 0 of 1 (it scopes the write instead, which is a valid alternative), `check-verify.mjs:100` and `check-e2e-verify.mjs:115` refuse their single axis, `css-prune-sweep.mjs` refuses nothing. **These are incomplete copies of one rule, each missing whichever axis its author was not holding in mind.** That is not a knowledge gap, so a reminder or a review checklist cannot close it — the person already knows the rule and applied it correctly elsewhere in the same function's neighbourhood. A structure that makes the write path consume the same narrowing set the compare path already computes can.

**So: my count supports the helper on mechanism grounds, not frequency grounds.** If the helper's design is "declare your axes once, and both the write guard and the staleness filter read that declaration", `matrix/run.mjs` is the case that justifies it — it would have made the write path inherit `measuredTargets` for free.

## The measurement trap, since it produced this error and one before it

I nearly published a second wrong answer from the same shape. My first pass grepped `refus|exit\(2\)` per file and reported `css-prune-sweep.mjs` as **guarded** — those tokens appear elsewhere in it. The same loose pattern reports `matrix/run.mjs` as guarded too, for a subtler reason: it genuinely refuses, just a different axis.

**The proxy answers *"does this file contain refusal machinery"* when the question is *"does it refuse **this**"*.** The check is only meaningful when the refusal's *condition references the selector variable in question* — evaluated **per axis, not per file**. Anyone re-running this scan should treat a file-level verdict as unsound by construction; the unit is (script × selector).

Corrected table now in `compatibility/gate-coverage.md` (#2456), with the per-axis breakdown and this reasoning.

Issue #2452, opened 2026-08-06
test: "Validator 333/333" is unfalsifiable — the report compares warning counts and the 207-entry ratchet is one-sided


baseballyama:
`AGENTS.md` reports `Validator 333/333` and "All in-scope fixtures pass (100.0%)" while `compatibility/validator-known-failures.json` holds **207** real divergences against a 332-fixture floor — ~62% of the suite is on the forgive list. Two independent mechanisms produce that, and they need separate fixes.

## Mechanism 1 — the report compares warnings by count

`crates/rsvelte_devtools/tests/compatibility_report.rs:751`:

```rust
let warnings_match = actual_count == expected_warnings.len();
```

Never the code, never the message, never the span. Emitting `a11y_missing_attribute` where upstream emits `a11y_img_redundant_alt` on the same sample matches. This test is the source of the Test Status table in `AGENTS.md` and `README.md` (`pnpm run compatibility-report`).

`crates/rsvelte_core/tests/validator.rs` *does* check the code — for the 125 fixtures not on the ratchet.

**Fix:** compare the warning code multiset in the report. Low cost. The published number will drop, which is the point — the current number is not measuring what its label says.

## Mechanism 2 — the validator ratchet is one-sided

`crates/rsvelte_core/tests/validator.rs:454-458` asserts `regressions.is_empty()` only. `fixed` is computed at `:431` and merely printed at `:433-442`.

Positive control that two-sidedness is the house rule and this is an omission, not a design choice:

- `crates/rsvelte_core/tests/sourcemaps_gate.rs:1061` — staleness fatal
- `scripts/compat-corpus/css-prune-sweep.mjs:514` — staleness fatal
- `scripts/compat-corpus/verify.mjs:585-603` states the reasoning outright: *"a large `now PASS` delta on the next PR is indistinguishable from noise, so a real regression can hide inside it"*

Consequence: the 207-entry list is under no shrink pressure, and an entry that starts passing silently re-covers a future regression on that same fixture.

**Fix:** assert `fixed.is_empty()` and re-baseline in the same PR, exactly as `verify.mjs` requires for the corpus ratchets.

## Why the pair is worth one issue

[...]

Issue #2490, opened 2026-08-06
a11y_figcaption_index points at the <figure> instead of the <figcaption> — the only confidently-wrong warning location in 14k files


baseballyama:
Diagnosis only — no fix proposed here.

`a11y_figcaption_index` is the **only** place in a 14,131-entry corpus run where rsvelte points a
warning confidently at the **wrong** source location. Every other position divergence (2,082 of
2,085 diverging `(entry, target, warning)` triples) reports no location at all.

## The divergence

`packages/svelte/tests/validator/samples/a11y-figcaption-wrong-place/input.svelte`:

```svelte
1  <figure>
2  	<img src='foo.jpg' alt='a foo'>
3
4  	<figcaption>
5  		a foo in its natural habitat
6  	</figcaption>
7
8  	<p>this should not be here</p>
9  </figure>
```

```
official:  a11y_figcaption_index @ 4:1     <- the <figcaption>
rsvelte:   a11y_figcaption_index @ 1:0     <- the <figure>
```

The `(dline, dcol) = (-3, -1)` delta is **not** a coordinate-base bug. `<figure>` starts at column
0 and `<figcaption>` at column 1 because of the tab. The delta is just the distance between two
different nodes, and reading it as an off-by-one would send someone to the wrong place.

## Mechanism

`crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/shared/a11y/mod.rs:602-621` runs the
check while visiting the **`<figure>`**, locates the offending child by index, and then constructs
the warning with no argument and therefore no span:

```rust
"figure" => {
    let children: Vec<_> = node.fragment.nodes.iter().filter(...).collect();
    let index = children.iter().position(|child| {
        matches!(child, TemplateNode::RegularElement(el) if el.name == "figcaption")
    });
    if let Some(idx) = index && idx != 0 && idx != children.len() - 1 {
        warnings.push(w::a11y_figcaption_index());   // <- :620, no node, no span
    }
}
```

Upstream passes the node it just found —
`packages/svelte/src/compiler/phases/2-analyze/visitors/shared/a11y/index.js:532`:

```js
w.a11y_figcaption_index(children[index]);
```

The caller's fallback then stamps the **visited element**'s span onto any warning that has none.

## Why this is worth its own issue despite sharing a root cause

**Correcting my own first reading:** this is *not* a different bug in kind. The root mechanism is
identical to the other 2,082 — a warning constructed without its node's span. What differs is the
[...]

PR #2496, opened 2026-08-06
test(validator): make the validator ratchet two-sided (#2457)

### LINKED ISSUE #2457: validator-known-failures.json is the only one-sided ratchet, and 26 of its 207 entries already pass
`compatibility/validator-known-failures.json` is the only ratchet in the repo that never re-checks
its own entries, and it has silently gone stale: **26 of its 207 entries (12.6%) already pass on
`origin/main`.**

## The defect

`crates/rsvelte_core/tests/validator.rs:430-431` computes both directions:

```rust
let regressions: Vec<&str> = failing.difference(&known_set).copied().collect();
let fixed: Vec<&str> = known_set.difference(&failing).copied().collect();
```

`fixed` is then printed (`:433-442`) and **dropped**. The only assertion is on `regressions`
(`:454-458`). So an entry that starts passing stays listed forever, and nothing fails.

Every other ratchet in the repo is two-sided. `matrix/run.mjs:227-234` fails on
`fixed.length`; the PR under #2418 makes its new message ratchet fail on `now_passing.is_empty()`.
Per AGENTS.md, *"Every ratchet is two-sided: a new failure **and** a listed entry that already
passes both fail CI."* This one is the exception, and it is the largest.

## The consequence, measured

```
cargo test -p rsvelte_core --test validator -- --nocapture \
  | grep -a -A99 'ratchet entries now pass'
```

On `origin/main` (73aef74f) that prints **26**:

```
a11y-alt-text                                          a11y-no-abstract-roles
a11y-anchor-is-valid                                   a11y-no-access-key
a11y-aria-activedescendant                             a11y-no-autofocus
a11y-aria-props                                        a11y-no-interactive-element-to-noninteractive-role
a11y-aria-proptypes-boolean                            a11y-no-noninteractive-element-to-interactive-role
a11y-aria-proptypes-integer                            a11y-no-redundant-roles
a11y-aria-proptypes-number                             a11y-not-on-components
a11y-aria-proptypes-token                              a11y-role-has-required-aria-props
a11y-aria-proptypes-tokenlist                          a11y-role-supports-aria-props
a11y-aria-proptypes-tristate                           a11y-scope
a11y-aria-role                                         a11y-tabindex-no-positive
a11y-aria-unsupported-element                          inline-new-class-2
a11y-autocomplete-valid
a11y-heading-has-content
```

(27 with #2451 applied; that PR removes its own entry, `a11y-anchor-in-svg-is-valid`, and leaves
these 26.)

**The paired `.md` states something that is now false.** Its header claims:

> every listed fixture is a real divergence from the last confirmed test run, not a placeholder.

A ratchet that never re-checks cannot support that claim for any entry, and empirically it fails
for 12.6% of them. The cluster counts in that file (141 span-plumbing + 53 span-only + 12 wording)
are equally unverified — 26 entries are distributed through those buckets and are not divergences
at all, so every number in that document is an upper bound of unknown tightness.

This also makes the file useless for sizing work: "207 validator divergences" has been quoted as a
backlog figure, and the real number is at most 181.

## Suggested fix, in two steps

1. **Make the gate two-sided** — assert on `fixed` as well as `regressions` in `validator.rs:454`.
   One-line change, and it is what makes step 2 stay true.
2. **Verify and remove the 26 per entry.** They cannot be bulk-deleted: an entry may be listed for
   a reason the `.md` never names, and removing it without checking *which* dimension now matches
   would be trading one unverified claim for another. Each needs the `code`/`message`/`start`/`end`
   comparison actually inspected, and the `.md` cluster counts re-derived from what remains rather
   than decremented.

Step 1 before step 2, so the list cannot re-rot between the audit and the merge.

## Ordering note

This will fight for the same files as #2418 (adds a sibling ratchet and edits `validator.rs`) and
#2451 (shrinks `validator-known-failures.json` by one). Worth landing after both.

Found while fixing #2413; the A/B that isolated that fix (26 now-passing before, 27 after) is what
surfaced the 26.

baseballyama:
Closes #2457. **Also closes mechanism 2 of #2452** — see the last section, so nobody later reads #2452 as untouched.

## The fix is the second assertion, not the deletion

`validator-known-failures.json` failed the run on a new failure and not on a listed entry that had started passing. The suite *printed* those and moved on, so the ratchet could never shrink on its own. 26 of 207 entries were already passing.

Deleting the 26 alone would have restored the same state in a few months. The gate now asserts `fixed.is_empty()` too, mirroring `sourcemaps_gate.rs`, which has been two-sided all along — so this is aligning with an in-repo precedent rather than inventing a convention.

## 26, measured

```
cargo test -p rsvelte_core --test validator test_validator -- --nocapture
  Total: 151/332 passed
  26 ratchet entries now pass
```

207 known − 181 failing = 26, and the run also reports 0 regressions, so the failing set is a subset of the known set and the subtraction is sound. I did not take 26 from the issue text.

## Re-clustered by measurement, not by subtraction

The doc's counts were hand-maintained. I classified each failure block the suite prints, by what actually diverges:

| cluster | doc said | measured now |
|---|---:|---:|
| error span not populated | ~141 | **141** |
| warning span-only | 53 | **36** |
| warning content | 13 | **4** |
| | 207 | **181** |

17 + 9 = 26, so the removals reconcile — but the reconciliation is a *check* on the old split, not how I got the new one.

## Four entries were absorbing something other than what the doc claimed

This is the part worth reviewing. An entry suppresses everything about its fixture, not the one divergence its justification names, so I checked each removal against its recorded reason. Two of the findings are on entries that are *still listed*:

**`a11y-anchor-in-svg-is-valid` — a live correctness bug in no cluster's list.** `a11y/mod.rs:490` passes the literal `"href"`; upstream (`a11y/index.js:408`) passes `href.name`. So `<a xlink:href=''>` inside `<svg>` is reported as an invalid **`href`** attribute — naming an attribute that is not in the source. It had no justification of any kind behind it.

**`unknown-code` — filed as span-only, actually an ordering bug.** The span-only cluster's stated property is that code and message match. Under the suite's ordered comparison they do not: rsvelte emits the three `svelte-ignore` comment-code warnings as a batch ahead of the a11y warnings, where upstream interleaves all six in source order (lines 3, 5, 8, 10, 13, 14). The multisets are equal; the sequences are not. I checked that neither compiler sorts its warning list, so **the span fix the justification promised would have left it failing.**

**`invalid-node-placement-5` and `module-script-reactive-declaration`** were cited as examples of the error-span cluster *and* given wording bullets under the content cluster, while the three counts summed to 207 as if each entry were counted once. Both are span-only failures today.

**3 of the 26 removed** — `a11y-alt-text`, `a11y-aria-role`, `a11y-no-noninteractive-element-to-interactive-role` — were named nowhere in the doc.

## Do the 26 pass for the reason recorded?

For the 3 carrying a *content* claim, which is checkable in source independently of the fixture: yes, all 3, and I checked rather than assumed. `a11y_unknown_aria_attribute` and `a11y_missing_attribute` now match upstream's format strings verbatim, `a11y_incorrect_aria_attribute_type_tokenlist` likewise, and `a11y_no_interactive_element_to_noninteractive_role` is called with `(element, role)` in upstream's order.

For the other 20, the recorded cause is a *span*, and the only evidence available that it was the cause is that the fixture now matches on spans. That is not independent, so the doc does not claim it. Stating the limit rather than letting 23-of-23 read as uniformly verified.

One near-miss worth recording: I first wrote that `regular_element.rs` "now says `cannot be a child of`, matching upstream", on a grep that returned only child-forms. rsvelte **does** have a descendant-form, at `regular_element.rs:948`, assembled from a separate string fragment my pattern could not match. The claim came out of the doc.

## Verification

```
cargo test -p rsvelte_core --test validator test_validator      PASS (181 known, 0 fixed, 0 regressions)
```

Mutant, to confirm the new assertion can fail and fails for the predicted reason — restore the 207-entry ratchet and re-run:

```
26 stale entries in compatibility/validator-known-failures.json (they already pass)
```

Predicted "26 stale", got "26 stale". Without the new assertion the same tree passes, which is the defect.

## Closes mechanism 2 of #2452

#2452 asks how `Validator 333/333` could ever move. Mechanism 2 was that it could not move *downward* on a fixed fixture: a fixture that started passing stayed on the ratchet and stayed out of the headline. It now cannot. **Mechanism 1 — `compatibility_report.rs:752` comparing warning *count* rather than warning *codes* — is untouched here** and is a separate change, since it needs a full `compatibility-report` run to regenerate the published headline.

## Merge note

This edits the `## Current baseline:` line of `validator-known-failures.md`, and #2488 edits the same line to add the JSON filename. Trivial conflict either order; the resolution is the #2488 wording with **181**. I wrote the line in #2488's format so it satisfies that PR's checker once both land.


baseballyama:
Heads-up on a small collision, and a request that costs you one line.

#2548 adds a `Partition of …` line to every ratchet doc that states per-cluster counts, checked by `known-failures-md-check.mjs`: the doc supplies the addends, the checker computes the total from the JSON. For `validator-known-failures.md` it uses today's numbers —

```
Partition of `validator-known-failures.json` by cluster: `141 + 53 + 13`
```

— which this PR replaces with 141/36/4 over 181. The conflict is that one line plus the sentence above it; keep yours and update the addends, and the check will then hold you to them.

The reason it is worth keeping rather than dropping: the doc you are rewriting is the instance #2500 was filed from. `141 + 53 + 13 = 207` stayed true while `invalid-node-placement-5` and `module-script-reactive-declaration` were cited under cluster 1 and counted under cluster 3 — every number correct, the clusters not a partition. Since you are re-measuring the classification anyway, you are the one PR that can state the new counts and have them checked in the same change.

One thing the sum check cannot see, which your measurement can: whether an entry is cited under exactly one cluster. #2500 says so explicitly and does not ask for it. But you now have the per-fixture classification in hand, so if the example fixture names in each cluster's prose come from that classification rather than from memory, the citation half is fixed for free — and that is the half that was actually wrong here.

baseballyama:
Merged `origin/main` and re-baselined. Three things landed here that are worth calling out, because one of them is currently breaking **every** open PR:

**1. `validator-message-known-failures.json` was undeclared — `main` is red.**

#2418 added that ratchet at the same time #2488 changed `known-failures-md-check.mjs` to require every ratchet JSON on disk to appear in a declared `RATCHETS` list. Neither PR conflicted with the other, so both merged green and `main` now fails:

```
[known-failures-md-check] validator-message-known-failures.json is a ratchet on disk
  but is not declared in RATCHETS — add it with the doc that justifies it
```

That check is the first step of the corpus jobs, so it fails on every PR built against current `main`. Declaring it (plus a `Current baseline: … 1 entry` line in the paired `.md`, which the count check needs) is in this PR. It is exactly the failure mode the declared-list design was meant to catch — the gate worked; the two PRs just raced.

**2. `validator-known-failures.json`: 181 → 179.** Both removals are entries `main` fixed while this PR was open, and both come out of clusters the doc already names:

- `a11y-anchor-in-svg-is-valid` — one of the four *content* divergences; #2451 now passes `&a.name`, so the message names `xlink:href`. Content cluster 4 → 3.
- `a11y-figcaption-wrong-place` — the "plausible wrong span" case under *warning span-only*; #2490 landed. Span-only 36 → 35.

The JSON conflict was resolved as a **set** union-of-removals, not by line, then verified by running the suite — which is what named the one entry the set merge had left behind.

**3. `validator-message-known-failures.json`: 2 → 1**, same cause as (2): #2451 fixed `a11y-anchor-in-svg-is-valid`, and its own two-sided ratchet reported it stale. Its justification section is removed with it.

Local: `cargo test -p rsvelte_core --test validator` → 3/3, `known-failures-md-check` → 19 declared ratchets across 16 docs, `test-known-failures-md-check` → all cases pass.

Issue #2397, opened 2026-08-06
chore(compiler): the legacy text-based `$:` reorder machinery in transform_legacy.rs is unreachable (~450 lines)


baseballyama:
Found while fixing #2373 (PR #2394), which was itself split out of the #2357 scan sweep.

The legacy text-based `$:` reorder machinery in
`crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs`
is **unreachable from component compilation**. It has been superseded by the AST
port of upstream's `order_reactive_statements` in
`crates/rsvelte_core/src/compiler/phases/3_transform/server/ast/script.rs`
(≈ lines 2946-3061), which is what actually orders `$:` statements for SSR today.

This is not a tidiness item. **Three PRs today touched this machinery** — #2366,
#2370 and #2394 — so we are spending review and CI time on code that cannot run,
and bugs "found" in it cannot be reproduced through the compiler.

## The call chain terminates at a caller that cannot supply a `$:`

```
extract_simple_assignments
  <- extract_reactive_lhs_vars
  <- sort_reactive_statements_topologically
  <- sort_reactive_in_place  (and the second call site in the same function)
  <- reorder_reactive_statements_after_functions      transform_legacy.rs:899
  <- transform_script_content_inner                   transform_script.rs:266  (the `!is_module` branch)
  <- transform_script_content_with_imports_and_derived transform_script.rs:20  (the only entry; `transform_script_content_inner` has exactly one caller)
  <- visit/try_async_declaration_tag                  server/ast/visitors/declaration_tag.rs:346
```

`declaration_tag.rs` builds `script_input` from a single `DeclarationTag`'s
`node.declaration` source range. Phase 1 rejects anything but a
`VariableDeclaration` there (`declaration_tag_invalid_type`,
`phases/1_parse/state/tag.rs:76,146,205`), so the text handed to
`reorder_reactive_statements_after_functions` can never contain a line whose
`trim()` starts with `$:`. Its first guard —
`if !has_reactive { return script.to_string(); }` — therefore always fires.

## Measurement, not inference

Instrumented `reorder_reactive_statements_after_functions` to print on
`has_reactive == true`, then ran every `rsvelte_core` integration suite whose
sources contain a `$:` statement — 18 binaries, 42 tests
(`compile_both_parity`, `component_bind_getter_setter`,
`dev_prop_mutation_disambiguation`, `dev_ownership_validator_mutation`,
[...]

PR #2530, opened 2026-08-07
fix(compiler): keep a legacy `$:` statement whole across a `//` line

### LINKED ISSUE #2387: compiler: the two legacy `$:` accumulation loops disagree on `//` as a statement boundary
## Summary

`server/transform_legacy.rs` has two loops that accumulate the lines of a multi-line `$:`
statement. They must agree on where the statement ends — if they disagree, the first loop routes
a block the second loop then collects differently. Their break-sets do **not** match:

| loop | break-set |
|---|---|
| `needs_reorder` decision scan (~`:909`) | `is_empty()` \| `"$:"` \| `"function "` |
| collection scan (~`:1056`) | those **plus** `"//"` |

The comment at `:848` used to assert the sets match. It was corrected in #2366 to a conditional
that holds as written, because the assertion was false. The underlying asymmetry was **not**
touched, deliberately, and this issue records why and what would be needed to touch it.

## Why it was not fixed in #2366

Three inputs were built specifically to exercise the difference — a `//` line inside a `$:`
continuation, with and without a following reactive statement — and all three produced output
**byte-identical** to `main`. With no constructible divergence, unifying the sets would be a
behaviour change nobody can show is correct: exactly the blind edit the #2357 sweep exists to
avoid.

## What is actually unknown

Not "does the difference matter" — that is answered for the three shapes tried. What is unknown
is whether a shape exists at all where the two loops' disagreement is observable in output. The
candidates the three probes did **not** cover:

- a `//` line that is the *first* line after the `$:` header (the collection loop breaks before
  its `stmt_lines.push`, so the two loops would disagree about whether the statement has a body
  at all);
- a `//` line inside a `$:` **block** rather than a continuation expression, where brace depth
  keeps the loop running past the break the other loop takes;
- interaction with the template-literal state added in #2366 — both guards now gate on
  `!acc_template`, so a `//` inside a template is no longer a boundary for either loop, which
  removes one way the sets could have diverged.

## Definition of done

Either a failing input, then unify the sets and pin it — or a written argument that no such
input exists, in which case the *sets themselves* should be unified anyway and the comment
upgraded from a conditional to the assertion it originally claimed to be. Both outcomes are
acceptable; what is not acceptable is changing the sets without doing one of them.

Related: #2366, #2357.

baseballyama:
## Which loop matched official

`reorder_reactive_statements_after_functions` accumulates the continuation lines of a `$:`
statement twice, and the two break-sets disagreed on `//`:

| loop | break-set (before) |
|---|---|
| `needs_reorder` decision scan | `is_empty()` \| `"$:"` \| `"function "` |
| collection scan | those **plus** `"//"` |

Official was run to settle it (`submodules/svelte` @ 5.56.8, `generate: 'server'`) on the two
shapes #2387 lists as unprobed — a `//` as the first line after the header, and a `//` in the
middle of a continuation:

```svelte
<script>
	export let a = 1;
	export let b = 2;
	let total;
	$: total =
		a +
		// second operand
		b;
	let after = 3;
</script>
```

```js
	// second operand
	let after = 3;

	$: total = a + b;
```

The comment is relocated and the statement stays whole: `//` is **not** a statement boundary.
The **decision scan is the one that matched official**; the `"//"` entry in the collection scan
is the defect. Same result for the first-line-after-the-header shape.

## The fix

Both loops now go through one `classify_continuation_line`, so a break-set can no longer move in
one loop without moving in the other:

```rust
enum ContinuationLine { Boundary, Comment, Code }
```

Removing `"//"` from the collection scan is necessary but **not sufficient** — see below.

## What changes for which input

`let foo = 1;` / `$: total =` / `// pick the sum` / `a + b;` / `let after = 2;`

| | statement collected |
|---|---|
| `origin/main` | `$: total =` — header only, comment and body demoted to ordinary code |
| `"//"` removed from the collection set only | `$: total =` + `// pick the sum` — still no body |
| this PR | `$: total =` + `// pick the sum` + `a + b;` |

The middle row is a second layer #2387 does not predict, and it is why the unification the issue
proposes is not on its own the fix. Once `//` stops being a boundary the comment line is
consumed — and then the loop's "does this line complete the statement?" test runs *on the
comment*: its last character is not a continuation operator and the next line does not start with
one, so the loop breaks there instead. A comment line carries no continuation signal of its own,
so `ContinuationLine::Comment` also suppresses that completeness test and lets the next code line
decide. Both loops apply it, so they stay in step.

## How the tests fail before the fix

`line_comment_does_not_end_a_reactive_continuation` calls the function directly (as the six
existing tests at `transform_legacy.rs:2082-2290` do) and asserts the whole statement survives.
Measured, not predicted — with `origin/main`'s collection-scan break restored:

```
---- line_comment_does_not_end_a_reactive_continuation stdout ----
statement split at the comment: let foo = 1;
// pick the sum
a + b;
let after = 2;

$: total =
```

and with only the break-sets unified (the middle row above):

```
statement split at the comment: let foo = 1;
a + b;
let after = 2;

$: total =
// pick the sum
```

`both_reactive_loops_share_one_line_classification` pins the shared classification, which kills
the *opposite* unification — adding `"//"` to the decision scan instead of removing it from the
collection scan. That mutant fails both tests. It is worth having its own test because for
well-formed input the decision scan's treatment of `//` never changes `needs_reorder` on its own:
the accumulation only reaches a comment while the statement is still incomplete, so a code line
always follows it and forces `needs = true` either way.

Mutant results, all measured:

| mutant | `line_comment_…` | `…share_one_line_classification` |
|---|---|---|
| `origin/main` (collection scan only breaks on `//`) | FAIL | pass |
| break-sets unified, completeness test unchanged | FAIL | pass |
| `//` added to the shared boundary set (opposite unification) | FAIL | FAIL |

## Ratchets

**No regeneration needed, and no compiled output moves.**
`reorder_reactive_statements_after_functions` is `pub(crate)` and its only live caller is
`{@const}` lowering (`server/ast/visitors/declaration_tag.rs:346` → `transform_script.rs:266`),
which passes a declaration; `has_reactive` is therefore false and the function returns before
either loop. No `$:` reaches it in a shipping compile, so
`known-failures.{client,server,client-dev}.json` and the warning ratchets are untouched. Nothing
was hand-edited.

## What was run

`cargo fmt --check`; `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings`;
`cargo test --release -p rsvelte_core --lib` (1463 passed) plus the fixture-free `{@const}` /
legacy-`$:` suites — `ssr_const_scope_2059`, `ssr_const_instance_shadow_2144`,
`const_tag_pattern_pin`, `const_tag_multibyte`, `reactive_block_comment_delimiter_2351`,
`legacy_pre_effect_comment_2355`, `reactive_label_comment_ownership`,
`legacy_reactive_member_compound_2373`, `compile_both_parity`, `labeled_statement_preserved` (36
passed).

The fixture suites (`ssr`, `runtime`, `compiler_fixtures`) were **not** run locally: this
worktree has no generated `fixtures/`, and generating them was not possible here. They panic
rather than silently pass when fixtures are absent, so there is no false green — CI covers them.

## Out of scope, found on the way

Two adjacent defects in the same accumulation, left alone:

- A comment line also blocks the *lookahead*: in `$: total = a` / `// wait` / `+ b;`, the
  `next_starts_continuation` probe reads the comment, not the `+` line, so accumulation never
  starts. Both loops agree here, so it is not this issue.
- `sort_reactive_in_place` (`:1211`) has a third copy of the break-set. It agrees on `//`, but it
  never threads the template-literal state, so a `$:`, blank or `function ` line *inside* a
  template literal is still a boundary there.

Fixes #2387

### DIFF
--- a/.changeset/reactive-continuation-boundary.md
+++ b/.changeset/reactive-continuation-boundary.md
@@ -1,0 +1,10 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix(compiler): keep a legacy `$:` statement whole across a `//` line
+
+The two accumulation loops in the server legacy `$:` reorder disagreed on
+whether a `// …` line ends a continuation. They now share one line
+classification, and a comment neither ends the statement nor completes it, so
+`$: total =` / `// c` / `a + b;` stays one statement as official emits it.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_legacy.rs
@@ -823,6 +823,33 @@ fn scan_reactive_line(line: &str, mut depth: i32, in_template: bool) -> (i32, bo
     (depth, false)
 }
 
+/// How one line is treated while accumulating the continuation lines of a `$:`
+/// statement.
+enum ContinuationLine {
+    /// Ends the statement.
+    Boundary,
+    /// Part of the statement, but carries no continuation signal of its own, so
+    /// whether the statement is complete is decided by the next code line.
+    Comment,
+    Code,
+}
+
+/// Both accumulation loops below must share this classification, or the decision
+/// scan mis-routes a statement the collection scan then splits differently.
+fn classify_continuation_line(trimmed: &str, in_template: bool) -> ContinuationLine {
+    // Inside a template literal every line is literal text.
+    if in_template {
+        return ContinuationLine::Code;
+    }
+    if trimmed.is_empty() || trimmed.starts_with("$:") || trimmed.starts_with("function ") {
+        ContinuationLine::Boundary
+    } else if trimmed.starts_with("//") {
+        ContinuationLine::Comment
+    } else {
+        ContinuationLine::Code
+    }
+}
+
 // The only live caller is `{@const}` lowering, whose input is a declaration, so no `$:`
 // reaches this in a shipping compile.
 pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> String {
@@ -845,8 +872,6 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
         let mut needs = false;
         let mut in_reactive_multiline = false;
         let mut reactive_depth: i32 = 0;
-        // Threaded like the collection loop below; if the two disagree on where a
-        // block ends, this scan mis-routes.
         let mut in_template = false;
         let mut i = 0;
         while i < lines.len() {
@@ -906,13 +931,8 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
                         i += 1;
                         while i < lines.len() {
                             let nt = lines[i].trim();
-                            // Inside a template literal these are literal text, not
-                            // statement boundaries.
-                            if !acc_template
-                                && (nt.is_empty()
-                                    || nt.starts_with("$:")
-                                    || nt.starts_with("function "))
-                            {
+                            let kind = classify_continuation_line(nt, acc_template);
+                            if matches!(kind, ContinuationLine::Boundary) {
                                 break;
                             }
                             (acc_depth, acc_template) =
@@ -945,7 +965,12 @@ pub(crate) fn reorder_reactive_statements_after_functions(script: &str) -> Strin
                             } else {
                                 false
                             };
-                            if !is_cont && !following_starts && acc_depth <= 0 && !acc_template {
+                            if !matches!(kind, ContinuationLine::Comment)
+                                && !is_cont
+                                && !following_starts
+                                && acc_depth <= 0
+                                && !acc_template
+                            {
                                 break;
                             }
                         }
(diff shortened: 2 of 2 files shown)


baseballyama:
Flagging a collision before this is merged: **this PR fixes ~450 lines that #2397 exists to
delete**, and both are open.

I verified the reachability claim independently rather than taking either issue's word for it —
every reference to the entry point outside this file's own `#[cfg(test)]` block:

```
$ grep -rn reorder_reactive_statements_after_functions crates --include "*.rs" | grep -v "fn "
…/transform_legacy.rs:2081,2089,2229,2242,2254,2289   <- all inside #[cfg(test)]
…/transform_script.rs:266                             <- the only production caller

$ grep -rn transform_script_content_inner …                 -> transform_script.rs:26 (one caller)
$ grep -rn transform_script_content_with_imports_and_derived -> declaration_tag.rs:346 (one caller)
```

That terminates at `{@const}` lowering, which Phase 1 restricts to a `VariableDeclaration`
(`1_parse/state/tag.rs:76,146,205`) — so the text reaching this scan can never contain a line
starting `$:`, and the function's first guard (`if !has_reactive { return … }`) always fires.
This PR's own report reaches the same conclusion from the other direction ("no `$:` reaches
either loop in a shipping compile and no compiled output moves"), and #2397 reached it by
instrumentation over 18 binaries / 42 tests. Three independent routes agree.

So the defect fixed here is real but **unobservable through the compiler**, and the new tests
necessarily drive the function directly.

Suggested sequencing, for whoever merges:

1. If #2397 lands, this PR's subject disappears with it — merge #2397 and close #2387 as moot,
   quoting the call chain rather than the "no repro" shorthand.
2. If #2397 is blocked or deferred, this is the right interim fix and should go in: the shared
   `classify_continuation_line` is what stops the two loops from drifting apart again, and the
   PR's mutant table shows the break-set unification alone is **not** sufficient — a detail
   #2387 does not predict and that would be re-derived from scratch later.

What should not happen is both landing independently, or #2530 landing and then being reverted
by a deletion nobody connected to it.

Also worth carrying into whichever PR wins: this PR reports a **third** copy of the break-set in
`sort_reactive_in_place` (`:1211`) that agrees on `//` but never threads template-literal state.
If the machinery is deleted, that copy goes with it; if it survives, that is the next drift.

PR #2537, opened 2026-08-07
perf: stop parsing the same script three times for the store scan

baseballyama:
Stacked on #2519.

Two independent reductions in redundant parsing, both verified byte-identical.

## 1. One parse per statement, not one per store

Store member mutations ran a parse + esrap re-print **per subscribed store**,
threading each iteration's text into the next — while the rewriter already
matches every store in one traversal and looks `prop_store_names` up by name.
The prop and state update-expression passes are likewise the same visitor
called with complementary argument lists, and its classifier already tries
props first, which is exactly what running one before the other did.

Honest framing: this removes parses deterministically, but **paired interleaved
runs came back inside noise on all four corpora** (−1.9% … +4.8%, 3–5/8 wins).
It is not a measured wall-clock win.

## 2. No third TypeScript parse for the store scan

`detect_store_subscriptions` reads the script with type-only syntax blanked out
so byte positions survive, and built that copy with a full `oxc_parser`
TypeScript parse of its own — the **third** parse of a script the compiler had
already parsed into `retained_scripts` and stripped of TypeScript.

The guard is **byte equality, not the pointer identity** its sibling in
`ScriptContent::from_script_with_ts` uses. That matters: this caller slices
`analysis.source`, a *copy* of the component source, so the addresses never
match even for the same script. With pointer identity the reuse path fired
**zero** times — and because the fallback is the original implementation, every
output stayed byte-identical and every test stayed green. A hit counter is what
caught it, and `a_typescript_script_is_blanked_without_reparsing_it` now pins
both halves (same bytes out, zero re-parses).

Redundant TypeScript parses, counted deterministically:

| corpus | before | after | bytes no longer re-parsed |
|---|---|---|---|
| Huly plugins | 1,384 | **0** | 3.02 MB |
| open-webui | 361 | **1** | 1.38 MB |
| SMUI | 393 | **0** | 0.52 MB |
| carbon-components-svelte | 0 | 0 | — (no TS scripts: the negative control) |

## Verification

**14,036 compiled outputs byte-identical** — four corpora × {client, server} ×
{prod, dev}. The four target columns are asserted to differ from one another
first: the previous baseline's `server+dev` column turned out to be a duplicate
of `client prod` (the flags never reached the binary), so it had been comparing
client output to itself.


baseballyama:
Added a third commit: **let the `await` probe decide when the `$` probe cannot.**

The gate on the analyze-phase await / rune-reference walk was
`has '$' || has "await"`. Every rune name starts with `$`, so the first probe
passes on most components and the second — true for about 1% of them — never
gets a say. `$` is only *informative* while rune detection is on: once runes
mode is already decided, the walk's only surviving output is `has_await`, which
an `await`-free source settles without walking anything.

Observationally equivalent — the `has_rune_reference` half of the result is read
only under `needs_rune_detection`, and the module walk already carried that
condition. All 14,036 outputs stay byte-identical, and
`a_rune_looking_source_only_needs_the_walk_while_rune_detection_is_on` pins the
truth table with a positive control on each half.

A note on measurement for all three commits: this machine has been running at a
load average of 50–215 from unrelated work for the duration, so wall-clock A/B
is not usable here (the same paired protocol reported −18.3% and +4.4% on
neighbouring corpora in one run). Every claim above is either a deterministic
count or byte identity; none of them is a measured wall-clock win.

baseballyama:
Superseded by #2576, which was the tip of this stack and contained every commit and changeset from this PR. Verified before merging: all four substantive commits (`002b84a3`, `e02426cc`, `bafb5835`, `1c77e55c`) and all five changesets were ancestors of the merged branch.

Closing rather than merging because squash-merging a stack rewrites the base commits and would have forced a rebase-and-rerun of the whole chain for no additional content.

Issue #2399, opened 2026-08-06
codegen (client): the `.svelte.(js|ts)` module path keeps top-level comments that official drops


baseballyama:
## Summary

In the `.svelte.(js|ts)` module path, the official compiler **drops every
top-level comment**; rsvelte **keeps them all**. Target: `client` (and
`client-dev`).

**Pre-existing, not caused by #2380.** Verified by forcing the pre-#2380
handwritten text-printer path with a mutant: it keeps the comments too. Both
print routes keep them, so this is upstream-fidelity, not a printing-route
regression.

## Repro

`m.svelte.js`, `generate: 'client'`, `dev: false`. Every position tested is
dropped by official and kept by rsvelte:

| source | official | rsvelte |
|---|---|---|
| `// lead` before a declaration | dropped | **kept** |
| comment between two top-level statements | dropped | **kept** |
| comment after an `import` | dropped | **kept** |
| `/** doc */` before an exported function | dropped | **kept** |
| license header above `import` + a rune | dropped | **kept** |
| comment **inside a function body** | kept | kept |

```js
// input
// top level comment
export const a = 1;
```

```js
// official
/* m.svelte.js generated by Svelte v5.56.8 */
import * as $ from 'svelte/internal/client';

export const a = 1;
```

```js
// rsvelte
/* m.svelte.js generated by Svelte v5.56.8 */
import * as $ from 'svelte/internal/client';

// top level comment
export const a = 1;
```

Official regenerates the module's top-level body from the AST — it adds the
`$` import even for a module with no runes — so Program-level comments do not
survive, while function bodies are carried through and keep theirs.

## Why this is filed now

It is the blocker for writing #2307's acceptance test honestly. That test was
briefed as "Program-level comments survive the routed module path". Against
**upstream's** output the correct assertion is the opposite — that they are
dropped — and it **fails today**. Written the other way it would pin this
divergence, which is the same mistake `ctor_private_state_set_paren_2253.rs`
made and that #2380 had to unwind.

## The gate can see it, which raises a second question

Comment presence is **not** absorbed by the corpus normalizer — measured
through `flattenTemplateHoles` → oxfmt 0.62.0 → `stripBlankLines` with
`compatibility/.oxfmtrc.json`, reformat control confirming oxfmt rewrote both
sides:

| difference | result |
|---|---|
| comment position (own line vs line end) | **VISIBLE** |
| comment presence | **VISIBLE** |
| Program-level comment presence | **VISIBLE** |
[...]

PR #2543, opened 2026-08-07
test(compiler): pin the module comments upstream keeps, not only the ones it drops (#2399)

baseballyama:
## #2399 is already fixed — this PR is what was left over

**Do not merge this expecting a fix.** #2399 was fixed on `main` by #2522
(`335989a3`, merged 2026-08-07 23:13 JST) while this branch was being written. I
verified that from primary data, not from the commit message: the issue's own
repro is `MATCH` on `main` under `scripts/compat-corpus/one.mjs --raw`, and the
ten-shape table below passes on an unmodified `77b4f8bc` build (binary mtime
checked against source mtime).

What was still missing is the **other half of the rule**.

## The mechanism, and why one half of it was untested

`compileModule` prints with esrap, and `client_module`
(`transform-client.js:704`) returns a freshly built `{ type: 'Program', body }`
with no `loc`. esrap's `Program` visitor calls `body()`, whose first act is
`reset_comment_index(node)` (`esrap/src/languages/ts/index.js:257`):

```js
if (!node.loc) {
	comment_index = comments.length;   // park past every comment
	return;
}
```

So nothing flushes at Program level. The **only** thing that rewinds that single
stateful cursor is a nested body that *does* carry a `loc` — a function block, a
class body — because `'BlockStatement|ClassBody'` calls `body()` again. Two
consequences follow, and they are equally load-bearing:

| | |
|---|---|
| a comment before any located body (header, JSDoc, between two statements, end of file) | **dropped** — the cursor never comes back |
| a comment **inside** a located body | **kept** — the rewind lands on it |
| a top-level comment **after** a located body | **kept** — the same rewind lands on it, and the next statement's leading flush emits it |

`tests/module_comment_ownership.rs` (shipped with #2522) covers rows 1 and 2. Row
3 was covered nowhere, and it is not a corner: the measurement in
[#2399's follow-up comment](https://github.com/baseballyama/rsvelte/issues/2399#issuecomment)
found official **preserves** the inserted comment in **112 of 192** generated
slots. A fix built on "drop them all" is wrong in the majority of positions, and
nothing on `main` would have caught one.

## What the test adds

`crates/rsvelte_core/tests/module_toplevel_comments_2399.rs` — ten shapes ×
`client` and `client-dev`, each asserted against the official compiler's
**byte-exact whole-file output** (`module_comment_ownership.rs` uses `contains`
on `dev: true` only). Seven shapes are drops (`lead_before_decl`,
`between_stmts`, `after_import`, `jsdoc_before_fn`, `license_header`,
`trailing_same_line`, `end_of_file`); three are keeps (`inside_fn_body`,
`after_fn_toplevel`, `inside_fn_then_two_top`).

Discriminating power, measured rather than asserted. Mutant: park the cursor at
the **end of a located body** — the natural over-drop, and exactly what a
blanket-drop fix degenerates into.

```
module_comment_ownership.rs        5 passed; 0 failed      ← survives the mutant
module_toplevel_comments_2399.rs   FAILED, 4 divergences:
  --- after_fn_toplevel [client] / [client-dev]
  --- inside_fn_then_two_top [client] / [client-dev]
```

No other row moves — the drop rows and `inside_fn_body` still pass — so the two
new rows are carrying the whole difference.

For the record, before #2522 landed, 14 of these 20 assertions failed on
`744b0c27`, every one by rsvelte emitting a comment official omits:

```
--- lead_before_decl [client]
 import * as $ from 'svelte/internal/client';

+// lead
 export const a = 1;
```

## Two findings this turned up, reported not fixed

**`generate: 'server'` still diverges, and #2522 did not touch it.** All ten
shapes diverge on `server`, and not only on comments — the margins are wrong too
(`after_fn_toplevel` and `inside_fn_then_two_top` are byte-identical to official
on `client` but lose the blank line before the surviving comment on `server`).
The cause is structural: `transform_server_module` splices the transformed source
as text (`parts.join("\n")`) instead of printing an AST, so no printer-level fix
reaches it. Routing it through `to_oxc` + esrap would fix comments and margins
together, with its own blast radius. The test documents the exclusion in its
module doc rather than pinning the current output.

**The issue's framing of the component path is wrong in both directions.**
Official *drops* `<script module>` top-level comments in the component path too
(verified against 5.56.8) — the `component_block.loc = instance.loc` trick at
`transform-client.js:393` rewinds the cursor onto the *instance* script, which is
after the module script, so the module script's comments stay parked. rsvelte
already matches; `client_script_comments_pin::module_and_instance_script_comments`
pins it.

## Ratchets

None regenerated, none needed. `node scripts/compat-corpus/matrix/run.mjs`:
`match 1665 / js-mismatch 330`, no regressions and no stale entries — the same
330 as the baseline. (The 240 module-file entries the issue's follow-up comment
counts come from #2435, which is not merged; `main`'s matrix generates no
`.svelte.(js|ts)` cases.)

Refs #2399


baseballyama:
main が #2522 で入れた `options.unlocated_program` と、この PR の `has_loc(span.start)`
による自動判定が同じ挙動に別経路で到達していたため、コンフリクト解消では**両方を残しました**
（フラグ利用側の挙動を変えずに、この PR の忠実なルールを足す形）。

ローカル確認: `module_toplevel_comments_2399` / `sourcemaps_gate` / `client_script_comments_pin`
/ `print` / `ssr` / `module_block_margins_2341` / `rsvelte_esrap` いずれも緑です。

どちらか一方に寄せるべきなら、この解消をたたき台にしてください。

PR #2548, opened 2026-08-07
test(compat): assert ratchet-doc cluster counts partition the ratchet (#2500)

### LINKED ISSUE #2500: test(compat): ratchet-doc cluster counts can sum correctly while double-citing an entry
Several `*-known-failures.md` docs justify their ratchet by **clusters with counts** rather than per entry. Nothing checks that those cluster counts partition the ratchet, so an entry can be cited under two clusters while the counts still sum to the file length — and every number involved looks correct.

## The instance

`compatibility/validator-known-failures.md`, before #2496:

- cluster 1 (*error spans not populated*, ~141) cited `invalid-node-placement-5` and `module-script-reactive-declaration` as examples
- cluster 3 (*content differs*, 13) gave those same two fixtures their own wording bullets
- the three counts summed to 207, exactly the ratchet length, as if each entry were counted once

Both statements were written in good faith and both described something real — the fixtures did have two divergences each. But the counts can only sum to the file length if the clusters are a *partition*, and a doubly-cited entry means they are not. So "the counts sum correctly" and "the clusters describe the file" were both true of *different populations*, and neither is checkable against the other.

Measured today (#2496), those two fixtures are span-only failures, and both wording defects are gone. The example lists had also drifted: they were cited under the error-span cluster while being warning failures.

## Why the existing check does not catch it

#2488 added `known-failures-md-check.mjs` coverage for every ratchet doc, asserting the **stated total** matches the JSON length. That check passes here: the doc said 207 and the file had 207. A double-citation is invisible to it, because it only ever compares one number to one number.

## Suggested check

Where a doc states per-cluster counts, assert they sum to the stated total, and require the total to match the JSON length (already checked). That turns "the clusters partition the ratchet" from a convention into something that fails.

The harder half — that each entry is cited under exactly one cluster — needs the clusters to name their members, which most docs deliberately do not do at 141 entries. A sum check is the cheap 80%: it catches double-counting whenever the cluster totals were also adjusted to match, which is the case that currently looks correct from every angle.

## Alternative worth considering

`validator-known-failures.md` after #2496 derives its cluster counts by classifying the failure report the suite prints, rather than maintaining them by hand:

| cluster | entries |
|---|---:|
| error span not populated | 141 |
| warning span-only | 36 |
| warning content | 4 |
| | **181** |

If the classification lived in the test rather than in a one-off script, the doc's table could be generated and the partition property would hold by construction instead of by assertion. That is a larger change and is not required to close this issue.

baseballyama:
Closes #2500.

## What the gap was

`known-failures-md-check.mjs` (from #2488) asserts each doc's **stated total** equals its JSON length. That check passes on a doc whose clusters double-cite an entry, because it only ever compares one number to one number. The issue's instance: `validator-known-failures.md` cited two fixtures under the error-span cluster *and* gave them their own bullets under the wording cluster, while `141 + 53 + 13` still summed to 207.

## What this adds

**A declared partition, whose total comes from the data.** The doc states the addends; the checker computes the total from the JSON — length, or the ids under a prefix for a sub-population (`comment-slot`'s 160 is not the matrix ratchet's 162).

```
Partition of `lint-known-failures.json` by rule: `36 + 16 + 6 + 9 + 7 + 2 + 2 + 2`
Partition of `matrix-known-failures.json` entries under `comment-slot/` by seed: `56 + 56 + 24 + 8 + 8 + 8`
```

Reading the total from the doc would not have worked: a sum checked against a total the same author typed agrees with itself whenever both were adjusted together, which is precisely the case that already looks correct from every angle.

Nine partitions across five docs. All nine sum correctly today — this is a regression guard, not a burn-down. `fmt-known-failures.md` already had the property and said why (an id carrying two clusters' divergences is filed under its dominant one); that tie-break is now stated next to the numbers it makes meaningful, and `validator-known-failures.md` gets the same sentence.

Declared in the script rather than scanned out of the docs, for the reason `RATCHETS` is: a scan cannot tell "this doc states no clusters" from "someone deleted the line it stated". Both directions fail — an undeclared `Partition of` line, and a declared one that is missing.

## Two checks that were narrower than their subject

- **`summing to N (\`a + b + NxM\`)` ran on `warning-known-failures.md` only.** A convention checked in one file is one the next doc to adopt it gets for free and unverified. Now scanned in every ratchet doc.
- **The count parser stopped at the first line stating a ratchet's size.** #2490 re-baselined 529 → 528; the header moved and a sentence 45 lines above did not, and the check bound the header. Every occurrence is read now and they must agree. The restatement in `warning-known-failures.md` is reflowed so its key and its count share a line — that is what brings it under the check at all.

## Verification

The self-test was unit tests on the parser. The checker itself had only ever run on inputs it passes, which says nothing about whether it can fail. It now also runs end-to-end against a mutated copy of the real `compatibility/` docs — one property broken per case, `KNOWN_FAILURES_DIR` pointing at the copy — with the **unmutated copy as the control**:

| case | mutation |
|---|---|
| a stale cluster count fails | one addend −1 |
| **a double-cited entry fails** | one addend +1 with the others left alone, i.e. the #2500 shape |
| a deleted partition line fails | line removed |
| a sub-population partition is bound to its prefix | `comment-slot` sum moved to 162 — the *ratchet's* length |
| an undeclared partition fails | label changed |
| a stale restatement fails | second copy of a count moved to 529 |

19/19 pass. That control earned its keep immediately: the restatement sentence as I first wrote it wrapped between its key and its number, so the mutation was invisible and the "stale restatement" case passed for the wrong reason. Reflowing the sentence is the actual fix.

## Not in scope

The harder half the issue names — that each entry is cited under **exactly** one cluster — needs the clusters to name their members, which a 141-entry cluster deliberately does not. The sum check is the cheap 80%: it catches double-counting whenever the cluster totals were adjusted to match.

`validator-known-failures.md` will be rewritten by #2496 (207 → 181, clusters 141/36/4). The partition line here uses today's numbers; #2496 replaces them and the check then forces its own to sum. Noted on that PR.


baseballyama:
Holding this one until #2593 lands. Its partition assertions were derived from a **162-entry** baseline; `main` is at 290 and #2593 takes it to 234, and the module-path cases #2435 added mean the two `comment-slot` partitions (`100 + 32 + 28 + 0` by what diverges, `56 + 56 + 24 + 8 + 8 + 8` by seed) no longer cover the family — they describe the `.svelte` template seeds only, which is now a strict subset. Since `mergeStateStatus` is UNSTABLE rather than DIRTY, git would merge these lines cleanly and the checker this PR adds would then fail **on `main`** — the same shape as the failure #2593 is fixing. Rebase onto post-#2593 `main` and re-derive from the JSON, and I will merge.

baseballyama:
Concrete numbers for the rebase, measured from `main` + #2593 (234 entries), so this doesn't need re-deriving:

```
Partition of `matrix-known-failures.json` by family: `2 + 232`
```

The two `comment-slot/` partitions need more than a number change. Both of that family's tables describe the **`.svelte` template seeds only** (160), but the `prefix: 'comment-slot/'` population is now **232** — #2435 added 192 `.svelte.(js|ts)` module cases, 72 of which are ratcheted. The ids are `comment-slot/<seed>__L<nn>__<kind>.svelte[.js|.ts]`, so the template subset is a *suffix* distinction and no prefix can select it.

Cleanest fix is to let both tables cover the whole family rather than teach the checker a second selector — the module entries are a real cluster, not an exception:

| by what diverges | entries |
|---|---:|
| rsvelte drops a comment official keeps | 100 |
| the comment survives but lands somewhere else | 32 |
| rsvelte emits a comment more than once | 28 |
| **official drops a trailing Program-level comment, rsvelte keeps it** (#2399, module path) | **72** |
| anything other than the comment itself | 0 |

→ `100 + 32 + 28 + 72 + 0`

| by seed | entries |
|---|---:|
| legacy-reactive | 56 |
| module-script | 56 |
| await-block | 24 |
| module-class-state | 24 |
| module-rune-exports | 24 |
| module-ts-extension | 24 |
| class-private-state | 8 |
| class-static-block | 8 |
| snippet-render | 8 |

→ `56 + 56 + 24 + 24 + 24 + 24 + 8 + 8 + 8` = 232

Verified against the JSON:

```
family        { binding-position: 2, comment-slot: 232 }
comment-slot  { template: 160, module: 72 }
module        by target { client: 24, server: 24, client-dev: 24 }
```

Worth noting for this PR's own thesis: the module rows were *invisible* to both tables while the header count was correct, which is exactly the double-citation-adjacent failure the partition assertion exists to catch — a doc whose numbers all look right because the population they describe silently stopped being the whole file.

baseballyama:
### Merge note: the doc check is **line-sensitive**, and that is easy to break silently

Resolving this branch against `main` turned up a trap worth recording for whoever merges
the next one of these.

`known-failures-md-check.mjs` recognises a **restatement** of a ratchet's count only when
the backticked filename and its `N entries` appear **on the same line** — `statedCounts`
scans line by line and slices to the right of the filename. `main`'s prose for
`warning-known-failures.md` wraps the filename and the count onto separate lines:

```
`warning-known-failures.client.json`, `.server.json` and `.client-dev.json` hold
the same 28 entries; …
```

Taking that wording verbatim is not a formatting preference — it **deletes two
machine-checked restatements**, and every check still reports green afterwards, because a
restatement that is not on one line is not a restatement to the checker. That is the same
shape as a ratchet entry suppressing everything about its subject: the thing that would
have caught the regression is the thing the change removes. This branch therefore keeps
the single-line wording.

The self-test's mutation anchor was re-pointed (`528` → `74`) so its
`a stale restatement fails` case is discriminating again; it now correctly reports
`states 75 entries`.

### Both conflicting counts were wrong, so neither side could be picked

- `warning-known-failures.md`: `main` said **51**, this branch said **528**. Measured from
  the JSON with `jq 'length'`: **28** codes and **74** positions.
- `validator-known-failures.md`: the partition line summed to `141 + 35 + 3` = **179**
  against a **172**-entry ratchet. Set to `141 + 30 + 1`, matching the table `main`
  already carried.

No ratchet JSON was hand-edited: all 35 `compatibility/*known-failures*.json` /
`*excluded*.json` blobs in the merged tree hash **identical to `origin/main`**, so no
regeneration run was owed.

PR #2555, opened 2026-08-07
fix(compiler): reflow the dev `$.tag(...)` wrap when its value carries a comment

### LINKED ISSUE #2495: dev `$.tag(...)` is not reflowed when its value carries a comment
Found while fixing #2437 (PR #2494), which deliberately left this out of scope.

When a rune class field's **value** carries a leading comment, official's dev output breaks
`$.tag(value, name)` across lines. rsvelte keeps the call on one line and leaves the comment
before it.

## Repro A — public field, comment above it (`client-dev`)

```js
export class C {
	// c
	n = $state(0);
}
```

```js
// official
	#n = $.tag(
		// c
		$.state(0),
		'C.n'
	);

// rsvelte
	#n = // c
	$.tag($.state(0), 'C.n');
```

## Repro B — private field, comment after the `=` (`client-dev`)

```js
export class C {
	#n = // c
		$state(0);
}
```

```js
// official
	#n = $.tag(
		// c
		$.state(0),
		'C.#n'
	);

// rsvelte
	#n = // c
	$.tag($.state(0), 'C.#n');
```

Both are `client-dev` only. On `client` both cases already match — `$.state(0)` with a comment
in front of it needs no reflow.

## Severity

Formatting only. Both outputs are **valid, semantically identical JS** — the comment is in a
position where it ends the line either way. This is not the invalid-output class of #2467.

## Why this is not one fix

The two repros reach that output through **different rsvelte code paths**, so a fix at either
site covers one of them:

- **Repro A** goes through `emit_class_field` in
  `crates/rsvelte_core/src/compiler/phases/3_transform/client/class_transforms.rs`, via the
  `comment_infix` branch that PR #2494 kept for public fields.
- **Repro B never reaches `emit_class_field` at all.** The rune-field scanner tests
  `has_pattern` against a *single* line, so `#n = // c` matches neither `= $state(` nor
  `=$state(`; the line is classified as a plain non-rune member and the `$state(` → `$.state(`
  and `$.tag` rewrites happen in later text passes. `parse_state_field` also sets
  `trailing_comment: None` and discards anything between the `=` and the rune, so even if the
  line were accumulated the comment would not survive to the emitter.

Naming both here for the reason #2485 and the `track_reactivity_loss` case (#2375) call out:
a one-sided fix that looks complete is the failure mode.

## What would have to change

Official's shape is what esrap produces when an argument carries a leading comment — the call
cannot stay on one line without the `//` swallowing the rest of it. rsvelte applies the dev
`$.tag` wrap textually, after the comment has already been placed, so it has no opportunity to
reflow. Matching it means the `$.tag` wrap has to know whether its value argument carries a
comment.

## Gate visibility

Visible to the generated shape matrix (`comment-slot`), which is where repro A's family lives.
Repro B's shape — a comment between `=` and the rune — is not currently generated by any axis;
the comment axis inserts at line boundaries, and this needs a mid-expression slot. Not visible
to the collected corpus: all three `known-failures.*.json` ratchets are empty, so no real-world
corpus file has either shape.

baseballyama:
## Still broken on today's `main`

Measured first, on `c193616e` (after #2522 / #2536 landed today), with a freshly built
`librsvelte_napi.dylib` copied into `.corpus-cache/rsvelte.node`:

```
$ node scripts/compat-corpus/one.mjs reproA.svelte.js --target client-dev --raw
 export class C {
-	#n = $.tag(
-		// c
-		$.state(0),
-		'C.n'
-	);
+	#n = // c
+	$.tag($.state(0), 'C.n');
```

Repro B produced the same diff with the `'C.#n'` label. So the fix was still needed.

## Cause

`tag_class_field_ast::handle_property_definition` / `handle_this_assignment` replaced exactly
the value's span with `$.tag(<value>, '<label>')`. When a comment sits between the `=` and the
value, that leaves the comment ahead of the `$.tag(` — which is why the whole call ends up on
the line *after* the comment.

Upstream's shape is what esrap produces for a call whose argument carries a leading comment:
it cannot stay on one line, because the `//` would swallow the rest of the call.

## Fix

One edit builder (`tag_edit`) shared by both handlers. It looks at the gap between the `=`
(located with a comment-skipping scan, so an `=` inside a comment is not mistaken for it) and
the value; when that gap holds a comment, the edit starts at the `=` instead of at the value
and emits the multi-line form, indented from the field's / assignment's own line.

Both handlers go through it, so the class-field shape (repro A) and the `this.#field =` shape
are covered by one change rather than one each — the one-sided-fix failure mode the issue
calls out.

Note the issue predicted repro B would need a second fix in `class_transforms.rs`. It does
not: repro B does bypass `emit_class_field`, but the later text passes leave it in the *same*
intermediate shape (`#n = // c` ⏎ `$.state(0);`) that repro A reaches, so both are fixed at the
single `$.tag`-wrap site. `class_transforms.rs` is untouched.

## Verification

Repro A, repro B and the no-comment control all `MATCH` against the official compiler on
`client-dev`. A wider sweep (`$state.raw`, a proxied object value, a `this.#n =` constructor
assignment, a comment above a *private* field) also matches.

Negative control — the new test file on the pre-fix `tag_class_field_ast.rs`:

```
test result: FAILED. 3 passed; 4 failed
failures:
    a_this_assignment_with_a_comment_after_the_equals_reflows
    private_field_with_a_comment_after_the_equals_reflows
    public_field_with_a_comment_above_it_reflows
    raw_and_proxied_values_reflow_too
```

The three controls pass before *and* after, which is what makes the four repro tests
discriminating rather than a shared property of every input.

Every expected string in `crates/rsvelte_core/tests/tag_reflow_comment_2495.rs` was taken by
running `compileModule(src, { generate: 'client', dev: true })` from `submodules/svelte`, not
transcribed from the issue.

Suites, `--release`: `runtime` (5/5, 150 s), `ssr` (2/2), `print` (2/2), `sourcemaps` (2/2),
`rsvelte_core --lib` (1488/1488). `cargo fmt` + `cargo clippy --all-targets --all-features -D
warnings` clean.

## Ratchets

`pnpm run corpus:matrix`: **162 known divergences before, 162 after** — `match 1833 /
js-mismatch 162`, identical to `compatibility/matrix-known-failures.json` (162 entries). No
regressions and no shrink, so no re-baseline. Repro A's exact shape turns out not to be
generated by the `comment-slot` axis after all (the axis inserts at line boundaries, and a
comment above a *public* rune field is what lands in the infix slot).

The three collected-corpus ratchets are all **empty**. That is itself the argument that this
change cannot move them: if any corpus file had a comment in the value's leading position, it
would already be diverging today (flat vs. reflowed) and would be listed. CI's corpus gate
confirms.

## Out of scope (reported, not fixed)

Two neighbouring shapes still diverge, both before and after this PR:

- A **block** comment above a public rune field (`/* c */` ⏎ `n = $state(0)`). rsvelte emits it
  as a prefix line, not in the infix slot, so it never reaches the `$.tag` wrap. That is a
  comment-*placement* bug in `emit_class_field`, upstream of this one.
- `$derived` with a comment above it. Official emits `$.derived((// c` ⏎ `) => 1)` — esrap
  attaches the comment inside the arrow's parameter list. This PR moves rsvelte to
  `// c` ⏎ `$.derived(() => 1)` inside the reflowed call, which is closer but still not equal.

## Merge-order note

`crates/rsvelte_core/.../client/class_transforms.rs` is busy (#2531, #2541, #2491), but this PR
does not touch it. The only source file changed is `client/tag_class_field_ast.rs`, which none
of those three PRs touch — no conflicting hunk with any of them.

Fixes #2495

### DIFF
--- a/.changeset/tidy-pugs-jump.md
+++ b/.changeset/tidy-pugs-jump.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Reflow the dev `$.tag(...)` wrap of a class state field whose value carries a leading comment
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/tag_class_field_ast.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/tag_class_field_ast.rs
@@ -311,9 +311,91 @@ fn handle_property_definition<'a>(
         None => format!("{}.#{}", class_name, field_name),
     };
 
+    replacements.push(tag_edit(
+        source,
+        prop.span().start,
+        prop.key.span().end,
+        init_span,
+        tag_fn,
+        &label,
+    ));
+}
+
+/// The `$.tag(...)` wrap for one value, reflowed when a comment sits between the
+/// `=` and the value: esrap cannot keep such a call on one line, because a `//`
+/// comment would swallow the rest of it.
+fn tag_edit(
+    source: &str,
+    stmt_start: u32,
+    lhs_end: u32,
+    init_span: Span,
+    tag_fn: &str,
+    label: &str,
+) -> Edit {
     let init_text = &source[init_span.start as usize..init_span.end as usize];
-    let rewrite = format!("{}({}, '{}')", tag_fn, init_text, label);
-    replacements.push((init_span.start, init_span.end, rewrite));
+    let flat = (
+        init_span.start,
+        init_span.end,
+        format!("{}({}, '{}')", tag_fn, init_text, label),
+    );
+    let Some(eq) = assignment_eq_offset(source, lhs_end, init_span.start) else {
+        return flat;
+    };
+    let comment = source[eq as usize + 1..init_span.start as usize].trim();
+    if comment.is_empty() {
+        return flat;
+    }
+
+    let indent = line_indent(source, stmt_start);
+    let arg_indent = format!("{}\t", indent);
+    let comment = comment
+        .lines()
+        .map(str::trim)
+        .collect::<Vec<_>>()
+        .join(&format!("\n{}", arg_indent));
+    // The value moves one level deeper, so its own continuation lines follow.
+    let init_text = init_text.replace('\n', "\n\t");
+    (
+        eq + 1,
+        init_span.end,
+        format!(
+            " {tag_fn}(\n{arg_indent}{comment}\n{arg_indent}{init_text},\n{arg_indent}'{label}'\n{indent})"
+        ),
+    )
+}
+
+/// Offset of the `=` separating a field / assignment target from its value,
+/// skipping comments so an `=` inside one is not mistaken for it.
+fn assignment_eq_offset(source: &str, from: u32, to: u32) -> Option<u32> {
+    let bytes = source.as_bytes();
+    let end = to as usize;
+    let mut i = from as usize;
+    while i < end {
+        match bytes[i] {
+            b'/' if bytes.get(i + 1) == Some(&b'/') => {
+                while i < end && bytes[i] != b'\n' {
+                    i += 1;
+                }
+            }
+            b'/' if bytes.get(i + 1) == Some(&b'*') => {
+                i += 2;
+                while i < end && !(bytes[i] == b'*' && bytes.get(i + 1) == Some(&b'/')) {
+                    i += 1;
+                }
+                i += 2;
+            }
+            b'=' => return Some(i as u32),
+            _ => i += 1,
+        }
+    }
+    None
+}
+
+fn line_indent(source: &str, pos: u32) -> &str {
+    let line_start = source[..pos as usize].rfind('\n').map_or(0, |i| i + 1);
+    let line = &source[line_start..];
+    let width = line.len() - line.trim_start_matches([' ', '\t']).len();
+    &line[..width]
 }
 
 fn walk_method_stmt_for_this_assigns<'a>(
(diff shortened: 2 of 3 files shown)


baseballyama:
Rebased onto `origin/main` (`723df2ff`). Both red checks on the previous push were `main`
breakages my base predated, not defects in this patch.

**1. `Compiler parity` → "Verify known-failures.md matches the JSON ratchets".** Fixed by
#2552 (`ea28ace7`), which is now in this branch's history.

**2. `Test (ubuntu-latest, 1)` → "Run tests".** The log (job 92937289847) shows one failure:

```
crates/rsvelte_core/tests/validator.rs:569
2 entr(ies) in compatibility/validator-message-known-failures.json now match — remove them
(and their justification in the paired .md): ["a11y-anchor-in-svg-is-valid", "svelte-self-deprecated"]
```

That is the two-sided message ratchet reporting a *shrink*, not a regression, and it is
unrelated to dev `$.tag` codegen. Controlled locally rather than argued:

| tree | result |
|---|---|
| this branch, rebased onto `8c53ac49` | FAIL, same two ids |
| identical tree with my only source edit reverted to `origin/main`'s version | FAIL, byte-identical message |

So it reproduces with the patch fully removed — the change is not implicated.

"2 entries now match" has two possible causes, and they need different fixes: the messages were
genuinely fixed, **or** those fixtures stopped reaching the text comparison at all (the test
`continue`s when codes or counts disagree, so a *code* regression removes a fixture from
`diverged` and looks exactly like a shrink). Deleting the entries under the second cause would
permanently hide the regression. It was the first: the ratchet's two entries name #2413 and
#2411, and both are closed — by #2451 (`a11y_invalid_attribute` naming `href` where the source
spells `xlink:href`) and #2477 (`svelte_self_deprecated` capitalising the suggested import
path) — neither of which re-baselined the ratchet.

#2565 has since landed on `main` and emptied the ratchet, so this is resolved by the rebase.

Re-verified after the rebase (which now also carries #2531 and #2541, both in the client
transform area): `validator` 3/3, `tag_reflow_comment_2495` 7/7, `tag_class_field_ast` unit
tests 20/20, `cargo fmt --check` clean. The corpus and shape-matrix gates run in CI on the
merge commit; disk pressure on the dev machine made a local re-run of those inadvisable, and
the pre-rebase measurement (matrix 162 → 162, no regression, no shrink) still stands for the
patch itself.

Issue #2554, opened 2026-08-07
ci: a repo-wide invariant can go red on main with both PRs green — the merge ref is stale and the push-to-main run is cancelled


baseballyama:
`node scripts/compat-corpus/known-failures-md-check.mjs` is **red on `main`** as of `52ac3896`:

```
validator-message-known-failures.json is a ratchet on disk but is not declared in
RATCHETS — add it with the doc that justifies it
```

#2548 fixes the breakage. This issue is about **why neither safety net caught it**, because both failures are structural and will recur.

## The measurement

| when | what |
|---|---|
| 13:57:37Z | #2418's last commit pushed (adds `validator-message-known-failures.json`) |
| 14:14:04Z | #2488 merges — this is the PR that added the "every ratchet on disk is declared" assertion |
| 14:53:31Z | #2418's `Compiler parity` **completes SUCCESS** |
| 16:14:36Z | #2418 merges |

Neither PR is at fault on its own. #2418 added a ratchet under a rule that did not exist when its CI event was created; #2488 added a rule that no existing PR was re-checked against. Each was green for a base that was true when its event fired.

## Net 1: the merge ref is frozen at event-creation time

A `pull_request` job checks out `refs/pull/N/merge`, computed when the **event** is created — not when the job starts. #2418's job *completed* 39 minutes after #2488 merged and still tested a base without it. Nothing re-runs a PR because a different PR merged, so "green" can describe a base an hour or more stale. The deeper runner queue gets, the wider that window.

This is the general hazard for any **repo-wide invariant** — "every X on disk is declared", "every gate has a row in `gate-coverage.md`", "every ratchet has a paired `.md`". A per-file check is safe under a stale base because the PR carries the file it checks. A repo-wide check is not: the property is over the union of both branches, and neither branch's CI ever sees that union.

## Net 2: the push-to-`main` run is cancelled by the next merge

`corpus-compat.yml` does run on `push: branches: [main]` with `compatibility/**` in its paths, and its very first step is this check, deliberately placed to fail fast. That should have caught it minutes after the merge. It did not:

```
2026-08-07T16:14:40Z  (running, still queued 31 min later)  52ac3896
2026-08-07T16:14:31Z  cancelled                             f0c8f3da
2026-08-07T16:08:02Z  cancelled                             5a722059
2026-08-07T16:07:54Z  cancelled                             f390acc3
2026-08-07T15:58:21Z  cancelled                             c193616e
```

[...]

PR #2558, opened 2026-08-07
test(corpus): gate compiler error message and position, on ratchets of their own

### LINKED ISSUE #2446: corpus: compiler error parity compares the code only — message text and span are never recorded
The corpus gate compares compiler **errors** by `code` alone. Neither the message text nor the span is ever recorded on either side, so an error with the right code at the wrong position, or with a message naming the wrong construct, is scored `error-parity`.

## Where

`scripts/compat-corpus/compile.mjs:85-96` builds everything the gate will ever know about an error:

```js
const errorInfo = (e) => {
    const message = String(e?.message ?? e);
    let code = e?.code ?? null;
    ...
    return { code, message: message.split('\n')[0] };
};
```

No `start`, no `end`, no frame. The first message line is written to `error.json` but is never read back: `scripts/compat-corpus/verify.mjs:331-348` compares only `e.code !== a.code`.

## Why this one matters more than it looks

The largest cluster in `compatibility/validator-known-failures.json` (207 entries) is ~141 entries of *"error spans not populated — `start`/`end` come back `None..None`"*, per `compatibility/validator-known-failures.md`.

So the class is real and known — and the gate that runs over ~14,025 real-world entries **cannot see it at all**. All burn-down pressure for it comes from a 332-fixture suite whose ratchet is additionally not two-sided (filed separately).

## Secondary: a code-less error on either side degrades to parity

`verify.mjs:332` guards with `e.code && a.code &&`. If either side's `code` is `null`, no mismatch is recorded and the verdict falls through to `error-parity` at `:335-337`. `compile.mjs:90-94` leaves `code` `null` when the error carries no `code` property *and* the message matches neither `svelte.dev/e/<code>` nor `code: "<code>"`.

Whether a `null` code is reachable across the NAPI boundary is **unmeasured** — I did not construct one. Recorded so the guard is not mistaken for a total.

## Fix

Record `start` (and ideally `end`) in `errorInfo`, then gate error positions on a ratchet of their own — the same split `verify.mjs` already uses for warnings (`warning-known-failures.*` vs `warning-position-known-failures.*`), and for the same reason: a wrong code is a semantic bug, a wrong span is one systemic cause, and folded together the larger span backlog would hide every semantic regression.

Message text is the same argument as #2403 (upstream rewords prose on minor bumps) and can be decided with it.

Found while building `compatibility/gate-coverage.md`.

baseballyama:
Closes #2446.

The corpus compared a compile error by `code` and nothing else. This adds the message text and the `start` position, each on its own shrink-only, two-sided ratchet.

## The measured blind spot

One full corpus run, 14,131 entries. 948 are rejected by both compilers, giving **2,843 `(id, target)` pairs** with two errors to compare.

| compared field | diverging pairs | distinct ids |
|---|---|---|
| `code` (pre-existing) | **0** | 0 |
| `message` (new) | 362 | **121** |
| `start` `(line, column)` (new) | 1,209 | **403** |

The `code` row is the argument. It is saturated — not one pair in 2,843 disagrees — so no amount of corpus growth could ever have moved this gate, while the two new columns were failing on 121 and 403 entries the whole time.

Nor did the fixture suites cover it: **33** of the 121 message divergences and **120** of the 403 position divergences are `svelte/packages/svelte/tests/compiler-errors/samples/…`, entries the 145/145-passing Compiler Errors suite compiles. That suite parses the sample's expected `message` and `position`, then asserts only `error_code_matches` (`compiler_errors.rs:272`); the parsed `message` field is `#[allow(dead_code)]`.

Sample of what was invisible:

- `dollar_binding_invalid` ×76 — rsvelte says ``Cannot use `$` as a variable name``, upstream says `The $ name is reserved, and cannot be used for variables and imports`. 73 of the 76 reach it from an *import*, so the user is told about a rule that does not appear to apply.
- `props_invalid_placement` ×3 — the two sentences say the same thing in different words, and neither is upstream's.
- `date-picker-svelte/src/lib/DateInput.svelte` — rsvelte reports `296:0` where upstream reports `262:11`. 34 lines off, column 0.

## Ratchet shape

Two families, split for the reason `warning-known-failures.md` sets out: wrong prose is a semantic bug fixed one string at a time, a wrong span is one systemic cause, and folded together the larger span backlog would hide every semantic regression.

| ratchet | entries | dominant cause |
|---|---|---|
| `error-message-known-failures.{client,client-dev}.json` | 121 | `dollar_binding_invalid` (76), `js_parse_error` (15, oxc's phrasing vs acorn's) |
| `error-message-known-failures.server.json` | 120 | — |
| `error-position-known-failures.<target>.json` | 403 | 349 rsvelte reports **no span at all**; 35 same line different column; 19 different line |

Unlike warning positions, the two are compared **independently** rather than chained behind agreeing codes: there is exactly one error per entry and target, so there is no pairing problem, and chaining would make a PR that fixes a message surface a position failure that was merely masked. Every per-entry justification is in `compatibility/error-known-failures.md`.

Regenerate with `node scripts/compat-corpus/verify.mjs --no-fmt --update-error-baseline` (documented in the doc). The flag touches only these six files; `scripts/dev/test-corpus-verify-baseline-flags.mjs` is extended to assert the three families stay disjoint.

## Why the binding had to change first

rsvelte's NAPI `compile` threw a `GenericFailure` whose `message` was a Rust `Debug` dump — `Parse(SvelteError { code: "attribute_duplicate", message: "…", span: (11, 12) })` — with no `code`, `start` or `end` property. Turning either comparison on against that would have produced a ratchet listing 100% of entries for one plumbing gap: no discriminating power at all.

So `compile` / `compileBoth` / `compileModule` now throw an object shaped like the official compiler's `CompileError`:

```
name: 'CompileError', code, message, filename,
start: {line, column, character}, end: {…}, position: [start, end]
```

Byte offsets are converted to JS (UTF-16) indices via the existing `Utf8ToUtf16` table, so a non-ASCII prefix does not shift the column — pinned by `diagnostic_span_is_a_utf16_position_not_a_byte_offset`. This is also what the vendored `@rsvelte/vite-plugin-svelte` already reads (`utils/error.js` builds `rollupError.loc` from `error.start.line/column`), so it was a shipped drop-in gap independently of the gate. Changeset included.

`compile.mjs` keeps its old `code: "<code>"` message-regex fallback, so an older binding still scores code parity instead of reporting `null` everywhere.

## Negative control

The gate was shown failing, not only passing. On the green tree: mutate one entry's rsvelte-side `error.json` (client `line` 14 → 99; server `message` → different prose) and add one already-passing id to a ratchet.

```
[verify] error parity:
  match                      13717
  error-message-mismatch     122
  error-position-mismatch    292

[verify] ❌ 1 error baseline entries already PASS — the ratchet is stale.
  node scripts/compat-corpus/verify.mjs --no-fmt --update-error-baseline

[verify] ❌ 2 NEW error failures (not in baseline); first 2:
  - layerchart/…/padding.svelte [error-message-mismatch] (server)
      expected: Cannot use `await` in deriveds and template expressions, …
      actual:   Cannot use await here
  - layerchart/…/padding.svelte [error-message-mismatch] (client)
      expected: 14:4
      actual:   99:4
```

exit 1. Both sides of the ratchet fire, and reverting all three mutations returns exit 0.

## Also in this PR

`compatibility/validator-message-known-failures.json` (added by #2418, the commit this branches from) was never declared in `known-failures-md-check.mjs`'s `RATCHETS`, so that check fails on `main` for any PR touching `compatibility/**`. Declared, with the count line its doc was missing. Unrelated to #2446 but it blocks this PR's CI.

## Out of scope, worth its own issue

- **The error `end` span is still uncompared.** rsvelte's `end` is frequently `start + 1` where `start` agrees — `attribute_duplicate` on `<div a="1" a="2">` reports `[11, 12]` against upstream's `[11, 16]` (pinned by a unit test). A wrong highlight *length* stays invisible; folding it in would add a third failure mode with its own cause to the position ratchet.
- **`frame` is neither captured nor compared**, and `compileWithCssHash` (async, not used by the corpus) still reports failures as a `Debug` message with no `code`/`start`.

Both are recorded as blind spots 4a in `compatibility/gate-coverage.md`, which this PR rewrites for gate 4 — including the answer to the reachability question #2446 left `[U]`: a `null` error code occurs on **1** of 2,843 pairs and on **0** one-sidedly, so the output verdict's `e.code && a.code &&` guard has never degraded a real divergence in this corpus.

## Verification

- full corpus `verify.mjs` (oxfmt-normalized): exit 0, output/warning/error ratchets all green
- `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test -p rsvelte_core --lib`: 1488 passed (3 new)
- `node scripts/compat-corpus/known-failures-md-check.mjs`, `test-known-failures-md-check.mjs`, `test-corpus-verify-baseline-flags.mjs`: pass

CI wiring confirmed by path, not by script name: `verify.mjs` runs at `.github/workflows/corpus-compat.yml:251` and `known-failures-md-check.mjs` at `:175` / `:540`, and the workflow's PR `paths` filter includes `crates/**`, `scripts/compat-corpus/**` and `compatibility/**` — all three of which this PR touches.


baseballyama:
Heads-up on a file this collides with, and a suggested resolution order.

#2548 rewrites `scripts/compat-corpus/known-failures-md-check.mjs` fairly heavily and **also** adds the `validator-message-known-failures.json` declaration you added here — we both hit the same break on `main` (measured, then filed as #2554 for the CI mechanism that let it through: the merge ref is frozen at event-creation time, and the push-to-`main` net is cancelled by the next merge, so both PRs were green for a base that never contained the other).

Whichever of the two lands second should **rebase, not hand-resolve**, because three things in that file changed shape in #2548:

1. `statedCount` → `statedCounts`, returning **every** line that states a ratchet's size rather than the first. A doc restating a count elsewhere is now checked too, so if either of your new docs states its entry count twice, both copies must agree.
2. A `PARTITIONS` list. A doc that splits its ratchet into counted clusters must state the addends on a `Partition of \`x.json\` by <label>: \`a + b + c\`` line, and the checker computes the total from the **JSON**, never from the doc. If `error-position-known-failures.md` states its 349 / 35 / 19 split — which from your report it should — that is exactly a partition (`349 + 35 + 19 = 403`), and declaring it means a later burn-down cannot update two of the three numbers and stay green.
3. `summing to N (\`a + b + c\`)` is now scanned in **every** ratchet doc, not just `warning-known-failures.md`. If you wrote that phrasing anywhere, it is now checked.

Also worth knowing for your `RATCHETS` additions: the per-target `<target>` placeholder form is what the checker expects for a three-file family (`error-message-known-failures.<target>.json`), and it asserts all three JSONs agree on the count — which will catch the day client and server diverge. Your report has message at 121/120/121, so that family needs **separate entries per target**, not the placeholder, until they converge.

Nothing here changes your measurements; it is only about the merge.

PR #2562, opened 2026-08-07
refactor(transform): state each private-class-assign decision once (#2485)

### LINKED ISSUE #2485: dual-run: `private_class_assign_ast` states every codegen decision twice, and the gate defends the behaviour but not the comments
Noticed while adding the deliberate-divergence comment in #2479 / #2483.

## The shape

`crates/rsvelte_core/src/compiler/phases/3_transform/client/private_class_assign_ast.rs`
implements the same rewrite twice, because the dual-run gate compares a spliced
path against an in-place path:

| decision | spliced path | in-place path |
|---|---|---|
| update expressions → `$.update` / `$.update_pre` | `visit_update_expression` (~L504) | `rewrite_update` (~L1036) |
| proxy flag (`, true`) | `visit_assignment_expression` (~L460) | `rewrite_assignment` (~L960) |
| compound read form (`q.v` vs `$.get(q)`) | `field_read_text` call (~L471) | `field_read` (~L1048) |
| nested-function depth reset | `visit_function` / `visit_arrow_function_expression` (Visit impl) | same pair on the `VisitMut` impl |

Every one of those is a semantic decision that has to be made identically in two
places. The deliberate-divergence comment added in #2479 therefore appears
**twice, verbatim**, once above each update-rewrite site.

## Why this is worth a ticket rather than a shrug

The risk is not the duplication, it is **drift**: someone edits one copy, misses
the other, and the stale copy still reads as authoritative — a comment that
confidently explains a rule the code next to it no longer follows is worse than
no comment.

This repo has already been bitten by exactly this shape:
`track_reactivity_loss` exists in two implementations and only one goes through
`ast_state_transform`, so a one-sided fix covered two of three paths while
looking complete.

The dual-run gate does defend the *behaviour* — if the two paths disagree on
output it fails — so a drifted decision is caught **when the paths are both
exercised on an input that distinguishes them**. It does not defend the
*comments*, and it cannot see a decision the two paths happen to agree on for
every input in the corpus. Note the deliberate divergence above is precisely such
a case: no corpus entry and no official fixture contains a non-`this` private
receiver at all, so both paths could drift together and nothing would go red.

## Not proposing a consolidation here

Merging the two implementations is the dual-run architecture's whole open
question and is well out of scope for a comment. What would help cheaply:

1. A note in the module header naming the paired sites, so the next person to
   edit one is told the other exists (this issue is the interim version of that).
2. If/when the in-place path becomes the only path, delete the spliced one and
   the problem goes away by itself — worth checking whether
   `private_class_assign_ast` is still dual-run-gated or whether in-place has
   already won and the spliced copy is dead weight.

## Sites, for grep

- `private_class_assign_ast.rs` — `PrivateClassAssignCollector` (`Visit`) vs
  `PrivateClassAssignRewriter` (`VisitMut`).
- Same pattern exists in the sibling `*_ast.rs` passes under
  `3_transform/client/` that carry a `transform_*_in_place` twin; this issue only
  audits `private_class_assign_ast`.

baseballyama:
Closes #2485.

`private_class_assign_ast` implements one rewrite twice — a text-splicing
`Visit` collector and an in-place `VisitMut` rewriter, picked between by
`ast_rewrite::dual_run::resolve`. Four codegen decisions were stated
separately in each: the compound-operator table, the `, true` proxy flag,
the `$.get(q) <op> rhs` value shape, and the `$.update` / `$.update_pre`
callee plus its `-1` step. Editing one copy and missing the other is the
`track_reactivity_loss` shape this repo has been bitten by before.

## What changed

Each of those decisions now lives in exactly one function that both paths
call:

| decision | shared as |
|---|---|
| is there anything to rewrite | `target_present` |
| `$state` vs other rune type | `classify` (already shared) |
| compound operator → binary / logical | `compound_of` + `Compound` |
| proxy flag | `needs_proxy` |
| update callee + step | `update_call` |

`compound_of` is now derived from oxc's own
`AssignmentOperator::to_logical_operator` / `to_binary_operator` instead of
a 16-arm table written out twice, and the text path spells operators with
`BinaryOperator::as_str()` / `LogicalOperator::as_str()` rather than a
second hand-maintained list of `"+"`, `"<<"`, `"??"`. There is no
behavioural change; the two paths already agreed.

## Both paths are tested, and each test fails when only its copy is stale

`shared_decision_tests::DECISIONS` is a 14-row table covering every unified
decision (one row per mutant: operator spelling, binary-vs-logical class,
proxy flag, update callee, update step). Two tests run it —
`spliced_path_lowers_every_decision` calls
`transform_private_class_assign_spliced`, `in_place_path_lowers_every_decision`
calls `transform_private_class_assign_in_place`. Direct calls are required:
which path answers is a process-wide `LazyLock` over `RSVELTE_AST_SPLICE`,
so the public entry point cannot exercise both in one process.

The tests were checked for discrimination against `origin/main`, where the
two copies still exist. With the same two decisions made stale in one copy
at a time (`--q` losing its `-1`, logical compounds losing the proxy flag):

| `origin/main` + | `spliced_path_…` | `in_place_path_…` |
|---|---|---|
| no mutation | ok | ok |
| text copy stale | **FAILED** | ok |
| in-place copy stale | ok | **FAILED** |

Neither test passes vacuously, and neither covers for the other.

## Output is byte-identical

Every `.svelte` fixture under `submodules/svelte/packages/svelte/tests`
(3839 that compile) was compiled before and after this change, in four
configurations — {prod, dev} × {default in-place, `RSVELTE_AST_SPLICE=1`}.
0 of 15,356 outputs differ. Positive control: appending one byte to a
single output file makes the comparison report exactly 1 difference.

Also run: `cargo test -p rsvelte_core --lib` (1487 passed), `--test
compiler_fixtures` (snapshot, 3 passed), `--test client_transforms_pin`
(3 passed), `--test runtime` (release), and the `dual_run_tally` harness
over all 4459 fixtures — `private_class_assign_ast:inplace` scores 22 runs,
0 mismatches, 0 unverified.

No ratchet moves.

## Scope of that sweep

The 15,356-comparison sweep covers `.svelte` and `.svelte.js` fixture
sources only. It did **not** reach `.svelte.ts` — `compile_module` rejects
TypeScript (callers strip it first) and this harness does not strip it — and
in the four real-world corpus repos `0` of the `705` `$state`-bearing
`.svelte` files contain `.#` at all, so this pass's real-world population is
almost entirely `.svelte.ts` modules.

That is a limit of **this measurement**, not of the gates: the corpus
output-equality gate does compile the `.svelte.ts` population on all three
targets (of `299` `.svelte.ts` sources in the collected corpus manifest,
`117` contain `.#`). So the `.svelte.ts` population is gated; it is only my
fixture-side sweep that skipped it.


baseballyama:
Excellent measurement, and the 2×2 discrimination matrix is exactly the right artefact — each test failing precisely when its own copy drifts is what makes "unified" a checked claim rather than a description.

One correction to the **Unmeasured** section, because it reads wider than it is.

> In the 4 real-world corpus repos, 0 of 705 `$state`-bearing `.svelte` files contain `.#` at all — this pass's real-world population is entirely in `.svelte.ts` modules, which no gate in this measurement reached.

The first half is a real and useful finding. The second half is true of **your sweep**, not of the gates. Measured just now on this checkout's collected corpus:

```
manifest: 13,679 .svelte + 299 .svelte.ts + 153 .svelte.js
of the .svelte.ts sources: 123 contain $state, 117 contain `.#`
```

So the corpus output-equality gate does compile that population on all three targets. What did not reach it is the fixture-side sweep you ran, because `compile_module` rejects TS and that harness does not strip it first.

Worth separating in the PR body, since "no gate reached it" and "my sweep did not reach it" size follow-up work very differently — the first reads as a coverage hole worth filing, the second as a scope note on one measurement. Everything else in the report stands.

Your out-of-scope `store_assign_ast` finding is now #2564. I added the framing that makes it urgent rather than tidy: since `prefer_in_place()` is a `LazyLock` over `RSVELTE_AST_SPLICE`, the two copies disagreeing means **compiler output depends on an environment variable** for an input that is in the corpus today — not merely that two code paths differ.

PR #2563, opened 2026-08-07
perf(transform): answer expression predicates from the typed AST, not JSON

baseballyama:
Stacked on #2537.

Five predicates in the client transform each answered a yes/no question about an
expression by materializing the whole expression as a `serde_json::Value` tree and
walking that. The tree is built for the question and immediately discarded.

`alloc_callers` attribution over the Huly corpus (2,123 files) put `JsNode::to_value`
at **15.83% of every allocation the compiler makes** — ~750 allocations per file.

## What changed

| predicate | file |
|---|---|
| "does this expression call anything" | `visitors/shared/events.rs` |
| "does it read reactive state" | `visitors/shared/utils.rs` |
| "is this a `$store` member expression" | `visitors/shared/component.rs` |
| expression-tag metadata flags | `visitors/shared/element.rs` |
| analyze-phase feature walk child iteration | `2_analyze/mod.rs` |

Each grew a typed front end and **kept its JSON walk** as the fallback for the shapes
the typed walk cannot reach (opaque `type_annotation` / `type_parameters` /
comment blobs). Each typed walk has a test asserting it agrees with the JSON walk
it replaces, case by case.

## Measurements

| | before | after |
|---|---|---|
| `as_json` calls | 49,776 | **15,056** (-70%) |
| JSON materializations | 27,488 | **12,089** (-56%) |

Output: **14,036 comparisons byte-identical** (open-webui/src 650, carbon/src 287,
smui 449, huly/plugins 2,123) x {client, server} x {prod, dev} — the four target
columns asserted mutually distinct first.

## What this does not claim

Wall-clock is not reported. The dev machine ran at load average 50-215 from parallel
worktrees throughout, and paired A/B under that load returned -18.3% and +4.4% on
neighbouring corpora in a single run. The allocation counters above are per-allocation
and load-independent; they are the evidence here, and they say the work is gone, not
how much time it was taking.

Remaining materialization is 70.3% concentrated at one site — `expression_has_reactive_state`'s
MemberExpression fallback in `utils.rs`.


baseballyama:
Superseded by #2576, which was the tip of this stack and contained every commit and changeset from this PR. Verified before merging: all four substantive commits (`002b84a3`, `e02426cc`, `bafb5835`, `1c77e55c`) and all five changesets were ancestors of the merged branch.

Closing rather than merging because squash-merging a stack rewrites the base commits and would have forced a rebase-and-rerun of the whole chain for no additional content.

Issue #2420, opened 2026-08-06
probe: does in_component_attribute have the same identity-vs-boolean defect?


baseballyama:
## The question

`in_component_attribute` is a **boolean** that grants the same dev `$.assign` exemption the event-attribute path grants, in the same condition:

```rust
if (is_event_attribute_arrow || context.state.in_component_attribute) && body_is_assignment
```

The event-attribute half of that condition was just replaced with node identity, because upstream's predicate is an identity test and a boolean cannot express it — a boolean says "somewhere inside", never "this exact node". That produced three divergences at once, in both directions.

**The same argument applies to `in_component_attribute` on its face.** Its upstream counterpart (`AssignmentExpression.js:204-215`) is also path-shaped and also node-specific:

```js
path.at(-1) === 'BindDirective' ||
path.at(-1) === 'Component' ||
path.at(-1) === 'SvelteComponent' ||
(path.at(-1) === 'ArrowFunctionExpression' &&
  (path.at(-2) === 'BindDirective' ||
   (path.at(-2) === 'Component' && path.at(-3) === 'Fragment') ||
   (path.at(-2) === 'SequenceExpression' &&
     (path.at(-3) === 'Component' || path.at(-3) === 'SvelteComponent' || path.at(-3) === 'BindDirective'))))
```

Note it is *more* structured than the event-attribute case — three levels of path, and a `SequenceExpression` branch for the `bind:prop={getter, (v) => (...)}` form.

**This is a probe request, not a bug report.** I have **no reproduced divergence** for it and did not change it, deliberately: the fix it sits next to was justified by measurement, and widening it on a structural hunch would have been the same mistake as the "namespaced SVG child components" cause that was inferred rather than derived.

## How to probe it

Compile with both compilers, `dev: true`, `generate: 'client'`, and diff the `$.assign*` calls. **The right-hand side must be non-primitive** — upstream's `should_transform` requires `!scope.evaluate(right).is_primitive`, so a literal RHS makes every case silent on both sides and the comparison proves nothing. Include a positive control that *does* emit on both, or a null result is unreadable.

Shapes worth covering, chosen so each can distinguish a boolean from an identity test:

- `<Comp prop={() => (o.x = obj)} />` — the arrow is the attribute's expression
- `<Comp prop={f(() => (o.x = obj))} />` — arrow nested in a call, `path.at(-2)` is not `Component`
[...]

PR #2569, opened 2026-08-07
fix(compiler): scope the `bind:` dev `$.assign` exemption to the directive's own arrows

### LINKED ISSUE #2484: compiler: dev $.assign exemption for the bind: SequenceExpression arm is wrong in both directions (not the in_component_attribute boolean)
Found while probing #2420. Filed standalone because it is **not** the `in_component_attribute` boolean that #2420 is about, and a fix for that boolean would not close this — but would very likely make some of these shapes go green by coincidence, leaving the real cause alive behind a closed issue.

Measured against official `svelte@5.56.8` (the version `submodules/svelte` pins), `dev: true`, `generate: 'client'`, diffing `$.assign*` calls in the generated code. GitHub Actions is in a major outage; none of this went through CI.

## Upstream's predicate

`3-transform/client/visitors/AssignmentExpression.js:204-217` — the `SequenceExpression` arm is the last three lines:

```js
if (
    path.at(-1) === 'BindDirective' ||
    path.at(-1) === 'Component' ||
    path.at(-1) === 'SvelteComponent' ||
    (path.at(-1) === 'ArrowFunctionExpression' &&
        (path.at(-2) === 'BindDirective' ||
            (path.at(-2) === 'Component' && path.at(-3) === 'Fragment') ||
            (path.at(-2) === 'SequenceExpression' &&
                (path.at(-3) === 'Component' ||
                 path.at(-3) === 'SvelteComponent' ||
                 path.at(-3) === 'BindDirective'))))
) {
    should_transform = false;
}
```

This is the `bind:prop={getter, (v) => (...)}` form. Note the arm requires the arrow to be a **direct element of the SequenceExpression**, and separately requires the sequence's parent to be `Component` / `SvelteComponent` / `BindDirective`.

## Result: wrong in both directions

`scroll` / `other` / `obj` are all `$state({...})` so the RHS is non-primitive — a primitive RHS silences `should_transform` on both sides and proves nothing.

| shape | official | rsvelte | |
|---|---|---|---|
| S0 `<form onsubmit={wrap(() => (scroll.x = obj))}>` | 1 | 1 | **positive control** |
| S1 `<Comp bind:value={() => scroll.x, (v) => (scroll.y = obj)} />` | 0 | 0 | same |
| **S2 `<Comp bind:value={() => scroll.x, wrap((v) => (scroll.y = obj))} />`** | **1** | **0** | **under-wraps** |
| **S3 `<input bind:value={() => scroll.x, (v) => (scroll.y = obj)} />`** | **0** | **1** | **over-wraps** |
| S4 `<input bind:value={() => scroll.x, wrap((v) => (scroll.y = obj))} />` | 1 | 1 | same |
| **S5 `<Comp bind:value={() => scroll.x, (v) => (scroll.y = wrap(() => (other.d = obj)))} />`** | **1** | **0** | **under-wraps** |
| S6 `<Comp bind:value={() => scroll.x, (v) => { scroll.y = obj; }} />` | 0 | 0 | same |

**S3 is the cleanest one.** It is an `<input>` — a regular element — so `in_component_attribute` is false there regardless. rsvelte wraps an assignment upstream exempts through `path.at(-2) === 'SequenceExpression' && path.at(-3) === 'BindDirective'`. rsvelte appears to have no equivalent of that arm for element bind directives at all.

Because S2/S5 under-wrap while S3 over-wraps, "the exemption is missing" is not a complete description: one direction needs an exemption **added**, the other needs one **removed**.

## Attribution: not `in_component_attribute`

Re-running every shape with `context.state.in_component_attribute` forced to `false` (the mutation that fixes two of #2420's four divergences):

| shape | official | flag ON | flag OFF | |
|---|---|---|---|---|
| S0 | 1 | 1 | 1 | ok under both |
| S1 | 0 | 0 | 0 | ok under both |
| **S2** | 1 | **0** | **0** | **diverges under both** |
| **S3** | 0 | **1** | **1** | **diverges under both** |
| S4 | 1 | 1 | 1 | ok under both |
| **S5** | 1 | **0** | **0** | **diverges under both** |
| S6 | 0 | 0 | 0 | ok under both |

All three divergences reproduce identically with the flag on and off. Whatever handles the sequence arm is a different code path.

One loose end for whoever takes this: S1 is correct with the flag **off** as well, so the exemption that makes the component case right is also not the flag — yet the identical shape on an element (S3) is wrong. So component and element bind directives are reaching different code, and that asymmetry is probably where the root cause is.

## Reachability

All produced by the public `compile()` with `dev: true, generate: 'client'`; I ran it rather than inferring it. The consequence is dev-only: a missing `$.assign` wrap means upstream's ownership/staleness validation does not run for that assignment (S2, S5), and an extra one means rsvelte validates an assignment upstream deliberately exempts (S3), which can surface a dev-mode warning upstream would not emit. Production output is unaffected.

## Why not fold this into #2420

#2420's fix replaces the `in_component_attribute` boolean with a node-identity test. That rewrite will touch the same predicate and could easily flip S2 or S5 green as a side effect while leaving S3 — an element path the boolean never covered — untouched. If the two are landed together, the sequence arm gets no root-cause pass and the issue that would have pointed at it is closed.

Related: #2420, #2421, #2064.

baseballyama:
Fixes #2484.

## The predicate upstream asks

`submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/AssignmentExpression.js:204-218`:

```js
// special case — ignore `bind:prop={getter, (v) => (...)}` / `bind:value={x.y}`
if (
	path.at(-1) === 'BindDirective' ||
	path.at(-1) === 'Component' ||
	path.at(-1) === 'SvelteComponent' ||
	(path.at(-1) === 'ArrowFunctionExpression' &&
		(path.at(-2) === 'BindDirective' ||
			(path.at(-2) === 'Component' && path.at(-3) === 'Fragment') ||
			(path.at(-2) === 'SequenceExpression' &&
				(path.at(-3) === 'Component' ||
					path.at(-3) === 'SvelteComponent' ||
					path.at(-3) === 'BindDirective'))))
) {
	should_transform = false;
}
```

The assignment's **parent** must be an arrow that is the directive's own expression or a
**direct element** of its `SequenceExpression`. rsvelte asked a different question: a sticky
`state.in_bind_directive` boolean, set for the whole conversion of a *component* bind
expression and never set at all for an *element* bind directive. Wrong question, not a
mis-tuned threshold — which is why it was wrong in both directions at once, and why this PR
replaces the predicate instead of adding a second condition to it.

`state.bind_exempt_arrows` now carries the start offsets of exactly those arrows for the
duration of the directive's conversion, and the arrow converter feeds them into the same
`event_handler_arrow_body_level` mechanism the event-attribute exemption already uses — which
fires only when the arrow's body *is* the assignment, and which every nested arrow clears.
Both arrow converters (the typed `JsNode` path and the `serde_json` fallback) were updated,
and the three sites that convert a `bind:` expression — element directives
(`bind_directive_inner`), component directives (`process_bind_directive`) and `bind:this`
(`unified_build_bind_this`) — now go through one entry point.

## Both directions, measured

`$.assign(` occurrences in `compile(src, { generate: 'client', dev: true })`, oracle = official
`svelte@5.56.8` (the `submodules/svelte` pin). "before" is `origin/main` at `ddc8be40`, i.e.
**with #2421 already merged** — the issue predates it, so every row was re-derived by running
both compilers rather than copied from the issue.

### Direction A — rsvelte wrapped where upstream exempts (element bind directives)

| shape | official | before | after |
|---|---|---|---|
| `<input bind:value={() => s.x, (v) => (s.y = o)} />` | 0 | **1** | 0 |
| `<input bind:value={() => s.x, (v) => (s.y ??= o)} />` | 0 | **1** | 0 |
| `<svelte:window bind:scrollY={() => s.x, (v) => (s.y = o)} />` | 0 | **1** | 0 |

rsvelte had no equivalent of `path.at(-3) === 'BindDirective'` for elements at all.

### Direction B — rsvelte exempted where upstream wraps (component bind directives)

| shape | official | before | after |
|---|---|---|---|
| `<Comp bind:value={() => s.x, wrap((v) => (s.y = o))} />` | 1 | **0** | 1 |
| `<Comp bind:value={() => s.x, (v) => (s.y = wrap(() => (d.e = o)))} />` | 1 | **0** | 1 |

The sticky flag exempted the directive's whole subtree, so an arrow handed over by a call and
an arrow nested inside the setter body were both silenced.

### Controls that must not move (all unchanged, before = after = official)

| shape | official |
|---|---|
| `<form onsubmit={wrap(() => (s.x = o))}>` (no `bind:` involved) | 1 |
| `<Comp bind:value={() => s.x, (v) => (s.y = o)} />` | 0 |
| `<input bind:value={() => s.x, wrap((v) => (s.y = o))} />` | 1 |
| `<Comp bind:value={() => s.x, (v) => { s.y = o; }} />` / same on `<input>` | 0 |
| `<svelte:component this={C} bind:value={() => s.x, (v) => (s.y = o)} />` | 0 |
| `<input bind:this={...}>` / `<Comp bind:this={...}>` sequence form | 0 |
| `<Comp bind:value={s.x} />` / `<input bind:value={s.x} />` | 0 |
| `<input bind:value={() => s.x, (v) => (s.y = o, d.e = o)} />` | **2** |

The last one is the control that pins the *mechanism* rather than the flag: the arrow's body is
a `SequenceExpression`, so the assignments' parent is not the arrow and upstream wraps **both**.
An implementation that exempted "the setter arrow's subtree" would report 0 here.

## Pre-fix failure output

`crates/rsvelte_core/tests/dev_assign_bind_sequence.rs` on `origin/main` (source unchanged,
test file added): **6 passed, 5 failed**, each for the predicted count and direction —

```
test an_element_bind_sequence_setter_arrow_is_exempt ... FAILED           left: 1  right: 0
test a_coercive_operator_in_a_setter_arrow_is_exempt ... FAILED           left: 1  right: 0
test svelte_component_and_svelte_window_bindings_follow_the_same_rule ... FAILED  left: 1  right: 0
test an_arrow_reached_through_a_call_is_not_a_direct_sequence_element ... FAILED  left: 0  right: 1
test an_arrow_nested_inside_the_setter_body_is_wrapped ... FAILED         left: 0  right: 1
```

All 11 pass after.

## Gates

- `cargo test -p rsvelte_core` on `runtime` (legacy + runes + hydration + browser), `ssr`, `css`,
  `validator`, `compiler_fixtures`, `compiler_errors`, `print` and all seven `dev_assign_*` /
  `dev_bind_setter_ownership` suites: clean.
- `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- `pnpm run corpus:matrix`: `match 1833 / js-mismatch 162 / error-mismatch 0` — identical to the
  162-entry baseline, so no re-baseline. Expected: `matrix/axes.mjs` generates no `bind:`
  directive shapes, so that gate cannot observe this class at all.

### DIFF
--- a/.changeset/bind-sequence-assign-exemption.md
+++ b/.changeset/bind-sequence-assign-exemption.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+fix(compiler): scope the `bind:` dev `$.assign` exemption to the directive's own arrows
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/types.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/types.rs
@@ -1882,9 +1882,12 @@ pub struct ComponentClientTransformState<'a> {
     /// This is used to skip rest_prop → $$props transformation for direct property assignments.
     pub in_direct_assignment_lhs: bool,
 
-    /// Flag indicating if we're inside a bind directive expression.
-    /// Used to skip coercive assignment transforms ($.assign_nullish, etc.) for bind setters.
-    pub in_bind_directive: bool,
+    /// Start offsets of the arrows a `bind:` directive exempts from the dev
+    /// `$.assign` wrap. Upstream asks a path question — the arrow must *be* the
+    /// directive's expression or a direct element of its `SequenceExpression`
+    /// (`AssignmentExpression.js`) — so an arrow reached through a call, or one
+    /// nested inside a setter body, keeps the wrap.
+    pub bind_exempt_arrows: Vec<u32>,
 
     /// Set while converting a component attribute / `on:` directive value: upstream
     /// exempts every such arrow body from the `$.assign` wrap, not just event ones.
@@ -2206,7 +2209,7 @@ impl<'a> ComponentClientTransformState<'a> {
             module_level_snippets: Vec::new(),
             snippet_names: ImHashSet::new(),
             in_direct_assignment_lhs: false,
-            in_bind_directive: false,
+            bind_exempt_arrows: Vec::new(),
             in_component_attribute: false,
             parent_is_regular_element: false,
             state_declarator_name: None,
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/bind_directive.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/bind_directive.rs
@@ -18,7 +18,9 @@ use rustc_hash::FxHashSet;
 use crate::ast::js::Expression;
 use crate::ast::template::{Attribute, AttributeValue, AttributeValuePart, BindDirective};
 use crate::compiler::phases::phase3_transform::client::types::*;
-use crate::compiler::phases::phase3_transform::client::visitors::expression_converter::convert_expression;
+use crate::compiler::phases::phase3_transform::client::visitors::expression_converter::{
+    convert_bind_expression, convert_expression,
+};
 use crate::compiler::phases::phase3_transform::js_ast::JsArena;
 use crate::compiler::phases::phase3_transform::js_ast::builders as b;
 use crate::compiler::phases::phase3_transform::js_ast::nodes::*;
@@ -59,10 +61,7 @@ pub fn unified_build_bind_this(
         apply_transforms_to_expression, apply_transforms_to_expression_with_shadowed,
     };
 
-    let saved_in_bind = context.state.in_bind_directive;
-    context.state.in_bind_directive = true;
-    let raw_expr = convert_expression(expression, context);
-    context.state.in_bind_directive = saved_in_bind;
+    let raw_expr = convert_bind_expression(expression, context);
 
     let (getter_expr, setter_expr) = if let JsExpr::Sequence(ref seq) = raw_expr {
         if seq.expressions.len() == 2 {
@@ -389,7 +388,7 @@ fn bind_directive_inner(
 
     // Visit the expression to transform it using the full expression converter
     // (supports ArrowFunctionExpression, MemberExpression, etc.)
-    let expression = convert_expression(&node.expression, context);
+    let expression = convert_bind_expression(&node.expression, context);
 
     // In dev mode with runes, validate binding to non-reactive properties.
     // Reference: BindDirective.js lines 26-41
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/expression_converter.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/expression_converter.rs
@@ -151,6 +151,33 @@ pub fn convert_expression(expr: &Expression, context: &mut ComponentContext) -> 
     convert_js_node(&node, context)
 }
 
+/// Convert a `bind:` directive's expression, exempting the arrows upstream
+/// exempts from the dev `$.assign` wrap.
+///
+/// Upstream's test is positional (`AssignmentExpression.js`): the assignment's
+/// parent arrow must be the directive's own expression or a direct element of
+/// its `SequenceExpression`, so an arrow handed over by a call — or one nested
+/// inside a setter body — is not covered.
+pub fn convert_bind_expression(expr: &Expression, context: &mut ComponentContext) -> JsExpr {
+    let node = expr.as_node();
+    let exempt: Vec<u32> = match node.node_type() {
+        Some("ArrowFunctionExpression") => node.start().into_iter().collect(),
+        Some("SequenceExpression") => context
+            .state
+            .parse_arena
+            .get_js_children(node.expressions())
+            .iter()
+            .filter(|child| child.node_type() == Some("ArrowFunctionExpression"))
+            .filter_map(|child| child.start())
+            .collect(),
+        _ => Vec::new(),
+    };
+    let saved = std::mem::replace(&mut context.state.bind_exempt_arrows, exempt);
+    let converted = convert_js_node(&node, context);
+    context.state.bind_exempt_arrows = saved;
+    converted
+}
+
 /// Convert a JsNode directly to JsExpr via pattern matching, bypassing serde_json::Value
 /// for simple expression types. Complex types fall back to convert_json_value.
 fn convert_js_node(node: &JsNode, context: &mut ComponentContext) -> JsExpr {
@@ -841,7 +868,9 @@ fn convert_js_node(node: &JsNode, context: &mut ComponentContext) -> JsExpr {
                 .analysis
                 .event_attribute_arrow_starts
                 .contains(arrow_start);
+            let is_bind_exempt_arrow = context.state.bind_exempt_arrows.contains(arrow_start);
             context.state.event_handler_arrow_body_level = if (is_event_attribute_arrow
+                || is_bind_exempt_arrow
                 || context.state.in_component_attribute)
                 && body_is_assignment
             {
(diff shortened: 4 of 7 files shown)


baseballyama:
**These two PRs fix overlapping halves of the same upstream `if`, by different mechanisms. Only one should land as written.** My fault for briefing #2484 and #2420 as separate work — the issues read as neighbours and are in fact one predicate.

Shared source files (5 of #2569's 6):

```
3_transform/client/types.rs
3_transform/client/visitors/bind_directive.rs
3_transform/client/visitors/expression_converter.rs
3_transform/client/visitors/fragment.rs
3_transform/client/visitors/shared/component.rs
```

#2577 additionally touches Phase 2 (`2_analyze/visitors/{component,svelte_component,svelte_element,bind_directive,shared/attribute}.rs`), because it records the qualifying nodes during analysis; #2569 stays in Phase 3 and keeps a set of qualifying arrow start offsets. Those are not two patches to combine — they are two answers to "where does this predicate live".

What each measured, so the comparison is on the record:

| | #2569 (#2484) | #2577 (#2420) |
|---|---|---|
| shapes | 11 | 49 |
| divergences found on `main` | 5 | 12 |
| after | 0/11 | 0/49 |
| predicate derived from | reading `AssignmentExpression.js:204-218` | **patching the official compiler to dump `context.path` per shape** |

#2577's derivation method is the stronger one and it produced three corrections to the issue's reading of the source — including that `path.at(-3) === 'Fragment'` is **not** vacuous (`RegularElement` is the one container reached without an interposed `Fragment`), which a source read had called redundant.

But breadth does not settle it, because #2569 carries a case that discriminates between two implementations agreeing everywhere else:

```svelte
<input bind:value={() => s.x, (v) => (s.y = o, d.e = o)} />
```

Upstream wraps **both** assignments (count 2), because the arrow's body is a `SequenceExpression` so neither assignment's parent is the arrow. An implementation that exempts "the setter's whole subtree" reports 0 there and looks correct on every other row.

**Resolution protocol**, so this is decided by measurement rather than by merge order: run the union of the two shape sets against both branches. If #2577 also gives 2 on the shape above and 0 divergences on #2569's other 10, it subsumes #2569 and #2569 should close in its favour, keeping its test file. If it does not, #2577 has the wider coverage and the wrong predicate, and the fix is #2569's rule at #2577's site.

I have asked #2577's author to run that comparison and report both numbers.

baseballyama:
Closing in favour of #2577, decided by the measurement I asked for rather than by merge order — and the discriminating case you built is what made it decidable, so it is worth being precise about what it settled.

**Your case does not separate them.** `<input bind:value={() => s.x, (v) => (s.y = o, d.e = o)} />` gives `["$.assign","$.assign"]` on official, on #2577 **and** on this branch. #2577 reaches it by a different route — its Phase-2 recorder marks the *arrow*, and Phase 3 clears the wrap only when that arrow's body **is** an `AssignmentExpression`; a `SequenceExpression` body is not, so both assignments wrap. Different mechanism, identical behaviour on the case designed to catch the wrong one.

**The union separates them, one way only.** 72 shapes (your 18, #2577's 49, plus 5 coercive-operator shapes), both branches against the same oracle:

| branch | divergences |
|---|---|
| #2577 | 2 — both the `bind_invalid_expression` gap, neither a `$.assign` divergence |
| this branch | 13 |

All 11 rows where the two disagree run the same direction: official matches #2577 and not this branch. **There is no row in the other direction.** The cause is structural rather than a tuning difference — this branch replaces `in_bind_directive` and leaves `in_component_attribute` in place, so the component half keeps the original identity-vs-boolean defect (`<Comp prop={f(() => (o.x = obj))} />`, `<Comp prop={(o.x = obj)} />`, `<svelte:component>`, `<svelte:self>` ×2, `{@attach}`, the `<div>`-wrapped component, …).

Your test file is ported to #2577 unchanged (`741388b9`) and all 11 pass there, so nothing this PR pinned is lost.

One thing from your side that the other branch's author checked because you wrote it: `assigns()` counts the literal `"$.assign("`, and `get_coercive_assign_callee` maps `??=` to `$.assign_nullish`, which that counter cannot see. Five shapes were added putting `??=` / `||=` / `&&=` in wrapped positions to test whether the counter was blind there. It is not — both compilers emit plain `$.assign` on this path — so the negative result stands, but it was worth the look, and a counter matching one literal callee name is worth flagging wherever it appears next.

baseballyama:
Superseded by #2577 (measurement above); test file ported.

PR #2571, opened 2026-08-07
bench(corpus): add the two shipped-code shapes the fixtures were missing

baseballyama:
## The problem

CodSpeed is this repo's only load-independent performance signal — instruction
counts under Valgrind, deterministic, head-vs-base per PR. Its workload is the
pinned corpus in `benches/corpus/`, 9 files chosen to *cover features*: one
distinct compiler slice each.

Feature coverage and population representativeness are different properties, and
the benchmarks aggregate over the corpus, so it is the population that the numbers
report. Measured over 3,509 `.svelte` files from four shipped projects
(huly/plugins 2,123, open-webui 650, carbon-components-svelte 287, SMUI 449):

| axis | shipped | fixtures 01-09 |
|------|---------|----------------|
| uses `$state` / `$derived` | **8.0%** | **88.9%** |
| uses `$:` | 0-60% by project | 11% |
| uses `export let` | 0-90% by project | 11% |
| `lang="ts"` | 0-99% by project | 11% |
| p90 source size | 4.9-15.7 KB | 2.6 KB |

Per-project runes share: huly 0.4%, open-webui 0.3%, carbon 0.0%, SMUI 59.9%.

The corpus is close to the inverse of shipped code on every axis that selects a
compiler hot path. A change that only pays off on legacy, TypeScript, or
store-heavy components reads as 0% — and that is most of what ships.

## What this adds

| file | shape |
|---|---|
| `10-legacy-typescript-props.svelte` | `export let` with type annotations, `$:` chains, `$store` reads, typed `createEventDispatcher` — the modal shipped component, and the one shape with no fixture at all (`05-legacy-reactive` is the closest and is untyped) |
| `11-store-heavy-legacy.svelte` | `$store` autosubscription across script *and* markup, `$store` assignment, `getContext` stores, `$:` over store reads |

Both compile clean — zero warnings — under the official compiler on all four
targets (client/server x prod/dev). Adding files does not disturb existing
CodSpeed history: benchmark identity is the filename, and these are new names.

I stopped at two deliberately. Each new file costs bench time on every PR, and
these two are the axes with zero current representation; further additions should
be argued the same way rather than accumulated.

### DIFF
--- a/benches/corpus/10-legacy-typescript-props.svelte
+++ b/benches/corpus/10-legacy-typescript-props.svelte
@@ -1,0 +1,127 @@
+<script lang="ts">
+	import { onMount, createEventDispatcher } from 'svelte';
+	import { writable, derived, type Readable } from 'svelte/store';
+
+	type Severity = 'info' | 'warning' | 'error';
+
+	interface Notice {
+		id: string;
+		severity: Severity;
+		message: string;
+		at: number;
+	}
+
+	export let title: string = 'Notifications';
+	export let notices: Notice[] = [];
+	export let maxVisible: number = 5;
+	export let dismissable: boolean = true;
+	export let emptyLabel: string = 'Nothing to show';
+	export let severityFilter: Severity | null = null;
+
+	const dispatch = createEventDispatcher<{ dismiss: { id: string }; seen: void }>();
+
+	const now = writable(Date.now());
+	const unreadCount = writable(0);
+	const label: Readable<string> = derived(unreadCount, (n) => (n === 0 ? 'read' : `${n} unread`));
+
+	let expanded = false;
+	let hovered: string | null = null;
+	let container: HTMLDivElement;
+
+	$: filtered = severityFilter ? notices.filter((n) => n.severity === severityFilter) : notices;
+	$: visible = expanded ? filtered : filtered.slice(0, maxVisible);
+	$: hiddenCount = filtered.length - visible.length;
+	$: errorCount = filtered.filter((n) => n.severity === 'error').length;
+	$: heading = errorCount > 0 ? `${title} (${errorCount})` : title;
+	$: ages = visible.map((n) => Math.max(0, Math.round(($now - n.at) / 1000)));
+
+	$: if (errorCount > 0 && !expanded) {
+		expanded = true;
+	}
+
+	function dismiss(id: string): void {
+		notices = notices.filter((n) => n.id !== id);
+		unreadCount.update((n) => Math.max(0, n - 1));
+		dispatch('dismiss', { id });
+	}
+
+	function severityClass(severity: Severity): string {
+		return `notice notice--${severity}`;
+	}
+
+	onMount(() => {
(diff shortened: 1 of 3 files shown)


baseballyama:
### On the CodSpeed failure

`CodSpeed Performance Analysis` reports **-38.89%** on `parallel_parse::corpus` (8.7 ms -> 14.3 ms). **This is a workload change, not a slowdown**, and it needs a one-time acknowledgement rather than a fix.

`bench_parallel_parse` (`crates/rsvelte_devtools/benches/parser.rs:41`) collapses the entire corpus into a single benchmark ID:

```rust
let files = corpus::load();
group.bench_function("corpus", |b| {
    b.iter(|| parse_parallel(black_box(sources.clone()), ParseOptions::default()));
});
```

so one iteration parses every fixture. This PR adds ~5.6 KB to a ~19 KB corpus, and the reported change tracks that. The 22 `new benchmark` rows in the same report are the per-file IDs for the two added fixtures; the 126 untouched rows are every pre-existing per-file ID, none of which moved.

The corpus README claimed "Adding a new file is free". CI has just shown that is false for the two whole-corpus benchmarks, so I have corrected it in this PR rather than leaving the next person to rediscover it.

I cannot acknowledge the alert myself — it needs the CodSpeed UI. Please acknowledge it (or tell me to drop the two fixtures) before merging.

baseballyama:
CodSpeed reports `❌ 1 regressed benchmark`, and it is **not a regression** — it is this PR's own input growing.

```
parallel_parse::corpus   8.7 ms → 14.3 ms   (-38.89%)
```

`parallel_parse::corpus` parses **every file in `benches/corpus/`** in one benchmark. This PR adds two of them, so the benchmark that CodSpeed matched by name is not measuring the same workload on both sides of the comparison. The other 126 benchmarks are untouched and the 22 new ones are the two fixtures × 11 benchmark kinds.

The magnitude checks out as input growth rather than a slowdown:

| | before | after |
|---|---:|---:|
| corpus files | 9 | 11 |
| corpus bytes | 17,370 | 24,385 (**+40%**) |
| `parallel_parse::corpus` | 8.7 ms | 14.3 ms (**+64%**) |

+64% time for +40% bytes is what these two shapes should cost: `11-store-heavy-legacy` (3,873 B) and `10-legacy-typescript-props` (3,142 B) are the two most expensive per byte in the set — `compile_both` puts them at 20 ms and 15 ms, above every existing fixture — because legacy `$:` dependency extraction is superlinear in the number of reactive statements. Adding the slowest shapes raises the aggregate more than their byte share.

CodSpeed also flags the comparison itself: *"Different runtime environments detected — some benchmarks with significant performance changes were compared across different runtime environments."*

So there is nothing to fix in the compiler here. The signal is only that a whole-corpus aggregate benchmark cannot be compared across a commit that changes the corpus. Worth noting for the next fixture addition: any per-fixture benchmark stays comparable (they are keyed by fixture name), but `parallel_parse::corpus` will report a "regression" every time the corpus grows.

PR #2572, opened 2026-08-07
fix(compiler): re-home a `$:` statement's comments instead of dropping them (#2368)

### LINKED ISSUE #2368: codegen: a comment inside a `$:` statement is dropped instead of re-homed onto a surviving successor
Split out of #2355 (fixed by PR #2365), which covered only the case where **nothing survives
after** the reactive statement. This is the complementary case, and it is a member of the
#2336 comment-cursor epic rather than a standalone fix — see "Why this belongs to #2336".

## Repro

```svelte
<script>
	let bar;
	$: {
		/* inner */
		bar = []
	}
	let z = 1;
	console.log(z);
</script>
```

`generate: 'client'`. Official **re-homes** `/* inner */` onto the surviving successor:

```js
	let bar = $.mutable_source();

	/* inner */
	let z = 1;

	console.log(z);

		$.set(bar, []);
	});
```

rsvelte drops it outright:

```js
	let bar = $.mutable_source();
	let z = 1;

	console.log(z);
```

Note the comment's absence also costs a **blank line**: official has a margin before
`/* inner */`, rsvelte runs `let bar` and `let z` together. So this is two bytes-worth of
divergence, not one.

## Second shape: `$: if (…)` emits it twice

```svelte
<script>
	let a = 1;
	let b;
	$: if (a) {
		/* inner */
		b = 1;
	}
	let z = 1;
	console.log(z, b);
</script>
```

Official prints `/* inner */` **twice** — re-homed onto `let z = 1`, *and* again inside the
effect:

```js
	let a = 1;
	let b = $.mutable_source();

	/* inner */
	let z = 1;

	console.log(z, $.get(b));

		if (a) {
			/* inner */
			$.set(b, 1);
		}
	});
```

rsvelte emits it zero times. The duplication is not a quirk of the `if` — it is
`reset_comment_index` searching from index 0 and therefore **rewinding**: `b.block(body)`
has no `loc`, so the cursor parks past the end, but the `if (a) { … }` inside it is an
untouched source node that still carries its `loc`, so the next `body()` call finds the
comment again. That is exactly the #2262 sign of the same mechanism.

## Why rsvelte drops it

Not the printer. `strip_reactive_statement_comments`
(`crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs:207`) removes the
comment from the **source text** before codegen. `reactive_statement_spans` returns the span
`(label, end)` when a successor survives and `(leading_run, end)` when none does — either way
comments *inside* the statement are stripped. A text pass cannot move a comment to a later
statement, so re-homing is not expressible there.

The existing unit test `drops_comments_inside_a_reactive_if_block` (`:1336`) asserts the
current stripping behaviour and is therefore **wrong against upstream** — it pins a
divergence. It should be re-baselined as part of this work, not before.

## Why this belongs to #2336

Upstream never strips anything. It replaces the reactive statement with `b.empty`
(`3-transform/client/visitors/LabeledStatement.js:63`) and lets the comment cursor decide:

- a successor with a `loc` follows → the cursor flushes the comment there (re-homed);
- nothing follows → the `loc`-less `b.thunk(b.block(body))` parks the cursor at
  `comments.length` and it is lost.

So **both halves of #2355 fall out of one rule** — the very rule #2336 item 1 names
(`crates/rsvelte_esrap/src/printer.rs:655` returns early where official sets
`comment_index = comments.length`). Once #2336 lands, `strip_reactive_statement_comments`
should be deleted, not extended: the no-successor case that PR #2365 fixes by stripping is
the same case the cursor would handle for free, and the has-successor case is only fixable
that way.

**Do not attempt this in isolation.** #2336 records that porting the exhaust rule alone
silently drops comments across every component, because rsvelte's synthesized wrapper blocks
lack the span official's carry — its four changes must land together. This issue is a fifth
consumer of that work, not a fifth change.

## Scope

Independent of #2355 as filed (that one is fixed and closed by #2365). The shape predates
#2365 — it behaved identically when the predecessor ended in `;`, which is the path #2365 did
not touch. Sibling of #2262 and #2307. Parent: #2336.

Measured against the official compiler at the pinned `submodules/svelte`, `generate: 'client'`,
and against rsvelte at `origin/main` + #2365.

baseballyama:
## The bug

A comment inside a legacy `$:` statement was deleted from the client script text before
codegen, so it never reached the output. Upstream deletes nothing.

Measured on `origin/main` @ `c193616e` (i.e. **after** today's #2522 and #2536 — neither
touches this), `generate: 'client'`, `dev: false`, against the pinned `submodules/svelte`:

```svelte
<script>
	let bar;
	$: {
		/* inner */
		bar = []
	}
	let z = 1;
	console.log(z);
</script>
```

```diff
 	let bar = $.mutable_source();
-
-	/* inner */
 	let z = 1;
```

## The rule upstream actually applies

`3-transform/client/visitors/LabeledStatement.js:63` returns `b.empty` and moves the body
into a synthesized `$.legacy_pre_effect(...)`. Placement is then entirely esrap's monotonic
comment cursor, and it does **two** independent things:

1. the next statement that survives has a `loc`, so the cursor flushes the comment as that
   statement's leading trivia — **re-homed**;
2. `b.block(body)` discards the `$:` body's own outermost block span, but any
   `BlockStatement` *nested* inside the body is an untouched source node that still carries
   its `loc`, so `reset_comment_index` rewinds into it and prints the comment **a second
   time, in place**.

Both are reproduced here. rsvelte's client path emits the effect body from the same text,
so "keep in place" costs nothing — the pass now moves the comment out and re-inserts it
past the statement, and skips the removal when the comment sits strictly inside a nested
block. `strip_reactive_statement_comments` becomes `rehome_reactive_statement_comments`.

The nested-block question is answered by **parsing the statement with oxc**, not by
classifying braces lexically: an object literal's `{` and an arrow body's `{` are one byte
apart and go opposite ways (`$: { bar = { /* c */ a: 1 } }` re-homes only;
`$: { const f = () => { /* c */ … } }` also keeps it). The parse runs only for a `$:`
statement that actually holds a comment, and an unparseable statement yields no ranges,
i.e. exactly the previous behaviour.

## Why the earlier "cannot be fixed in isolation" verdict does not hold

The issue thread concluded a fix needs #2336 first, on the grounds that official prints the
comment **twice** for `$: if (…)` while rsvelte prints it **zero** times, so "one comment
token relocated once cannot produce two occurrences". That is true of a *relocation*. The
measurement that changes it: with the strip disabled, rsvelte's text path **already**
carries the comment into the effect body — 1 occurrence, in exactly the position official's
second one occupies. So the operation is not "move once" but "insert a copy at the re-home
point, then decide independently whether the in-place one survives", and the second
decision is a property of the `$:` body's AST alone.

## Shapes measured against the official compiler

20 shapes, raw (unformatted) client output, official vs rsvelte:

| | before | after |
|---|---|---|
| byte-identical | 4 / 20 | 18 / 20 |

The two residuals are unrelated to this change and unchanged by it:

* `svelte-ignore` comments are still left where they are — official re-homes them too, but
  later text passes locate them by scanning backwards from the node they annotate
  (`expression_converter.rs:4587`), so moving them is a separate change with its own risk;
* one dependency thunk prints `() => $.get(bar)` where official prints `() => ($.get(bar))`
  — a pre-existing parenthesisation difference, and one the corpus normalisation erases.

Also found and **not** fixed here (both filed as observations, not regressions):

* a comment trailing the `$:` statement's closing `}` is kept by upstream *inside* the
  effect as well (`flush_trailing_comments`); rsvelte drops that second copy. Different
  mechanism, different pass.
* on `server` every one of these shapes still diverges, for the unrelated SSR
  interior-comment reason already recorded in the issue thread.

## Tests

`crates/rsvelte_core/tests/reactive_statement_comment_rehome_2368.rs` — 6 tests, every
expectation taken by running the official compiler on the same source at the pinned
submodule. Verified to fail **for the predicted reason** with the two source files reverted
to `origin/main`: 5 fail with the comment simply absent from the output, and

* `still_drops_a_block_body_comment_with_no_successor` — the #2355 / PR #2365 case, where
  nothing survives after the statement — **passes both before and after**. That is the
  control: re-homing must not resurrect it.
* `keeps_a_nested_block_comment_with_no_successor` is its counterweight — no successor, but
  a nested block, so upstream *does* print it. It failed before. "Nothing survives" and
  "nothing is printed" are not the same claim, and only having both distinguishes them.

Plus 5 new / 6 re-baselined unit tests in `formatting.rs`, including the object-literal
negative case.

## Ratchet

`pnpm run corpus:matrix`: **162 → 134**, all 28 in the shrink direction, none the other
way. The cleared entries are the `client` / `client-dev` halves of
`legacy-reactive__L06__*` and `legacy-reactive__L07__*` (14 cases × 2 targets), which is
every comment slot inside that seed's `$:` block body. Re-baselined with
`--update-baseline` in this PR; `matrix-known-failures.md` updated (family split, per-seed
split, the drops row of the classification table 100 → 72, and the server share).

Green in `--release`: `--lib` (1489), `runtime` (legacy + runes + hydration + browser),
`ssr`, `compiler_fixtures`, `sourcemaps`, `sourcemaps_gate`, `client_script_comments_pin`,
`client_transforms_pin`, `reactive_block_comment_delimiter_2351`, and the new suite.
`cargo fmt` + `cargo clippy --all-targets --all-features -D warnings` clean.

## Overlap with #2530

#2530 also touches legacy `$:`, but on the **server** text path
(`server/transform_legacy.rs`); this PR touches only `client/formatting.rs` and one call
site in `client/mod.rs`. No file overlap.

Fixes #2368


baseballyama:
`Compiler parity` is failing with **6 NEW corpus failures**, and they look like real regressions from this change rather than ratchet drift — two distinct symptoms across three entries, on `client` and `client-dev` alike ([run](https://github.com/baseballyama/rsvelte/actions/runs/31203937192/job/92950302509)):

**A derived read loses its call** — 2 entries, 4 comparisons:

```
layercake/src/lib/LayerCake.svelte (client line 822, client-dev line 1003)
  expected:  data: $_data(),
  actual:    data: $_data,

svelte-table/src/SvelteTable.svelte (client line 270)
  expected:  } else if (filterSelections()[f] === "") {
  actual:    } else if (filterSelections[f] === "") {

svelte-table/src/SvelteTable.svelte (client-dev line 325)
  expected:  } else if ($.strict_equals(filterSelections()[f], "")) {
  actual:    } else if ($.strict_equals(filterSelections[f], "")) {
```

Both are `$:`-declared values read elsewhere in the instance. The `()` that turns the derived into a read is gone, so the emitted code passes the derived itself where the value was meant — silently wrong at runtime, not a crash.

**Output that does not parse** — 1 entry, 2 comparisons:

```
svelte-ux/packages/svelte-ux/src/lib/components/Icon.svelte
  client:      Unexpected token
  client-dev:  A 'return' statement can only be used within a function body.
```

`return` escaping its function body suggests the re-homing moved a comment across a `}` and the surrounding text-splice offsets were computed against the pre-move source.

Both symptoms are consistent with one cause: the re-home changes the byte offsets that the downstream `$:`-read scans and the statement splices index into, and those consumers still use offsets taken before the move. Worth checking whether the re-home happens before or after the derived-read pass, and whether the offsets it hands downstream are the post-move ones.

The corpus keeps its artifacts on a failing run, so `compatibility/expected` vs `compatibility/actual` for those three ids should diff directly.

baseballyama:
I reduced the corpus failures this PR introduces. Two of them were **not** this PR's doing, and I have split them out into #2586.

Once a `$:` statement's comments survive to codegen, they reach two hand-rolled quote scanners — `is_inside_string_literal` (`client/state_transforms.rs`) and the walk in `transform_prop_reads_in_expr` (`client/props_transforms.rs`) — neither of which knew what a comment is. An apostrophe in `// it's not defined` opens a string nothing closes, so every store/prop read after it is emitted **uncalled**: `filterSelections[f]` where official has `filterSelections()[f]`. That is what `layercake` and both `svelte-table` divergences were.

Both defects reproduce on current `main` without this PR (measured with a `main`-built binding — the store case through any ordinary multi-line statement, the prop case through a `svelte-ignore` comment, which is the one comment kind `$:` statements already preserve). So they are pre-existing; this PR only widens the population that reaches them. Rebase on #2586 and those seeds should stop failing.

What is left after that is genuinely this PR's subject matter — comment **placement**, not comment loss:

- comments dropped from an object literal that this PR now carries into codegen;
- a comment re-homed to a position that changes the emitted statement, e.g. `var // comment\n\ttable = root_12();`.

I did not re-examine `svelte-ux/src/lib/components/Icon.svelte` (`js-unparseable`) after the two scanner fixes — that one may or may not have the same root cause, and is worth re-running before treating it as separate work.

baseballyama:
Pushed a fix for the `Compiler parity` failure — it was a real regression, not ratchet drift.

```
[verify] ❌ 2 NEW failures (not in baseline)
  - svelte-ux/.../components/Icon.svelte [js-unparseable] (client)
  - svelte-ux/.../components/Icon.svelte [js-unparseable] (client-dev)
```

## What happens

`reactive_statement_end` returns at the brace that closes the body's **first** block, so

```js
$: if (typeof data === 'object') {
  // Font Awesome
  …
} else if (typeof data === 'string') {
```

was recorded as ending before the `else`. On `main` that was harmless — the pass only blanked comments, and a short span just meant the `else` branch's comments were left alone. **Re-homing inserts text at `span.end`**, so the short span put `\n` + indent + the comment between `}` and `else`:

```
	  // Font Awesome else if (typeof data === 'string') {
```

which split the chain — the `else if` escaped the effect and became top-level text, and the output stopped being parseable JavaScript. Minimal repro (11 lines, no svelte-ux) reproduces it exactly.

This is the [adjacency-guard shape](https://github.com/baseballyama/rsvelte/issues/2368) again: **an in-place insertion turns a latent off-by-one in a span into a code-destroying edit**, and the span was wrong before this PR.

## The fix

`reactive_statement_end` now looks past the closing brace, skipping whitespace and comments, and continues when the next token is `else`, `catch` or `finally` — none of which can *begin* a statement.

`while` is deliberately excluded: `do … while` would also be a continuation, but `while` begins a statement too, so absorbing it would swallow the statement after a plain `$: { … }`. `a_while_after_the_block_is_not_absorbed` is the negative control for exactly that, and it is the test a "skip any keyword after `}`" fix would fail.

## Verification

- `Icon.svelte` client **and** client-dev output is now byte-identical to the official compiler's after oxfmt + blank-line stripping (both were checked; the dev target failed with a different message, `A 'return' statement can only be used within a function body`, so a fix verified only on `client` would not have covered it).
- 3 new tests in `reactive_statement_comment_rehome_2368.rs` — `else if` chain (exact official output), `try/catch/finally`, and the `while` control. 9/9 pass, as do the 5 in `reactive_block_comment_delimiter_2351.rs`.

Note for the rebase: `Shape matrix parity` is red on `main` for an unrelated reason (#2593), and this branch's matrix ratchet is 262 = main's 290 − the 28 this PR clears. After #2593 lands the target becomes 234 − 28 = 206, so re-derive rather than keeping 262.

baseballyama:
Ran the whole corpus locally on this branch with the fix, freshly built binding:

```
[verify] results:
  match            13183
  error-parity     948
  js-mismatch      0
  js-unparseable   0        ← was 2 (Icon.svelte, client and client-dev)
  css-mismatch     0
  error-mismatch   0

[verify] ✅ all corpus outputs identical after normalization
[verify] ✅ no warning regressions (556 known warning failures remain)
[verify] ✅ no error regressions (413 known error failures remain)
```

14,131 entries × 3 targets. So the fix clears both new failures and introduces none, and neither the warning nor the error ratchet moves.

Issue #2573, opened 2026-08-07
codegen (client): a private `$derived` field updated at a constructor root emits unparseable JS (`$.get(this.#d)++`)


baseballyama:
Found while settling #2483 (which asked whether *upstream* was the wrong side for
a non-`this` receiver — it is). Sweeping the neighbouring shapes with both
compilers turned up the mirror image: for a private **`$derived` / `$derived.by`**
field, a plain **`this`** receiver, at the **constructor root**, rsvelte emits
JavaScript that does not parse, where official's output is valid.

Unlike #2483's shape this needs no unusual receiver — `this.#d++` in a
constructor is mainstream code.

## Repro

`A.svelte.js`, `generate: 'client'`, `dev: false` (dev is identical):

```js
export class R {
	#a = $state(1);
	#d = $derived(this.#a * 2);

	constructor() {
		this.#d++;
	}
}
```

official:

```js
constructor() {
	$.update(this.#d);
}
```

rsvelte:

```js
constructor() {
	$.get(this.#d)++;
}
```

`oxc_parser` and acorn both reject that with `Assigning to rvalue`; V8 parses it
and throws `ReferenceError: Invalid left-hand side expression in postfix
operation` when the constructor runs. Vite/Rolldown reject the module.

## Full sweep

`#d` declared as `$derived(...)` and as `$derived.by(...)`, every update and
every assignment operator, constructor root vs method body. **Method bodies are
correct for every operator** — this is constructor-only.

### rsvelte emits unparseable JS (10 shapes × 2 field kinds = 20)

| source | official | rsvelte |
|---|---|---|
| `this.#d++` | `$.update(this.#d);` | `$.get(this.#d)++;` |
| `this.#d--` | `$.update(this.#d, -1);` | `$.get(this.#d)--;` |
| `++this.#d` | `$.update_pre(this.#d);` | `++$.get(this.#d);` |
| `--this.#d` | `$.update_pre(this.#d, -1);` | `--$.get(this.#d);` |
| `this.#d &= 5` | `$.set(this.#d, $.get(this.#d) & 5);` | `$.get(this.#d) &= 5;` |
| `this.#d \|= 5` | `$.set(this.#d, $.get(this.#d) \| 5);` | `$.get(this.#d) \|= 5;` |
| `this.#d ^= 5` | `$.set(this.#d, $.get(this.#d) ^ 5);` | `$.get(this.#d) ^= 5;` |
[...]

PR #2574, opened 2026-08-07
docs(compat): record and pin the non-`this` private-receiver divergence (#2483)

### LINKED ISSUE #2483: deliberate divergence (client): keep `$.update(inst.#n)` for a non-`this` receiver — matching upstream would emit unparseable `$.get(inst.#n)++`
**This is a decided, deliberate divergence, not a bug. Please do not "fix" it
toward upstream without reading this first.**

Split out of #2467 / #2479, where it was the one shape that stayed different
after the non-`this` receiver work.

## The divergence

`A.svelte.js`, `generate: 'client'`, `dev: false`:

```js
export class R {
	#n = $state(0);

	constructor(s) {
		const inst = this;
		inst.#n++;
		--inst.#n;
	}

	m(s) {
		const inst = this;
		inst.#n++;
	}
}
```

| | official | rsvelte (kept) |
|---|---|---|
| `inst.#n++` in a constructor | `inst.#n.v++;` | `$.update(inst.#n);` |
| `--inst.#n` in a constructor | `--inst.#n.v;` | `$.update_pre(inst.#n, -1);` |
| `inst.#n++` in a method | `$.get(inst.#n)++;` | `$.update(inst.#n);` |

A `this` receiver is **not** affected — `this.#n++` is `$.update(this.#n)` in
both, and that is plain parity.

## Why we do not match

`phases/3-transform/client/visitors/UpdateExpression.js:14-19` gates the
`$.update` form on the receiver:

```js
if (
	argument.type === 'MemberExpression' &&
	argument.object.type === 'ThisExpression' &&
	argument.property.type === 'PrivateIdentifier' &&
	context.state.state_fields.has('#' + argument.property.name)
) {
```

For any receiver that is not literally `this`, upstream falls through to
`context.next()` — a plain visited `MemberExpression` — which in a **method body**
yields `$.get(inst.#n)++`. That assigns to a CallExpression and **is not valid
JavaScript**; acorn rejects it with `Assigning to rvalue`, and Vite/Rolldown
reject the module.

So byte-equality and validity point in opposite directions here. This project's
rule is that output must match the official compiler exactly, and that rule holds
because upstream is the specification — but it does not extend to reproducing
output that is not valid JavaScript. Emitting a parse error on purpose to score a
byte match would trade a real defect for a cosmetic one. Exactness serves
correctness; where they conflict, correctness wins.

rsvelte's `$.update(inst.#n)` is valid, and is the same helper upstream itself
uses for the `this` case, so the semantics are upstream's own.

## Why this needs to be written down: no gate can see it

- **The corpus gate cannot.** `compatibility/known-failures.{client,server,client-dev}.json`
  are all `[]`, so no corpus entry contains the shape.
- **The fixture suites cannot.** Grepping the whole official fixture corpus for a
  non-`this` private-field receiver returns *nothing at all* — not one
  assignment, not even one read (measured in #2479).
- **A parse-validity gate cannot**, because our output is the *valid* side. If
  anything, `ast_gate_preconditions` would go red if someone "corrected" us
  toward upstream — but only if a fixture existed, and none does.

That combination — an unobserved surface plus a plausible local reason ("we
should match upstream", which is normally correct) — is exactly how a future
contributor ends up reintroducing a parse error while believing they are
improving parity. The only thing standing in the way is this record and the
comment at the rewrite site.

## Where it is recorded in the code

`crates/rsvelte_core/src/compiler/phases/3_transform/client/private_class_assign_ast.rs`,
one comment at each of the two rewrite sites (the spliced collector's
`visit_update_expression` and the in-place `rewrite_update`), plus a pointer under
the mapping table in the module header.

## If upstream fixes it

Svelte would presumably extend the `ThisExpression` check to any receiver, making
official emit `$.update(inst.#n)` too — at which point this divergence closes by
itself and both the comments and this issue should be deleted. Worth reporting
upstream: `$.get(x.#n)++` is an unambiguous codegen bug on their side.

baseballyama:
Closes #2483.

The issue asked for a document. Before writing one I ran both compilers on its
input (`submodules/svelte` 5.56.8 vs the NAPI binding) and checked every output
with a parser, because "deliberate" is a claim about which side is wrong.

## Verdict: the divergence the issue names is real, but the issue mis-scopes it

| position (`#n = $state(0)`) | official | parses | rsvelte | parses |
|---|---|---|---|---|
| method body, `o.#n++` | `$.get(o.#n)++;` | **no** | `$.update(o.#n);` | yes |
| static method, `o.#n++` | `$.get(o.#n)++;` | **no** | `$.update(o.#n);` | yes |
| nested fn in constructor, `inst.#n++` | `$.get(inst.#n)++;` | **no** | `$.update(inst.#n);` | yes |
| constructor root, `inst.#n++` | `inst.#n.v++;` | yes | `$.update(inst.#n);` | yes |
| constructor root, read `inst.#n` | `inst.#n.v` | yes | `$.get(inst.#n)` | yes |
| any position, `this.#n++` | `$.update(this.#n);` | yes | `$.update(this.#n);` | yes |

The unparseable rows are upstream's bug: `UpdateExpression.js:14-19` gates the
helper on `argument.object.type === 'ThisExpression'`, while the visitor it
falls through to (`MemberExpression.js:11-19`) does not check the receiver at
all, so outside a constructor root the fallthrough puts a CallExpression in
assignment position. acorn and `oxc_parser` both say `Assigning to rvalue`; V8
parses it and throws `ReferenceError: Invalid left-hand side expression in
postfix operation` when the method runs. Reported upstream as
**sveltejs/svelte#18621** (open).

Two corrections to the issue, both from running the thing:

* the **constructor-root** rows are not that case — `inst.#n.v++` parses. rsvelte
  is still right there, but for a different reason (`.v++` writes the source
  without notifying, and upstream's receiver test does not establish that the
  receiver is the object under construction), and the record says so separately
  instead of folding it into "upstream emits invalid JS".
* the issue's claim that grepping the fixture corpus for a non-`this` private
  receiver "returns nothing at all — not one assignment, not even one read" is
  **false**. Three samples use one: `private-identifiers-not-this`,
  `class-private-fields-reassigned-this`, `class-state-derived-private`. What is
  actually absent is an *update expression* through one (grepped with a
  `this.#count++` positive control), and a constructor-root read through one.
  The issue also describes in-code comments at the rewrite sites that do not
  exist on `main`; this PR adds them.

## What lands

* `compatibility/deliberate-divergences.md` — the record: input, both outputs,
  the parse verdicts, upstream file+line, the upstream report, the condition
  under which the entry disappears, and the per-gate evidence for why nothing
  sees it.
* `crates/rsvelte_core/tests/private_field_non_this_receiver_2483.rs` — enforces
  the choice for client and client-dev. It carries a **positive control** that
  `oxc_parser` rejects `$.get(o.#n)++`; without it the parse assertions would be
  a non-discriminating test.
* one WHY comment at each of the three sites that emit the helper, plus a pointer
  under the mapping table in `private_class_assign_ast.rs`.

No compiler behaviour changes.

## Separate finding, filed rather than fixed here

The same measurement swept the neighbouring shapes and found **rsvelte** emitting
unparseable JS for a plain `this` receiver: a private `$derived` field updated at
a constructor root lowers to `$.get(this.#d)++`. 20 unparseable shapes plus 6 that
read `.v` off a call result — filed as #2573 with the full table. It is a
behaviour fix and does not belong in a documentation PR.


baseballyama:

```
```

baseballyama:
CodSpeed's one regression here is **non-discriminating for this PR**, and I am merging over it.

The regressed benchmark is `crates/rsvelte_devtools/benches/parser.rs::bench_parallel_parse::corpus`, whose body is a single call to `parse_parallel(...)` — Phase 1 only. This PR's code diff is two files in `3_transform/client` (`class_transforms.rs`, `private_class_assign_ast.rs`) plus two `compatibility/*.md`. Nothing the parse benchmark executes is reachable from that diff.

Three things narrow it further:

- CodSpeed itself attached its **"Different runtime environments detected"** banner to this comparison, which is its own statement that the two sides were not measured comparably.
- `parallel_parse` is the rayon benchmark, so it carries scheduling noise the single-threaded ones do not.
- 148 benchmarks are untouched and 1 moved. A real Phase-1 regression from a Phase-3 edit would not be able to pick exactly the one parallel case.

What this is **not**: a claim that CodSpeed is broken, or a general licence to ignore it. Code layout alone has been measured to move unrelated benchmarks by ~5% in this repo, and 10% is above that floor — so this rests on the unreachability argument above, not on the size of the number. If the same benchmark regresses on a PR that *can* reach Phase 1, that is a real signal.

Issue #2404, opened 2026-08-06
compiler: offset newtypes do not catch a mismatched *length* — `char_offset + name.len()` still compiles


baseballyama:
## Summary

#2386's `ByteOffset` / `CharOffset` make a mismatched **offset** a compile error. They do nothing
about a mismatched **length**, because both types implement `Add<usize>` and a length is a
`usize`.

This is not a corner: it is the shape of the largest cluster the epic has found.

## Demonstration

```rust
let i = CharOffset::new(4);
let after = i + "$名前".len();   // compiles
assert_eq!(after.get(), 11);     // passes — and 11 is wrong; the answer is 7
```

`"$名前"` is 3 characters and 7 bytes. `i` is a char offset. The sum is neither unit.

That is exactly #2389's defect: `client/store_transforms.rs` advanced a char index by
`store_sub.len()` at **nine sites**, dropping four characters of source text after every match.
**Converting `i` to `CharOffset` would not have caught a single one of them** — all nine still
compile and still produce the same wrong value.

## Why the offset types cannot fix it

The wrong operand is not the offset. `i` is correct, `+` is correct, and the result type is
correct. The error is in `"$名前".len()`, whose type says `usize` and whose meaning is *bytes*.
An offset newtype has no way to object, because it never sees a unit on the right-hand side.

## Proposed shape

Give lengths types too, and admit only the matching pair:

```rust
struct ByteLen(usize);
struct CharLen(usize);

impl Add<CharLen> for CharOffset { type Output = CharOffset; }
impl Add<ByteLen> for ByteOffset { type Output = ByteOffset; }
// and no `Add<usize>` for either offset
```

Removing `Add<usize>` is the load-bearing half; adding the length types is what makes the
remaining arithmetic expressible. `str::len()` and `chars().count()` then need named wrappers
(`ByteLen::of(s)` / `CharLen::of(s)`) so the unit is chosen at the point of measurement rather
than inferred at the point of use.

## Cost, stated up front

This is wider than #2386's slices. Dropping `Add<usize>` breaks every stepping expression
(`i + 1`, `close + 1`) until each is given a unit — and most of those really are unit-agnostic
"advance by one element", which argues for keeping a deliberate `step()`/`next()` rather than
forcing a `CharLen(1)` at every site. That design question should be settled before any
conversion starts.

## Sequencing

**Not to be started yet.** #2386's slices should land first; this widens the design and would
[...]

PR #2577, opened 2026-08-07
fix(compiler): decide the component/bind $.assign exemption by node identity (#2420)

### LINKED ISSUE #2420: probe: does in_component_attribute have the same identity-vs-boolean defect?

baseballyama:
Closes #2420.

## The probe

Answer: **yes, `in_component_attribute` has the same defect** — and so does its
neighbour `in_bind_directive`, which implements the other half of the same
upstream `if`.

49 shapes were compiled with both compilers at Svelte 5.56.8, `dev: true`,
`generate: 'client'`, and the `$.assign*` calls diffed. The RHS is a non-primitive
(`obj = {}`) throughout, so `should_transform`'s `!is_primitive` conjunct never
silences the comparison, and the set includes a positive control that emits on
both sides (`<button onclick={() => { f(o.x = obj); }}>` → `["o.x"]`).

**12 of the 49 shapes diverged on `origin/main`, in both directions:**

| shape | official | rsvelte @ main |
|---|---|---|
| `<Comp prop={f(() => (o.x = obj))} />` | `["o.x"]` | `[]` |
| `<Comp prop={() => (o.x = f(() => (c.d = obj)))} />` | `["c.d"]` | `[]` |
| `<Comp prop={(g, (h, () => (o.x = obj)))} />` | `["o.x"]` | `[]` |
| `<Comp bind:value={() => o.x, (v) => (o.y = f(() => (c.d = obj)))} />` | `["c.d"]` | `[]` |
| `<svelte:component this={C} prop={() => (o.x = obj)} />` | `["o.x"]` | `[]` |
| `{#if obj}<svelte:self prop={() => (o.x = obj)} />{/if}` | `["o.x"]` | `[]` |
| `{#if obj}<svelte:self bind:value={() => o.x, (v) => (o.y = obj)} />{/if}` | `["o.y"]` | `[]` |
| `<Comp prop={(o.x = obj)} />` | `[]` | `["o.x"]` |
| `<svelte:component this={C} prop={(o.x = obj)} />` | `[]` | `["o.x"]` |
| `<input bind:value={() => o.x, (v) => (o.y = obj)} />` | `[]` | `["o.y"]` |
| `<svelte:window bind:scrollY={…, (v) => (o.y = obj)} />` | `[]` | `["o.y"]` |
| `<svelte:element this={tag} bind:clientWidth={…, (v) => (o.y = obj)} />` | `[]` | `["o.y"]` |

After this change: **0 of 49**.

The upstream predicate was not inferred from the source alone — the official
compiler was instrumented at the `if` to dump `context.path` per shape, which is
what showed that `OnDirective` and `AttachTag` are *absent* from the path for a
component (so `<Comp on:foo={arrow}>` is exempt) while `BindDirective` and
`SpreadAttribute` are *present*, and that `RegularElement` is the only container
whose children are reached without an intervening `Fragment` node.

## The fix

Mirrors the #2421 precedent. Phase 2 records the qualifying arrow starts
(`assign_exempt_arrow_starts`) and assignment starts
(`assign_exempt_assignment_starts`) on `ComponentAnalysis`; Phase 3 tests
identity. `in_component_attribute`, `in_bind_directive` and
`parent_is_regular_element` are deleted — the last survives as a Phase-2 local,
because upstream's lone-arrow arm alone carries `path.at(-3) === 'Fragment'`.

## Tests

`crates/rsvelte_core/tests/component_attribute_assign_exemption.rs`, 11 tests.
Every expectation in it was checked against the official compiler at the same
version. On `origin/main`, 7 of the 11 fail, each for the predicted reason:

```
---- an_assignment_that_is_the_attribute_expression_is_exempt stdout ----
assertion failed: assigns(&src("<Comp prop={(o.x = obj)} />")).is_empty()

---- a_component_attribute_exempts_only_its_own_arrow stdout ----
  left: []
 right: ["c.d"]

---- an_arrow_in_a_nested_sequence_is_not_exempt stdout ----
  left: []
 right: ["o.x"]

---- a_lone_arrow_on_svelte_component_or_svelte_self_is_not_exempt stdout ----
  left: []
 right: ["o.x"]

---- a_getter_setter_pair_on_svelte_self_is_not_exempt stdout ----
  left: []
 right: ["o.y"]

---- an_arrow_nested_in_a_component_attribute_is_not_exempt stdout ----
  left: []
 right: ["o.x"]

---- an_element_bind_setter_is_exempt stdout ----
expected no wrap for `<input bind:value={() => o.y, (v) => (o.y = obj)} />`, got ["o.y"]

test result: FAILED. 4 passed; 7 failed
```

The 4 that pass on main are the pins the fix must not break — the positive
control, the exemptions that were already right, `<div><Comp prop={arrow}/></div>`
(the `Fragment` conjunct, #2404's case) and the spread attribute.

83 existing `rsvelte_core` test binaries covering `dev_*`, `client_*`,
`component_*`, `bind*`, `ssr_*`, `legacy_*`, `module_*`, `state_*`, `attribute_*`
and the directive suites pass; `cargo fmt` and
`cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` are
clean. The full suite and the corpus gates are left to CI: this machine's free
disk fell below 4 GiB mid-run.

## Out of scope, found while probing

`<Comp bind:value={o.x = obj} />` and `<Comp bind:this={o.x = obj} />` are
accepted by rsvelte; official raises `bind_invalid_expression` ("Can only bind to
an Identifier or MemberExpression or a `{get, set}` pair"). rsvelte does raise it
for the same shape on an element. Unrelated to this predicate, not touched here.


baseballyama:
**These two PRs fix overlapping halves of the same upstream `if`, by different mechanisms. Only one should land as written.** My fault for briefing #2484 and #2420 as separate work — the issues read as neighbours and are in fact one predicate.

Shared source files (5 of #2569's 6):

```
3_transform/client/types.rs
3_transform/client/visitors/bind_directive.rs
3_transform/client/visitors/expression_converter.rs
3_transform/client/visitors/fragment.rs
3_transform/client/visitors/shared/component.rs
```

#2577 additionally touches Phase 2 (`2_analyze/visitors/{component,svelte_component,svelte_element,bind_directive,shared/attribute}.rs`), because it records the qualifying nodes during analysis; #2569 stays in Phase 3 and keeps a set of qualifying arrow start offsets. Those are not two patches to combine — they are two answers to "where does this predicate live".

What each measured, so the comparison is on the record:

| | #2569 (#2484) | #2577 (#2420) |
|---|---|---|
| shapes | 11 | 49 |
| divergences found on `main` | 5 | 12 |
| after | 0/11 | 0/49 |
| predicate derived from | reading `AssignmentExpression.js:204-218` | **patching the official compiler to dump `context.path` per shape** |

#2577's derivation method is the stronger one and it produced three corrections to the issue's reading of the source — including that `path.at(-3) === 'Fragment'` is **not** vacuous (`RegularElement` is the one container reached without an interposed `Fragment`), which a source read had called redundant.

But breadth does not settle it, because #2569 carries a case that discriminates between two implementations agreeing everywhere else:

```svelte
<input bind:value={() => s.x, (v) => (s.y = o, d.e = o)} />
```

Upstream wraps **both** assignments (count 2), because the arrow's body is a `SequenceExpression` so neither assignment's parent is the arrow. An implementation that exempts "the setter's whole subtree" reports 0 there and looks correct on every other row.

**Resolution protocol**, so this is decided by measurement rather than by merge order: run the union of the two shape sets against both branches. If #2577 also gives 2 on the shape above and 0 divergences on #2569's other 10, it subsumes #2569 and #2569 should close in its favour, keeping its test file. If it does not, #2577 has the wider coverage and the wrong predicate, and the fix is #2569's rule at #2577's site.

I have asked #2577's author to run that comparison and report both numbers.

baseballyama:
## Settling the overlap with #2569 by measurement

#2569 and this PR fix overlapping halves of the same upstream `if`
(`AssignmentExpression.js:204-215`) and share 5 source files, so only one should
land. Both branches were run against the **union** of the two shape sets — my 49
plus #2569's 18 from `dev_assign_bind_sequence.rs`, plus 5 coercive-operator
shapes added below — 72 shapes, official compiler at Svelte 5.56.8, `dev: true`,
`generate: 'client'`, comparing the **list of `$.assign*` callees** rather than a
substring count.

**#2569's discriminating case, `<input bind:value={() => s.x, (v) => (s.y = o, d.e = o)} />`:**

| | official | #2577 | #2569 |
|---|---|---|---|
| `$.assign*` calls | `["$.assign", "$.assign"]` | `["$.assign", "$.assign"]` | `["$.assign", "$.assign"]` |

Both give **2**. This branch is not the "exempt the setter's whole subtree"
implementation the case was built to catch: the Phase-2 recorder marks the arrow,
and Phase 3 only clears the wrap when the arrow's body *is* an
`AssignmentExpression`. A `SequenceExpression` body is not, so both assignments
wrap.

**Over the 72-shape union:**

| branch | divergences |
|---|---|
| #2577 (this PR) | **2** — both the `bind_invalid_expression` gap below, not `$.assign` |
| #2569 | **13** |

Every row where the two branches disagree:

| shape | official | #2577 | #2569 |
|---|---|---|---|
| `<Comp prop={f(() => (o.x = obj))} />` | `["$.assign"]` | ok | `[]` |
| `<Comp prop={() => (o.x = f(() => (c.d = obj)))} />` | `["$.assign"]` | ok | `[]` |
| `<Comp prop={(o.x = obj)} />` | `[]` | ok | `["$.assign"]` |
| `<svelte:component this={C} prop={() => (o.x = obj)} />` | `["$.assign"]` | ok | `[]` |
| `<svelte:component this={C} prop={(o.x = obj)} />` | `[]` | ok | `["$.assign"]` |
| `{#if obj}<svelte:self prop={() => (o.x = obj)} />{/if}` | `["$.assign"]` | ok | `[]` |
| `{#if obj}<svelte:self bind:value={…, (v) => (o.y = obj)} />{/if}` | `["$.assign"]` | ok | `[]` |
| `<Comp prop={(g, (h, () => (o.x = obj)))} />` | `["$.assign"]` | ok | `[]` |
| `<Comp {@attach () => (o.x = obj)} />` | `[]` | ok | `["$.assign"]` |
| `<div><Comp prop={(o.x = obj)} /></div>` | `[]` | ok | `["$.assign"]` |
| `<Comp prop={wrap(() => (o.y ??= obj))} />` | `["$.assign"]` | ok | `[]` |

There is no row in the other direction: no shape in the union where #2569 matches
official and #2577 does not. #2569 leaves `in_component_attribute` in place and
only replaces `in_bind_directive`, so the component half of the predicate keeps
the original defect.

**#2569's 11 tests are ported here unchanged** (`dev_assign_bind_sequence.rs`,
741388b9) and pass, so its cases survive the close. Combined with this PR's own
11 and the two pre-existing files, 37 tests now cover the predicate.

### One blind spot checked and cleared

`dev_assign_bind_sequence.rs` counts the literal `"$.assign("`, and rsvelte has a
`get_coercive_assign_callee` that maps `??=` to `$.assign_nullish` — which that
counter cannot see. Measured with 5 added shapes putting `??=`, `||=` and `&&=` in
wrapped positions (`<input bind:value={() => s.x, wrap((v) => (s.y ??= obj))} />`
and four others): both compilers emit `$.assign` in every one. The `_nullish`
callee is not reached on this path, so the counter is discriminating for those
tests. Reporting it because a negative result found by looking is worth more than
one assumed.

### Adjacent defect, filed separately

Same input, three compiles:

```svelte
<Comp bind:value={o.x = obj} />
<Comp bind:this={o.x = obj} />
<input bind:value={o.x = obj} />
```

Official raises `bind_invalid_expression` — "Can only bind to an Identifier or
MemberExpression or a `{get, set}` pair" — for all three. rsvelte raises it only
for the element, and only with the text "Invalid binding expression". The two
component forms compile, emitting `return o.x = obj;` as the getter and
`o.x = obj = $$value` as the setter. Not touched here.

PR #2585, opened 2026-08-07
perf(ast): shrink JsNode from 144 to 80 bytes

baseballyama:
## Why

Every JS expression node in Phase 1 is individually heap-boxed
(`Expression::Typed(Box<TypedExpr>)`), so `size_of::<JsNode>()` is paid on every
`malloc` and on every `Vec<JsNode>` memcpy. It was 144 bytes — and
`-Zprint-type-sizes` showed that 144 came from exactly **two** of the 87
variants:

| variant | size | why |
| --- | --- | --- |
| `Literal` | 143 | two inline `RegexValue`s (`value` 48 + `regex` 48) |
| `Program` | 127 | four 24-byte fields, on a node there is one of per script |
| `ExportNamedDeclaration` | 79 | — |
| `Identifier` and ~75 others | ≤ 55 | — |

So the modal node — `Identifier`, 55 bytes of real payload — was stored at 144
because a regex literal *might* appear.

## What

Two representation-only commits:

1. `LiteralValue::Regex(Box<RegexValue>)` and `Literal.regex: Option<Box<RegexValue>>`.
2. `Program`'s `leading_comments` / `trailing_comments` / `ignore_comment_map`
   move into a `ProgramMetadata` behind one `Box`.

Nothing else changes: the serializer already reached both regex payloads through
`if let Some(..)` / a match arm, and the `Program` arm through named bindings, so
boxing leaves every emitted key, order and null-vs-absent decision identical.

## Sizes

`cargo +nightly rustc -p rsvelte_core --lib -- -Zprint-type-sizes`

| | before | after |
| --- | --- | --- |
| `JsNode` | 144 | **80** |
| `TypedExpr` | 152 | **88** |

Top variants after:

```
type: `ast::typed_expr::JsNode`: 80 bytes, alignment: 8 bytes
    discriminant: 1 bytes
    variant `Literal`: 79 bytes
    variant `ExportNamedDeclaration`: 79 bytes
    variant `TemplateElement`: 71 bytes
    variant `Program`: 63 bytes
    variant `ImportDeclaration`: 63 bytes
    variant `Comment`: 63 bytes
    variant `Identifier`: 55 bytes
```

`Literal` and `ExportNamedDeclaration` now tie at 79, so the next byte would cost
a third indirection on two unrelated variants at once. This is where it stops
paying.

## Allocated bytes

`crates/rsvelte_devtools/src/bin/alloc_count.rs`, release, over the 1,296
`.svelte` files of `submodules/flowbite-svelte`:

| target | bytes before | bytes after | Δ |
| --- | --- | --- | --- |
| client | 482,494,867 | 458,289,569 | **−24,205,298 (−5.02%)** |
| server | 324,343,421 | 300,334,317 | **−24,009,104 (−7.40%)** |

Allocation *count* moves the other way by exactly the amount predicted — +2,649
(client) / +2,544 (server) over 1,296 files, i.e. the ~2 `ProgramMetadata` boxes
per file. That is the intended trade: two 72-byte boxes per file to take 64 bytes
off every node in the arena.

No wall-clock timing was taken; this PR reports static sizes and deterministic
counters only.

## Verification

- `cargo test -p rsvelte_core --no-fail-fast`: **2,623 passed, 1 failed, 8
  ignored** across 246 test binaries. The single failure is
  `validator::validator_warning_messages_match_official`, a two-sided-ratchet
  complaint ("2 entr(ies) … now match") that reproduces **byte-identically on
  unmodified `origin/main`** — verified by re-running the same test with
  `git checkout origin/main -- crates/`. Pre-existing, unrelated.
  (`tests/ast_gate_preconditions` overflows the rayon worker stack in debug on
  this machine, also on unmodified `origin/main`; run under
  `RUST_MIN_STACK=33554432`.)
- `cargo test -p rsvelte_bindings_support -p rsvelte_esrap -p rsvelte_lint`: 0 failures.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean. No
  `clippy::large_enum_variant` allow existed on `JsNode`, so there was none to remove.
- `cargo fmt --all`: clean.

### DIFF
--- a/.changeset/shrink-jsnode-to-80-bytes.md
+++ b/.changeset/shrink-jsnode-to-80-bytes.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Shrink `JsNode` from 144 to 80 bytes by boxing the payloads of its two outlier variants (`Literal`'s regex values and `Program`'s comment/ignore metadata). Compiler output is unchanged.
--- a/crates/rsvelte_bindings_support/src/napi_raw_parse.rs
+++ b/crates/rsvelte_bindings_support/src/napi_raw_parse.rs
@@ -1228,7 +1228,7 @@ fn write_js_node<W: Writer>(w: &mut W, node: &JsNode, arena: &ParseArena) -> std
             write_typed_loc(w, loc.as_deref());
             write_literal_value(w, value);
             write_str(w, raw.as_str());
-            write_regex(w, regex.as_ref());
+            write_regex(w, regex.as_deref());
         }
         JsNode::BinaryExpression {
             start,
--- a/crates/rsvelte_bindings_support/src/napi_raw_parse.rs
+++ b/crates/rsvelte_bindings_support/src/napi_raw_parse.rs
@@ -1627,17 +1627,14 @@ fn write_js_node<W: Writer>(w: &mut W, node: &JsNode, arena: &ParseArena) -> std
             loc,
             body,
             source_type,
-            leading_comments,
-            trailing_comments,
-            // Internal analyze-only metadata; not part of the serialized AST.
-            ignore_comment_map: _,
+            metadata,
         } => {
             write_preamble(w, JS_PROGRAM, *start, *end);
             write_typed_loc(w, loc.as_deref());
             write_id_range(w, *body, arena)?;
             write_str(w, source_type.as_str());
-            write_opt_inline_json(w, trailing_comments.as_ref())?;
-            write_opt_inline_json(w, leading_comments.as_ref())?;
+            write_opt_inline_json(w, metadata.trailing_comments.as_ref())?;
+            write_opt_inline_json(w, metadata.leading_comments.as_ref())?;
         }
         JsNode::ExpressionStatement {
             start,
--- a/crates/rsvelte_core/src/ast/typed_expr.rs
+++ b/crates/rsvelte_core/src/ast/typed_expr.rs
@@ -58,7 +58,9 @@ pub enum LiteralValue {
     Number(f64),
     Bool(bool),
     Null,
-    Regex(RegexValue),
+    /// Boxed: a regex payload is two `CompactString`s, and inlining it would
+    /// widen every `JsNode` by 24 bytes for a variant that is vanishingly rare.
+    Regex(Box<RegexValue>),
 }
 
 impl Serialize for LiteralValue {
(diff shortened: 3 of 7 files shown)


baseballyama:
CodSpeed's one regression here is the workload change, not a slowdown.

It compares against `main` at `723df2f`, which predates #2571 — that is also why 22 benchmarks show up as "new". The regressed ID is `parallel_parse::corpus`, the aggregate that parses **every** fixture in one iteration, so adding two fixtures to a ~19 KB corpus raises it by construction. This is the case `benches/corpus/README.md` now documents ("Adding a new file is free for the per-file benchmark IDs, but not for the two that aggregate the whole corpus into a single ID").

The signal worth reading in the same report is the two per-file IDs, whose workload did *not* change:

| benchmark | base | head | |
|---|---|---|---|
| `parse[synthetic-state-heavy]` | 3.9 ms | 3.4 ms | **+16.95%** |
| `parse[synthetic-legacy-state-heavy]` | 3.9 ms | 3.4 ms | **+17.08%** |

That is the `JsNode` 144 B → 80 B shrink paying off. Acknowledging the corpus alert.

baseballyama:
Merging with CodSpeed red. Its own table refutes the verdict:

| | benchmark | BASE | HEAD | |
|---|---|---|---|---|
| ❌ | `parallel_parse::corpus` | 9.5 ms | 12.6 ms | −24.49% |
| ⚡ | `parse[synthetic-legacy-state-heavy]` | 3.9 ms | 3.4 ms | **+17.08%** |
| ⚡ | `parse[synthetic-state-heavy]` | 3.9 ms | 3.4 ms | **+16.95%** |

The two *per-fixture* parse benchmarks — same phase, same code path, fixed input — get **17% faster**, which is what shrinking `JsNode` should do. Only the aggregate regresses.

`parallel_parse::corpus` measures `benches/corpus/` as a whole, and #2571 grew that directory from 9 files to 11 (the `🆕 22` new benchmarks in the same report are those two fixtures). Its wall-clock is a different measurand on either side of that commit, so the comparison is not a comparison.

The number is not specific to this PR either: #2586 saw the same benchmark fall −27.1% from the same 9.5 ms base, and #2587 −33.61%, and neither of those touches phase 1 at all. `main`'s Benchmark runs keep being cancelled by the merge rate, so CodSpeed has no post-#2571 baseline and every PR inherits the same phantom.

PR #2586, opened 2026-08-07
fix(transform): stop a comment apostrophe from suppressing store and prop read rewrites

baseballyama:
An apostrophe in a comment makes rsvelte emit a store or prop read **uncalled**. The output parses, so nothing that checks parseability sees it; it is simply wrong at runtime.

```svelte
<script>
	import { writable } from 'svelte/store';
	const s = writable(1);

	const obj = {
		// it's fine
		v: $s
	};
</script>
```

| | |
|---|---|
| official | `v: $s()` |
| rsvelte | `v: $s` |

Delete the apostrophe and rsvelte is correct.

## Cause

Two scans rewrite reads — `$store` → `$store()` and `prop` → `prop()` — and both track quote state so they do not rewrite identifiers inside string literals. Neither knew what a comment was:

- `is_inside_string_literal` (`client/state_transforms.rs:1078`), which guards `transform_store_reads_client`, has **no `/` handling at all**.
- the walk in `transform_prop_reads_in_expr` (`client/props_transforms.rs:162`) has its own copy of the same state machine, with the same gap.

The `'` in `doesn't` opens a string that nothing closes, so every position after it answers "inside a string" and its rewrite is skipped — for the rest of the statement in the prop scan, and for the rest of the string handed to the store scan.

Both now skip `//` and `/* */`. The template-literal semantics are unchanged on purpose: `${…}` contents stay *code*, so `` `${$store}` `` still becomes `` `${$store()}` ``. That is why this does not just call `shared/js_scan.rs::skip_opaque`, which treats a whole template including its interpolations as opaque — `server/helpers.rs:1396` already keeps a separate branch for the same reason.

## Reachability

Both shapes reproduce on **current `main`**, measured by building the NAPI binding from `main` and from this branch and running both compilers on the same input:

- **store read** — any ordinary multi-line statement. Nothing strips comments outside a `$:` statement, so the comment reaches the scan directly.
- **prop read** — a `svelte-ignore` comment inside a `$:` body. Reactive statements have their comments stripped before the visitors run, *except* `svelte-ignore`, which is deliberately left in place because later text passes locate it by scanning back from the node it annotates. So it is the one comment kind that always reaches these scans from a reactive statement.

Found while investigating #2572, which stops stripping `$:` comments and so turns the first class into a corpus failure on `layercake`, `svelte-table` and `svelte-ux`. It is split out here because both defects predate that PR and are reachable without it — `git log` will otherwise attribute a bug to the change that merely uncovered it.

## Verification

`crates/rsvelte_core/tests/comment_quote_read_scans.rs` — the line-comment store case, the `svelte-ignore` prop case, a block-comment case (the other branch), and a **control** asserting an identifier inside a real string literal is still left alone, so the fix reads as "comments are not strings" rather than "quote tracking was dropped".

Locally green: `--lib` 1508, `runtime` (release) 5/5, `ssr`, `print`, `css`, `validator`, `client_transforms_pin`, `compiler_fixtures`, `cargo clippy --all-targets --all-features -- -D warnings`.

Against the corpus seeds this was reduced from, using the gate's own `flattenTemplateHoles → oxfmt → stripBlankLines` chain: `layercake/src/lib/LayerCake.svelte` goes from mismatching to **MATCH**, and `svelte-table` loses both `filterSelections()` divergences.

https://claude.ai/code/session_01Sb9ucum4Cxu99K1WCeyGC8

### DIFF
--- a/.changeset/comment-quote-read-scans.md
+++ b/.changeset/comment-quote-read-scans.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Stop an apostrophe in a comment from suppressing the store and prop read rewrites. `// it's fine` opened a string literal that nothing closed, so every `$store` / prop read after it was emitted uncalled — code that parses and is silently wrong at runtime.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/props_transforms.rs
@@ -276,6 +276,31 @@ pub(super) fn transform_prop_reads_in_expr(expr: &str, prop_vars: &[String]) -> 
                 }
             }
 
+            // A quote inside a comment is text: an apostrophe in `// it's not
+            // defined` would otherwise open a string that nothing closes, and
+            // every identifier after it would be left untransformed.
+            if c == '/' && i + 1 < chars.len() && (chars[i + 1] == '/' || chars[i + 1] == '*') {
+                let line = chars[i + 1] == '/';
+                new_result.push(c);
+                new_result.push(chars[i + 1]);
+                i += 2;
+                while i < chars.len() {
+                    if line {
+                        if chars[i] == '\n' {
+                            break;
+                        }
+                    } else if chars[i] == '*' && i + 1 < chars.len() && chars[i + 1] == '/' {
+                        new_result.push(chars[i]);
+                        new_result.push(chars[i + 1]);
+                        i += 2;
+                        break;
+                    }
+                    new_result.push(chars[i]);
+                    i += 1;
+                }
+                continue;
+            }
+
             // Check for string literal start
             if c == '\'' || c == '"' || c == '`' {
                 in_string = Some(c);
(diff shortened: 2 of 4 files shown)


baseballyama:
Merging with CodSpeed red. The one benchmark it names is `parallel_parse::corpus`, 9.5 ms → 13 ms, and this PR cannot reach it: the diff is `client/props_transforms.rs`, `client/state_transforms.rs`, one test and one changeset — all phase 3. `bench_parallel_parse` is phase 1.

What actually moved is the corpus. #2571 added `10-legacy-typescript-props.svelte` and `11-store-heavy-legacy.svelte` to `benches/corpus/`, and `parallel_parse::corpus` is an **aggregate over that directory** — 9 files before, 11 after, so its wall-clock is a different measurand on either side of that commit. The same report lists `🆕 22` new benchmarks, which are those two fixtures' 11 cases each; it also says it compared against `5f6de88` because no successful run existed on `main`, and warns "different runtime environments detected".

`main`'s Benchmark runs are being **cancelled** by the merge rate (five consecutive `cancelled` runs at the time of writing), so CodSpeed has no post-#2571 baseline to compare against and every PR inherits the same phantom −27%. That is a property of the merge queue, not of any of these PRs. Worth fixing separately — either by exempting the aggregate benchmark from the gate across a corpus change, or by letting `main`'s benchmark run finish.

PR #2587, opened 2026-08-07
perf(compiler): build a warning frame from the shared line index

baseballyama:
A warning's code frame quotes five source lines. `generate_frame` reached them with

```rust
let lines: Vec<&str> = source.split('\n').collect();
```

— a full scan and allocation of the whole source, **once per warning that has a position**. A file that warns a lot pays O(source × warnings) to print O(warnings × 5) lines.

The line starts it needs are already computed: the caller builds one `legacy::Utf8ToUtf16` per compile to turn warning byte offsets into line/column, and that table holds `line_starts_byte`. The frame now takes the table and slices its five lines straight out of `source`.

## Measured

3200 `` lines (44.8 KB, 3200 warnings), release, whole `compile` call, warm — steady-state of three runs each:

| | 400 | 800 | 1600 | 3200 |
|---|---|---|---|---|
| before | 18.1 ms | 25.1 ms | 46.8 ms | 126 ms |
| after | 3.4 ms | 2.8 ms | 5.4 ms | 13 ms |

Before grows ~2.7× per doubling; after grows ~2×. The absolute numbers are noisy (shared machine), the shape is not.

This is why I opened it separately rather than folding it into #2402: that PR gives spans to five warnings that previously had none, which moves them from "no frame" to "one full source scan each", and CodSpeed scored it a **-12% regression on all three `synthetic-state-heavy` benchmarks**. The quadratic is not #2402's — it is reachable on `main` today through any warning that already carries a position, which is most of them.

## Byte-identical output

Not argued, measured. Both implementations were run against the same eight shapes — the ones where slicing by offset can drift from `split('\n')`:

CRLF (the `\r` belongs to the line, not the terminator), multi-byte characters before the warning, the only line, the last line with no trailing newline, the phantom empty line a trailing newline produces, empty context lines, a window clamped at neither end, leading tabs.

`diff` is clean on all eight. As a control that the harness observes this function at all, the old `split` was poisoned to return `"POISON"` per line and the dump changed on 25 lines.

Those eight shapes are pinned in `crates/rsvelte_core/tests/warning_frame_line_index.rs`, with the expected strings taken from the **pre-change** implementation rather than written by hand.

Also green: `--lib` 1508, `compiler_fixtures`, `validator`, `css`, `cargo clippy --all-targets --all-features -- -D warnings`.

https://claude.ai/code/session_01Sb9ucum4Cxu99K1WCeyGC8

### DIFF
--- a/.changeset/warning-frame-line-index.md
+++ b/.changeset/warning-frame-line-index.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Build a warning's code frame from the shared line index instead of splitting the whole source once per warning, so a file with many spanned warnings no longer costs O(source × warnings).
--- a/crates/rsvelte_core/src/compiler/legacy.rs
+++ b/crates/rsvelte_core/src/compiler/legacy.rs
@@ -156,6 +156,24 @@ impl Utf8ToUtf16 {
         // Return column as offset from line start in UTF-16
         abs_utf16_pos.saturating_sub(line_start_utf16)
     }
+
+    /// Number of lines, counted the way JavaScript's `split('\n')` does — a
+    /// trailing newline yields a final empty line.
+    pub fn line_count(&self) -> usize {
+        self.line_starts_byte.len()
+    }
+
+    /// The 0-indexed `line` of `source`, without its terminator. `source` must
+    /// be the string this table was built from.
+    pub fn line_text<'s>(&self, source: &'s str, line: usize) -> &'s str {
+        let start = self.line_starts_byte[line];
+        let end = match self.line_starts_byte.get(line + 1) {
+            // The next line starts one byte past this line's `\n`.
+            Some(&next) => next - 1,
+            None => source.len(),
+        };
+        &source[start..end]
+    }
 }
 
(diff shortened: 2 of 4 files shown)


baseballyama:
Merging with CodSpeed red, for the same reason as #2586.

The one benchmark it names is `parallel_parse::corpus` (−33.61%), which this PR cannot reach: the diff is `compiler/mod.rs`, `compiler/legacy.rs` and a test — `bench_parallel_parse` is phase 1 and never builds a warning frame.

`parallel_parse::corpus` is an **aggregate over `benches/corpus/`**, and #2571 grew that directory from 9 files to 11. Its wall-clock is therefore a different measurand on either side of that commit, and the same report lists the two new fixtures' `🆕 22` benchmarks to prove the corpus moved. `main`'s Benchmark runs are being cancelled by the merge rate, so CodSpeed has no post-#2571 baseline and every PR opened since inherits the same phantom.

Not a reason to keep any of these open, but the gate is currently non-discriminating and should be fixed — either exempt a corpus-aggregate benchmark from the comparison when the corpus changes, or let `main`'s benchmark run finish before the next merge cancels it.

PR #2590, opened 2026-08-07
fix(compiler): treat a trailing `=>` as a statement continuation (client)

baseballyama:
## The defect

A legacy `export let` whose default is an arrow function with the body on the **next line** lost its body, and the emitted module was not parseable JavaScript.

```svelte
<script>
  export let fmt = (n) =>
    `${n} items`;
</script>
```

| | client output |
|---|---|
| official | ``let fmt = $.prop($$props, 'fmt', 8, (n) => `${n} items`);`` |
| rsvelte (before) | `let fmt = $.prop($$props, 'fmt', 8, (n) =>);` ← **syntax error** |

The instance-script loop accumulates source lines until the statement looks complete, and its continuation test at `client/mod.rs` excluded `=>`. So `export let fmt = (n) =>` closed the declaration; the body was re-emitted as a separate top-level statement on the following line.

The `!t.ends_with("=>")` guard it was written as was already dead — `"=>".ends_with('=')` is `false`, since the last char is `>`. The actual bug is the **absence** of a positive clause, so this PR removes the dead guard and adds one.

## Why the server path is the control

`server/transform_legacy.rs:149` and `server/transform_script.rs:6373` already count `=>` as a continuation. The two targets disagreed and the server one was right — confirmed on the repro, where server output already matched official byte-for-byte while client did not. The test asserts **both**, so a fix that regressed the server cannot pass.

## Blast radius

`corpus_hash` over 14,851 `.svelte` files — 19 submodules + `compatibility/pattern-corpus` + four shipped app corpora (carbon-components-svelte, huly/plugins, open-webui, SMUI) — × client-prod / server / client-dev. 144 columns, none empty, all 72 (corpus × target) pairs compared:

- **5 files change, all client-only.** Server and dev-server byte-identical everywhere.
- All 5 previously produced output **esbuild rejects**; all 5 parse now.
- 3 in carbon (`ToolbarBatchActions`, `MultiSelect`, `Pagination`), 2 in open-webui (`ModelMenu`, `AccessControl`) — every one of them has a source line ending in `=>`.
- `ModelMenu.svelte` now matches the official compiler **byte-for-byte**; it did not before.

Across carbon's 287 files, exact-match-vs-official is unchanged at 134/287 and nothing regressed — the fix is strictly additive.

## How this was found

Not by a gate. Every ratchet compares rsvelte against official *after* normalization, so a divergence is scored as a mismatch regardless of whether the output is even valid JavaScript — "wrong text" and "text that cannot be parsed" are the same row. Feeding every corpus output through a JS parser is a different question, and it is the one that surfaced this.


baseballyama:

```
```

baseballyama:
The `Shape matrix parity` failure here is **inherited from main, not caused by this PR**.

Measured with a freshly built binary at three commits — 56 baseline entries already pass at `043943b6` (#2435's own merge commit), at `ddd9c718` (main + #2584), and at main + this PR. Identical set each time. The 56 already passed before this branch existed.

#2595 re-baselines the ratchet (pure shrink, 56 deletions, 0 additions). Once that lands I will rebase this PR and the job should go green.

PR #2591, opened 2026-08-07
test(corpus): gate that the generated output is JavaScript at all

baseballyama:
## The hole

Compiling 10,389 real-world `.svelte` files with rsvelte and feeding each generated module to a
JS parser: **30 produce output no parser accepts.** The official compiler compiles all 30 and
its output parses, so all 30 are rsvelte defects.

Every ratchet in this repo compares rsvelte's text against official's *after normalization* and
scores the result match/mismatch. **"Wrong text" and "text that is not JavaScript" produce the
same row**, so the corpus can sit at zero known failures while the compiler ships unparseable
modules. `AGENTS.md` states the rule this violates: ask what the oracle does not look at, not
only what the input does not contain.

## What this adds

A parse gate inside `verify.mjs`, on its own shrink-only ratchet
`compatibility/parse-known-failures.<target>.json` with a paired `.md`, exactly like the warning
and error families.

**Why inside `verify.mjs` rather than a sibling script.** It already has both output trees, the
manifest, the target descriptors, the artifact lifecycle and the two-sided ratchet machinery in
hand, so the marginal cost is one `Parser.parse` per module and ~50 lines. A sibling would have
to re-derive all of that *and* be sequenced before `verify.mjs` deletes the trees on a green run
— i.e. it would live inside the same CI job anyway, with a second copy of the plumbing. It slots
into the existing `DIAGNOSTIC_FAMILIES` loop, so it inherits the FALSE-SHRINK guard, the
"asked to rewrite and wrote nothing" check, and `--targets` scoping for free.

**Why acorn and not OXC.** rsvelte parses JavaScript with OXC, and both existing "does it parse"
checks in the repo — `ast_equiv_batch` and `tests/ast_gate_preconditions.rs` — re-use OXC. An
acceptance quirk in the parser rsvelte itself depends on is invisible to all of them. acorn is
the parser upstream Svelte uses, is a separate implementation, and is already a root
devDependency.

Three things make it see what the output ratchet cannot:

1. **Its own ratchet.** A ratchet entry suppresses *everything* about its entry. An entry listed
   in `known-failures.<target>.json` for a text mismatch would otherwise absorb a later
   regression to unparseable output silently.
2. **It runs before normalization.** The claim is about what the compiler emitted, not what
   survived `flattenTemplateHoles` + oxfmt. (It is therefore meaningful under `--no-fmt`.)
3. **Its population includes entries official rejected.** `compile.mjs` writes rsvelte's
   `<target>.js` whenever rsvelte compiles, regardless of official's status; those entries are
   `error-mismatch` and the byte comparison never looks at rsvelte's text for them.

Official's output is parsed too, but only as the oracle's own control — a rejection there exits
`2` as a harness misconfiguration and can never become a ratchet entry. And because a gate whose
population is "modules rsvelte produced" gets *greener* the more the compiler refuses to compile
(the failure mode recorded for gate 15a), the run aborts if rsvelte produced fewer than 90% as
many modules as official did. That denominator is official's count, which no rsvelte change can
move, so numerator and denominator cannot shrink together.

## Measured, not estimated

**Calibration.** Compiling 3,509 real-world components (huly, open-webui, carbon, SMUI) with the
**official** compiler across all three targets produced 10,464 modules. acorn under the gate's
options rejected **0**.

**Discrimination.** Of the 30 components for which rsvelte emits output esbuild rejects, acorn
rejects **30**.

**The test can move.** `scripts/dev/test-corpus-parse-gate.mjs` drives `verify.mjs` over a
synthetic corpus. Two of its cases are chosen so a plausible-but-wrong implementation fails
them, and both were confirmed by running the suite against a deliberately mutated `verify.mjs`:

| Mutant | Case that flipped |
|---|---|
| `parseBaseline` → `baseline` (fold into the output ratchet) | "the OUTPUT ratchet does not suppress it" — exit 1 → exit 0 |
| skip entries where official produced no output | "caught where OFFICIAL rejected the input" — exit 1 → exit 0 |

It also asserts the ratchet is two-sided, that an unparseable *official* output exits 2 rather
than being ratcheted, that a collapsed population exits 2, and that `--update-parse-baseline`
rewrites only its own family.

## Baseline: 0, and why that is the expected number

The corpus output ratchets are all at 0 today, and an unparseable rsvelte output is necessarily
byte-different from official's parseable one — so it would already have surfaced as
`js-unparseable` and been listed there. The 30 known defects are **not** in this ratchet because
huly / open-webui / carbon / SMUI are not corpus sources (`corpus-sources.json`).

Stated plainly, and repeated in `parse-known-failures.md`: **this is a regression gate for the
class, not a burn-down of it.** It closes gate-coverage blind spots 15d (wrong population) and
the oracle half of 15b. Enrolling those repositories is what would change the number, and corpus
size is the axis `AGENTS.md` calls saturated — which is why this PR grows *what we compare*
instead.

The 30 files are being fixed in parallel; they will re-baseline here if any of them turn out to
be reachable from a corpus source.

## The gate found something on its first CI run

acorn rejected **2 OFFICIAL modules**: it enforces the duplicate-lexical-declaration early
error, and official's client codegen for
`compiler-errors/samples/const-tag-snippet-invalid-reference-1` declares `foo` twice in one
scope. The gate exited 2 with the "oracle, not rsvelte" message rather than blaming the
compiler — which is the control doing its job.

It also falsifies my calibration's scope. The options were tuned on 10,464 modules from
real-world components; this corpus also contains Svelte's **deliberately-invalid fixtures**,
which is exactly where a compiler is most likely to emit something a strict parser refuses. A
calibration corpus reproducing a measurement only shows the method is sound on *its* population.

Widening `OPTIONS` would weaken the oracle for ~42,000 modules to accommodate two, so instead
those `(id, target)` pairs go on `compatibility/parse-oracle-excluded.json` and are skipped on
**both** sides: where the reference does not parse there is no claim to make about rsvelte.
Shrink-only in both directions — an unlisted oracle rejection exits 2, and a listed pair whose
official output now parses also exits 2.

That opens a hole (an excluded pair is checked on neither side), recorded as blind spot **19f**
and pinned by a test: the excluded pair is seeded with an unparseable *rsvelte* output too, and
the run must be green **and** must not count the pair in the parsed population.

## Blind spots (`compatibility/gate-coverage.md` § 19)

- **19a** — says nothing about whether the output is *correct*; the verdict is one boolean.
- **19b** — JS only. CSS is never handed to a CSS parser; `js.map` is not captured at all.
- **19c** — the population is inherited from the corpus, and the known defects are not in it.
- **19d** — one parser, so a construct acorn and a real engine disagree on is unobservable.
- **19e** — `--targets` narrows it silently; the FALSE-SHRINK guard counts entries, not
  entries × targets. Inherited from the existing diagnostic families, not introduced here.
- **19f** — an excluded pair is checked on neither side (2 pairs today).

## Not included

No changeset: harness-only, no `@rsvelte/compiler` behaviour change.


baseballyama:

```
```

PR #2592, opened 2026-08-07
fix(compiler): end a destructure assignment's RHS at the line break

baseballyama:
## What

The client lowers a destructuring assignment `[a] = rhs` into an IIFE, and found the end of `rhs` in `find_destructure_rhs_end` by scanning for `;`, `,` or an unbalanced closer. Semicolon-free source has none of those after the RHS, so the scan ran on through every statement that followed it and the emitted call was never closed.

## Minimal repro

```svelte
<script>
  export let selected

  function pick (result) {
    ;[selected] = result
    done = true
  }

  let done = false
</script>

<button onclick={() => pick([1])}>{selected}{done}</button>
```

Client output before (esbuild: `Expected ")" but found "done"`):

```js
	    ;(($$value) => {
		var $$array = $.to_array($$value, 1);
		selected = $$array[0];
		return $$value;
	})(result
	    done = true)}
```

After, the RHS ends at the line break and the rest of the function stays where it was — and the whole instance script now lowers through the same path the official compiler uses:

```js
	function pick(result) {
		((result) => {
			var $$array = $.to_array(result, 1);
			selected($$array[0]);
			return result;
		})(result);

		$.set(done, true);
	}
```

## The rule

A line break ends the RHS by the same ASI rule the source is relying on: it ends the expression unless the line ends with an operator (`= + && ?` …) or the next line opens with a token that continues it (`.`, `?`, `:`, `(`, `[`, an operator, …). The leading `;` these files write before a line starting with `(`/`[` exists for exactly this reason.

## Tests

`crates/rsvelte_core/tests/destructure_assignment_line_break.rs`, four cases, all parsing the generated module with `oxc_parser` — a compiler may emit output we would call wrong, it may never emit output that is not JavaScript.

Run against the unpatched compiler, the three client cases fail with the predicted diagnostic (`Expected `,` or `)` but found `Identifier``, i.e. `done = true` swallowed into the IIFE call argument list); the server case passes.

Controls that a wrong-but-plausible fix would break:

- **server_lowers_the_same_assignment_unchanged** — the server lowers this assignment without the scan and was already right. A fix that reached into the shared lowering would fail here.
- **a_line_break_inside_the_rhs_is_not_the_end_of_it** — "cut at the first newline" truncates a multi-line call argument list.
- **a_leading_dot_on_the_next_line_continues_the_rhs** — "cut at the first newline" silently drops `.filter(Boolean)` from the value while still emitting *parseable* JavaScript, so the parse assertion alone would not catch it.

## Measured

10,389 real-world `.svelte` files — 18 submodules + `compatibility/pattern-corpus` + carbon / huly / open-webui / smui — compiled with `corpus_hash` before and after, for client-prod, server and client-dev:

|---|---|---|
| client | 10389 | 17 |
| server | 10389 | **0** |
| client-dev | 10389 | 17 |

The 17 are the same set on client and client-dev, and are exactly the 17 files in this corpus whose client output esbuild previously rejected with `Expected ")" but found <token>`. All 17 now parse on all three targets. Every other file is byte-identical.


baseballyama:
`Shape matrix parity` fails here with **`56 baseline entries already PASS — the ratchet is stale`**, not with a new mismatch. Every one of the 56 is a `comment-slot/module-*` case on the **server** target, i.e. the `.svelte.(js|ts)` module cases enrolled by #2435 — the same staleness #2593 is re-baselining. It is not this change: this change is client-only, and the server output is byte-identical over the 10,389 corpus files above **and** over `submodules/svelte` + `submodules/svelte.dev` (40,696 further walked entries, 0 changed on client, server and client-dev alike).

`Compiler parity` — the corpus output-equality gate — passes.

I will rebase once #2593 lands.

baseballyama:

```
```

baseballyama:
`CodSpeed Performance Analysis` fails with **`Performance Regression: -0.87%`**, from a single benchmark: `crates/rsvelte_formatter/benches/formatter.rs :: svelte[synthetic-markup-heavy]`, 38.9 ms → 43.9 ms.

**That benchmark cannot execute this diff.** The only function this PR changes is `find_destructure_rhs_end`, which is `pub(super)` inside `phases/3_transform/client/` and is reachable only from the client code generator. `rsvelte_formatter` imports `rsvelte_core::{ParseOptions, parse}` and AST types — it never calls `compile`, and never runs phase 3 at all. So there is no execution path from a formatter benchmark into the changed code.

Three further signals in CodSpeed's own report point the same way:

- the second changed benchmark is `rsvelte_devtools/benches/parser.rs::corpus`, **+10.86% faster** — also code this PR does not touch, and a swing of the same magnitude in the opposite direction;
- CodSpeed flags *"Different runtime environments detected — some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results"*;
- the comparison base is `043943b`, not main's `ddd9c71` — *"No successful run was found on main during the generation of this report"* — so the report itself warns it may contain changes unrelated to this PR.

Two unrelated benchmarks moving ±11% in opposite directions, neither of which runs the changed function, is the signature of code-layout perturbation rather than a real cost.

I have re-run the benchmark job to get a second sample and will report whether it reproduces. If it does not, this is noise; if it does, I will bisect it rather than acknowledge it away.

baseballyama:
The `CodSpeed Performance Analysis` failure here is not attributable to this PR.

The single regressed benchmark is

```
crates/rsvelte_formatter/benches/formatter.rs::bench_format::format::svelte[synthetic-markup-heavy]
38.9 ms → 43.9 ms  (-11.35%)
```

and this PR changes exactly one non-test file: `3_transform/client/destructure_transforms.rs`. The formatter benchmark formats a Svelte source — it never runs phase-3 client codegen, so there is no path from the change to the measurement. The report also flags **"Different runtime environments detected"** itself, and the other moved row (`parallel_parse::corpus`, +10.86%) is equally unreachable from a client-transform edit.

Worth being precise about what that does and does not license: it makes the number **non-discriminating for this PR**, not "CodSpeed is broken". The base for this comparison is `043943b`, a real post-#2571 `main` run, so the earlier problem — no successful `main` baseline at all, which had every PR reading −25% to −34% on `corpus` from a 9.5 ms base — has resolved itself. #2594 keeps it resolved by stopping merges to `main` from cancelling each other's benchmark runs.

`Shape matrix parity` here is the inherited `main` failure (#2593).

baseballyama:
**CodSpeed: resolved, and it was not noise.** Reporting the whole sequence since I said I would either way.

The regression *reproduced* on a second, independent run of the benchmark job — `svelte[synthetic-markup-heavy]` 38.9 ms → 43.9 ms, then 38.6 ms → 44.4 ms. Two samples, same direction, ~+5.3 ms both times. In Simulation mode that is an instruction-count measurement, so "the runner was noisy" was not available as an explanation.

So I bisected rather than leaning on the structural argument. The cause was the **comparison base**, not the diff. CodSpeed's own report said it: *"No successful run was found on `main` (ddd9c71) during the generation of this report, so 043943b was used instead."* This branch was cut at `043943b6`; `main` had since gained, among others, [#2560](https://github.com/baseballyama/rsvelte/pull/2560) — `fix(parse): trim the template's trailing whitespace with the JS whitespace set`. The failing benchmark formats markup, and `rsvelte_formatter` reaches `rsvelte_core::parse` (and only `parse` — it never calls `compile`). So the branch was being compared against a baseline containing a parse change it did not have, on the one benchmark that measures parsing markup.

Merging `origin/main` in — diff unchanged, base current — and the gate **passes**. Diff held constant, base varied, signal disappeared.

Two things I got wrong along the way and want on the record: I first called this "the signature of code-layout perturbation", which the replication falsified; and I could not show the benchmark was chronically noisy (the two comparable PRs I sampled, #2590 and #2591, both passed), so that argument never had the empirical companion I implied it might.

`Shape matrix parity` also passes now that #2593 has landed, and `Compiler parity` passes with the new `pattern-corpus` entry included.

PR #2593, opened 2026-08-07
chore(corpus): re-baseline the shape matrix ratchet that #2435 enrolled against a stale merge base

baseballyama:
`main` is currently **red** on `Shape matrix parity (generated inputs)`, and has been since #2435 merged. Reproduced on `origin/main` at `ddd9c718` with a locally built binding:

```
```

So this is not a defect in any of the PRs that are failing on it (#2574, #2590, and every branch cut from `main` since) — they inherit it.

## Cause

#2435 added the `.svelte.(js|ts)` module cases and baselined 128 module entries. Its baseline was measured on a branch cut **before** #2566 (`fix: drop the comments a server .svelte.(js|ts) module cannot own`), which had already fixed 56 of them. #2435 merged with 56 entries that pass on the merged tree.

The run that would have caught it is `Corpus Compat` on `main` — every one of the last eight was **`cancelled`**, killed by the merge rate rather than failing:

```
31220338308 completed cancelled | test(corpus): generate .svelte.(js|ts) module cases … (#2435)
31220015778 completed cancelled | perf(ast): shrink JsNode from 144 to 80 bytes (#2585)
31219637556 completed cancelled | perf(compiler): build a warning frame … (#2587)
…
```

That is the same failure mode that made CodSpeed non-discriminating earlier in this queue: **a gate whose baseline run never completes cannot report anything, and the silence is indistinguishable from green.** Worth a follow-up on the concurrency group for `main`; not fixed here.

## What this PR does

`--update-baseline` on a full, formatted, all-families run: **290 → 234**. All 56 removals are `comment-slot` module-path entries on `server`.

The module block's justification is **re-measured, not subtracted**, per the doc's own rule — and the measurement corrects the previous framing:

| | was | now |
|---|---:|---:|
| `binding-position` | 2 | 2 |
| `comment-slot` → `.svelte` template seeds | 160 | 160 |
| `comment-slot` → module path | 128 | **72** |
| | 290 | 234 |

- The old text said "every one of these 128 is the same open bug #2399". That was wrong for 56 of them: those were rsvelte's own #2307 defect, fixed in #2566.
- The 72 that remain are **one slot per seed** — `module-rune-exports` L08, `module-class-state` L11, `module-ts-extension` L05, each the line *after* that seed's last statement — × 8 comment kinds × 3 targets. The diverging case set is **identical** on `client`, `server` and `client-dev`, so it is one target-independent rule, not three defects. Re-running the doc's mechanical classifier over the gate's artifacts puts all 72 in the "official drops, rsvelte keeps" bucket and **zero** in the other two.
- So the residue is not "Program-level comments" in general but a **comment trailing the end of the module** — the same shape as the `.svelte` server residual documented two sections above.
- The old claim that official "drops in **80** of 192 and preserves in 112" was that same server-only entry count read as a property of the official compiler. The honest number is **24 of 192 positions**.
- Separately, "Server still dominates (82 of these 160)" is corrected to **80**; no entry count supported 82, and the template entries are untouched by this change.

## Verification

```
node scripts/compat-corpus/matrix/run.mjs      → ✅ no regressions (234 known divergences remain)
                                               → 25 declared ratchets across 19 docs match their JSON
```

Section counts partition the total: 2 + (160 + 72) = 234.


baseballyama:
Re-ran the `Changeset` job: the `skip-changeset` label was applied after that run started, so it captured `SKIP: false`. Nothing to change on your side.

Flagging that this PR is the blocker for three others — #2590, #2592, #2596 all show the same red `Shape matrix parity` inherited from the stale entries you are re-baselining here. I'll rebase all three once this lands.

PR #2594, opened 2026-08-07
ci: stop each merge to main from cancelling the previous merge's verdict

baseballyama:
Follow-up to #2593, which fixes the symptom. This fixes the reason nobody saw it.

## What is broken

Every gate sets `cancel-in-progress: true` on a group keyed by `github.ref`:

```yaml
group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
cancel-in-progress: true
```

On a `pull_request`, `head_ref` is the PR branch, so this correctly collapses to one run per branch. On a **push to `main`**, `head_ref` is empty and `github.ref` is the same string — `refs/heads/main` — for *every* merge. So each merge cancels the previous merge's run.

At the merge rate this queue has been running at, that means `main` is never verified. All eight of its most recent `Corpus Compat` runs are `cancelled`:

```
31220338308 completed cancelled | test(corpus): generate .svelte.(js|ts) module cases … (#2435)
31220015778 completed cancelled | perf(ast): shrink JsNode from 144 to 80 bytes (#2585)
31219637556 completed cancelled | perf(compiler): build a warning frame from the shared line index (#2587)
31219325748 completed cancelled | fix(transform): stop a comment apostrophe from suppressing … (#2586)
31219086973 completed cancelled | fix: drop the comments a server `.svelte.(js|ts)` module cannot own (#2566)
31218661268 completed cancelled | test(compiler): pin the module comments upstream keeps … (#2543)
31218650492 completed cancelled | fix(compiler): skip comments when balancing brackets … (#2550)
```

**A cancelled run and a green run are indistinguishable in the branch header.** The gate was written, wired and passing on PRs, and reported nothing on the branch it exists to protect.

Two concrete costs, both already paid in this queue:

- **#2435 landed a stale ratchet.** Its baseline was cut before #2566, so 56 module entries it enrolled already passed on the merged tree. The shape-matrix gate went red *on `main`*, and stayed red until PRs #2402/#2464/#2574/#2590 inherited the failure and it got diagnosed by hand. #2593 re-baselines it.
- **CodSpeed had no baseline.** It reported "no successful run on main, using 5f6de88 instead" and measured every PR against a commit predating #2571's corpus change, producing −25% to −34% on `parallel_parse::corpus` for PRs that cannot touch phase 1. Three PRs were merged over that noise after proving it non-discriminating.

## The change

One line in `ci.yml`, `corpus-compat.yml`, `benchmark.yml` and `coverage.yml`:

```yaml
cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

Pull requests are **unchanged** — a new push to a PR still cancels the old run.

### Why not a unique group per merge commit

`group: …-${{ github.head_ref || github.sha }}` also gives every commit a verdict, but it starts a full `Corpus Compat` per merge *in parallel*, which at this merge rate would starve the PR queue of runners. Disabling cancellation instead keeps `main` at one concurrent run: GitHub holds at most **one pending run per group** and cancels an older *pending* one when a newer arrives, so the in-progress run always finishes and the tip is always the one scheduled next. The cost is a lag between the tip and its verdict, not a runner per commit — and a lag of one cycle would still have caught #2435.

Left alone deliberately: the `release-*` and `publish-*` workflows already use `cancel-in-progress: false` with hand-written group names, and `deploy-docs`'s `pages-build` group is *supposed* to collapse.

`AGENTS.md` gains the two rules this cost us: re-baseline **after** rebasing onto `main`, never before; and a cancelled run is not a green run.


baseballyama:

```
```

PR #2595, opened 2026-08-07
chore(corpus): re-baseline the shape matrix ratchet (290 -> 234)

baseballyama:
## Why

`Shape matrix parity` is **red on main**. 56 listed entries already pass, and the ratchet is two-sided by design, so every PR based on current main inherits the failure. #2590 hit it first.

## Attribution — measured, not assumed

I ran the matrix at three commits, each with a freshly built binary (the staged `.corpus-cache/rsvelte.node` was two days stale and silently reports garbage — that cost me two bogus runs first):

| commit | result |
|---|---|
| `043943b6` — #2435's own merge commit | 56 baseline entries already PASS |
| `ddd9c718` — main with #2584 | 56 baseline entries already PASS |
| main + #2590 | 56 baseline entries already PASS |

The 56 already pass **at the commit that added them**, before #2584. So:

- **#2584's byte-identity claim stands** — it did not cause this.
- **Nothing was fixed to clear these.** They were listed as failing when they were not.
- #2435's CI could not see it: it ran against a merge ref whose `main` half predated a fix that the merge commit itself contains. That is precisely the gap a two-sided ratchet exists to catch on the next PR, and it did.

CI's log and my macOS run name the **identical 56 ids**, so this is not the platform sensitivity that governs the fmt corpus.

## What changed

Pure shrink — 56 deletions, 0 additions — from a full run (no `--no-fmt`, no `--families` subset; the tool refuses both for the FALSE-SHRINK reason).

All 56 are on `server`: 32 `module-rune-exports`, 16 `module-ts-extension`, 8 `module-class-state`.

The paired `.md` is brought back into agreement with the JSON, since the counts in it were load-bearing for #2399's framing: header 290→234, `comment-slot` 288→232, module path 128→72, a by-seed table for the three `.svelte.(js|ts)` seeds, and an explicit note that nothing was fixed to clear these so nobody reads the shrink as progress. Verified against the JSON: template seeds 160 + module path 72 + `binding-position` 2 = 234.


baseballyama:
Duplicate of #2593 (opened ~7 min earlier) — closing this one, but the measurement is the more valuable half of it.

**The two JSON baselines are byte-identical**, reached independently on two machines from separately built bindings:

```
git diff fix/matrix-ratchet-rebaseline chore/matrix-rebaseline -- compatibility/matrix-known-failures.json
(no output)
```

That is a stronger result than either PR could claim alone, and it rules out the platform sensitivity that governs the fmt corpus.

Your per-seed breakdown of the 56 (32 `module-rune-exports`, 16 `module-ts-extension`, 8 `module-class-state`) and the per-seed table are folded into #2593, along with one thing your framing gets right that mine understated: **nothing was fixed to clear these**, so the shrink must not be read as progress.

Two corrections #2593 carries that this PR's doc keeps as-is, both checkable against the JSON:

- **`Server still dominates (82 of these 160)`** — the template entries split 40 / 40 / **80** by target. No entry count supports 82, and this PR does not touch the template entries either way.
- **`official drops in 80 of 192 and preserves in 112`** — that 80 *is* the server-only entry count, read back as a property of the official compiler. The three seeds have 8 + 11 + 5 = **24 insertion slots**, so 24 × 8 kinds = 192 cases, and official drops in **3 of the 24 slots** (the trailing one of each seed) = 24 of 192 cases. The residue is not "Program-level comments" in general but *a comment trailing the end of the module* — the same shape as the `.svelte` server residual documented one section above.

Also relevant to your `.corpus-cache/rsvelte.node` note: #2594 fixes why none of this was caught on `main` — every gate set `cancel-in-progress: true` on a group keyed by `github.ref`, which is the same string for every push to `main`, so main's last eight Corpus Compat runs are all `cancelled` and the branch has carried no verdict at all.

PR #2596, opened 2026-08-07
fix(compiler): strip TypeScript through an OXC-only parse diagnostic

baseballyama:
## What

`strip_typescript` returned the source **unstripped** whenever OXC produced any diagnostic. OXC enforces TypeScript rules the official parser does not — "a required parameter cannot follow an optional parameter", "type annotation cannot appear on a constructor declaration" — and reports them *while parsing*, even though it built a complete AST. A component carrying one kept its type annotations.

## Minimal repro

```svelte
<script lang="ts">
	export let overlay = false;

	const f = (a: string, b?: string, c: string) => {
		return a + (b ?? '') + c;
	};
</script>

<p>{f('x')}{overlay}</p>
```

`c: string` is required and follows the optional `b` — a TypeScript *type* rule; it is not a syntax error, and `acorn-typescript` does not enforce it. Two symptoms:

**client** — annotations go straight into the module (esbuild: `Expected ")" but found ":"`):

```js
	const f = (a: string, b?: string, c: string) => {
```

and with `export let n: number = 0` the declared type ends up inside the prop *name*: `$.prop($$props, 'n: number', 0, 0)`.

**server** — the script lowering parses the stripped text as JavaScript and returns no statements when that fails, so the component silently loses its **entire instance script** while still emitting parseable output:

```js
export default function A($$renderer, $$props) {
	$$renderer.push(`<p>${$.escape(f('x'))}${$.escape(overlay)}</p>`);
	$.bind_props($$props, { overlay });
}
```

Both now match the official compiler: `const f = (a, b, c) => {`.

## The rule

The bail is on `panicked` — a stub AST, from which nothing can be stripped. Every malformed script I tried panics (`const a: Foo<Bar = 1`, `let a: = 1`, `const a = 1 as;`, `function f(a,, b)`, `export type = 2`), so that alone would do. The strip result is *additionally* re-parsed as JavaScript before a diagnostic-bearing parse is trusted, because a recovered AST can also carry TypeScript this stripper does not lower — a constructor parameter property — and that case must keep returning the source rather than emit a new broken shape.

## Tests

`crates/rsvelte_core/tests/typescript_strip_recoverable_parse_error.rs` (3 cases, each parsing the generated module with `oxc_parser`) plus two `strip_typescript` unit tests. Run against the unpatched compiler the three integration cases fail with exactly the predicted output: `(a: string, b?: string, c: string)` in the client module, and a server module with no instance script.

Controls that a wrong-but-plausible fix would break:

- **a_different_oxc_only_diagnostic_also_strips** — a return type on a constructor is a *different* OXC-only rule and must strip identically, so recognising the one diagnostic that appears in the corpus does not pass. (Fails unpatched, on the `'n: number'` prop name.)
- **a_strip_that_would_not_yield_javascript_is_refused** — the control for the validation re-parse, and it can move: this input parses without panicking *and* carries a diagnostic, and its strip leaves `constructor(private a)` behind. Gating on `panicked` alone would emit that non-JavaScript into the module; the test pins the verbatim source instead.

## Measured

10,389 real-world `.svelte` files — 18 submodules + `compatibility/pattern-corpus` + carbon / huly / open-webui / smui — hashed before and after for client-prod, server and client-dev:

|---|---|---|
| client | 10389 | 1 |
| server | 10389 | 1 |
| client-dev | 10389 | 1 |

The same single file on all three (`open-webui/src/lib/components/chat/ChatControls/Embeds.svelte`), which is the only file in the corpus that hits this, and which goes from output esbuild rejects to output it accepts. Every other file is byte-identical on all three targets.


baseballyama:
`Shape matrix parity` fails here with **`56 baseline entries already PASS — the ratchet is stale`**, not with a new mismatch — the same staleness #2593 is re-baselining, from the `.svelte.(js|ts)` module cases #2435 enrolled.

Evidence that it is not this change: the matrix tallies are **identical** on this PR and on #2592, which is a completely different change to a different file — `match 2337 / error-parity 0 / js-mismatch 234 / error-mismatch 0`, and the same 56 entries, all `comment-slot/module-*` on the **server** target. Two unrelated diffs cannot produce the same result unless the result is independent of both.

Corroborating: over the 10,389 corpus files in the description exactly one file changes on each target, and over `submodules/svelte` + `submodules/svelte.dev` (40,696 further walked entries) nothing changes on client, server or client-dev.

I will rebase once #2593 lands.

baseballyama:

```
```

PR #2599, opened 2026-08-07
fix(compiler): keep an `else` that starts its own line attached to its `if`

baseballyama:
## Minimal repro

```svelte
<script>
	let a = 1;
	let b = 0;
	$: if (a > 0) b = 1
	else b = 2
</script>

<p>{b}</p>
```

Client output (before):

```js
export default function A($$anchor, $$props) {
	$.push($$props, false);

	else b = 2                       // <- Unexpected "else"

		if (a > 0) b = 1;
	});
	…
}
```

The braced form with `else` on its own line (`}` newline `else {`) is broken the same way.

## Mechanism

The client instance-script line accumulator closes a statement as soon as the brace/paren depths
balance. `$: if (a > 0) b = 1` balances on its own line — its consequent is brace-less and sits
on the same line — so the statement is emitted and the `else` line starts a new one.

`transform_instance_script_for_visitors` already has a lookahead for exactly this shape: if the
next non-empty line starts with `.`, `?`, `:`, `&&`, `||` or `??`, the expression continues.
`else` belongs in that list and was missing. It is unconditionally safe there because **no
JavaScript statement may begin with `else`** — a line that does is always the tail of the `if`
above it.

The existing `ends_with_braceless_control_header` path handles the *other* half of this shape
(`$: if (cond)` with the body on the next line); it cannot help here because the `if` is already
complete.

## Predicted failure, then checked

Written against unpatched code first. The two witnesses fail with the predicted message
(`an else was emitted at statement position`); the two controls pass:

| test | unpatched |
|---|---|
| `a_braceless_else_on_the_next_line_stays_with_its_if` | FAIL (witness) |
| `a_braced_else_on_its_own_line_stays_with_its_if` | FAIL (witness) |
| `the_server_path_is_unaffected` | pass (control) |
| `an_identifier_beginning_with_else_is_not_a_continuation` | pass (control) |

**Controls that can move.** The server path does not share this accumulator and was already
correct. The second is the reason `starts_with_else_keyword` exists rather than
`starts_with("else")`: a line beginning `elsewhere = 2` is a separate statement, and a prefix
match would glue it onto the one above.

## Byte-identity gate

`corpus_hash` over carbon/src, huly/plugins, open-webui/src, smui and
`compatibility/pattern-corpus` × client / server / client-dev — **10,830 outputs**, measured
against the same tree without this change.

| corpus | client | server | client-dev |
|---|---|---|---|
| carbon (287) | 0 | 0 | 0 |
| huly (2123) | 1 changed | 0 | 1 changed |
| open-webui (650) | 0 | 0 | 0 |
| smui (449) | 0 | 0 | 0 |
| pattern-corpus (101) | 0 | 0 | 0 |

**2 of 10,830 outputs changed; the other 10,828 are byte-identical.** Both are
`huly/…/setting-resources/src/components/Settings.svelte` — the file this fixes. Server is
untouched, as predicted.

That file moves from unparseable to **byte-identical to official**, measured as the line-multiset
distance to the official compiler's output:

| target | before | after |
|---|---|---|
| client | 289 | **0** |
| client-dev | 311 | **0** |

## Corpus-wide effect

Re-running the parse check over the 30 known-bad components × 3 targets, this change moves
`huly/…/setting-resources/src/components/Settings.svelte` from unparseable to parsing on client
and client-dev.


baseballyama:

```
```

baseballyama:
### Where the fix lives (for the in-flight `collect_vars` / `line_loop` rewrite)

This fix is **one clause in `next_continues`**, the lookahead that asks whether the
*next* line continues the current statement:

- `crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs` — `starts_with_else_keyword(future_trimmed)`

It is deliberately the sibling of the clause `#2590` added to **`trailing_operator`**,
which asks whether the *current* line ends mid-statement:

- same file — `t.ends_with("=>")`

The two answer different halves of the same question, so a rewrite of
`collect_vars` / `line_loop` needs to carry **both** forward. Merging `main` here
auto-merged the file without a textual conflict, which says only that the hunks did
not overlap — both clauses were re-confirmed by reading the merged file.

### What pins each clause

Two independent suites, one per clause, each of which fails if its clause is dropped:

| suite | clause it pins |
|---|---|
| `crates/rsvelte_core/tests/reactive_else_on_next_line.rs` | `next_continues` / leading `else` |
| `crates/rsvelte_core/tests/export_let_arrow_line_break.rs` | `trailing_operator` / trailing `=>` |

They are two separate inputs with two separate controls on purpose. A single combined
test would pass with either rule missing, and would not tell a later refactor which
rule it broke.

PR #2602, opened 2026-08-07
perf(transform): measure the SemanticBuilder share of compile()

baseballyama:
## Why

#2588 answered *which* pass builds a `Semantic` and how often, but not *how
much of `compile()` that is* — and the consolidation go/no-go rests on the
second question. This PR measures it, and the answer is a no-go.

## What

`profile::semantic_build` folds the existing per-site count with an optional
timer around the build. The timer is gated behind a new
`measure-semantic-build` feature: the counts are asked for on every compile, the
`Instant` pair is asked for once. Default builds are unchanged.

`semantic_sites --time` reports the share of `compile()` wall time. Numerator
and denominator are collected in the same pass so both see the same machine
load; the pass with the smallest total is reported and the spread across passes
is printed, so a number that should be distrusted says so.

## Result: share of `compile()`

`--reps 7`, two independent runs for `client-prod`, machine at load 22–35
throughout. The two prod runs are shown as `a / b`.

| corpus | builds/file | client-prod | client-dev |
|---|---|---|---|
| svelthree | 240.48 | 3.14 / 3.34 % | 1.56 % |
| smelte | 7.73 | 2.00 / 1.96 % | 1.74 % |
| powertable | 8.07 | 1.88 / 1.92 % | 1.59 % |
| attractions | 7.51 | 1.91 / 1.84 % | 1.60 % |
| sveltestrap | 8.03 | 1.50 / 1.47 % | 1.18 % |
| svelte-calendar | 4.71 | 1.15 / 1.19 % | 1.07 % |
| svelte-ux | 6.10 | 1.13 / 1.24 % | 0.62 % |
| layercake | 1.57 | 0.24 / 0.23 % | 0.11 % |
| flowbite-svelte | 0.09 | 0.28 / 0.21 % | 0.38 % |
| bits-ui | 0.02 | 0.10 / 0.11 % | 0.05 % |
| **shadcn-svelte** (control) | 0.02 | **0.08 / 0.08 %** | **0.03 %** |

Every corpus reproduces between the two prod runs to within 0.2 points.

**Negative control holds.** shadcn-svelte, at 0.02 builds/file, reads 0.08 %
prod and 0.03 % dev — the instrument is measuring the builds and not something
that happens to scale with compile time.

**Mean cost per build** is 0.9–2.4 µs on the per-statement sites. The timer pair
is ~100 ns, so the numerator is inflated by roughly 10 % at most — and that bias
points *toward* consolidation being worth it, which is the safe direction for a
no-go.

## Result: dev is not this mechanism

Dev was the worst target in earlier profiling (6.59×) and it is 2–3× slower here
too (svelthree 194 → 567 ms, shadcn 510 → 1166 ms). But the **build count is
identical** between prod and dev — 5050 on svelthree in both — so the timer share
*falls* in dev: the extra dev cost is entirely somewhere other than these builds.
Dev being the worse target is true and is an argument against this particular
consolidation, not for it.

## Go / no-go: **no-go, and now final**

The ceiling is 3.3 % at 240 builds/file — twelve times carbon's 19.38/file rate.
Interpolating the curve, carbon sits near 2 %. Consolidation cannot take that to
zero, because the passes still need *a* `Semantic`; halving the builds saves
roughly 1 % on app-code corpora and 0.05 % on component libraries. That does not
justify sharing one `Semantic` across passes that each own a thread-local
`Allocator`, use different `ParseOptions`, and mutate the text between them.

The attribution and the timer are the durable output. They cost nothing when
off, and they mean the next person asking this question measures it in one
command instead of rediscovering the 14 sites.

## Verification

- **Byte-identity**: `corpus_hash` at this branch vs `origin/main`, 17
  repositories + `compatibility/pattern-corpus` × client / server / dev —
  **134 421 outputs, zero differing hashes**.
  Columns were checked for silent duplication and emptiness before the
  comparison was read: 51 of 54 carried data and were mutually distinct; the 3
  `runed` columns were empty because that submodule is not checked out in this
  worktree, and are excluded rather than counted as agreement.
- `cargo test -p rsvelte_core` (debug, `RUST_MIN_STACK=33554432`): 256 suites,
  **2709 passed, 0 failed, 8 ignored**.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean — which
  compiles the `measure-semantic-build` path too.
- `cargo fmt --all`: clean.


baseballyama:
Relaying the context this PR settles, since the measurement here closes a thread that has been open across several PRs.

The SemanticBuilder consolidation question started from a number I got wrong: I reported `oxc_ast_visit::walk_*` at ~15-16% of samples and sized the work against it. That was a symbolication artifact — 28.9% of samples live in other libraries and my aggregator resolved every address through `nm` on the main binary, aliasing them onto whatever symbol preceded them. Corrected, `oxc_ast_visit` totals 3.06% and `walk_function` is absent from the top 35.

I then challenged the resulting no-go on the grounds that 64.3% of 9,388 shipped files are legacy. That was also wrong: the census shows legacy syntax does not predict the build rate at all — flowbite-svelte is 56% legacy at 0.09 builds/file while carbon is 100% legacy at 19.38. The split is app code vs component library.

This PR is the number neither of those arguments had: **~3.3% ceiling at 240 builds/file, which is 12x carbon's rate**, with shadcn holding at 0.08% as a negative control that the instrument measures the builds rather than anything scaling with compile time. carbon interpolates to ~2%.

The dev column is the part I'd underline. I asked for it expecting it to strengthen the case, since dev is the worse target overall. It does the opposite — the build count is identical between modes (5050 on svelthree either way), so the share *falls* in dev and dev's extra cost is entirely elsewhere. A column requested to support a hypothesis that instead argues against it is worth more than the ones that agree.

No-go accepted as final. The per-site attribution (#2588) and this instrumentation are the durable output.

PR #2611, opened 2026-08-08
refactor(compiler): answer the control-header probe from the last line

baseballyama:
## What

Two things, both inside the instance-script line loop in `3_transform/client/mod.rs`.

**1. Load-independent counters for the text machinery.** `compile_profile` grows a `COUNTERS` row:
lines walked, control-header probe calls and bytes materialised, fast-pathed statements, whole-script
scan passes and bytes in `collect_vars`.

This is not decoration. Wall-clock attribution is currently unusable on a loaded box — two runs of the
same binary on the same input put `line_scan` at **2.08ms and 26.84ms** (13x). Every counter below was
bit-identical across the same pair of runs. A rewrite here can be byte-identical, green, and a complete
no-op; the counter is what rules that out.

**2. The brace-less-control-header probe now answers from the last line.**

`ends_with_braceless_control_header` only ever reads a *suffix* — `else` / `do`, or a `)` whose matching
`(` it scans back to. The caller was joining the entire accumulated statement to ask it, at **every line
where the depth counters read balanced**, and computing it eagerly even when a trailing comma or
operator already settled the boundary.

Now: evaluate lazily, and answer from the final line whenever that line settles it. The join survives
only for the case it exists for — a header whose `(` opens on an earlier line, or a last line that is
empty after comment stripping.

## Counter evidence

Bytes materialised for the probe, whole corpus, output unchanged:

| corpus | before | after |
|---|---:|---:|
| flowbite-svelte | 605,481 | **0** |
| bits-ui | 213,319 | 270 |
| svelte-ux | 246,676 | 1,233 |
| smelte | 175,988 | 730 |

Probe calls also fall slightly (laziness): smelte 2020 → 2018, svelte-ux 2872 → 2813, bits-ui 1524 → 1502.

## Why it is byte-identical

`braceless_control_header_from_last_line` returns `None` — deferring to the joined path — in exactly the
two cases where earlier lines can change the answer:

- the last line is empty after trimming, or
- it ends with `)` whose matching `(` is not on this line, or is at the very start of it (so the keyword
  before the paren could span the line break).

Otherwise the verdict provably depends only on the final line: a keyword match that would span the join
would have to include the `\n`, and `ends_with_keyword` cannot match across it.

## Sizing note

The brief that prompted this sized `line_loop` at 22.7% of `script_text_transform`. That timer brackets
`process_accumulated` too — `compile_profile` already prints the split, and the character scanner alone
is 0.4–5.3% of total across four corpora. The map and the corrected numbers are in the follow-up PR's
doc commit. This PR is a mechanical reduction, not the removal.

## Gates

Local `cargo test --release` in flight at time of opening; CI is authoritative for the corpus,
matrix and formatter-parity gates. No ratchet re-baselined.

### DIFF
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/expression_utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/expression_utils.rs
@@ -301,6 +301,31 @@ pub(super) fn ends_with_braceless_control_header(prefix: &str) -> bool {
     false
 }
 
+/// [`ends_with_braceless_control_header`] answered from the statement's final
+/// line alone, or `None` when the verdict genuinely depends on earlier lines.
+///
+/// The full predicate only ever reads a suffix, so joining the accumulated
+/// lines is wasted unless the header's `(` opens on an earlier line.
+pub(super) fn braceless_control_header_from_last_line(last: &str) -> Option<bool> {
+    let t = last.trim_end();
+    if t.is_empty() {
+        return None;
+    }
+    if !t.ends_with(')') {
+        return Some(ends_with_keyword(t, "else") || ends_with_keyword(t, "do"));
+    }
+    let open = matching_open_paren(t)?;
+    let before = t[..open].trim_end();
+    if before.is_empty() {
+        return None;
+    }
+    Some(
+        ["if", "for", "while", "switch", "catch"]
+            .iter()
+            .any(|kw| ends_with_keyword(before, kw)),
+    )
+}
+
 pub(super) fn find_statement_end_client(s: &str) -> usize {
     let mut depth = 0;
     let mut in_string = false;
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -4506,6 +4506,22 @@ fn stage<'a>(
     }
 }
 
+/// Whether the statement accumulated so far ends in a brace-less control-flow
+/// header, materialising the joined text only when the last line cannot settle it.
+fn accumulated_ends_with_control_header(accumulated: &[&str], last: &str) -> bool {
+    if let Some(verdict) = expression_utils::braceless_control_header_from_last_line(last) {
+        super::profile::record_st_ctrl_header(0);
+        return verdict;
+    }
+    let mut acc = accumulated[..accumulated.len() - 1].join("\n");
+    if !acc.is_empty() {
+        acc.push('\n');
+    }
+    acc.push_str(last);
+    super::profile::record_st_ctrl_header(acc.len() as u64);
+    expression_utils::ends_with_braceless_control_header(&acc)
+}
+
 fn transform_instance_script_for_visitors(
     script: &str,
     analysis: &ComponentAnalysis,
@@ -4663,6 +4679,7 @@ fn transform_instance_script_for_visitors(
     // Pre-filter state_vars to only include variables that actually appear in the script.
     // This avoids O(M*N) scanning in downstream transforms for variables that can't match.
     // Uses O(text_len) identifier extraction instead of O(N*text_len) substring search.
+    super::profile::record_st_collect_scan(script_rest.len() as u64);
     utils::text_retain_matching_identifiers(&script_rest, &mut state_vars);
 
     // Ensure reactive import names are included in state_vars for $.get()/$.mutate() wrapping.
@@ -4715,6 +4732,7 @@ fn transform_instance_script_for_visitors(
     // top-level bindings take precedence for scope-level transforms. For example,
     // if there's a top-level `const multiplier = () => { let multiplier = $state(2); ... }`,
     // the inner `multiplier` should NOT cause the outer `multiplier` to be wrapped with $.get().
+    super::profile::record_st_collect_scan(script_rest.len() as u64);
     let local_reactive_vars = extract_local_reactive_vars(&script_rest);
     let top_level_binding_names: rustc_hash::FxHashSet<&str> = analysis
         .root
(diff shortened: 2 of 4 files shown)


baseballyama:
## Ceiling on the speedup, computed from counters only

Wall clock on this box is unusable (2.08ms vs 26.84ms for the same measurement),
so instead of an A/B, bound the win from the deterministic counters and a
per-unit cost. This bound survives a loaded machine because both inputs are
load-independent; only the per-unit constant is estimated, and it is estimated
generously (in the change's favour).

### #2611, control-header probe

Removed per full flowbite-svelte run (1,296 files): 4,899 `String` allocations
and 605,481 bytes of `join` + `push_str` copying.

Generous per-unit costs: 50ns per malloc/free pair, 1 GB/s for the copy.

    4,899 x 50ns                      = 0.24ms
    605,481 bytes / 1 GB/s            = 0.61ms
                                 total ~ 0.85ms

Against a flowbite total of 930ms, the ceiling is **~0.09% of compile**, and the
copy half of that is optimistic — 605KB in 4,899 pieces averages 124 bytes each,
where allocation dominates and bandwidth is irrelevant.

So this change is **not** a measurable speedup on typical input, and should not
be sold as one. What it removes is a **quadratic**: the join was O(statement
length) evaluated once per balanced line, so its cost grows with the square of a
statement's line count. Typical corpus statements are short, which is why the
ceiling is 0.09%; a pathological long statement is where it would have shown.
The honest claim is "removes a latent quadratic and 605KB of copying", not "makes
the compiler faster".

### #2612, runes fast path in dev

Removed: 1,252 of 4,891 `process_accumulated` calls on flowbite dev (25.6%).

Per-call cost from the prod profile: `process_accum` 11.75ms over 3,630
statements = 3.24us/statement. The skipped statements are by construction the
cheapest ones (no `$`, so the rune stage returns `Cow::Borrowed` immediately;
every other stage is `!analysis.runes`-gated), so 3.24us is an over-estimate.

    1,252 x 3.24us = 4.06ms

Against a flowbite total of 930ms, the ceiling is **<=0.44% of compile**, dev only.

### Why the ceiling is worth stating

A sibling agent removed 951 parses on carbon, verified by call count, and
measured **+4.19% slower** on huly. Work provably removed is not time removed.
Neither of these changes has been shown to speed anything up; both have been
shown to do strictly less work, and the bound above says how much that could
possibly be worth if it converts perfectly.

baseballyama:
## Byte-identity: measured, not argued

The PR body argues structurally that this is byte-identical. That argument is the same *kind* of
argument that failed six times in this file, so here is the measurement. Run locally on this branch
with a freshly built addon (`rm -f .corpus-cache/rsvelte.node`, `cargo build --release -p rsvelte_napi
--lib`, restage).

### Corpus output equality vs **official**, all three targets

    14,132 entries x 3 targets (client, server, client-dev)

    match            13184
    error-parity       948
    js-mismatch          0
    js-unparseable       0
    css-mismatch         0
    error-mismatch       0

    warning parity:  no regressions (102 known remain, unchanged)
    error parity:    no regressions (413 known remain, unchanged)
    output parseability: 14,132 match, 0 unparseable; all 39,551 generated modules parse

### Shape matrix (generated inputs, not collected)

    857 cases x 3 targets = 2,571 comparisons
    match 2337 / js-mismatch 234 / error-mismatch 0
    ✅ no regressions (234 known divergences remain)

Two-sided: neither a new failure nor a listed entry that started passing.

### Pattern-corpus regression files

All 33 files under `compatibility/pattern-corpus/issues/` compared against official on all three
targets: **33 match on all three, 0 divergences.**

No ratchet was re-baselined; `git diff HEAD -- compatibility/` is empty.

### One correction worth recording

I was warned that the corpus holds the ~24 known non-JavaScript outputs as *known failures*, so
reproducing the same wrong bytes would stay green. That trap does not apply here, because **there are
no known failures to hide behind**: `known-failures.{client,server,client-dev}.json` and
`parse-known-failures.{client,client-dev}.json` are all 3 bytes — `[]`. Output parseability reports 0
unparseable across 14,132 entries x 3 targets and all 39,551 generated modules parsing.

So this corpus does not contain those components, and this run says nothing about them either way —
neither that they are fixed nor that they are not. They must live in a different population (the
ecosystem repos, or the mutation fuzz). Whoever is tracking the 24 should know that the ratcheted
corpus is not where they will be observed.

### What this evidence does *not* establish

That either change makes the compiler faster. Neither has a timing A/B, because this box returned
2.08ms and 26.84ms for the same measurement. See the ceiling comment above: bounded from counters at
~0.09% (#2611) and <=0.44% of dev compile (#2612), and a sibling agent removed 951 parses by call count
and measured **+4.19% slower** on one corpus. Work provably removed is not time removed.

baseballyama:
### Correction: the first gate run used a build I could not vouch for — re-run, same result

I built the addon while a rebase moved commits between these two branches. Afterwards
`cargo build --release -p rsvelte_napi --lib` **recompiled `rsvelte_core`**, which proves the staged
`.corpus-cache/rsvelte.node` did not correspond to the tree I measured. A stale addon reports garbage
silently, so the numbers I posted above were not trustworthy when I posted them.

Re-run end to end with a build verified fresh (a second `cargo build` reporting `Finished` in 0.23s
with nothing to compile), addon re-staged from scratch, corpus re-collected and re-compiled:

    corpus, 14,132 entries x 3 targets (client, server, client-dev)
      match 13184 / error-parity 948 / js-mismatch 0 / js-unparseable 0
      css-mismatch 0 / error-mismatch 0
      warning parity: no regressions (102 known remain)
      error parity:   no regressions (413 known remain)
      output parseability: 14,132 match, 0 unparseable; all 39,551 modules parse

    shape matrix, 857 cases x 3 targets = 2,571 comparisons
      match 2337 / js-mismatch 234 / error-mismatch 0
      ✅ no regressions

    pattern-corpus issue files: 33/33 match official on all three targets

    counters, verified-fresh profiler:
      flowbite prod  ctrl_header 4,899 calls / 0 bytes   fastpath 1,261
      flowbite dev   ctrl_header 4,899 calls / 0 bytes   fastpath 1,252
      bits-ui  prod  ctrl_header 1,502 calls / 270 bytes fastpath   199
      bits-ui  dev   ctrl_header 1,502 calls / 270 bytes fastpath   197

Same numbers as before, so the earlier run happened to be right — but it was not *known* to be right,
and "the two runs agreed" is only evidence because I re-ran it. Recording the catch because the failure
mode is silent: a stale binary produces a confident, plausible, wrong reading.

PR #2612, opened 2026-08-08
fix(compiler): let the runes fast path run in dev

baseballyama:
Stacked on #2611 — review that first; this branch contains its commits.

## What

The per-statement fast path in the instance-script loop was disabled in dev **twice over**:

```rust
let runes_fastpath_eligible = analysis.runes && !dev && prop_mutation_vars.is_empty();
```

`prop_mutation_vars` is itself populated only `if dev`, so either conjunct alone kills it. Consequence:
in dev, *every* statement went through the full per-statement chain — including ones with no `$`, no
`export` and no reactive destructure, i.e. ones the chain provably cannot touch. Dev is the population
editors and `vite dev` actually hit.

## Why the guard was there, and why it can go

`prop_mutation_vars` feeds `wrap_prop_mutation_validation`, which runs over the **whole result after the
loop**, never over a single statement. It has no bearing on whether a statement can skip the processor.

Under `analysis.runes`, every stage in `process_accumulated` is gated on `!analysis.runes` except one:
the dev console wrap. And `transform_console_calls_dev_fragment` early-returns `None` unless the source
contains `console.`. So a single token guard is sufficient, and the bytes emitted are the statement
either way. Walking the chain stage by stage:

- `$:` branch, `transform_let_with_reexported_props`, state assigns, member mutations, prop update /
  source-read / assignment, store transforms, legacy destructure and state declarations, state reads,
  rest-prop and read-only-prop rewrites — all `!analysis.runes`.
- `transform_client_runes_with_skip_and_state` returns `Cow::Borrowed` immediately when the statement
  has no `$` (`rune_transforms.rs`).
- the `$state.snapshot` dev block requires `$state.snapshot(`, hence `$`.
- the export branches require a leading `export `, which `needs_export` already excludes.
- `console_dev` — the one dev-only stage, now guarded.

The fast path's predicate is otherwise untouched, so prod behaviour is unchanged by construction.

## Counter evidence

Statements skipping the per-statement chain in **dev**, output unchanged:

| corpus | before | after | of total |
|---|---:|---:|---|
| flowbite-svelte | 0 | **1,252** | 4,891 (25.6%) |
| bits-ui | 0 | 197 | 1,499 (13.1%) |

Legacy corpora (smelte, svelte-ux) are unaffected — the fast path is runes-only, and their counters are
byte-for-byte unchanged in both arms, which is the negative control.

The residual gap to prod (flowbite 1,252 vs 1,261; bits-ui 197 vs 199) is exactly the statements
containing `console.`.

## Also here

A doc commit recording the job map for `collect_vars` and `line_loop`, the blocked/unblocked split, and
two corrections: `line_loop`'s timer includes `process_accumulated` (so the character scanner is 0.4–5.3%
of total, not 22.7% of `script_text`), and the M1 progress table claimed 1 of 12 passes ported when 11
of 12 use `with_program_mut`.

The blocked group, for the record: `index_const_state_decls`, `index_reassigned_vars` and the
`has_legacy_export_let` scan all read string literals and comments as if they were code —
`index_const_state_decls` documents that it deliberately preserves a substring match firing on
`aconst x = $state(`. An AST replacement cannot see those. Whether that difference occurs on real input
has to be measured before the move; if it does, the AST side is right and the text side is a latent bug
that may currently be ratcheted green.

## Gates

Local `cargo test --release` in flight at time of opening; CI is authoritative for the corpus, matrix
and formatter-parity gates. No ratchet re-baselined. This changes dev-mode output paths, so
`client-dev` is the gate that matters most here.


baseballyama:
## Ceiling on the speedup, computed from counters only

Wall clock on this box is unusable (2.08ms vs 26.84ms for the same measurement),
so instead of an A/B, bound the win from the deterministic counters and a
per-unit cost. This bound survives a loaded machine because both inputs are
load-independent; only the per-unit constant is estimated, and it is estimated
generously (in the change's favour).

### #2611, control-header probe

Removed per full flowbite-svelte run (1,296 files): 4,899 `String` allocations
and 605,481 bytes of `join` + `push_str` copying.

Generous per-unit costs: 50ns per malloc/free pair, 1 GB/s for the copy.

    4,899 x 50ns                      = 0.24ms
    605,481 bytes / 1 GB/s            = 0.61ms
                                 total ~ 0.85ms

Against a flowbite total of 930ms, the ceiling is **~0.09% of compile**, and the
copy half of that is optimistic — 605KB in 4,899 pieces averages 124 bytes each,
where allocation dominates and bandwidth is irrelevant.

So this change is **not** a measurable speedup on typical input, and should not
be sold as one. What it removes is a **quadratic**: the join was O(statement
length) evaluated once per balanced line, so its cost grows with the square of a
statement's line count. Typical corpus statements are short, which is why the
ceiling is 0.09%; a pathological long statement is where it would have shown.
The honest claim is "removes a latent quadratic and 605KB of copying", not "makes
the compiler faster".

### #2612, runes fast path in dev

Removed: 1,252 of 4,891 `process_accumulated` calls on flowbite dev (25.6%).

Per-call cost from the prod profile: `process_accum` 11.75ms over 3,630
statements = 3.24us/statement. The skipped statements are by construction the
cheapest ones (no `$`, so the rune stage returns `Cow::Borrowed` immediately;
every other stage is `!analysis.runes`-gated), so 3.24us is an over-estimate.

    1,252 x 3.24us = 4.06ms

Against a flowbite total of 930ms, the ceiling is **<=0.44% of compile**, dev only.

### Why the ceiling is worth stating

A sibling agent removed 951 parses on carbon, verified by call count, and
measured **+4.19% slower** on huly. Work provably removed is not time removed.
Neither of these changes has been shown to speed anything up; both have been
shown to do strictly less work, and the bound above says how much that could
possibly be worth if it converts perfectly.

baseballyama:
## Byte-identity: measured, not argued

The PR body argues structurally that this is byte-identical. That argument is the same *kind* of
argument that failed six times in this file, so here is the measurement. Run locally on this branch
with a freshly built addon (`rm -f .corpus-cache/rsvelte.node`, `cargo build --release -p rsvelte_napi
--lib`, restage).

### Corpus output equality vs **official**, all three targets

    14,132 entries x 3 targets (client, server, client-dev)

    match            13184
    error-parity       948
    js-mismatch          0
    js-unparseable       0
    css-mismatch         0
    error-mismatch       0

    warning parity:  no regressions (102 known remain, unchanged)
    error parity:    no regressions (413 known remain, unchanged)
    output parseability: 14,132 match, 0 unparseable; all 39,551 generated modules parse

### Shape matrix (generated inputs, not collected)

    857 cases x 3 targets = 2,571 comparisons
    match 2337 / js-mismatch 234 / error-mismatch 0
    ✅ no regressions (234 known divergences remain)

Two-sided: neither a new failure nor a listed entry that started passing.

### Pattern-corpus regression files

All 33 files under `compatibility/pattern-corpus/issues/` compared against official on all three
targets: **33 match on all three, 0 divergences.**

No ratchet was re-baselined; `git diff HEAD -- compatibility/` is empty.

### One correction worth recording

I was warned that the corpus holds the ~24 known non-JavaScript outputs as *known failures*, so
reproducing the same wrong bytes would stay green. That trap does not apply here, because **there are
no known failures to hide behind**: `known-failures.{client,server,client-dev}.json` and
`parse-known-failures.{client,client-dev}.json` are all 3 bytes — `[]`. Output parseability reports 0
unparseable across 14,132 entries x 3 targets and all 39,551 generated modules parsing.

So this corpus does not contain those components, and this run says nothing about them either way —
neither that they are fixed nor that they are not. They must live in a different population (the
ecosystem repos, or the mutation fuzz). Whoever is tracking the 24 should know that the ratcheted
corpus is not where they will be observed.

### What this evidence does *not* establish

That either change makes the compiler faster. Neither has a timing A/B, because this box returned
2.08ms and 26.84ms for the same measurement. See the ceiling comment above: bounded from counters at
~0.09% (#2611) and <=0.44% of dev compile (#2612), and a sibling agent removed 951 parses by call count
and measured **+4.19% slower** on one corpus. Work provably removed is not time removed.

baseballyama:
### Correction: the first gate run used a build I could not vouch for — re-run, same result

I built the addon while a rebase moved commits between these two branches. Afterwards
`cargo build --release -p rsvelte_napi --lib` **recompiled `rsvelte_core`**, which proves the staged
`.corpus-cache/rsvelte.node` did not correspond to the tree I measured. A stale addon reports garbage
silently, so the numbers I posted above were not trustworthy when I posted them.

Re-run end to end with a build verified fresh (a second `cargo build` reporting `Finished` in 0.23s
with nothing to compile), addon re-staged from scratch, corpus re-collected and re-compiled:

    corpus, 14,132 entries x 3 targets (client, server, client-dev)
      match 13184 / error-parity 948 / js-mismatch 0 / js-unparseable 0
      css-mismatch 0 / error-mismatch 0
      warning parity: no regressions (102 known remain)
      error parity:   no regressions (413 known remain)
      output parseability: 14,132 match, 0 unparseable; all 39,551 modules parse

    shape matrix, 857 cases x 3 targets = 2,571 comparisons
      match 2337 / js-mismatch 234 / error-mismatch 0
      ✅ no regressions

    pattern-corpus issue files: 33/33 match official on all three targets

    counters, verified-fresh profiler:
      flowbite prod  ctrl_header 4,899 calls / 0 bytes   fastpath 1,261
      flowbite dev   ctrl_header 4,899 calls / 0 bytes   fastpath 1,252
      bits-ui  prod  ctrl_header 1,502 calls / 270 bytes fastpath   199
      bits-ui  dev   ctrl_header 1,502 calls / 270 bytes fastpath   197

Same numbers as before, so the earlier run happened to be right — but it was not *known* to be right,
and "the two runs agreed" is only evidence because I re-ran it. Recording the catch because the failure
mode is silent: a stale binary produces a confident, plausible, wrong reading.

baseballyama:
## `process_accumulated` as a share of **total compile**, per corpus

First, a correction to the composition, because it changes the arithmetic but not the conclusion:
**every row under `Script-text xform` in `compile_profile` is already a share of total compile**
(`compile_profile.rs`: `val / ms(total) * 100.0`). Multiplying `script_text% x process_accum%` divides
by the total twice. `process_accum`'s share of compile is the row's own number.

The guess landed near the right answer for smelte by luck — 73.7% x 75.1% = 55%, and the measured
value is 55.3% — but via an invalid route, and on runes corpora the same method gives 0.03% where the
measured value is 1.7%, a 57x error.

Measured, prod, quiet machine, 8 corpora. `legacy_reactive_stmts` counts top-level `$:` and is
load-independent, so the mode label is measured rather than inferred from the repo name.

| corpus | mode | files | `$:` stmts | script_text | **process_accum** | line_scan | Σstages/parent |
| shadcn-svelte | **runes** | 1682 | **0** | 7.6% | **1.1%** | 0.8% | 1.08 |
| skeleton | **runes** | 686 | **0** | 10.5% | **1.8%** | 1.3% | 1.13 |
| layerchart | **runes** | 1400 | **0** | 11.6% | **1.5%** | 1.3% | 1.03 |
| bits-ui | **runes** | 617 | **0** | 12.5% | **1.7%** | 1.2% | 1.12 |
| flowbite-svelte | **runes** | 1296 | **0** | 16.7% | **2.4%** | 1.7% | 1.16 |
| svelte-ux | **legacy** | 198 | 355 | 20.4% | **29.8%** | 1.4% | ⚠ 1.76 |
| sveltestrap | **legacy** | 129 | 225 | 25.9% | **35.1%** | 2.2% | ⚠ 1.70 |
| smelte | **legacy** | 74 | 194 | 33.1% | **55.3%** | 3.1% | ⚠ 2.01 |

Raw counts alongside, so a row that rounds small can be told from a row never reached: every runes row
has `process_accum` > 0 with thousands of statements processed (shadcn 2,103; skeleton 1,553;
layerchart 3,469; bits-ui 1,300; flowbite 3,630). **None of these is an unreached-code zero** — the
machinery runs on runes code, it is just cheap there.

**Your guess is confirmed: this is a legacy-only lever.** 1.1–2.4% on runes, 29.8–55.3% on legacy.
15–50x, with nothing in between — the same bimodality your file-count proxy found, now visible on the
time side with a load-independent discriminator (`$:` count is exactly 0 for all five runes repos).

### ⚠ One caveat that lands on the impressive rows specifically

The legacy percentages are not trustworthy at face value. Σstages / parent is 1.03–1.16 on runes and
**1.70–2.01 on legacy**. On svelte-ux, `process_accum` (37.04ms) **exceeds its own parent**
`script_text` (25.29ms), which is impossible for a sub-stage.

This is not load noise: it reproduces to ±0.1ms over 3 reps, and every consistency check the profiler
has passes — `nested 0`, `entries == parent_calls`, `entries_outside_parent 0`. `in_function`
(18.97 / 26.18 / 44.63ms) agrees with Σstages and disagrees with the parent, so two instruments agree
against one and **the parent `record_script_text` is under-measuring on legacy input**. Cause not
identified; I did not want to guess at it.

So: read the legacy numbers as one significant figure. The qualitative conclusion is robust because the
legacy/runes gap shows in the parent, the sub-stage, *and* `in_function` independently.

### The honest headline

Not one number — **two numbers and a condition**: removing `process_accumulated` is worth **30–55% on
legacy components and 1–2% on runes components**, conditioned on whether the component uses `$:`.
Modern large shipping repos are at zero.

Per your instruction I have **not** averaged these. A blended number would describe our corpus mix,
which is a choice we made.

### And the part that does not move

**The correctness case is unaffected by all of this.** Comments read as code (#2351, #2347, #2590,
#2599) is a defect class that does not depend on how many milliseconds the machinery costs. A shrunken
perf number is not a reason to leave the defects in — they are two independent justifications, and
deflating one does nothing to the other. That is now stated in the handoff doc next to the table, so
the next reader can't take the small number as permission.

baseballyama:
## Re-derived your census independently — your 12.3% reproduces, and then moves again

Per your own instruction, I re-derived rather than took it. Same matcher, `submodules/` whole,
`node_modules` excluded: **13,121 files / 15,151,873 bytes.**

| marker | files | files% | bytes% | mean size |
|---|---|---|---|---|
| `^\s*\$:` | 478 | 3.64% | **12.29%** | **3,898 B** |
| `^\s*export let` | 2,004 | 15.27% | 16.73% | 1,265 B |
| **union** | **2,157** | **16.43%** | **18.24%** | 1,281 B |
| non-legacy (complement of union) | 10,964 | 83.57% | 81.76% | 1,129 B |

**Your corrected numbers reproduce exactly** — 3.64% vs your 3.65%, 12.29% vs your 12.34%, mean 3,898 B
vs your 3,898 B. The file-count-to-bytes correction was right.

### But `$:` is the wrong marker for "legacy", and that moves it a fourth time

`export let` also forces legacy mode, and thousands of files have it *without* any `$:`. Counting only
`$:` **undercounts the legacy population by about a third**: legacy by bytes is **18.24%, not 12.29%**.

This matters for weighting because the two markers select different work. On smelte, `process_accum`
splits into `reactive_stmt` 31.3% and `pa_rest` 23.9% of compile — roughly half each. `$:` files carry
both; `export let`-only legacy files carry the `pa_rest` half (every `!analysis.runes` text stage) and
none of the `reactive_stmt` half. So the right weight is genuinely two-part, not one number.

### One thing to *stop* generalising: "legacy files are large"

That is true of the `$:` subset (3,898 B, **3.5x** the non-legacy 1,129 B) and **not** of legacy
overall. `export let` files average 1,265 B — ordinary size. The union averages 1,281 B, only **1.13x**
non-legacy. The 3.7x you found is a property of deep-legacy `$:` files specifically, so it does not
license "legacy code is bigger, therefore the byte weight understates further" as a general claim.

### On the superlinear exponent

Confirmed in the direction you said — `script_text` exponent is 1.989 on smelte, and it is the only
bucket above 1.0. So 18.24% is a **lower** bound on the time weight. I have deliberately **not**
multiplied the byte share by an exponent to produce a number; compounding an unmeasured quantity is the
exact route that produced three wrong figures today, and the compile-share table settles this without
needing the composition.

### Population label

Relabelled in the doc as **library corpus (`submodules/`)**, not "shipped Svelte source", with your
application-corpus rates (huly 0.4% / open-webui 0.3% / carbon 0.0% / SMUI 59.9% runes) recorded next
to it and the note that **runes rate and `$:` rate are not complements** — a file can use neither, so
the application `$:` share has to be measured directly rather than inferred. Flagged that 18.24% is
plausibly a distant floor for application source, and that published libraries often ship pre-compiled,
so application source is closer to real compile volume.

### On your track record note

Taken, and it cut both ways here — your correction was right and my check found a further error in it.
That's the process working, not a reason for either of us to defer. Worth noting the general form of
what your `$:` matcher and my first counter had in common: **both measured a proxy that was named after
the thing it was standing in for.** "legacy" measured as `$:`, "scan happened" measured as "scan was
called". The name made the gap invisible until someone asked what the proxy excludes.

baseballyama:
## (a) Found it. It was not legacy, and it was not re-entrancy.

`compile_profile` warms up on the first 100 files, then drains **`take_breakdown()` only**. The
script-text stage timers live in a separate accumulator set (`take_script_text_breakdown()`), which was
never drained. So the warmup's 100 files were charged to **every stage timer, `in_function`, `statements`,
and every counter I added** — while the parent `SCRIPT_TEXT` was correctly reset.

Predicted ratio: **(100 + N) / N**. Measured Σstages/parent, all eight:

| corpus | N | predicted | measured |
|---|---|---|---|
| smelte | 74 | 2.00 | 2.01 |
| sveltestrap | 129 | 1.78 | 1.70 |
| svelte-ux | 198 | 1.51 | 1.76 |
| bits-ui | 617 | 1.16 | 1.12 |
| skeleton | 686 | 1.15 | 1.13 |
| flowbite | 1296 | 1.08 | 1.16 |
| layerchart | 1400 | 1.07 | 1.03 |
| shadcn | 1682 | 1.06 | 1.08 |

**"Legacy" was a coincidence of corpus size** — my three legacy repos were simply the smallest (74, 129,
198), so a fixed 100-file warmup contaminated them most. Your re-entrancy hypothesis was the right shape
of question but the wrong mechanism, and `nested 0` was already telling us so.

Fixed by draining the second accumulator set. After the fix, all eleven corpora close:
`prologue+earlyout` +0.05 to +1.18 ms, `in_function <= parent`, Σ <= parent, on every row.

**The lesson I'd keep: reproducibility is not correctness.** ±0.1 ms over three reps and every built-in
consistency check passing (`nested 0`, `entries == parent_calls`, `entries_outside_parent 0`) — a
systematic accounting error reproduces perfectly and passes checks aimed at a different failure.

## (b) Measured — and it falsifies my "nothing in between"

`svelte-heroicons`: 294 files, **293 `export let`, 1 `$:`**. That is the population you asked for.

**`process_accum` = 12.1% of compile.** Intermediate, and much closer to the legacy pole.
`layercake` (50 `$:`) lands at **4.4%**, filling in more.

Corrected table, warmup-fixed, 11 corpora:

| corpus | `$:` | `export let` | script_text | **process_accum** | line_scan |
|---|---|---|---|---|---|
| shadcn-svelte | 0 | 0 | 7.4% | **1.0%** | 0.8% |
| bits-ui | 0 | 0 | 12.3% | **1.4%** | 1.0% |
| svar-core | 0 | 2 | 11.6% | **1.5%** | 1.2% |
| layerchart | 0 | 5 | 11.6% | **1.5%** | 1.3% |
| skeleton | 0 | 0 | 10.9% | **1.7%** | 1.2% |
| flowbite-svelte | 0 | 2 | 16.3% | **2.0%** | 1.4% |
| layercake | 50 | 1 | 12.6% | **4.4%** | 1.1% |
| **svelte-heroicons** | **1** | **293** | 16.5% | **12.1%** | 1.8% |
| svelte-ux | 196 | 103 | 19.7% | **16.1%** | 0.8% |
| sveltestrap | 126 | 66 | 26.0% | **20.2%** | 1.3% |
| smelte | 97 | 47 | 32.3% | **26.6%** | 1.4% |

**So: it is a continuum, not two poles.** 1.0 → 1.5 → 2.0 → 4.4 → 12.1 → 16.1 → 20.2 → 26.6%.
My "15–50x, nothing in between" was an artifact of sampling only the extremes *and* of the warmup bug
inflating the small end. **Retracted.**

Note also that the legacy numbers roughly halved: 29.8–55.3% → **16.1–26.6%**.

**Consequence for the marker:** `$:` is not a good discriminator. `export let` alone — no `$:` at all —
already reaches 12%. What predicts the share is *how thoroughly runes are disabled*, not one token.

**Caveat I'd attach:** svelte-heroicons is 294 tiny icon wrappers. One point should not carry the whole
`export let` population. The next measurement is an `export let`-heavy repo with *large* files.

I have **not** computed a blended epic figure, and the doc now says why not: (a) is fixed but (b) is
sampled once, so where inside the 18.24% byte share the mass sits is still open.

## Counters in these PRs were contaminated too — corrected

Same root cause. Re-measured clean:

| | before-fix (reported) | uncontaminated |
|---|---|---|
| flowbite ctrl_header calls | 4,899 | 4,246 |
| flowbite ctrl_header bytes | 0 | 0 |
| flowbite fastpath (dev) | 1,252 | 1,109 |
| bits-ui ctrl_header bytes | 270 | 270 |
| bits-ui fastpath (dev) | 197 | 184 |

**The deltas and every conclusion survive** — both arms carried identical contamination, so the
before/after comparison was always valid; only the absolute counts were inflated. The headline
(605,481 → 0 materialised bytes) is unchanged in kind: it is now 494,981 → 0 on the clean measurement,
still every join eliminated.

## Also recorded in the doc

Your point about proxies, generalised — this session produced **four** instances of the same error, and
they are now tabulated: "legacy" measured as `$:`, "scan happened" measured as "scan was called",
"`line_loop` time" measured with `process_accumulated` inside it, "measured-pass time" measured with
warmup left in. In each case **the proxy shared its name with the target, so it never read as a proxy**.
The doc now asks for a one-line "what does this quantity exclude?" on any new measurand.

## Merges

Both branches are rebased and retitled. I'll take them in on green with step 13 confirmed `success`
rather than `skipped`. The profiler fix rides on the docs commit in #2612; it is devtools-only and
cannot affect compiler output.

baseballyama:
## Application corpus measured — and it does **not** confirm 16%

Cloned at the versions this repo cites (open-webui v0.11.0; carbon/src, huly/plugins, smui/packages),
file counts within a few of the recorded ones: 291 / 650 / 2,126 / 449 = **3,516 files**.

`process_accumulated` as a share of **total compile**, warmup-fixed profiler, self-check closing on
every row (`prologue+earlyout` +0.1 to +1.0 ms, `in_function <= parent`):

| repo | files | legacy(bytes) | script_text | **process_accum** | line_scan | `$:` stmts |
|---|---|---|---|---|---|---|
| carbon/src | 291 | 87.9% | 36.3% | **30.2%** | 1.5% | 765 |
| open-webui/src | 650 | 70.3% | 30.7% | **25.7%** | 1.1% | 577 |
| huly/plugins | 2,126 | 74.9% | 27.8% | **22.5%** | 1.2% | 3,312 |
| **SMUI** (control) | 449 | **0.00%** | 17.6% | **2.1%** | 1.7% | **0** |

**Aggregate 22.8%** — and labelled: huly alone is **55.8% of this corpus's compile time**, so the
aggregate is a statement about corpus assembly. Excluding huly gives 23.3%, so it is not fragile to
that one repo, but four repos is four repos.

### The negative control passed

SMUI is application source at **0.00% legacy** and its `process_accum` is **2.1%** — sitting with the
runes libraries (1.0–2.0%), not with the other three applications. **Marker and timer agree.** So the
split is tracking the legacy/runes axis, not "applications differ from libraries". Had SMUI landed at
20%, the marker and the timer would have been in conflict and that would have outranked the headline.

### On the 16%

**It is not confirmed, and I'd avoid the symmetry entirely.** Two reasons:

1. **The measurement is 22.8%, not ~16%.** Nothing landed on it.
2. **Even a match would have been a coincidence of quantities.** The old 16% referred to
   `script_text` minus `ast_transforms` — the whole text machinery — while this measures
   `process_accumulated` alone. They are different sets. A number agreeing with a number about a
   different quantity is not confirmation.

So: this is the **first direct measurement** of the epic on the population that determines compile
volume. The old 16% still has no surviving provenance and should not be cited — including by me, and
including now that a nearby number exists.

### `line_scan` on applications: 1.1–1.7%

Consistent with every library corpus. **The character scanner is cheap in every population measured.**
Changing the population does not rescue the perf case for killing the scanner; that case is and stays a
correctness case.

### Gate-coverage finding — recorded in `gate-coverage.md`, not here

Per your instruction it went to `compatibility/gate-coverage.md` as a new section, since it is about
what our instruments cannot see:

> Every performance gate we own points at the runes end. CodSpeed fixtures are 8/9 runes; the library
> corpus is 12.34% legacy by bytes; application source is **68.89%**. The gates are aimed at the 1–2%
> end of a 1–30% range. A change removing most of `process_accumulated` would read as **flat on
> CodSpeed and nearly flat on the library corpus** while being worth 22–30% on application source — and
> symmetrically, a regression confined to the legacy path is invisible to every perf gate we run.

Marked `[U]`: whether these four repos represent application Svelte generally. They were chosen because
prior work already cited them, which is a selection that document cannot justify.

baseballyama:
## Decomposed — and **neither prediction holds**, because the bucket is three-way

`compile_profile.rs:276` takes `pa_rest` as the residual of `process_accumulated` after `runes_xform`
**and** `reactive_stmt`. So it splits three ways, not two, and the dichotomy "legacy branch vs rune
branch" leaves out the piece that turns out to be the majority.

Share of **total compile**, application corpus, warmup-fixed profiler:

| repo | process_accum | runes_xform | reactive_stmt | **pa_rest** |
|---|---|---|---|---|
| **SMUI** (control) | 2.0% | 0.3% | **0.00%** | 1.7% |
| carbon/src | 30.3% | 0.1% | 13.4% | **16.8%** |
| open-webui/src | 26.3% | 0.1% | 3.8% | **22.4%** |
| huly/plugins | 22.6% | 0.1% | 8.6% | **14.0%** |
| **aggregate** | **23.0%** | **0.08%** | **7.0%** | **15.9%** |

Within the bucket: `runes_xform` **0.3%**, `reactive_stmt` **30.6%**, `pa_rest` **69.1%**.

### Control passed first, as you asked

**SMUI `reactive_stmt` = 0.00 ms exactly.** The counter measures what its name says, so the other three
rows are readable.

### Three results

1. **The rune branch is negligible: 0.08% of compile, 0.3% of the bucket.** Your 50/50 outcome is
   **disconfirmed**. And the corollary you were most interested in — a previously unexamined ~10% runes
   target — **does not exist**. SMUI's *entire* `process_accum` is 2.0%. There is no hidden mass on the
   majority population; that is a real negative result and it closes the question.
2. **But `reactive_stmt` doesn't dominate either** — 30.6% of the bucket, not ~90%. So the strong case
   as you framed it (legacy branch ≈ the whole thing) is also not what the numbers say.
3. **The majority is `pa_rest`, 69.1%** — which is neither `$:` work nor rune work. It is the rest of
   the per-statement chain: store transforms, state reads, prop transforms, legacy declarations, member
   mutations.

### The honest limit on attributing `pa_rest`

Nearly every stage inside `pa_rest` is `!analysis.runes`-gated (the `stage(...)` calls in
`client/mod.rs`), so it *should* be legacy work — but **SMUI, which is pure runes, still shows
`pa_rest` = 1.7%**. So it is not purely legacy, and "23.0% is all legacy-attributable" is not
established.

That attribution is a **structural argument from code, not a measurement** — the middle tier of
`gate-coverage.md`'s own evidence taxonomy. Converting it needs per-stage timers inside `pa_rest`. The
`stage()` wrapper already exists but its `measure-stmt-chain` feature records *whether a stage rewrote*,
not how long it took, so it does not answer this. **I have not added those timers**, and I'd flag that
as the prerequisite for anyone who wants a defensible legacy/runes split of the 23%.

So the epic's value on application source is **~23% of compile, of which ~2 points are reachable in
pure runes mode** and the remainder is legacy-gated by code structure but not yet measured as such.

### Recorded

In the handoff doc next to the application table, including the negative result about the runes side and
the explicit "structural argument, not discriminating case" label.

baseballyama:
Adding `skip-changeset`, and recording why rather than just unblocking the gate.

The gate is correct to ask: the title says `fix(compiler)`, and `fix`/`feat` require a
changeset. The title is also correct — the runes fast path was disabled twice over
(`!dev`, plus a `prop_mutation_vars.is_empty()` guard that is itself dev-only and feeds
a whole-result post-pass), so it never ran in dev. That is a real defect.

But **the defect produced byte-identical output**: 14,132 corpus entries × client /
server / client-dev, `js-mismatch 0`, `css-mismatch 0`. No user ever saw wrong output
from it; what they got was the slow path. So there is nothing to describe in a release
note, and a changeset would publish a fix for a bug no consumer could observe.

`fix(` is the accurate word for the defect. `skip-changeset` is the accurate statement
that nothing user-facing changed. Those are two different questions and this PR
genuinely answers them differently.

Worth noting the label was needed only because of a retitle: this PR was opened as
`perf(compiler)` and renamed because its measured ceiling (≤0.44% of dev compile) sits
two orders of magnitude under the ~5% code-layout floor, so no A/B in this repo could
confirm a speed claim. The evidence that stands is the work-avoided counter — dev
statements skipping the transform chain went flowbite **0 → 1,252 of 4,891** and
bits-ui **0 → 197**, with legacy unchanged as the negative control.

baseballyama:
Superseded by #2633 — this PR was auto-closed when its base branch was deleted on #2611's merge, and a closed PR with a missing base can be neither reopened nor retargeted. Same two commits, rebuilt on current `main` by cherry-pick (#2611 was squash-merged, so the originals are not ancestors), verified byte-identical to the tree the combination test ran against.

The measurement history in this thread is the part worth keeping: the warmup-contamination root cause, the application-corpus three-way decomposition, and the census re-derivation. #2633 links back here rather than restating it.

PR #2613, opened 2026-08-08
docs: record where the `to_value` cost actually is (one site, not the lazy cache)

baseballyama:
Docs only. No code change, no ratchet touched.

**This PR started life claiming the opposite of what it now says.** It opened as "close `JsNode` → `serde_json::Value` as a performance target, ceiling 0.1–0.6%". That claim was measured over the wrong population and has been replaced. The history is left intact because the way it broke is the most transferable part.

## What is actually true

`JsNode::to_value` has **54 call sites**. One is the lazy JSON cache in `TypedExpr::as_json`; the other 53 bypass it. Every materialization figure this project has ever quoted — the **27,488 → 12,089 → 3,649** series across #2510 / #2570 / #2576 — counts *only* the cache, because `MATERIALIZATIONS` increments inside the cache's `record()`.

The bypassing population had never been measured. When it was, 98% of it was one site: **`instance_labeled_statements_json`** (`2_analyze/mod.rs`), the legacy `$:` serializer.

| corpus | cache calls | cache objects | `$:` site calls | its objects | its share of all objects / map entries |
|---|---|---|---|---|---|
| huly/plugins | 3,649 | 21,016 | 3,303 | **71,293** | **77.0% / 77.2%** |
| carbon | 899 | 2,756 | 725 | **13,790** | **81.7% / 82.1%** |
| open-webui | 4,998 | 26,627 | 587 | **15,452** | **33.6% / 33.4%** |
| SMUI | 973 | 5,828 | **0** | **0** | **0%** |

All other direct sites combined: 70 calls / 306 objects on huly.

So on huly the lazy cache — the thing three merged PRs optimized — is **22.7%** of the JSON the compiler builds. One never-instrumented site is 77%.

## Cross-validated by a second, independent instrument

An allocator-event sampler (1-in-64, inclusive attribution) run by a parallel investigation, sharing no assumption with the object counter:

- **8.91 / 8.87 / 8.82 allocations per object** on huly / carbon / open-webui
- **1.78 allocations per map entry** on huly — exactly what `preserve_order` predicts (`serde_json::Map` is an `IndexMap`: one key `String` malloc plus amortised table growth per entry, and one hash insert)
- site share of `to_value` allocation cost: **84.7 / 91.5 / 36.0 / 0%** — same ordering and magnitude as the object share, consistently a few points higher, which is the expected direction
- the SMUI zero reproduces as a zero on both instruments

## The remedy is not a cache

The single caller already serializes once and shares across all three legacy `$:` passes — #2510 did that dedup. The fix is porting the three consumers (`check_reactive_declaration_cycles`, `populate_legacy_dependencies`, `collect_reactive_statement_dependencies`) to typed traversal, which deletes the entries outright. One producer, three consumers, no scope-dependent constant folder.

Gated on legacy `$:` density: ~80% on Svelte-4-era corpora, 0% on runes-only code, and therefore invisible to the CodSpeed population (8/9 runes).

## No time share is claimed

Deliberately. A wall-clock timer over all 54 sites spanned **5.07 / 2.89 / 0.58%** of compile on three runs of identical deterministic work; an allocator-model estimate was retracted by its own author for a circular correction factor. Both are recorded as unresolved, and the doc says not to quote a percentage without a new measurement. A SipHash→FxHash probe measured huly −4.08% (8/8 paired wins), which sizes the hashing component but sits at the ~5% layout floor and was not shipped.

## Two rules this cost us

1. **Count a function's call sites before trusting a per-function measurement.** A timer at one call site reports that site, not the function, and nothing in the output says so.
2. **Attribute a memoised value by reader *set*, not first reader.** Under a per-node cache, `#[track_caller]` names only the first reader, so converting the site it blames moves the count by **zero** — it points at the wrong site. It blamed `extract_metadata_from_tag` for 66.7% of materializations; converting it left the count unchanged and `expression_has_reactive_state` rose 1,933 → 20,264.

The sharpest illustration of both: #2510 replaced whole-script serialization with per-`$:`-statement serialization, won −20% on huly, and reported materializations falling 27,488 → 12,089. All true. It also **created the largest single JSON producer in the compiler**, and the instrument built to validate it could not see it.

Instrumentation is on `tools/measure-json-instrumentation` (`e4f47227`), behind the existing `measure-json` feature, deliberately unmerged.


baseballyama:
**Draft — do not merge. The ceiling table in this PR is measured over the wrong population, and the closure claim as written is wrong in scope.**

An independent allocation-attribution measurement put `JsNode::to_value` at ~2.5% of non-kernel CPU against my 0.18–0.61% of compile. Chasing that 4–10x disagreement found the defect, and it is in my instrument, not theirs:

`JsNode::to_value` has **54 call sites**. Exactly **one** is the lazy JSON cache in `TypedExpr::as_json` (`crates/rsvelte_core/src/ast/js.rs:45`). My timer was placed there, so it measured *the cache path only*. The other 53 call `to_value` directly and bypass the cache entirely — Phase 1 `1_parse/read/expression.rs`, Phase 3 `client/visitors/expression_converter.rs`, Phase 2 `visitors/{class_body,await_expression,each_block,variable_declarator}.rs`.

The allocation profile attributes by stack containing `to_value` and therefore catches all 54. So the two instruments were never measuring the same population, and mine was measuring a subset without saying so.

The same blind spot applies to the **materialization counter**: `MATERIALIZATIONS` only increments inside `record()`, which only the cache calls. "3,649 on huly" is cache materializations, not `to_value` invocations — and the historical 27,488 → 12,089 → 3,649 series inherits the same restriction.

This is the repo's own recorded failure mode (`feedback_probe_fidelity_is_separate_from_probe_result`): a re-implemented probe returns a confident small number because its input population is degenerate, and the positive control still passes.

Re-measuring with the timer inside `to_value` itself (depth-guarded so recursion is not double-counted), which covers all 54 sites. Will update this PR with the corrected table, or close it if the corrected number does not support a closure.

PR #2619, opened 2026-08-08
fix(compiler): keep the operand after a line-ending binary operator (client)

baseballyama:
## The defect

Four of the 30 real-world files whose client output no JS parser accepts fail
with `Unexpected ")"`, and each one ends a source line with a binary operator
whose right operand is on the next line:

```svelte
let isRestricted =
  a() !== undefined ||
  b() !== undefined
```

```svelte
$: isDocx =
  item?.meta?.content_type ===
    'application/vnd.openxmlformats-…' ||
  (item?.name && item.name.toLowerCase().endsWith('.docx'));
```

Two client text passes decide where such a statement ends, and neither looked at
the operator the line ends with:

- `find_statement_end_client` (`client/expression_utils.rs`), which reads the
  initializer that goes inside `$.mutable_source(…)`. At a depth-0 newline it
  consults only the *next* line's first character; `b` is not in its
  continuation set, so the statement ended after `||`.
- the instance-script line accumulator (`client/mod.rs`), whose
  `trailing_operator` test excludes `==` / `!=` / `<=` / `>=` — and therefore
  `===` and `!==` — from being continuations. Those exclusions belong to the
  *assignment* test (`ends_with('=')` must not fire on a comparison), but they
  were written as an exclusion from the whole disjunction.

## Minimal repros

```svelte
<script>
  let flag = foo(1) !== undefined ||
    foo(2) !== undefined
  function go() { flag = true }
</script>

<p on:click={go}>{flag}</p>
```

before: `let flag = $.mutable_source(foo(1) !== undefined ||)`

```svelte
<script>
  export let item;
  $: v = item.a ===
    1;
</script>

<p>{v}</p>
```

before: `$.set(v, item().a ===);`

after, both keep the right operand. The server target was already correct on
both, so it is the control.

## The fix

One helper, `ends_with_binary_operator`, used by both sites. It lists only the
operators that are unambiguous at the end of a line: `||`, `&&`, `==`, `!=`,
`<=`, `>=`, `?`, and `+` when it is not `++`. `-` is left out because `a--` ends
a statement, and `/` because it also closes a block comment. The helper tests the
code on the line, not the raw line, so an operator inside a trailing comment is
not read as a continuation.

## Tests

`crates/rsvelte_core/tests/client_trailing_binary_operator.rs`. Against
unpatched code, `client_keeps_the_operand_after_a_trailing_or` and
`client_keeps_the_operand_after_a_trailing_strict_equals` fail with exactly the
predicted output (`$.mutable_source(foo(1) !== undefined ||)` and
`$.set(v, item().a ===)`); the four other tests pass before and after.

Two controls a wrong-but-plausible fix would break:

- `an_operator_inside_a_comment_is_not_a_continuation` — testing the raw line
  instead of the code on it merges `export let a = 1; // wide || tall` with the
  declaration below it.
- `a_post_increment_still_ends_the_statement` — reading every trailing `+` as a
  continuation swallows the statement after `n++`.

## Byte-identity gate

10,389 `.svelte` files (18 submodule corpora + `compatibility/pattern-corpus` +
carbon / huly / open-webui / smui) × client-prod / server / client-dev, hashed
with `corpus_hash` before and after:

|---|---|---|
| client | 10389 | 4 |
| server | 10389 | 0 |
| client-dev | 10389 | 4 |

The four are the same four files in both client columns, and they are exactly the
four whose output a JS parser rejected with `Unexpected ")"`:

- `huly/…/card-resources/…/settings/ProperitiesSection.svelte`
- `huly/…/setting-resources/…/EditAttribute.svelte`
- `huly/…/view-resources/…/ViewOptions.svelte`
- `open-webui/…/common/FileItemModal.svelte`

Every other file is byte-identical on all three targets.

Baseline and candidate are both built from `73a0fbe1` (an earlier run against a
drifting `origin/main` reported nine changes; that was baseline drift, not this
change — re-measured with both sides pinned).

## Repro fixture

`compatibility/pattern-corpus/issues/2605-trailing-binary-operator.svelte` is
byte-identical to the official compiler's output on client and server.


baseballyama:
The fix is right and I am not asking for changes to it. But I measured what it leaves behind,
and the answer is larger than the PR implies: **#2619 takes the handled set from 4 operators to
8, out of 23. Fifteen still emit unparseable JavaScript.**

### Method

The shape matters — my first attempt used a runes component and all 23 parsed, because the runes
path never reaches `find_statement_end_client`. On the legacy shape from this PR's own fixture
(`export let` + `$:`), varying only the operator at the line break:

```svelte
<script>
	export let item;
	export let width = 1;

	$: kind =
		item.a <OP>
		item.b;

	function toggle() { width = 2; }
</script>
<button on:click={toggle}>{kind}{width}</button>
```

Compiled with `compile_one` and checked with `node --check`.

### Result

| | operators | count |
|---|---|---|
| handled on `main` | `\|\|` `&&` `??` `+` | 4 |
| **added by #2619** | `==` `!=` `<=` `>=` | **4** |
| **still unparseable** | `*` `%` `-` `/` `<` `>` `\|` `&` `^` `**` `<<` `>>` `in` `instanceof` `,` | **15** |

The harness moves in both directions, so it is not a vacuous pass: 19 fail on `main`, 15 fail on
this branch, and the four that changed are exactly the four the PR adds.

Same mechanism as #2605 — the right operand is dropped and the dangling operator is left:

```js
op_star  →  $.set(kind, item().a *);          // SyntaxError: Unexpected token ')'
op_le    →  $.set(kind, item().a <= item().b);  // fixed here
```

### The part I would change

`ends_with_binary_operator`'s doc comment says *"Only operators that are unambiguous at the end
of a line are listed: `-` is left out because `a--` ends a statement, and `/` because it also
closes a block comment."* Those two exclusions are correct and well argued. The problem is that
naming two exclusions with reasons makes the list read as **surveyed** — a reader concludes the
remaining thirteen were considered and found inapplicable, when they are simply absent.

Three of the survivors are worth noting because they look like oversights rather than judgement
calls: **`<` and `>` fail while `<=` and `>=` pass**, so a prefix of a handled operator is
broken; **`**` fails while `+` passes**; and **`,`** fails, which is ordinary formatting
(`const a = 1,\n b = 2`) rather than an exotic shape.

Suggestion, in ascending order of effort — I am not asking for any of them in this PR:

1. Change the comment to say what it means: these are the operators *currently* handled, `-` and
   `/` are known-ambiguous, and the rest are **unhandled and known to emit invalid output**. That
   is one line and it stops the list being read as complete.
2. Add the unambiguous survivors: `*` `%` `<` `>` `|` `&` `^` `**` `<<` `>>` `,` `in`
   `instanceof`. None carries `-`'s or `/`'s ambiguity.
3. The real fix is that this is a token-level question being answered by suffix matching on text,
   which is the epic's whole thesis.

Happy to open a follow-up issue with the matrix and the 23 fixtures attached. Flagging rather
than filing so this does not read as a merge blocker — **it is not one; #2619 strictly improves
the situation.**

Issue #2601, opened 2026-08-07
Corpus Compat / Compiler parity is red on main: the mutation full sweep has 43 new divergences and 481 stale entries


baseballyama:
`main` is red on `Corpus Compat / Compiler parity`, in the **mutation full sweep** — a stage that runs on `main` pushes only (PRs get a deterministic sample), so no PR can see it.

Run [31224615836](https://github.com/baseballyama/rsvelte/actions/runs/31224615836) at `09950a48`:

```
[mutate] results:
  match             21496
  error-parity       1967
  unparseable          43
  code-mismatch        44
  comment-mismatch  12924
  error-mismatch        0
  compiler-crash        0

[mutate] ❌ 43 NEW divergences (not in the baseline)
[mutate] ❌ 481 baseline entries already PASS — the ratchet is stale.
```

## Why this surfaced now, and not earlier

This is **the first `Corpus Compat` run on `main` to complete in a long time**. The eight before it are all `cancelled`: every workflow set `cancel-in-progress: true` on a group keyed by `github.ref`, which is the identical string for every push to `main`, so each merge killed its predecessor. #2594 fixes that. The mutation full sweep is the stage with the most to lose from it, because it is the one stage PRs never run.

So the honest framing is **"a gate nobody has seen finish since #2353 baselined it"**, not "a regression from the last merge". The 481 / 43 have accumulated over an unknown number of merges.

## The 43 are not all new breakage — some are re-labelled

`scroll-area__m0__line-with-paren.svelte.ts (server)` appears in **both** lists: as a NEW `[unparseable]` and as a stale `[code-mismatch]`. The ratchet key is `id [verdict] (target)`, so an entry whose verdict *changes* is scored as one removal plus one addition. Any triage that reads "43 new" as "43 newly broken files" will over-count, and any that reads "481 fixed" as progress will over-count in the other direction. **The two lists have to be joined on `id (target)` before either number means anything.**

That said, `unparseable` is a strictly worse verdict than `code-mismatch` — the compiler emitted something no JS parser accepts — so a `code-mismatch → unparseable` transition is a real degradation even though the entry count is flat.

## Sample of the `unparseable` set

All are delimiter-carrying comments (`line-with-semi`, `line-with-paren`, `block-with-paren`), which is the #2253 signature the mutation gate exists to find:

```
bits-ui/…/scroll-area__m0__line-with-paren.svelte.ts   (server)      Unexpected token
[...]

PR #2622, opened 2026-08-08
perf(analyze): answer the legacy `$:` passes from the typed AST

baseballyama:
Ports the legacy `$:` analysis off `serde_json::Value` and onto the typed AST, deleting the single largest JSON producer in the compiler.

## What this removes

`instance_labeled_statements_json` (`2_analyze/mod.rs`) serialized every top-level legacy `$:` `LabeledStatement` of the instance script via `JsNode::to_value()`, so three consumers could walk it as JSON. Measured per call site, that one line produced:

| corpus | its objects | its map entries | share of ALL `to_value` objects |
|---|---|---|---|
| huly/plugins | 71,293 | 356,896 | **77.0%** |
| carbon | 13,790 | 67,522 | **81.7%** |
| open-webui | 15,452 | 77,191 | **33.6%** |
| SMUI | 0 | 0 | **0%** |

With `serde_json`'s `preserve_order` feature (`Map` = `IndexMap`), each map entry is a key `String` malloc *and* a hash insert, so the entry column is allocation and hashing load in equal measure.

The producer now returns `Vec<&JsNode>` and the three consumers — `check_reactive_declaration_cycles`, `populate_legacy_dependencies`, `collect_reactive_statement_dependencies` — plus their private walkers traverse typed nodes through the shared `for_each_js_child`.

## Gate: deterministic counts, not a stopwatch

| corpus | JSON objects before → after | map entries before → after |
|---|---|---|
| huly/plugins | 92,615 → **21,322**  (−77.0%) | 462,157 → **105,261**  (−77.2%) |
| carbon | 16,874 → **3,084**  (−81.7%) | 82,205 → **14,683**  (−82.1%) |
| open-webui | 45,937 → **30,485**  (−33.6%) | 231,227 → **154,036**  (−33.4%) |
| **SMUI (negative control)** | 6,078 → **6,078**  (0) | 30,326 → **30,326**  (0) |

The producer's call site is **absent** from the after profile on all four corpora. Across the 3,509-file corpus that is **100,535 fewer JSON objects and 501,609 fewer map entries** — and with `preserve_order`, each of those entries was a key `String` allocation *and* a hash insert.

Both arms were measured at the same base commit with the same instrumented binary; counts are deterministic and identical across runs.

SMUI is the negative control: the producer short-circuits on `analysis.runes`, so it never called `to_value` there and the expected delta is exactly 0.

**No time share is claimed.** #2613 records that none is currently defensible — a wall-clock timer over all 54 `to_value` sites spanned 5.07 / 2.89 / 0.58% of compile across three runs on identical deterministic work, and an allocator-model estimate was retracted by its author. That has not changed.

### The affected population is the majority of application source

The win is gated on legacy `$:` density, so I measured it. Two populations, reported separately because **the gap between them is the finding** — a file that contains a top-level `$:` (heuristic: a line whose first non-whitespace token is `$:`):

**Libraries** (`submodules/`, 23 repos — bits-ui, flowbite, shadcn, skeleton, layerchart …):
```
files   13,078 total     478 legacy  =  3.65%
bytes   15,098,016   1,863,261       = 12.34%
```

**Applications** (the 4 real-world corpora, per-repo then aggregate):

| corpus | files | legacy files | bytes | legacy bytes |
|---|---|---|---|---|
| huly/plugins | 2,123 | 1,252 (58.97%) | 7,124,519 | 5,334,209 (**74.87%**) |
| open-webui | 650 | 215 (33.08%) | 3,612,860 | 2,538,330 (**70.26%**) |
| carbon | 287 | 173 (60.28%) | 941,662 | 827,741 (**87.90%**) |
| SMUI | 449 | **0 (0.00%)** | 951,109 | **0 (0.00%)** |
| **aggregate** | **3,509** | **1,640 (46.74%)** | **12,630,150** | **8,700,280 (68.89%)** |

**68.9% of application source by bytes**, against 12.3% for libraries — a 5.6x gap. Legacy files are 2.5x larger than non-legacy ones in the application corpus, so files-vs-bytes matters here too. Published libraries also frequently ship pre-compiled, so application source is the closer proxy for real compile volume.

This is not a legacy tail. **It is most of what actually gets compiled**, and the per-repo rows are bimodal — 0% or ≥33%, nothing between. SMUI's exact 0.00% is also an independent confirmation of the counter gate below, which measures 0 `to_value` calls at this site on SMUI by a completely different route.

### What CodSpeed will show — pre-registered

The bench corpus is often described as "8 of 9 runes". **That is stale** — fixtures 10 and 11 were added to close exactly that gap, and measuring it directly gives 3/11 legacy by files and **37.7% by bytes**. So the timing gates are under-weighted against applications by ~1.8x (37.7% vs 68.9%), not blind.

That makes this change falsifiable rather than invisible, so I am pre-registering the prediction:

- `05-legacy-reactive`, `10-legacy-typescript-props`, `11-store-heavy-legacy` should improve
- the eight runes fixtures should move by ~0
- `compile_both` should improve by roughly the legacy share

**A uniform result across all eleven is evidence against this change**, not a corpus artifact. A flat result on 05/10/11 specifically would mean the objects removed below cost less than I think, and I would want to know that.

The library-weighted gates are the badly-aimed ones: `submodules/` is 12.34% legacy by bytes against 68.89% for applications, a 5.6x gap. That population mismatch is filed as its own row (C6) in `compatibility/gate-coverage.md`, since it is a gate-coverage finding rather than a fact about this PR.

## Correctness gate, and what it is blind to

| corpus | files | client | server | client-dev |
|---|---|---|---|---|
| huly/plugins | 2,123 | IDENTICAL | IDENTICAL | IDENTICAL |
| open-webui | 650 | IDENTICAL | IDENTICAL | IDENTICAL |
| carbon | 287 | IDENTICAL | IDENTICAL | IDENTICAL |
| SMUI | 449 | IDENTICAL | IDENTICAL | IDENTICAL |

**10,527 comparisons, 0 differing.** Positive control: the same comparison run across client-vs-dev reports 2,111 differing lines, so it can see a change when there is one. The two arm binaries were confirmed to differ (`cmp`), so this is not a stale build compared against itself.

**Both arms were rebuilt at the same base and the base SHA is part of the result.** Measured with base = `c3968878` (current main) against this branch at `2bce2e60`.

That is not bookkeeping. The first run of this gate used base `23e68acf`; while it was in flight, #2426 landed on main and moved diagnostic source ranges. Re-running the base arm at current main and diffing the two baselines against each other gives:

```
old base 23e68acf vs new base c3968878, huly client:  8 files differ
```

**Main changed its own output on 8 files.** Had I diffed the merged port against the stale baseline, those 8 would have surfaced as differences *in this PR*, and the obvious next move — hunting my own diff for a behaviour change that isn't there — would have cost an hour and could plausibly have ended in a wrong "fix". A differential measurement is only meaningful with its base SHA attached, and on an actively-merging repo the base moves on the timescale of a single verification run.

#2426 also touched the same function this PR ports, adding a span read off the JSON node. Merging main produced a **textually clean merge that did not compile** — the span now comes off the typed accessors (`node.start().zip(node.end())`). Worth noting because CI builds `refs/pull/N/merge`: no local build on the pre-merge branch could have caught it, since the call site did not exist there.

The comparison is **rsvelte-before vs rsvelte-after**, not rsvelte vs the official compiler. That is strictly stronger than the official-parity ratchet for detecting *movement* — the ratchet only reports pass/fail against official and shows nothing when a change is parity-neutral in both directions — but it is blind to a case where both sides are wrong identically. That is the correct trade for a refactor whose entire thesis is "no behaviour change", and it is worth stating because a reader who knows the corpus gate compares against *official* will otherwise assume this one does too. No ratchet is re-baselined.

The hash covers `js.code`, `css.code`, **and the warning set** (code, message, position). `corpus_hash` previously hashed output only. That is the exact blind spot that let #2256 ship while the corpus scored the reproducing entry as `MATCH`, and these three consumers feed dependency sets — one of the few things that can move warnings without moving a byte of output. Compile errors are already distinguishable via the existing `ERR` line, which matters here because two of the three consumers can raise `reactive_declaration_cycle`.

## Two places where the typed walk and the JSON walk did not already agree

Porting surfaced two fields where `for_each_js_child` descends differently from the old JSON catch-all, which iterated every object key except `type`/`start`/`end`/`loc`. Both were found with discriminating cases, and **both behaviours are preserved**.

**1. `LabeledStatement.label`.** `for_each_js_child` deliberately skips it; all three JSON walkers visited it. It is observable through dependency *ordering*, not just membership — a `$: { outer: for (const x of xs) { s = x; break outer; } }` collects `outer` before `xs` under the JSON walk and after it under a naive typed walk, because `outer` otherwise survives only via `break outer`. An explicit `LabeledStatement` arm restores the JSON field order.

Recorded deliberately: **neither the old nor the new behaviour matches upstream here.** Upstream treats neither the label nor the break-label as a reference (`is_reference` is false for both). Walking `BreakStatement.label` is a pre-existing rsvelte deviation; this PR preserves it rather than fixing it, because that is a semantic change and belongs in its own PR where a reviewer can see it on its merits. An undocumented preserved bug is indistinguishable from an unnoticed one.

**2. TS `type_annotation` blobs on `ObjectPattern` / `ArrayPattern`.** The JSON walk descended into the serialized annotation and could pick up a type name as a runtime dependency; `for_each_js_child` does not. Reachability is narrow (TS script, an annotated *destructuring* declarator inside a `$:` body, and the type name must also name a value binding) but real — the discriminating case flips a `reactive_declaration_cycle` error on and off.

Preserved, by walking the annotation blob explicitly. **This does not reintroduce what the PR removes:** `type_annotation` is stored on the typed node as an `Option<Box<serde_json::Value>>` — an opaque blob the parser keeps verbatim — so it is *already* JSON and is not produced by `to_value()`. Walking it costs no serialization. Without that fact the cheap option looks expensive, and a surviving JSON walker in a PR titled "stop walking JSON" reads as an incomplete port rather than a deliberate fidelity choice.

The current behaviour is arguably a latent bug (upstream strips type annotations before scope building, so a type name should not be a reactive dependency). Fixing it is a separate change.


baseballyama:
## Re-measured after base drift

The byte-identity evidence in the PR body was measured at base `c3968878`. `main` has since
gained two commits that change `3_transform/client/mod.rs` — `5092d12d` (#2624) and `80abbe52`
(#2611) — so that base no longer describes the tree this PR would land on, and the old numbers
did not cover the combination. Re-measured at the current base.

**Base `80abbe52`, PR head `4afdcc56` (merge of `main` into the branch; tree verified identical
to my local merge of the same two commits).**

| corpus | files | client | server | client-dev |
|---|---|---|---|---|
| carbon | 1,324 | IDENTICAL | IDENTICAL | IDENTICAL |
| huly | 2,501 | IDENTICAL | IDENTICAL | IDENTICAL |
| open-webui | 650 | IDENTICAL | IDENTICAL | IDENTICAL |
| SMUI | 449 | IDENTICAL | IDENTICAL | IDENTICAL |

4,924 files × 3 targets = **14,772 comparisons, 0 divergences.** The hash covers generated
output *and* the warning set as `(code, message, start)`, so a changed warning set cannot pass
as identical.

Local combination test against the current `main` (the two binaries that could observe either
side): `private_field_non_this_receiver_2483` 4 passed, `import_comment_delimiter_2601` 6 passed.

### Two guards that actually fired

The first run of this gate reported **DIFFER on 100% of rows in all 12 cells**. That was not a
compiler divergence — the port-side harness emits a warning column that the base-side harness
did not, because the base binary was built from `main`'s older `corpus_hash.rs`. Every `w0` row
hashed identically; only rows carrying warnings differed. I rebuilt the base harness from the
*same* harness source so that only `rsvelte_core` varies between the two binaries.

Worth stating plainly because the failure is the interesting direction: had the two harness
versions been swapped, the extra column would have been absent on both sides and the run would
have reported IDENTICAL while silently comparing nothing about warnings. The run also serves as
this gate's positive control — it is now observed to be able to emit a non-identical verdict,
which a gate that has only ever printed IDENTICAL is not.

Second: the two harness binaries are required to differ by sha256 before any comparison runs
(`5abbabff…`/`87c38e0a…` base, `747afc08…` port), so a stale binary compared against itself
cannot produce a false IDENTICAL.

baseballyama:
## Re-measured again — base drifted a second time

`main` moved to `5f73abbd` (adding #2628, another `3_transform` fix), so the previous
measurement at `80abbe52` no longer described the tree this lands on. Re-measured at the
current base, this time through **#2627's own `scripts/dev/diff-corpus-hash.mjs`**, so the
label / target / file-count guards are active rather than a hand-rolled `diff`.

**Base `5f73abbd`, PR head `b36a3aed`.**

```
carbon      1,324 files   client IDENTICAL   server IDENTICAL   client-dev IDENTICAL
huly        2,501 files   client IDENTICAL   server IDENTICAL   client-dev IDENTICAL
open-webui    650 files   client IDENTICAL   server IDENTICAL   client-dev IDENTICAL
smui          449 files   client IDENTICAL   server IDENTICAL   client-dev IDENTICAL
```

4,924 files × 3 targets = **14,772 comparisons, 0 divergences**, all twelve cells exit 0.

One trap worth recording, because it bit the same way at both bases: this branch adds a
warning column to `corpus_hash.rs`, and `main` does not have it. Building the base binary from
`main`'s own harness therefore compares two **different output formats** and reports 100% of
rows differing — which is what the first run did. The base harness has to be built from *this
branch's* harness source so that only `rsvelte_core` varies between the two binaries. The
inverse ordering is the dangerous one: with the extra column absent on both sides the run would
report IDENTICAL while comparing nothing about warnings.

The two binaries are also required to differ by sha256 before any comparison runs
(`967856e6…` base, `12e16952…` port), so a stale binary cannot be compared against itself.

**Base drift is now the recurring cost here**, not the port: `main` has gained four
output-affecting `3_transform` commits (#2624, #2611, #2628, and #2601's follow-up) while this
PR sat in the queue, and each one invalidates a locally-measured identity result. The structural
fix is for this differential gate to run in CI on the merge ref — #2627 made it pin and print
its own base, which is the prerequisite, but nothing yet invokes it there. Filing that as a
follow-up rather than widening this PR.

PR #2623, opened 2026-08-08
docs(compat): record the perf gates' population as gate-coverage row C6

baseballyama:
Adds row **C6** to `compatibility/gate-coverage.md`: every corpus this repo gates on samples a population where legacy `$:` is a small minority, and application source is not.

## The finding

A file counts as legacy if any line's first non-whitespace token is `$:` (heuristic, stated in the row along with what it over- and under-counts).

| population | legacy by files | **legacy by bytes** |
|---|---|---|
| libraries — `submodules/`, 23 repos | 3.65% | **12.34%** |
| applications — huly/plugins | 58.97% | **74.87%** |
| applications — open-webui | 33.08% | **70.26%** |
| applications — carbon | 60.28% | **87.90%** |
| applications — SMUI | **0.00%** | **0.00%** |
| **applications, aggregate** | **46.74%** | **68.89%** |

A **5.6x gap by bytes**, and bimodal per repo — 0% or ≥33%, nothing between — so an aggregate over a library-weighted sample does not interpolate to an application. Legacy files are 2.5–3.7x larger than the rest, so files and bytes disagree by ~3x and bytes is the closer weight for anything scaling with script content. Libraries also frequently ship pre-compiled, making application source the better proxy for real compile volume.

## It is expensive, measured independently

A sibling investigation timed `process_accumulated` across the same corpora: carbon 30.2%, open-webui 25.7%, huly 22.5%, **aggregate 22.8%** — against **SMUI at 2.1%**, sitting with the runes libraries rather than its fellow applications. A 10.9x split on this row's axis, from an instrument unrelated to the object counters.

Attributed to what it actually measures: `process_accumulated` spans **both** rune and legacy statement transforms, so the split is *consistent with* the legacy branch dominating it but does not prove the whole 22.8% is legacy; and it is a Phase-3 script-text path, not the Phase-2 JSON serialization that #2622 removes. The row says so.

## A stale figure this repo repeats, now retired

The bench corpus is widely described as **"8 of 9 runes"**, including in `benches/corpus/README.md`'s own distribution table, which compares fixtures **01–09**. Fixtures 10 and 11 were added specifically to close that gap. Measured directly:

```
command grep -lE '^[[:space:]]*\$:' benches/corpus/*.svelte
  05-legacy-reactive  10-legacy-typescript-props  11-store-heavy-legacy
→ 3/11 files (27.3%), 9,195/24,385 bytes (37.7%)
```

So the timing gates are **under-weighted ~1.8x against applications, not blind**. That distinction matters: anyone reasoning from "8 of 9" concludes CodSpeed cannot see a legacy change at all and is then licensed to ignore any result it gives. It can see one, at roughly half the weight real application code would give it — which makes such a change **falsifiable** rather than invisible. The row records the disconfirming outcome (a uniform result across all eleven fixtures) so the prediction is a test.

The library-weighted gates, at 5.6x under, are the badly-aimed ones.

## Negative control

SMUI is **0.00%** legacy by the source regex **and** makes **0** `to_value` calls at the legacy-`$:` producer by the compiler counter — two mechanisms sharing no assumption agreeing on an exact zero. The regex is demonstrably capable of returning a real zero and did, on the one corpus where the compiler independently agrees; that clause is what separates a negative control from a regex that merely found no matches.

## Marked `[U]`

(a) the Phase-2 legacy `$:` path's own time share — two instruments tried, neither defensible; (b) the legacy-vs-runes decomposition of the 22.8%, which `st.reactive_stmt` could settle directly; (c) whether four repositories represent application Svelte generally — densities run 33–88%, so the direction is not in doubt, but the aggregate is four samples.


baseballyama:
Two checks against a fresh clone of `carbon-design-system/carbon-components-svelte`. The
conclusion holds both times; the **number in the headline** does not survive either one
unchanged, and the row currently reads as a census.

### 1. The carbon row is `src/` only, and the scope is unstated

The repo has **1,324** `.svelte` files, not 287:

```
525 tests/   425 docs/   291 src/   78 e2e/   5 examples/
```

`src/` alone is 291 files at 176 legacy (60.5%), which is the row as written
(287 / 173 / 60.28% — the small delta is the pin). Scored over the whole repo it is
**250 / 1,324 = 18.9%**, because `tests/` and `docs/` together are 950 files and are
mostly not legacy.

So an unstated scope choice moves that row by **3.2x**. The choice itself is defensible —
`tests/` and `docs/` are not what "compile volume in the field" means — but the table gives
a denominator without saying what it excludes, and 287 next to "carbon" reads as the repo.
State the filter in the row, the same way the `$:` heuristic already gets its caveat stated
two paragraphs up.

### 2. carbon is a component library, and it is in the applications bucket

It ships to npm as `carbon-components-svelte`; `src/` is library source, not application
code. It is not in `submodules/`, so nothing is double-counted — but the placement is what
sets the 5.6x. Moving carbon's `src/` to the library population:

| | libraries | applications | ratio |
|---|---|---|---|
| as written | 12.34% | 68.89% | **5.6x** |
| carbon → libraries | 16.78% | 67.36% | **4.0x** |

**The finding survives and the number does not.** 4.0x still says the performance gates are
aimed at the minority of compile volume, which is the claim the section exists to make. But
"5.6x" is currently doing headline work in this PR *and* in `AGENTS.md`, and it rests on a
classification the table never argues for.

Suggestion: keep carbon where it is, and add one line saying it is a component library
included on the application side because its `src/` is hand-written Svelte rather than a
published bundle — plus the 4.0x alternative, so the next reader sees the number's
sensitivity instead of re-deriving it. A number that moves 5.6 → 4.0 under a defensible
reclassification should carry that range, per this file's own rule that an unsupported
blind-spot claim is worse than a blank.

Neither point blocks merge. Both are one line each in the table.

Issue #2621, opened 2026-08-08
compiler(server): official re-homes a `$:` trailing comment onto the next statement; rsvelte drops it


baseballyama:
On the **server** target, the official compiler **re-homes** the trailing comment of a
`$:` statement onto the next surviving statement. rsvelte **drops** it.

```svelte
<script>
	let total = 10;
	$: half = total / 2; // halved
</script>
```

Official keeps `// halved` (attached to the next surviving statement); rsvelte emits
nothing for it. Client is unaffected — this is server-only.

### Provenance, and why it is filed rather than ratcheted

Found while writing the `pattern-corpus/issues/2604` fixture for #2618. The author
**removed the comment from the fixture** rather than seed a known-failure entry, which
was the right call for that PR — but it means the divergence is currently recorded in
**no** gate and in no ratchet. Filing it is what keeps it from being lost, because a
fixture that no longer contains the shape is indistinguishable from a fixture that
never diverged.

This is worth stating plainly: **a ratchet entry suppresses everything about the entry
it lists, so declining to add one was correct — but the alternative to a ratchet entry
is an issue, not silence.**

### Open question for whoever picks it up

**#2572 looks adjacent — check whether it covers the server side.** Do not assume it
does from the title; the client and server comment paths are separate, and server SSR
is pure-AST while client CSR still has a text fallback, so a client-side comment fix
has no automatic server counterpart.

Two things to establish before fixing:

1. **What official's rule actually is.** "Re-homes onto the next surviving statement"
   is inferred from one case. Whether it re-homes onto the next statement, onto the
   enclosing block, or only when the `$:` statement is fully elided, is unmeasured.
   Generate the cases; do not generalise from the one that was hit by accident.
2. **Whether the drop is unconditional.** A comment on a `$:` statement that survives
   into the output may behave differently from one whose statement is elided.

Related: `compatibility/gate-coverage.md` — server comment handling is a known thin
spot, and esbuild strips comments in ~68% of module-target entries, so comment
divergences are structurally under-observed by the corpus in the first place.

baseballyama:
## The rule stated in this issue is false, measured

Measured on `origin/main` @ `e62a6a02` against the pinned `submodules/svelte` (Svelte
5.56.8), `generate: 'server'`, `dev: false`, over a **generated** matrix — comment kind
(`// trailing`, `/* trailing */`, `// leading`) × `$:` body kind (`half = t/2`,
[...]

PR #2626, opened 2026-08-08
fix(compiler): keep comment delimiters out of the initializer scan (client)

baseballyama:
**Stacked on #2619** (`fix/client-trailing-binary-operator`) — it edits the same
function, so review that one first. Merge #2619 and this rebases to a two-line
delta.

## The defect

`find_statement_end_client` reads a legacy `let` initializer by hunting for `;`,
`)`, `}` byte by byte, with no knowledge of comments. A delimiter inside a
comment therefore ends the statement:

```svelte
<script>
  let x = a + // ; c
    b
  function go() { x = 2 }
</script>
```

before (client):

```js
let x = $.mutable_source(a + //); c
    b
```

The generated paren is spliced into the comment body and the continuation line is
severed — the #2253 signature, and output no JavaScript parser accepts.

after, byte-identical to the official compiler:

```js
let x = $.mutable_source(a + // ; c
	b);
```

## Two halves

1. The scan steps over strings, comments and regex literals through the shared
   `shared::js_scan::skip_opaque` — the scanner #907 / #2253 introduced for
   exactly this, which this function never adopted.
2. Even with the right expression, appending `)` to text that ends inside a line
   comment hides it. `let x = foo(1) // ; c` needs the paren on the next line;
   upstream breaks the line there, and `break_after_line_comment` now does too.
   Fixing only (1) leaves `$.mutable_source(foo(1) // ; c)` — still unparseable.

## Tests

`crates/rsvelte_core/tests/comment_delimiters_in_initializer.rs`. Against
unpatched code the three positive tests fail with exactly the predicted output —
`$.mutable_source(a + //); c`, `$.mutable_source(foo(1) //); c`,
`$.mutable_source(foo(1) //)) c` — and the two controls pass before and after:

- `a_real_semicolon_still_ends_the_initializer` — "skip every `;`" would pass the
  three positives and merge the declaration with whatever follows it.
- `the_server_still_keeps_both_operands` — the server never ran this scan, so a
  fix that reached it would move it.

All three minimal repros are byte-identical to the official compiler's client
output after the change.

## Byte-identity gate

10,389 `.svelte` files x client-prod / server / client-dev, `corpus_hash`
before and after (both sides built from the same tree, differing only by this
commit):

|---|---|---|
| client | 10389 | 0 |
| server | 10389 | 0 |
| client-dev | 10389 | 0 |

**Zero.** No shipped file in the corpus carries a delimiter inside a comment in
an initializer position, which is the point: the collected corpus cannot see this
class at any size. The mutation-fuzz gate ranks `line-with-semi` at 21.5
divergences per 1,000 mutants against `line` at 0.0 — these are the
discriminating inputs.

## Why there is no `pattern-corpus` fixture

I wrote one and withdrew it. Every shape here diverges from official on the
**server** target for an unrelated, pre-existing reason: the server drops a
comment that official keeps (`let mid = a + b;` against official's `let mid = a +
// ; c` / `b;`). Landing the fixture would mean seeding a `known-failures` entry,
which convention 6 of `pattern-corpus/README.md` exists to prevent. That server
comment-fidelity gap is #2621; when it is fixed, this shape should be added.


baseballyama:
Retargeted the base from `fix/client-trailing-binary-operator` to `main`.

This PR was stacked on #2619's head branch. When a base branch is merged and auto-deleted,
GitHub **closes** the PRs stacked on it — that is what happened to #2612 in this repo yesterday,
and it cost a rebuild-by-cherry-pick plus a new PR (#2633) to recover the review history.

Nothing about the content changes. Until #2619 merges, the diff shown here includes #2619's
commits; it narrows to this PR's own change automatically once #2619 lands.

baseballyama:
Resolved a conflict on this branch (and on #2619's) — #2618 merged and added its own row to
`compatibility/pattern-corpus/README.md`. Purely additive on both sides; I kept both rows in numeric
order and changed no wording. Both PRs are `MERGEABLE` again.

**One review point, non-blocking.** This PR ships a changeset and
`tests/comment_delimiters_in_initializer.rs`, but **no `compatibility/pattern-corpus/issues/` fixture
for its own defect** — the `2605-*.svelte` file here comes from #2619, which this branch carries.
#2618 and #2619 each landed one for theirs.

Normally that would be a style note. Here it is worth more than usual, because this defect's corpus
delta was measured at **0/0/0 across 10,389 files** — no shipped file in the collected corpus carries
the shape, so the corpus gate cannot observe a regression in it **at any corpus size**. The Rust test
is real protection and I am not claiming this is unguarded; the difference is that a pattern-corpus
fixture also runs the full pipeline across all three targets with the comparison-side normalization,
which is where the class actually manifests.

Rule 6 in that README explicitly allows "in the fix PR **or immediately after it merges**", so this
does not need to hold the merge. Flagging it so it happens rather than becoming the fixture nobody
adds — the PRs that closed the neighbouring shapes all have one, and an inconsistent set is how the
next person concludes the convention is optional.

Issue #2629, opened 2026-08-08
compiler: does a constructor-root private-field read follow the update divergence or upstream? (#2464 vs #2574)


baseballyama:
`main` was red for ~45 minutes because two merged PRs assert opposite behaviour for the
same expression. This issue records the substantive question that fell out; the red
itself is fixed separately.

### The collision

- **#2464** (`fix(compiler): mirror upstream's constructor read form and $state proxy
  gate for private-field $.set`) moved the constructor-root read of a private rune field
  onto upstream's `.v` shortcut **for every receiver**, and pins it:
  `nonthis_private_state_receiver_2467.rs` asserts `log(inst.#n.v);` and
  `const a = inst.#n.v;`.
- **#2574** (`docs(compat): record and pin the non-`this` private-receiver divergence`)
  recorded the opposite as a deliberate divergence and pinned *that*:
  `private_field_non_this_receiver_2483.rs` asserted `console.log($.get(inst.#n));` plus
  `!out.contains("inst.#n.v")`.

Both were green. #2574's branch was cut before #2464 landed and never rebased, so its
tests never ran against a tree containing #2464 — `Test (ubuntu-latest, N)` completed at
01:57–02:35Z and the PR merged at 04:30Z, two hours and several merges later.

Attribution was established by a three-value test, not by reading the diff: reverting
**only** `675b34d5` (#2464) on `c3968878` makes all four tests pass; without the revert
three pass and one fails; and the failure is the predicted one (`inst.#n.v` where the
test wants `$.get(inst.#n)`).

### The open question

**Should the constructor-root read follow the update, or upstream?**

The update rows diverge for a reason that does not reach the read: upstream's
`$.get(o.#n)++` **does not parse**, and rsvelte will not trade a valid module for an
invalid one. Both read forms parse, so that argument is silent here.

#2574's argument for `$.get`, which #2464 did not answer because it never saw it:

> `.v` is the untracked read, `$.get` the tracked one. Upstream's receiver check is
> purely syntactic — it does not establish that the receiver is the object under
> construction. `constructor(o) { … }` where `o` is an already-live instance reads
> `o.#n.v`, and a reaction that runs the constructor never subscribes.

Against it: upstream is the specification wherever its output is valid, and #2464 chose
parity deliberately.

**This is a semantics decision, not a cleanup.** It needs someone to run both compilers
on a constructor invoked from inside a reaction and say which observable behaviour is
right — not to pick whichever test is more convenient to keep.

[...]

PR #2632, opened 2026-08-08
fix(compat): record the constructor-root read as parity, not a divergence

baseballyama:
**`main` is red.** `rsvelte_core::private_field_non_this_receiver_2483::a_read_through_a_non_this_receiver_goes_through_get` fails on `c3968878`, and it fails for a docs-only PR too (#2613), which is how it surfaced — every PR cut from current `main` inherits it.

## Attribution, by A/B rather than by reading the diff

The test arrived with #2574. Copying its file onto two commits and running it:

| commit | verdict |
|---|---|
| `c84d227a` (#2594, immediately before #2464) | 4 passed |
| `675b34d5` (#2464) | 3 passed, **1 failed** |

Exactly one of the four flips, and it is the read test. #2464 is
"mirror upstream's constructor read form …", which adopts
`MemberExpression.js:15-18`'s `in_constructor` gate for **any** receiver.

## Why both PRs were green

#2574 was cut before #2464 merged, so its branch measured the pre-#2464 compiler and pinned that behaviour as a deliberate divergence. #2464's own
`nonthis_private_state_receiver_2467.rs` asserts `log(inst.#n.v);` — the exact opposite — two files away. Neither CI run could see the other. Green + green = red, which is the merge-order hazard, not a defect in either change.

## Which side stands, and why it is not a coin flip

**#2464's.** `compatibility/deliberate-divergences.md` opens by saying the file exists for cases where reproducing upstream's bytes would mean shipping *invalid JavaScript*. That is true of the update rows — official emits `$.get(o.#n)++`, which no parser accepts — and it is **not** true of the read row: `inst.#n.v` and `$.get(inst.#n)` both parse and both run.

What was left, once the parse argument does not apply, was "`.v` is the untracked read, so restricting the shortcut to a literal `this` only ever over-tracks". That is a judgement about which tracking behaviour is nicer, and where the two merely differ, upstream is the specification. An entry in this file has to carry more than a preference, because the file's whole purpose is to be the short list of places where the usual rule is suspended.

## The change

- `private_field_non_this_receiver_2483.rs`: the read test now pins **parity** (`console.log(inst.#n.v);` at a constructor root, `return $.get(inst.#n);` in a method body — the `in_constructor` boundary), renamed so it does not read as a divergence.
- `deliberate-divergences.md`: the read row is marked parity, the "why rsvelte's form is correct" section is scoped to the two update rows, and the retraction is written down rather than the old text quietly overwritten — including that it was reintroduced by two changes that were each green alone.
- No compiler code changes. The behaviour `main` already has is the behaviour that is now recorded.

## Verification

`private_field_non_this_receiver_2483` (4), `nonthis_private_state_receiver_2467` (8) and `ctor_private_state_read_and_proxy_2395` (6) all pass together — which they could not before, since the first contradicted the second.

### DIFF
--- a/compatibility/deliberate-divergences.md
+++ b/compatibility/deliberate-divergences.md
@@ -51,7 +51,7 @@ export class R {
 | nested fn in constructor, `inst.#n++` | `$.get(inst.#n)++;` | **no** | `$.update(inst.#n);` | yes |
 | constructor root, `inst.#n++` | `inst.#n.v++;` | yes | `$.update(inst.#n);` | yes |
 | constructor root, `--inst.#n` | `--inst.#n.v;` | yes | `$.update_pre(inst.#n, -1);` | yes |
-| constructor root, read `inst.#n` | `inst.#n.v` | yes | `$.get(inst.#n)` | yes |
+| constructor root, read `inst.#n` | `inst.#n.v` | yes | `inst.#n.v` — **parity** | yes |
 | any position, `this.#n++` | `$.update(this.#n);` | yes | `$.update(this.#n);` | yes |
 
 The parse column is acorn's verdict on the official output and `oxc_parser`'s on rsvelte's;
@@ -74,8 +74,8 @@ Reported upstream as **sveltejs/svelte#18621** (open as of 2026-08-08).
 
 ### Why rsvelte's form is the correct one
 
-The unparseable rows need no argument beyond the parse column. The three constructor-root
-rows are the ones that need one, because upstream's output there is valid:
+The unparseable rows need no argument beyond the parse column. The two constructor-root
+**update** rows are the ones that need one, because upstream's output there is valid:
 
 - `.v++` writes the source's value **without notifying**, and upstream's receiver check is
   purely syntactic — it does not establish that the receiver is the object under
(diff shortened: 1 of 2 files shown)


baseballyama:
Closing as a duplicate of #2630, which was opened ~15 minutes earlier for the same `main` breakage. Two sessions reached the same diagnosis independently, and the A/B attributions agree: the test passes on the tree immediately before #2464 (`c84d227a` here, `c3968878` with only `675b34d5` reverted there) and fails at #2464, with exactly the predicted failure mode.

**#2630 is the better of the two, on the point where they differ.** I wrote that "#2464's side stands" and argued it from the divergence file's own scope — the parse argument covers the update rows and not the read row. #2630 makes the same tree change but **does not** settle the semantics, filing it as #2629 instead, and its reason is one I did not account for: #2464's author had no way to see #2574's tracking argument, so treating #2464 as having answered it is wrong. Picking whichever test is cheaper to keep is not an answer to the question.

It is also stronger on method: it cleared two nearer suspects before reaching #2464, and it records a **broken control** it used on the way (swapping `crates/` wholesale to a pre-#2605 tree also removed #2574's own compiler change, so "still fails" told it nothing). Writing that down is worth more than the conclusion it did not support.

Nothing here is worth porting over. The only thing my version has that #2630's does not is the claim it deliberately declines to make.

PR #2633, opened 2026-08-08
fix(compiler): let the runes fast path run in dev

baseballyama:
Replaces #2612, which GitHub auto-closed when its base branch (`perf/script-text-line-loop`) was
deleted on #2611's merge. A closed PR whose base is gone cannot be reopened or retargeted, so this is a
fresh PR with the same two commits rebuilt on current `main`. **#2612's thread holds the full
measurement history** — the warmup root cause, the application-corpus decomposition, and the census
re-derivation — and is worth reading rather than repeating here.

## What

The per-statement fast path was disabled in dev **twice over**:

```rust
let runes_fastpath_eligible = analysis.runes && !dev && prop_mutation_vars.is_empty();
```

`prop_mutation_vars` is itself populated only `if dev`, so either conjunct alone kills it. It feeds
`wrap_prop_mutation_validation`, which runs over the **whole result after the loop**, never a single
statement — so it has no bearing on whether a statement can skip the processor.

Under `analysis.runes` every stage in `process_accumulated` is gated on `!analysis.runes` except the dev
console wrap, and `transform_console_calls_dev_fragment` returns `None` unless the source contains
`console.`. Guarding that one token is sufficient; the emitted bytes are the statement either way.

Labelled `fix`, not `perf`: the defect is a supposedly-live path being dead in half the compiler's
configurations. The time it buys is bounded from counters at **<=0.44% of dev compile** — under the ~5%
code-layout floor, so it is not separately measurable and is not claimed.

## Counter evidence (uncontaminated)

Statements skipping the per-statement chain, dev, output unchanged:

| corpus | before | after |
|---|---:|---:|
| flowbite-svelte | 0 | **1,109** |
| bits-ui | 0 | 184 |

Legacy corpora are unaffected — the fast path is runes-only, and their counters are byte-for-byte
identical in both arms. That is the negative control.

## ⚠ #2611 merged on a stale green, and this PR's combination test is the only thing that has checked it

Stating it plainly so no reader assumes CI covered it. **#2611 merged on a green whose run was created
at 03:32Z — four merges stale.** Its checks were computed against a `main` that no longer existed by the
time it landed, and one of the intervening merges (#2574) changed `client/mod.rs`, the same file #2611
changed.

The combination test below is the **only** verification of #2611's content against current `main`, and
it happened **after** the merge, not before. It came out clean — corpus, matrix and the three named
binaries all pass on the merged tree, and this branch contains #2611's content — so the outcome is fine.
But "verified clean" and "verified before merging" are different claims and only the first one is true
here.

## Verified against **current** main, not a stale one

#2611 and #2612's earlier greens were computed at 04:19, before #2574 merged at 04:30 — the run creation
time is when the merge ref is computed, so a job *completing* at 05:28 still tested a pre-#2574 tree.
#2574 touched `client/mod.rs`, **the same file this PR touches**, so "disjoint" was not available as an
argument.

Local combination test, `origin/main` (5092d12d) merged into this branch, textually clean:

- **Corpus 14,136 entries x 3 targets** (client, server, client-dev): `js-mismatch 0`,
  `js-unparseable 0`, `css-mismatch 0`, `error-mismatch 0`, "all corpus outputs identical after
  normalization". Warning and error ratchets: no regressions. All 39,563 generated modules parse.
- **Shape matrix**: 857 cases x 3 targets = 2,571 comparisons, no regressions.
- **The three tests main's recent commits added/pinned**, run individually rather than trusting one
  invocation: `private_field_non_this_receiver_2483` 4 passed, `import_comment_delimiter_2601` 6 passed,
  `client_transforms_pin` 3 passed.

The branch pushed here is **byte-identical** to the tree those gates ran against (`git diff --quiet`
against the validated worktree returns 0), rebuilt on `main` by cherry-pick after #2611 was
squash-merged.

**A retraction about that verification.** I first reported that the combined
`cargo test --test a --test b --test c` invocation silently ran only two of the three binaries and still
exited 0. **That was wrong.** Re-run to completion, the identical command line runs all three
(3 `Running` lines). What actually happened is that the run exceeded its foreground timeout, was moved
to the background, and I read its output file **while it was still being written** — treating a
mid-write log as a final one. The tell was in the artifact and I missed it: the partial log had two
`test result:` lines but only **one** `Running` line, and results without their own `Running` line mean
you are reading a log mid-write, not a runner that skipped work.

So there is no cargo defect here, and nothing about `--test` selection should be inferred from this PR.

What does survive, and is worth keeping: **count `Running` lines against binaries requested**, because a
test that never ran and a test that passed produce the same exit code. That check is sound whatever the
cause — it is what surfaced the discrepancy, even though my first explanation of the cause was wrong.
And separately: **do not read a backgrounded job's output as final until the job reports completion.**

## Also here

A docs commit recording the job map for `collect_vars` / `line_loop`, the blocked/unblocked split, the
warmup-contamination bug found in `compile_profile` (fixed in the same commit), the three-way
decomposition of `process_accumulated` on application source, and a table of the four proxy-naming
failures this work produced. `compatibility/gate-coverage.md` gains a section on every performance gate
we own pointing at the runes end of the population.

No ratchet re-baselined.


baseballyama:
## Folded in — all three, plus your direct-comparison result

### 1. `(100+N)/N` is a **lower bound**, not a prediction

You're right about the mechanism and it flips three rows from scatter to corroboration. Warmup calls
`rsvelte_core::compile` whole; the measured loop calls `transform_component` alone. Warmup's per-file
contribution is therefore the **larger** of the two, so the measured ratio should come in **at or above**
the bound:

| corpus | N | lower bound | measured |
|---|---|---|---|
| smelte | 74 | 2.00 | **2.01** (2.24 by your direct old/new comparison) |
| sveltestrap | 129 | 1.78 | 1.70 |
| svelte-ux | 198 | 1.51 | **1.76** ≥ |
| bits-ui | 617 | 1.16 | 1.12 |
| skeleton | 686 | 1.15 | 1.13 |
| flowbite | 1296 | 1.08 | **1.16** ≥ |
| layerchart | 1400 | 1.07 | 1.03 |
| shadcn | 1682 | 1.06 | **1.08** ≥ |

Your two-binary comparison is recorded next to it as the **independent** route, and the distinction is
stated: my argument was an inconsistency in the *uncorrected* quantity (Σstages/parent), yours measures
the *corrected* quantity directly. layercake landing on 1.57 against a 1.565 bound is the tighter of the
two confirmations.

### 2. `take_index_oracle()` now drained after warmup

`compile_profile.rs:362` reads it once after the loop; it was not in the reset. Fixed, with the reason
recorded in the comment — a raw count can't skew a percentage, but leaving it undrained makes the check
count cover a population no other number in the report covers, which is the same hazard one field over.
Verified after the change: smelte self-check still closes (−0.05 ms), oracle reports `0 checks` when
unset.

### 3. The confound is now named, above the table

This is the one that mattered most, because the **pre-fix version of that exact table** produced the
"line loop is a legacy cost" reading I then had to retract. The note says: contamination is monotone
decreasing in N; in this corpus **file count and legacy density are correlated** — legacy-dense repos
are the small ones (smelte 74, sveltestrap 129, svelte-ux 198), runes-pole repos the large ones
(shadcn 1682, layerchart 1400, flowbite 1296) — so before the fix "legacy is slow" and "small N means
more contamination" were **perfectly confounded** and indistinguishable. The rows below are post-fix and
not subject to it, but a reader can't tell that from the table alone. Anyone re-measuring must record N
and legacy density separately.

### Scope of this push

The only code change is `crates/rsvelte_devtools/src/bin/compile_profile.rs` — a dev binary the corpus
gates do not use (they run through the NAPI addon). `git diff --name-only` against the validated commit,
restricted to `crates/ scripts/ compatibility/`, returns that one file. **The compiler tree the
combination test validated is unchanged.**

### On the duplicate

No harm done from my side — and the duplicate produced a measurement I didn't have and couldn't have
gotten from my own route, since mine can only ever show the uncorrected quantity is internally
inconsistent. Two independent routes agreeing is worth more than either alone.

PR #2636, opened 2026-08-08
fix(transform): keep the newline after a declarator's line comment

baseballyama:
Third site of the #2601 class, after #2605 (`$effect` deletion) and #2624 (import terminator). It does not overlap #2618, #2619 or #2626.

## The defect

`split_comma_separated_declarations` accumulates a multi-line declaration onto one line before splitting it at top-level commas, and the accumulator joined continuation lines with a **space**:

```rust
while !is_declaration_complete(&full_decl) && line_idx + 1 < lines.len() {
    line_idx += 1;
    full_decl.push(' ');
    full_decl.push_str(lines[line_idx].trim());
}
```

When a continuation line is a `//` comment, the space folds every declarator after it into that comment.

```svelte
<script>
	let a,
		b = 1,
// c
		c;
</script>

<p>{a}{b}{c}</p>
```

rsvelte before this PR:

```js
let a;
let b = 1;
let // c c;
```

`let // c c;` is a `let` with no declarator — the output does not parse. Upstream:

```js
let a;
let b = 1;

let // c
c;
```

The fix joins with a newline whenever the text so far ends inside a line comment, which reproduces upstream's shape.

## Which input found it

`pattern/matrix/ts-declarations/definite-assignment-multi-declarator__m0__line-with-paren` (client, client-dev) — one of the 16 `unparseable` entries keeping main's mutation sweep red, and the one no other session had claimed.

**The `-paren` in the name is a coincidence, and my earlier attribution on #2601 was wrong.** Two probes:

- both comment kinds break identically (`line-with-paren` → `let // ) c c;`, plain `line` → `let // c c;`), so no delimiter scan is involved;
- the JS-only form (`let a,\n\tb = 1,\n// c\n\tc;`, no `lang="ts"`) breaks identically, so it is not the TS-declaration strip either.

I said on #2601 that this entry was "newline loss in the TS-declaration strip, not a delimiter scan". The second half was right for the wrong reason; the first half was wrong. It is newline loss in the *declaration accumulator*, shared by the TS and JS paths.

## Verification

6 tests in `crates/rsvelte_core/tests/declaration_continuation_comment_2601.rs`.

Negative control — `git checkout HEAD -- transform_script.rs`, re-run, restore:

```
test result: FAILED. 3 passed; 3 failed
    a_line_comment_before_the_first_continuation_line_keeps_it
    a_line_comment_between_declarators_does_not_swallow_the_next_one
    the_same_holds_for_a_typescript_definite_assignment_list
```

The 3 that pass either way are controls: a block comment must still fold onto one line (`let /* c */ c;` — asserted byte-for-byte, so a blanket "always newline" fix would fail it), a comment-free multi-line list must be unchanged, and `'// x'` inside a string must not read as a comment.

Locally: `--test ssr` and `--test runtime` (hydration + legacy + runes) green in release.

## Base

Branched on #2628, which adds the `js_scan::ends_inside_line_comment` helper this fix calls. Merge #2628 first, or this PR carries that commit in.

### DIFF
--- a/.changeset/declaration-continuation-line-comment.md
+++ b/.changeset/declaration-continuation-line-comment.md
@@ -1,0 +1,23 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Keep the newline after a line comment that sits between two declarators.
+
+A multi-line `let`/`const`/`var` list is accumulated onto a single line before
+it is split at its top-level commas. The accumulator joined continuation lines
+with a space, so a `//` comment on one of them swallowed every declarator that
+followed:
+
+```svelte
+<script>
+	let a,
+		b = 1,
+// c
+		c;
+</script>
+```
+
+emitted `let // c c;` — a `let` with no declarator, and output that does not
+parse. Continuation lines now join with a newline whenever the text so far ends
+inside a line comment, which is also the shape upstream prints.
--- a/.changeset/import-comment-delimiter.md
+++ b/.changeset/import-comment-delimiter.md
@@ -1,0 +1,18 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Find an `import` statement's terminator lexically, so a `;` inside a comment does not truncate the specifier list
+
+`extract_imports` accumulated a multi-line `import` until a line "closed" it, and both the
+close test (`trimmed.contains(';')`) and `import_statement_end` read raw bytes. A `// ; c`
+line inside the specifier list closed the import after the previous specifier, terminated it
+with the comment's own `;`, and routed the rest of the statement — starting mid-comment —
+into the component body, so the output stopped being JavaScript.
+
+The two tests are now one lexical scan, which is what kept them from disagreeing: comments,
+template literals and regex literals are opaque, and the open-block-comment state already
+carried by `ScanState` is consulted, so a `;` on the continuation line of a `/* … */` is text
+too. All four `contains(';')` sites are replaced — `extract_imports` and
+`extract_imports_with_projection` are two copies of the same loop, and fixing one would have
+left the other live depending on whether source projection is on.
--- a/.changeset/rune-argument-trailing-comment.md
+++ b/.changeset/rune-argument-trailing-comment.md
@@ -1,0 +1,15 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Close a rune field's `$.derived(…)` on a new line when its argument ends in a `//` comment
+
+The server class-field path splices the `$derived(…)` argument verbatim and then appends the
+closing paren. `value.trim()` removes the newline that ended a trailing `//` comment, so the
+paren landed inside the comment and the call was never closed — the emitted module stopped
+being JavaScript.
+
+An object-literal argument is worse, because it takes a wrapping-paren branch and loses two.
+
+The variant carrying a delimiter (`// ) c`) already worked: it bails to the AST path, which
+relocates the comment. It is the *plain* comment that was unguarded here.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,6 +40,23 @@ kept only as a fallback for comment-bearing / unsupported-node programs. The rem
 processing (client visitors building `Raw` strings, `shared/async_body.rs`, the `.svelte.js`
 module path) is internal IR construction with unchanged output — a maintainability cleanup only.
 
+### Where compile time goes ([`docs/phase3-ast-refactor-plan.md`](docs/phase3-ast-refactor-plan.md) § Findings 2026-08-08)
+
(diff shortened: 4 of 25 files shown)


baseballyama:
Verified independently against a control, because I had raised a structural doubt about this PR
and it turns out to be wrong. Recording both.

**The doubt.** #2636's only source change is in `server/transform_script.rs`, while the mutation-fuzz
survivors it is meant to close — `pattern/matrix/ts-declarations/definite-assignment-multi-declarator`
and `svelte-checkbox/Checkbox` — fail on **client and client-dev**. There are also four separate
`split_declarators` implementations in the tree (`server/transform_legacy.rs:458`,
`server/helpers.rs:1385`, `shared/async_body.rs:2193`, `client/props_transforms.rs:2396`), so
"the declaration splitter" is not one thing.

**The measurement.** Same fixture, `compile_one`, `node --check`, two builds:

```svelte
<script lang="ts">
	let a!: string,
		b = 1,
// ) c
		c!: number;
</script>
<p>{a}{b}{c}</p>
```

| | client | client-dev | server |
|---|---|---|---|
| main `5f73abbd` | `let // ) c c;` — **does not parse** | **does not parse** | fine |
| this PR | fine | fine | fine |

The control moves and it fails in the predicted way, so the attribution is clean: **#2636 closes all
four client/client-dev declarator survivors.** My structural argument was wrong — "which module the
function lives in" is not "which target reaches it", and I should have run this before raising it.

**One thing worth a follow-up, not a change here.** On main the **server** target already emits this
correctly; only client and client-dev were broken. So a fix in `server/transform_script.rs` is
resolving a client-only symptom, which means `split_comma_separated_declarations` is reached from the
client path despite its module. That is a naming hazard rather than a defect — it is exactly what made
me doubt a correct PR — and the next person auditing "does this reach target X" will trip on it the
same way. A one-line doc comment on the function saying which targets call it would pay for itself.

After this merges, the remaining mutation-fuzz survivors are the **two server `.svelte.ts`
arrow-object-literal cases** (bits-ui/toggle, layerchart/Circle.shared), which have a minimisation
but no PR yet.

baseballyama:
**Correcting my own comment above: #2636 closes two of the four declarator survivors, not four.**
I verified one fixture and generalised it to the family. The four are **two seeds × two targets**,
and I only tested one seed's shape.

Re-measured on this branch (`e6f32302`), both shapes, both client targets:

| shape | client | client-dev |
|---|---|---|
| `definite-assignment-multi-declarator` (the fixture in my comment above) | **fixed** | **fixed** |
| `svelte-checkbox/Checkbox` | still fails | still fails |

The Checkbox shape needs no delimiter at all — six lines, an ordinary `// x c`:

```svelte
<script>
  let a = "",
    // x c
    b = "";
  export { a, b };
</script>
<p>{a}{b}</p>
```

On this branch the client emits:

```js
let a = $.prop($$props, 'a', 8, "");
  let // x c
b("");            // SyntaxError: Unexpected token '('
```

`b`'s entire ` = $.prop($$props, 'b', 8, ` is gone. On current `main` the same input fails
differently — everything joins onto the comment line as `let // x c b = ""; export { a, b };` — so
**#2636 changes this input without fixing it**, which is precisely why a single-fixture check reads
as a clean kill.

**This is a second defect stacked behind the first, and it is a consequence of #2636 rather than a
flaw in it.** #2636 correctly joins with a newline, which splits the declaration across lines;
`transform_let_with_reexported_props` then runs per line, sees `let // x c` with no declarators after
comment-stripping, returns `None`, and the orphaned `b = ""` on the next line is picked up by the
prop-*assignment* rewrite as `b("")`. Removing the comment is correct; making it a **block** comment
is correct — so the trigger is line-comment-ness, not the `;`, and `line-with-semi` in the seed name
is incidental.

**#2636 should still land.** It is a strict improvement, its own fixture is a real fix, and the
Checkbox failure exists on `main` today in a different form. The family needs a second PR.

Also retracting the reachability note in my earlier comment: the four `split_declarators`
implementations I listed are a **different function** with an unrelated call graph — same-sounding
name. `split_comma_separated_declarations` is called from `client/mod.rs:4592` and `:4647`,
fully-qualified into `server::transform_script`. The doc-comment suggestion stands; the evidence I
gave for it did not.

baseballyama:
Subsumed by #2642, which merged as a squash of a branch that had this one merged into it (`39e42cdf`).

Verified by content rather than by SHA, since a squash preserves neither:

- `git grep -c ends_inside_line_comment origin/main -- .../server/transform_script.rs` → **2** (this PR's only source hunk plus its existing definition site).
- `crates/rsvelte_core/tests/declaration_continuation_comment_2601.rs` is present on `origin/main`.

Both halves matter: the test file alone would only show that *a* commit named for this defect landed, and the hunk alone would not show the regression test came with it.

PR #2641, opened 2026-08-08
perf(profile): split process_accumulated into named stages with work counters

baseballyama:
> ## The split, as min over n=6, per corpus
>
> | stage | smelte | carbon/src | open-webui |
> |---|---|---|---|
> | `reactive_stmt` | **14.86%** | **9.83%** | 2.18% |
> | `stores` | 0.14% | 0.37% | **2.41%** |
> | `state_assigns` | 1.09% | 1.42% | **2.12%** |
> | `legacy_state_declarations` | 0.64% | 0.66% | 1.54% |
> | `state_reads` | 0.76% | 1.15% | 1.38% |
> | **`export_let`** | **3.51%** | 1.94% | 0.51% |
> | `prop_assignments` | 1.65% | 1.80% | 0.92% |
> | `member_mutations` | 0.65% | 0.43% | 0.73% |
> | min total compile | 27.51 ms | 350.96 ms | 1654.75 ms |
>
> Interference can only *add* time to a stage, so the minimum over repeats is the least-contaminated observation and needs no identification of which runs were bad. Comparing two arms by their minima is only fair at equal n with comparable interference.
>
> **`min` upper-bounds a *time*; it does not bound a *share*.** With `P = p + δp` and `S = s + δs`, the measured share is below the true one iff `δs/s ≤ δp/p` — the sign is set by *relative* contamination. The denominator aggregates contamination from every stage while a stage absorbs only its own, so **for any stage that is not the main absorber these percentages are likely underestimates**, not caps. The ms columns are upper bounds; the % columns are estimates.
>
> **The estimator dissolves the "disjoint top rows" claim, which was itself contamination.** `reactive_stmt` is #1 on smelte and carbon and #2 on open-webui (0.2 points off the top). `member_mutations`, which looked like open-webui's top row, spans min 12.108 / max 100.579 ms (8.31×) — under min it falls to 7th at 0.73%. Changing the estimator made three corpora agree, which is evidence the estimator is working.
>
> ## ⚠️ The single-run share tables further below are WITHDRAWN
>
> Six consecutive runs per corpus (counters bit-identical) show per-stage share spreads of **1.54×** (smelte), **17.45×** (carbon/src) and **30.00×** (open-webui — with the parent stable to 1.39×, so the perturbation is per-stage and does not divide out of within-run shares).
>
> **Every share number below that comes from carbon/src or open-webui is a single-run point estimate and is not measurement.** So is the ordering-inversion finding, and so is `export_let`'s "3.6–14× its siblings". See [the withdrawal comment](https://github.com/baseballyama/rsvelte/pull/2641#issuecomment-5225398930).
>
> ### The result that does hold: a paired sign test over 18 runs
>
> | comparison | wins | p (two-sided) |
> |---|---|---|
> | `export_let` > `prop_source_reads` / `state_assigns` / `state_reads` / `member_mutations` / `destructure_assignments` / `legacy_state_declarations` | **18/18** each | 0.00001 |
> | `export_let` > `prop_assignments` | **17/18** | 0.00014 |
>
> **`export_let` costs more per byte than every sibling in the chain**, unit price **77.8–200.3 ns/byte**. Per-stage perturbation destroys magnitudes but not *direction* — carbon's one 1727.9 ns/byte outlier counts as a single win rather than a 14× ratio, which is the whole reason to use the sign.
>
> Negative control, same test on the withdrawn inversion: `member_mutations` > `prop_source_reads` is **4/6 on open-webui (p=0.69, not significant)** and **0/6 on smelte and carbon (p=0.031, significant the other way)**. The instrument supports one claim and kills the other.
>
> The instrumentation itself stands. The body is left intact below rather than rewritten, so the withdrawn claims stay visible.

Instrumentation only. No compiler behaviour changes, no optimisation.

*(Two corrections were made during review and are preserved in the comments below rather than edited away: I initially claimed the prior 69% figure was retracted — it is not — and I initially led with `export_let`, which is a real anomaly but a small one.)*

## What this adds

The `process_accumulated` closure in `client/mod.rs` (~620 lines) is split into **22 named stages**, each with a `Duration` and a **deterministic work counter** (call count, plus input bytes — accumulated lines for `join`), and an explicit `other` row so the shares are checkable against the parent instead of trusted.

Timers sit behind a new `measure-pa-split` feature (off by default, zero-cost when off), because the chain runs per statement and 22 `Instant` pairs land **inside the parent they are being compared to**. The profiler calibrates one pair in the same process and prints the bound as `PA-OVERHEAD`.

Also carries the `compile_profile` warmup-drain fix from #2633, because this branch is cut from `main` and the split cannot be measured on a contaminated binary. If #2633 lands first, that hunk resolves to the same content.

## The split, in share of compile

Share-of-parent is not reported: `pa_rest/parent = 1 − reactive_share − runes_share` by definition, so it restates its siblings and moves when they move. Absolute share is what sizes work.

| stage | smelte | svelte-ux | **carbon/src** | **open-webui** |
|---|---|---|---|---|
| `reactive_stmt` | **14.75%** | **8.20%** | **19.34%** | 1.92% |
| `member_mutations` | 2.51% | 2.32% | 0.50% | **5.59%** |
| `state_assigns` | 3.91% | 1.34% | **6.04%** | **4.65%** |
| `stores` | 0.55% | 1.30% | 0.53% | **2.71%** |
| `legacy_state_declarations` | 2.30% | 3.20% | 0.72% | **2.55%** |
| `state_reads` | 2.78% | 4.03% | 1.21% | 1.29% |
| `export_let` | 3.49% | 1.38% | 1.95% | 0.53% |
| `prop_assignments` | 6.06% | 4.62% | 1.95% | 1.03% |

**The two applications do not share a top row** — carbon is `reactive_stmt` at 19.3%, open-webui is `member_mutations` at 5.6% — so there is no single next target at this resolution.

Corpora measured: smelte (74 files / 765 statements), sveltestrap (129 / 1177), svelte-ux (198 / 1581), shadcn-svelte (1682 / 1951), carbon/src (291 / 6101), open-webui (650 / 7680), SMUI (449 / 1617).

## The prior 69% stands; the original numbers reproduce

| repo | `pa_rest` **orig** | `pa_rest` **now** | `reactive_stmt` orig → now | `process_accum` orig → now |
|---|---|---|---|---|
| SMUI | 1.7% | **1.9%** | 0.00 → 0.00 | 2.0 → 2.1 |
| carbon/src | 16.8% | **17.4%** | 13.4 → 19.3 | 30.3 → 36.8 |
| open-webui | 22.4% | **20.9%** | 3.8 → 1.9 | 26.3 → 22.9 |

`pa_rest` in share-of-compile reproduces on all three; the epic's aggregate 15.9% survives. carbon's apparent 8-point gap was in the **denominator**, not in `pa_rest`. No pin exists for the application corpora (`corpus-sources.json` has none; the notes record only file counts), so the revision caveat could not have been closed by checkout — but it is moot.

A `$:`-density model I recorded before measuring open-webui (predicting `pa_rest` 81–87% at 2–3% density) **failed by 29 points**: actual density 7.66%, actual `pa_rest` 91.4%, model prediction 62%. `$:` acts through absolute count against total compile work, not density per statement.

## `export_let`: anomalous unit price, small size

| corpus | ms | calls | bytes | **ns/byte** | sibling `prop_assignments`, same run | ratio |
|---|---|---|---|---|---|---|
| smelte | 0.954 | 369 | 12411 | **76.9** | 21.4 | 3.6× |
| sveltestrap | 1.109 | 369 | 10774 | **102.9** | 7.2 | 14.2× |
| svelte-ux | 1.878 | 647 | 22129 | **84.9** | 12.1 | 7.0× |
| carbon/src | 6.629 | 1665 | 54645 | **121.3** | 15.2 | 8.0× |

2.6–4.0µs for a ~30–34 byte statement. ~~3.6–14× its chain siblings per byte~~ — **this magnitude is withdrawn**; see the sign test in the banner above for the form that holds. Cause: `transform_export_let` → `state_pipeline_ast` runs an AST pass per statement. **But it is 0.53–3.49% of compile** — unit price and size are different questions.

## Unexplained and worth a counter

On open-webui, `member_mutations` + `state_assigns` + `stores` + `legacy_state_declarations` total **15.5% of compile**, and `member_mutations` alone runs **9.5× `prop_source_reads`** in the same run — while on smelte it is *cheaper* than `prop_source_reads`. The ordering inverts within a run, so load cannot explain it. open-webui statements average 220 bytes vs smelte's 69, and these are the stages that scan per state/store variable: suspected scaling in the **product** of statement size and variable-set size. `[S]`, unmeasured.

Runes pole: `destructure_assignments` is the only row above the instrument floor on two independent runes corpora (shadcn 34.4% of parent, SMUI 51.9%). It is the only chain stage without a `!analysis.runes` guard. Still `[S]` — the discriminating count comes first.

## How to read the numbers, and what I did not trust

- **Within-run shares only.** Re-running smelte moves `process_accum` 7.33 → 8.59ms (+17%) while `export_let`'s share moves 13.0% → 14.2%.
- **Cross-binary ON/OFF timing is unusable on this machine**: svelte-ux reported `process_accum` **41.30ms feature-off vs 21.06ms feature-on**.
- **The counters are the load-immune channel and agree exactly**: 765 / 1177 / 1581 / 1951 statements, identical feature-on and feature-off.
- **`other` is the instrument, not a residual.** 2.1–3.4% of parent on legacy corpora, 20.9–24.6% at the runes pole; shadcn's 1.469ms `other` matches its ON−OFF delta of 1.44ms.
- **Do not lean on `PA-OVERHEAD`.** Its calibration loop is descheduled under load: carbon reported 330.8ns/pair (elsewhere 43–95ns), a 27.4ms bound against a measured `other` of 2.66ms. `other` is tighter; the coverage claim rests on it.

## Verification

`cargo test --release -p rsvelte_core --test runtime --test validator` (2 binaries requested, 2 `Running` lines, all pass), plus `compiler_fixtures`, `ssr`, `client_transforms_pin`, `client_script_comments_pin`. `cargo fmt --check` clean; `clippy -D warnings` clean **both** with and without `measure-pa-split`.


baseballyama:
## Correction: the 69% is not retracted, and my framing of it was wrong

The PR body originally said the 69% figure was "retracted, not refined". That was an over-general claim built on a population difference, and measuring the application corpora settles it against me.

### `pa_rest`'s share is a function of `$:` density, not of library-vs-application

| corpus | kind | files | statements | `$:`/stmt | `process_accum` | `reactive_stmt` (of parent) | **`pa_rest`** |
| smelte | lib | 74 | 765 | 12.8% | 26.8% of compile | 55.0% | **44.8%** |
| sveltestrap | lib | 129 | 1177 | 10.7% | 21.6% | 48.7% | **51.1%** |
| svelte-ux | lib | 198 | 1581 | 12.4% | 15.5% | 53.0% | **46.8%** |
| **carbon/src** | **app** | **291** | **6101** | **12.5%** | **36.8%** | **52.5%** | **47.4%** |
| shadcn-svelte | lib (runes) | 1682 | 1951 | 0% | 1.1% | 0.0% | 89.5% |
| **SMUI** | **app (runes)** | **449** | **1617** | **0%** | **2.1%** | **0.0%** | **90.5%** |

`pa_rest/parent = 1 − reactive_share − runes_share` by definition, so this column is just a restatement of `$:` density. And **all four of my legacy corpora sit at 10.7–12.8% `$:` per statement — they are one point on that axis, not four.** carbon/src, an application at 87.9% legacy bytes, lands at 47.4%, inside the library range; SMUI, an application, lands at 90.5% next to the shadcn library. The library/application split does not predict this column; `$:` does.

**So the 69% aggregate stands.** open-webui's 85% implies a `$:` density of roughly 2–3% — an intermediate point that neither of my clusters (12% and 0%) reaches. carbon/src alone came out at 47.4% against the original 55%, and I cannot attribute that gap: warmup contamination in the original, an arithmetic error, or **revision drift** (the clone is today's HEAD, not the original pin) all remain open. Scope is ruled out — I measured `src/` (291 files), not the 1,324-file root whose legacy density differs 3.2×.

The narrow claim the evidence does support: **at ~12% `$:` density, `pa_rest` is 45–51% of the bucket and `reactive_stmt` is the majority** — which points the epic at #2622, and does not depend on resolving the gap.

### The number that does not move: `export_let`'s unit price

| corpus | ms | calls | bytes | **ns/byte** | sibling `prop_assignments`, same run | ratio |
|---|---|---|---|---|---|---|
| smelte | 0.954 | 369 | 12411 | **76.9** | 21.4 | 3.6× |
| sveltestrap | 1.109 | 369 | 10774 | **102.9** | 7.2 | 14.2× |
| svelte-ux | 1.878 | 647 | 22129 | **84.9** | 12.1 | 7.0× |
| carbon/src | 6.629 | 1665 | 54645 | **121.3** | 15.2 | 8.0× |

**2.6–4.0µs for a ~30–34 byte statement; 77–121 ns/byte.** One caveat on the comparison you drew: `script_text`'s ns/byte uses **whole-file bytes** as its denominator while this uses **statement bytes**, so "slower per byte than the pipeline it sits inside" isn't like-for-like. The like-for-like version is the last column — against its own chain siblings, which are measured with the same denominator in the same run, `export_let` is **3.6–14× more expensive per byte**. Cause is structural: `transform_export_let` → `state_pipeline_ast` runs an AST pass per statement.

### Runes pole

`destructure_assignments` now stands up on **two independent runes corpora**: shadcn 34.4% of parent (1950 × 1.05µs), SMUI **51.9%** (1617 × 1.58µs). Every other row on both sits at 0.02–0.05µs/call, which is the instrument floor. Still `[S]`, not `[D]` — the runes ∧ three-vectors-non-empty count comes before anything else there.

### One instrument caveat found on carbon

`PA-OVERHEAD` calibrates in-process but the calibration loop is itself descheduled under load: carbon reported 330.8ns/pair (other runs: 43–95ns), giving a 27.4ms bound against a measured `other` of 2.66ms. **`other` is the tighter of the two**, so the coverage claim should rest on `other`, not on the calibration.

baseballyama:
## open-webui closes the axis, and dissolves the disagreement: nothing was wrong with the original numbers

### The `$:` density model is falsified — prediction recorded first, missed by 29 points

Before measuring I recorded: if a `$:` statement's unit cost is constant, density `d` fixes `reactive_share`, and `d` = 2–3% → `pa_rest` 81–87%.

Measured: **`d` = 7.66%** (588 `$:` / 7680 statements) and **`pa_rest` = 91.4%**. The same model predicts **62%** at that density. The model is dead: `$:` does not act through density-per-statement. It acts through **absolute count against total compile work** — carbon has 765 `$:` in 291 files and `reactive_stmt` = 19.3% of compile; open-webui has 588 in 650 files and 1.9%.

Both my inference ("~12% is one point on the axis") and yours ("85% implies 2–3% density") were readings of a ratio that carries no independent information.

### Share of compile: the original numbers reproduce

| repo | `pa_rest` **orig** | `pa_rest` **now** | `reactive_stmt` orig → now | `process_accum` orig → now |
|---|---|---|---|---|
| SMUI | 1.7% | **1.9%** | 0.00 → 0.00 | 2.0 → 2.1 |
| carbon/src | 16.8% | **17.4%** | 13.4 → 19.3 | 30.3 → 36.8 |
| open-webui | 22.4% | **20.9%** | 3.8 → 1.9 | 26.3 → 22.9 |

**`pa_rest` in share-of-compile reproduces on all three.** carbon's 8-point "gap" was never in `pa_rest` (16.8 → 17.4, within noise); it was in the **denominator**, `reactive_stmt` moving 13.4 → 19.3. The epic's aggregate 15.9% of compile survives.

**The revision caveat is therefore moot, and I could not have closed it anyway**: no pin for carbon exists in the repo or the measurement notes — `corpus-sources.json` carries no application corpora, and the notes record only the file count (287; I measured 291). I've written that into the doc rather than leaving it as a live suspicion.

### Re-ranked in share of compile — this changes the answer

| stage | smelte | svelte-ux | **carbon/src** | **open-webui** |
|---|---|---|---|---|
| `reactive_stmt` | **14.75%** | **8.20%** | **19.34%** | 1.92% |
| `member_mutations` | 2.51% | 2.32% | 0.50% | **5.59%** |
| `state_assigns` | 3.91% | 1.34% | **6.04%** | **4.65%** |
| `stores` | 0.55% | 1.30% | 0.53% | **2.71%** |
| `legacy_state_declarations` | 2.30% | 3.20% | 0.72% | **2.55%** |
| **`export_let`** | 3.49% | 1.38% | 1.95% | **0.53%** |
| `prop_assignments` | 6.06% | 4.62% | 1.95% | 1.03% |

**`export_let` — which I led the PR with — is 0.53–3.49% of compile.** Its unit price is still anomalous (3.6–14× its siblings per byte, same-run same-denominator), but unit price and size are different questions and I conflated them by leading with it. Demoted in the doc and the PR body.

**The two applications do not share a top row**: carbon is `reactive_stmt` at 19.3%, open-webui is `member_mutations` at 5.6%. There is no single "next target" at this resolution.

### New, and the most interesting thing here

On open-webui, four stages — `member_mutations` + `state_assigns` + `stores` + `legacy_state_declarations` — total **15.5% of compile**, and `member_mutations` alone runs **9.5× `prop_source_reads`** in the same run. On smelte `member_mutations` is *cheaper* than `prop_source_reads`. **The ordering inverts within a run**, so load cannot explain it. open-webui statements average 220 bytes against smelte's 69, and these are the stages that scan per state/store variable — so the suspicion is scaling in the **product** of statement size and variable-set size. `[S]`, unmeasured; it needs a variable-set-size counter, which is one more counter beside the timers that are already there.

baseballyama:
## They are right. Within-run shares are not load-immune, and several of my claims above are withdrawn.

Six consecutive runs, same binary, same corpus, counters bit-identical in every run (`statements` / `pairs`: smelte 765 / 7465, carbon/src 6101 / 82833, open-webui 7680 / 110956).

### Per-stage share spread

| corpus | files | parent spread | **max share spread** | stage |
|---|---|---|---|---|
| smelte | 74 | 1.63× | **1.54×** | `stores` |
| carbon/src | 291 | 2.93× | **17.45×** | `export_let` (3.07% ↔ 53.57%) |
| open-webui | 650 | **1.39×** | **30.00×** | `rest_prop_member_access` (0.52% ↔ 15.73%) |

**The open-webui row settles it.** The parent moves only 1.39× while an individual stage's share moves 30×. A uniform slowdown cannot redistribute a stable total — the perturbation is per-stage, exactly as the other investigator found, and it does not divide out of within-run shares.

Both prior results were real: smelte at n=2 was stable because smelte *is* stable; carbon at n=5 was unstable because carbon *is* unstable. I generalised from the small corpus.

Mechanism, `[S]`: the perturbation is time-windowed (core migration, frequency scaling, competing processes) and corpus files are processed **in order**. Files are heterogeneous, so whichever stages are active during a window absorb its cost. Longer runs have more exposure — consistent with smelte stable and carbon/open-webui not.

### Withdrawn

**1. The ordering inversion.** I claimed `member_mutations` > `prop_source_reads` on open-webui and the reverse on smelte was "the one class of evidence machine load provably cannot manufacture." Tested at n=6:

| corpus | runs with `member_mutations` higher | ratio range |
|---|---|---|
| smelte | 0 / 6 | 0.45–0.61 |
| carbon/src | 0 / 6 | 0.04–0.36 |
| **open-webui** | **4 / 6** | **0.34–5.12 (15×)** |

**The ordering flips between runs of the same binary on the same corpus.** My claim rested on open-webui n=1. Ordinal comparisons are more robust than ratios but not sufficient: if the perturbation is per-stage, the ordering moves too.

**2. The 10–11× "disjoint top rows" headline.** carbon's `export_let` alone spans 17.45× across runs of one corpus, so a 10–11× cross-corpus difference sits *inside* single-corpus run-to-run spread. Not established.

**3. `export_let` at "3.6–14× its siblings."** That range was one run per corpus read as a cross-corpus range; it is run-to-run noise. On carbon the ratio spans 0.74–118.79× (spread 160×) and one run has `export_let` *cheaper* per byte.

### What survives

- **The counters.** Bit-identical across ON/OFF and across all 18 runs. This remains the only load-immune channel, and it is why the `calls` / `work` columns are worth more than the `ms` beside them.
- **`export_let`'s absolute unit price**: 77.8–200.3 ns/byte across 18 runs, 2.6–4.0µs for a ~30–34 byte statement. Notably **`export_let` is the stable term and the sibling is the noisy one** — carbon's `prop_assignments` spans 14.5–174.4 ns/byte. Ordinally, `export_let` > `prop_assignments` in **17 of 18 runs**.
- **smelte-scale shares** (≲100 files): max spread 1.54×, usable.

### Consequence for this PR

The instrumentation stands — the split closes, `other` is still the instrument, and the counters do their job. **The share tables do not**, for carbon and open-webui. I have marked them withdrawn in the doc rather than deleting them, with the reason attached, and I am **not** proposing this land with point-estimate shares.

The pre-registered bucketed discriminator now inherits a precondition it did not have: it is only viable on carbon and open-webui, which are precisely the two corpora with 17–30× spread. Every bucket cell must be reported as an n≥6 min/max range, and **overlapping ranges must be reported as `unmeasured`** — otherwise the discriminator inherits the defect it was designed to resolve.

baseballyama:
## n=5 repeats: your smelte reading reproduces, but it does not generalise to carbon

I was asked to sub-split `reactive_stmt` and hit a stability problem that bears directly on the shares published here. Short version: **I reproduce your smelte result, and I think your numbers are right for smelte.** What I dispute is only the scope of the "within-run shares cancel load" argument, which is not a claim you made — it is one that was given to you, and it happens to be corpus-dependent.

### Interleaved design

Five repeats of carbon (`carbon-components-svelte/src`, 291 files) and five of smelte (74 files), **alternating within one session**, same binary, seconds apart. Alternating is the point: it separates "these stages are load-sensitive" from "that half-hour was bad", which five consecutive runs of one corpus cannot.

`rs` = `reactive_stmt`; the share is parent-relative (`reactive_stmt / process_accum`), i.e. your readout.

| run | CARBON rs/proc_acc | CARBON rs_deps/rs | SMELTE rs/proc_acc | SMELTE rs_deps/rs |
|---|---|---|---|---|
| 1 | 44.9% | 79.5% | 51.6% | 46.9% |
| 2 | 53.1% | 82.4% | 57.1% | 50.2% |
| 3 | 27.8% | 47.8% | 51.2% | 49.2% |
| 4 | 18.7% | 47.4% | 56.5% | 50.2% |
| 5 | 27.5% | 34.1% | 54.3% | 55.8% |
| **min–max** | **18.7–53.1 (2.8x)** | **34.1–82.4 (2.4x)** | **51.2–57.1 (1.11x)** | **46.9–55.8 (1.19x)** |

Absolute `reactive_stmt`: carbon 30.18–156.79 ms (5.2x), smelte 4.12–6.90 ms (1.67x).

Every deterministic counter was **bit-identical in all ten runs**: carbon `pairs=23273 reactive_calls=765`, smelte `pairs=1641 reactive_calls=97`. So nothing about the work moved; only the clock did.

### What this means for #2641

1. **Your smelte numbers reproduce.** I get 51.2–57.1% where you report 55.0% → 54.4%. Both of your points sit inside my five-run range. On smelte, the parent-relative share really is stable, and a point estimate is defensible.
2. **The same readout is not stable on carbon** — 18.7–53.1% in the same alternating session. So the load-cancelling property is not a property of the *readout*; it is a property of the readout **on a small corpus**.
3. **The load-window confound is ruled out.** Smelte stayed stable while carbon swung 2.8x in the same minutes, measured alternately. "That half-hour was bad" does not survive the interleaving.
4. Within-run shares cancel a *uniform* slowdown. The perturbation here is **per-stage**: in my earlier consecutive-run set, one run inflated `rs_deps` while `rs_body` stayed flat and the next did the reverse. A ratio between two siblings that are perturbed independently does not cancel.

Hypothesis for the asymmetry, offered as a hypothesis and not a result: carbon's run is ~8x longer in wall time, so it intersects more of the concurrent `cargo build` spikes on this box, and its per-stage times are large enough to absorb multi-ms scheduler stalls. If that is right, **instability grows with corpus size** — the more realistic the corpus, the less trustworthy the timer, which is an uncomfortable direction.

### Suggested change to the PR

Not "withdraw the shares" — for smelte they are sound. Rather: report **min/max over n>=5 per corpus** rather than point estimates, and state the corpus each share was measured on, so a smelte-calibrated share is never read as a compiler-wide constant. Anything carbon-sized needs the range or it will be over-read.

### Unrelated but load-immune

The counters were the only thing that survived. `extract_reactive_statement_deps` does **30.4 variable-scans per `$:` statement** on carbon (23,273 over 765), and `ns/pair_byte` sits at 5.9–11.5 across five corpora while the timers vary 8x — so its cost is O(variables x body bytes) as a measurement, not a hypothesis. That makes `rs_deps` a fifth member of the same per-variable-scan family as `member_mutations` and `state_assigns`.

Caveat on my own table: the `rs_body/rs` and `rs/total` columns in my harness were broken by an awk defect and printed 0.0%. They are **unmeasured**, not zero, and I have left them out above.

baseballyama:
## Addendum: instability tracks run scale, not corpus identity — and the right estimator is the minimum

Follow-up to my previous comment. I ran the control that separates "carbon is unstable" from "long runs are unstable", and it changes the prescription — including part of what I said last time.

### Design

A **73-file subset of carbon/src**, sampled every 4th file so the mix is preserved. It has the *same density as full carbon* (0.75 `$:`/KB, identical) and the *same file count as smelte* (73 vs 74). So it holds corpus identity and density fixed and varies only scale. All three arms interleaved in one session, 5 repeats each.

Excluding each arm's first run (every arm's r1 is anomalous — cold page cache):

| arm | bytes | rs/procacc (r2-r5) | rs_deps/rs | rs_body/rs |
|---|---|---|---|---|
| smelte | 124 KB | 56.7-57.0 (**1.01x**) | 49.8-50.4 | 40.0-40.5 |
| carbon74 | 253 KB | 44.3-45.2 (**1.02x**) | 46.8-48.1 | 40.4-41.4 |
| carbonFULL | 1015 KB | 10.1-77.3 (**7.6x**) | 9.7-48.6 | 42.0-88.7 |

`pairs` was bit-identical within every arm across all runs (1641 / 7505 / 23273).

**carbon74 is the same code as carbonFULL and is as stable as smelte.** Instability is a property of scale, not of carbon. That argues for **chunking large corpora**, not for retreating to small ones.

### Two findings that matter more than the headline

**1. Wall time does not identify a contaminated run.** carbonFULL r2 (wall 1024 ms, rs=198.8 ms) versus r4 (wall 1004 ms, rs=30.2 ms): near-identical total duration, 6.6x different `reactive_stmt`. The perturbation is internal to stages and invisible in the total. So "discard the slow runs" does not work as a filter — I would have reached for it, and it would have silently kept the bad ones.

**2. Contamination is strictly one-sided, so the estimator should be the minimum, not the mean.** Scheduler interference can only *add* time to a stage; it can never subtract. Taking `min` over n>=5 rather than mean/median is therefore principled rather than ad hoc, and it is robust to the fact that the contaminated runs are not identifiable individually.

At each arm's minimum-`rs` run, **all three arms agree**:

- `rs_deps / reactive_stmt` = **47-50%**
- `rs_body / reactive_stmt` = **40-43%**

This is not circular: smelte and carbon74 are independently stable across 4/5 runs and calibrate the ratio; carbonFULL's minimum reproduces it. Its 3 contaminated runs are the ones that inflate `rs_body` to 85-89% and crush `rs_deps` to ~10%.

### Correcting myself

In my previous comment I said this bucket "cannot be sized" on this machine. **That was an over-claim.** With min-over-n>=5 it can be sized, and the answer is stable across three arms at two scales. The revised prescription for this PR:

- report **min over n>=5**, not a point estimate and not a mean;
- **name the corpus** each share was measured on;
- for anything above ~250 KB / ~300 ms, **chunk and aggregate** rather than measuring the whole corpus in one run.

Your smelte tables remain correct, and are now additionally corroborated: I get `rs_deps/rs` = 49.8-50.4% on smelte across four runs.

### Bearing on the lead

`rs_deps` being ~half of `reactive_stmt` — consistently, at two scales, on three corpora — strengthens the per-variable-scan lead rather than weakening it. `extract_reactive_statement_deps` does 30.4 variable-scans per `$:` statement on carbon, and `pairs` remains the load-independent instrument for scoring any fix to it.

Method note: the `rs_body/rs` column my harness previously printed as `0.0%` was an awk defect and I withheld it as unmeasured. It is fixed and measured above.

PR #2654, opened 2026-08-08
fix(transform): stop legacy $: order from depending on whitespace

### LINKED ISSUE #2646: Legacy $: statement order depends on whitespace: `{mid=seed*2}` and `{ mid = seed * 2 }` compile to different execution order
## Summary

Two legacy `$:` programs with **identical ASTs**, differing only in whitespace around `=`, compile to **different reactive statement execution order**. At most one can be correct, so this is a live reactivity defect independent of any performance work.

It is currently invisible to every gate.

## Reproduction

Both files below are semantically identical — the only difference is the spaces inside the block.

`A-nospace-assign.svelte`:

```svelte
<script>
  export let seed = 1;
  let mid = 0;
  let out = 0;
  $: out = mid + 1;
  $: {mid=seed*2}
</script>

<p>{out}</p>
```

`B-spaced-assign.svelte`:

```svelte
<script>
  export let seed = 1;
  let mid = 0;
  let out = 0;
  $: out = mid + 1;
  $: { mid = seed * 2 }
</script>

<p>{out}</p>
```

### Output (client, via `compile_one`)

**A — wrong order.** `out` is computed before `mid` is assigned, so it reads a stale `mid`:

```js
$.legacy_pre_effect(() => ($.get(mid)), () => {
	$.set(out, $.get(mid) + 1);
});

$.legacy_pre_effect(() => ($.deep_read_state(seed())), () => {
	$.set(mid, seed() * 2);
});
```

**B — correct order.** `mid` is assigned first:

```js
$.legacy_pre_effect(() => ($.deep_read_state(seed())), () => {
	$.set(mid, seed() * 2);
});

$.legacy_pre_effect(() => ($.get(mid)), () => {
	$.set(out, $.get(mid) + 1);
});
```

`out` depends on `mid` and `mid` is assigned by the other statement, so topological order requires the `mid` assignment first. B is correct; **A is wrong**.

## Cause

`sort_reactive_statements` is fed by `extract_reactive_statement_deps`
(`crates/rsvelte_core/src/compiler/phases/3_transform/client/reactive_transforms.rs:23`),
which detects "is this variable assigned anywhere in the body" by **text scanning**, not from the AST.

The assignment detection in `is_assigned_anywhere_in_body`
(same file, line 111) matches against literal operator strings that **require surrounding spaces**:

```rust
let assign_patterns = [
    " = ", " += ", " -= ", " *= ", " /= ", " %= ", " **= ", " &= ", " |= ", " ^= ", " <<= ",
    " >>= ", " >>>= ", " ??= ", " &&= ", " ||= ",
];
for assign_op in &assign_patterns {
    let pattern = format!("{}{}", var_name, assign_op);
    if let Some(pos) = body.find(&pattern) {
```

`mid=seed*2` contains no `" = "`, so the statement is not credited with assigning `mid`. The
dependency edge that would force the ordering is therefore never created, and the topological
sort leaves the two statements in source order.

Any assignment operator written without surrounding spaces is affected, not just `=`.

## Phase 2 already computes this correctly

`cycle_collect_assignments_and_deps`
(`crates/rsvelte_core/src/compiler/phases/2_analyze/mod.rs:1406`) walks the typed AST and routes
`AssignmentExpression` / `UpdateExpression` targets into `assignments`. It is whitespace-insensitive
by construction and gets both A and B right. Its result is currently used only for cycle detection
and then discarded.

So the fix direction is to feed the topological sort from the Phase-2 AST result rather than from
the text scan. Two details that will bite whoever does it:

- `cycle_collect_assignments_and_deps` pushes onto `reactive_stmts` **only when `assignments` is
  non-empty** (`mod.rs:1327`), so its ordinals do **not** line up with Phase 3's per-`$:` counter.
  It must push unconditionally to stay ordinal-aligned.
- The two use different population filters: Phase 2 keeps instance-scope bindings + root
  declarations, Phase 3 keeps `state_vars ∪ prop_vars ∪ store_sub_vars`.

Because the text scan is *wrong* on these shapes, switching should produce newly **passing** corpus
entries. The two-sided ratchet requires those be re-baselined in the same PR.

## What this issue is *not* claiming

Two further shapes were probed and came out **correctly ordered**; they are recorded here as
**non-discriminating as constructed**, not as evidence of additional defects:

- **Function-body dependency** — `$: out = [1,2].reduce((acc, n) => acc + mid * n, 0)` with
  `$: mid = seed * 2`. Phase 2's `cycle_collect_assignments_and_deps` skips function bodies (own
  scope) while the text scan sees them, so the two disagree on the dependency set — but the emitted
  order was correct either way here.
- **Comment-borne assignment** — `$: out = seed + 1; // mid = 999` with `$: mid = out * 2`. The
  comment contains the byte sequence `mid = `, which the text scan credits as an assignment and the
  AST cannot see. The phantom edge did not flip the order in this construction because source order
  already satisfied the constraint.

Both remain plausible divergence sources; they simply were not demonstrated by these probes.

## Environment note

Verified on `main` @ `39ba6489` (#2622, which moved the Phase-2 passes onto the typed AST and does
not touch `reactive_transforms.rs`). Reproduced with `cargo run --release -p rsvelte_devtools --bin
compile_one -- <file>`.

## Related, separate finding

`rehome_reactive_statement_comments`
(`crates/rsvelte_core/src/compiler/phases/3_transform/client/formatting.rs:229`) re-derives the
top-level `$:` byte spans lexically via its own `JsScan`, while Phase 2's typed AST already carries
those spans (`node.start().zip(node.end())`, used at `2_analyze/mod.rs:1328`). The trivia /
`svelte-ignore` / `has_successor` half of `rehome` is **not** in Phase 2 at all.

Recording this so it is not conflated with the fix above: anyone retaining Phase-2 data for
`rehome`'s benefit wants **spans**, not dependency sets. `rehome` shares no traversal, no data
structure and no call path with `extract_reactive_statement_deps`, so it is not a second consumer
of the dependency retention.

baseballyama:
Fixes #2646.

## What was wrong

`$: {mid=seed*2}` and `$: { mid = seed * 2 }` are the same program — identical
ASTs, differing only in whitespace — but they compiled to **different execution
order**:

| | before | after |
|---|---|---|
| `$: {mid=seed*2}` | `out` first, reading a stale `mid` — **wrong** | `mid` first ✅ |
| `$: { mid = seed * 2 }` (control) | `mid` first | `mid` first — **unchanged** ✅ |

`is_assigned_anywhere_in_body` decides which variables a `$:` statement assigns,
and that feeds the topological sort. It matched literal operator strings with
the spaces baked in:

```rust
let assign_patterns = [" = ", " += ", " -= ", /* … */];
let pattern = format!("{}{}", var_name, assign_op);
if let Some(pos) = body.find(&pattern) {
```

So the unspaced form was credited with assigning nothing, never acquired an
ordering edge, and ran before the statement whose value it consumes. Every
compound operator was affected, not just `=`.

## The fix

The scan now finds the name at a word boundary, skips any spaces/tabs, and
tests the operator run at that point. Comparisons stay excluded, and the
**longest** candidate wins so `<=` remains a comparison while `<<=` assigns;
`=` is rejected when followed by `=` or `>`.

## Evidence

**Both probes**, treatment and control — the control matters because it shows
the change fixes the broken shape without moving the shape that was already
correct. The two now emit byte-identical output modulo the component name.

`reactive_statement_order_ignores_whitespace_around_the_assignment` pins both.
Run against the unfixed scan it **fails, for the predicted reason** — it trips
the *unspaced* assertion, not the control:

```
unspaced ran out of order: ["$.set(out, $.get(mid) + 1);", "$.set(mid, seed() * 2);"]
```

`the_assignment_scan_separates_operators_from_comparisons` tables the operator
rule: `a=1`, `a **= 2`, `a<<=1`, `a >>>= 1` assign; `a==1`, `a=>1`, `a<=1`,
`a !== 1`, `ab=1`, `b.a=1` do not.

**Full corpus** (14,138 entries, all three targets), run against this exact
dylib:

```
match          13190     — identical to main's own figure
js-mismatch        0     js-unparseable 0     css-mismatch 0     error-mismatch 0
no warning regressions (32 known, unchanged)
no error regressions  (236 known, unchanged)
all 39569 generated modules parse
```

Zero movement on every gate. **This is the acceptance criterion, not merely a
good result**: the output-parity baseline is *empty* — main is at 0 known
output failures — so main already matches official on all 14,138 entries, and
any correct change here must produce exactly zero diff. There is no direction
in which a corpus entry could improve.

**Official fixtures**, release, `Running` lines counted against binaries
requested rather than trusting the exit code:

```
4 Running lines / 4 binaries requested
compiler_fixtures  3 passed  0 failed
runtime            5 passed  0 failed   (runtime-legacy 1209, runtime-runes 1013,
                                          hydration 80, runtime-browser 32)
ssr                2 passed  0 failed   (126 fixtures)
validator          3 passed  0 failed
```

Clippy `--all-targets --all-features` clean. No ratchet is regenerated: nothing
moved, so there is nothing to re-baseline.

### What those green suites do *not* say

They cannot vote on `$:` ordering. The official fixtures contain no `$:`
statement whose ordering depends on a reference inside a function body — during
an earlier attempt at this fix they stayed green through **44 corpus
regressions**. A suite that cannot express the shape is not weak evidence about
it; it is no evidence. Read them as "nothing unrelated broke."

## Deliberate deferral

This leaves the text scanner on the client instance-script path, which is the
defect class [the pipeline notes](../blob/main/CLAUDE.md) exist to remove.
That is a considered trade, not an oversight.

The principled fix is to sort on the AST sets upstream itself uses
(`order_reactive_statements` reads `reactive_statement.{assignments,dependencies}`).
That was implemented and measured here, twice:

- Sourcing them from `cycle_collect_assignments_and_deps` — the walk the
  reactive-**cycle** check performs and discards — reordered **44** corpus
  entries away from parity. That walk stops at function bodies: correct for
  cycle detection, wrong for ordering, because upstream's dependency set comes
  from `scope.references` and counts references inside functions.
- Sourcing dependencies from `reactive_statement_dependencies` and assignments
  from a visitor-mirroring walk brought it to **38**. Still a regression, and
  against an empty baseline every one of those is pure loss.

So the rewrite is real work with a real remaining defect, and it belongs with
the AST migration that deletes the scanner outright rather than being rushed in
beside a one-line correctness fix. #2646 gets fixed today at zero blast radius;
the ordering rewrite is tracked for that migration.

### DIFF
--- a/.changeset/reactive-order-whitespace.md
+++ b/.changeset/reactive-order-whitespace.md
@@ -1,0 +1,17 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Stop legacy `$:` statement order from depending on whitespace.
+
+`$: {mid=seed*2}` and `$: { mid = seed * 2 }` are the same program, but they
+compiled to different execution order. The scan that decides which variables a
+reactive statement assigns — which feeds the topological sort — matched the
+literal `" = "`, spaces included, so the unspaced form was credited with
+assigning nothing, never got an ordering edge, and ran before the statement
+whose value it produces. Anything reading `mid` then saw a stale value on first
+run. Every compound operator was affected too, not just `=`.
+
+The scan now finds the name and its operator separately, so any spacing (or
+none) is recognised. Comparisons are still excluded, and the longest operator
+wins, so `<=` stays a comparison while `<<=` assigns.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/reactive_transforms.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/reactive_transforms.rs
@@ -146,38 +146,58 @@ pub(super) fn is_assigned_anywhere_in_body(body: &str, var_name: &str) -> bool {
         }
     }
 
-    // Check for assignment operators: `var = ...`, `var += ...`, `var -= ...`, etc.
-    let assign_patterns = [
-        " = ", " += ", " -= ", " *= ", " /= ", " %= ", " **= ", " &= ", " |= ", " ^= ", " <<= ",
-        " >>= ", " >>>= ", " ??= ", " &&= ", " ||= ",
-    ];
-    for assign_op in &assign_patterns {
-        let pattern = format!("{}{}", var_name, assign_op);
-        if let Some(pos) = body.find(&pattern) {
-            // Verify the variable name is at a word boundary (not part of a longer name)
-            let before = if pos > 0 {
-                body.as_bytes()[pos - 1]
-            } else {
-                b' '
-            };
-            let before_ok = !before.is_ascii_alphanumeric()
-                && before != b'_'
-                && before != b'$'
-                && before != b'.';
-            if before_ok {
-                // Also make sure it's not `==` or `=>`
-                let after_eq = pos + var_name.len() + assign_op.len();
-                if assign_op == &" = " && after_eq < body.len() {
-                    let next = body.as_bytes()[after_eq - 1]; // the char after '='
-                    if next == b'=' || next == b'>' {
-                        continue;
-                    }
-                }
-                return true;
-            }
+    // Check for assignment operators: `var = …`, `var += …`, etc. The spacing
+    // between the name and its operator is not fixed, so this walks every
+    // occurrence of the name rather than matching `"<var> = "` literally —
+    // `$: {a=b}` assigns `a` exactly as `$: { a = b }` does.
+    let bytes = body.as_bytes();
+    let mut from = 0;
+    while let Some(rel) = memmem::find(&bytes[from..], var_name.as_bytes()) {
+        let pos = from + rel;
+        from = pos + var_name.len().max(1);
+
+        let before = if pos > 0 { bytes[pos - 1] } else { b' ' };
+        if before.is_ascii_alphanumeric() || before == b'_' || before == b'$' || before == b'.' {
+            continue;
+        }
+        let mut i = pos + var_name.len();
+        // The name has to end here, not run on into a longer identifier.
+        if let Some(&c) = bytes.get(i)
+            && (c.is_ascii_alphanumeric() || c == b'_' || c == b'$')
+        {
+            continue;
+        }
+        while matches!(bytes.get(i), Some(b' ' | b'\t')) {
+            i += 1;
+        }
+        if is_assignment_operator_at(bytes, i) {
+            return true;
         }
     }
 
+    false
+}
+
+/// Whether an assignment operator starts at `i`. `<=` and `>=` are comparisons
+/// while `<<=` and `>>=` are not, so the longest candidate has to win; `=` also
+/// begins `==`, `===` and `=>`, none of which assign.
+fn is_assignment_operator_at(bytes: &[u8], i: usize) -> bool {
+    const OPS: [&[u8]; 16] = [
+        b">>>=", b"<<=", b">>=", b"**=", b"??=", b"&&=", b"||=", b"+=", b"-=", b"*=", b"/=", b"%=",
+        b"&=", b"|=", b"^=", b"=",
+    ];
+    let Some(rest) = bytes.get(i..) else {
+        return false;
+    };
+    for op in OPS {
+        if rest.starts_with(op) {
+            if op == b"=" {
+                let next = rest.get(1).copied().unwrap_or(b' ');
+                return next != b'=' && next != b'>';
+            }
+            return true;
+        }
+    }
     false
 }
 
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/tests.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/tests.rs
@@ -1935,3 +1935,68 @@ fn a_store_call_is_not_inserted_into_a_longer_non_ascii_identifier() {
         "$c()(1); x\u{E0}$c(2)"
     );
 }
+
+/// `$: {a=b}` and `$: { a = b }` are the same program, so they must order the
+/// same. The assignment scan feeding the topological sort matched the literal
+/// `" = "`, so the unspaced form was credited with assigning nothing, lost its
+/// ordering edge, and ran before the statement it depends on.
+#[test]
+fn reactive_statement_order_ignores_whitespace_around_the_assignment() {
+    fn effect_order(source: &str) -> Vec<String> {
+        let js = crate::compiler::compile(
+            source,
+            crate::compiler::CompileOptions {
+                generate: crate::compiler::GenerateMode::Client,
+                filename: Some("order/index.svelte".to_string()),
+                ..Default::default()
+            },
+        )
+        .unwrap()
+        .js
+        .code;
+        js.lines()
+            .filter(|l| l.contains("$.set(mid,") || l.contains("$.set(out,"))
+            .map(|l| l.trim().to_string())
+            .collect()
+    }
+
+    let unspaced = effect_order(
+        "<script>\nexport let seed = 1;\nlet mid = 0;\nlet out = 0;\n$: out = mid + 1;\n$: {mid=seed*2}\n</script>\n\n<p>{out}</p>\n",
+    );
+    let spaced = effect_order(
+        "<script>\nexport let seed = 1;\nlet mid = 0;\nlet out = 0;\n$: out = mid + 1;\n$: { mid = seed * 2 }\n</script>\n\n<p>{out}</p>\n",
+    );
+
+    // `out` reads `mid`, so the statement assigning `mid` has to come first.
+    assert!(
+        unspaced[0].contains("$.set(mid,"),
+        "unspaced ran out of order: {unspaced:?}"
+    );
+    // The spaced form is the control: it was already correct and must stay so.
+    assert!(
+        spaced[0].contains("$.set(mid,"),
+        "spaced ran out of order: {spaced:?}"
+    );
+    assert_eq!(unspaced, spaced);
+}
+
+/// A comparison is not an assignment, and the longest operator has to win:
+/// `<=` is a comparison while `<<=` assigns.
+#[test]
+fn the_assignment_scan_separates_operators_from_comparisons() {
+    use super::reactive_transforms::is_assigned_anywhere_in_body;
+
+    for body in [
+        "a=1", "a = 1", "a  =  1", "a+=1", "a **= 2", "a<<=1", "a >>>= 1", "a??=1", "a++", "--a",
+    ] {
+        assert!(is_assigned_anywhere_in_body(body, "a"), "missed: {body:?}");
+    }
+    for body in [
+        "a==1", "a === 1", "a=>1", "a<=1", "a >= 1", "a!=1", "a !== 1", "ab=1", "b.a=1", "abc",
+    ] {
+        assert!(
+            !is_assigned_anywhere_in_body(body, "a"),
+            "false positive: {body:?}"
+        );
+    }
+}


baseballyama:
`Compiler parity` is red on this branch with **22 `js-mismatch`** entries against a ratchet that is at **0 on all three targets**, so every one is a regression this PR introduces. Pulling the report artifact from [that run](https://github.com/baseballyama/rsvelte/actions/runs/31270696397) so you don't have to re-derive it:

```
counts  match 13168  error-parity 948  js-mismatch 22  js-unparseable 0  css-mismatch 0  error-mismatch 0
parseCounts  match 14138  output-unparseable 0
```

The output still **parses** everywhere — this is a semantic change, not a splice accident, which is why the parse gate is clean and only the byte gate moved.

17 of the 22 are `svelthree`, the rest are `svelte-fa`, `svelte-table`, `svelte-toast`, `svelte-ux` and `sveltestrap`. Three shapes, by cluster count:

**1. A dependency is dropped from a `$:` dependency list (dominant).**

```
svelthree/src/lib/components/AmbientLight.svelte  client / client-dev
  expected: () => ($.deep_read_state(light()), $.get(light_uuid)),
  actual  : () => $.deep_read_state(light()),
```

Same shape with `camera_uuid` (OrthographicCamera), `$progress()` (ToastItem), `setCustomFontSize` (fa-layers-text). In each case the *last* element of the sequence is gone and the enclosing parens with it. `svelte-ux/Icon` is the variant where the dependency is not dropped but **substituted** — `$.deep_read_state(svgUrl())` became `$.deep_read_state(svg())`.

**2. Statements are reordered across a comment (server).**

```
svelthree/…/AmbientLight.svelte  server
  expected: // ---  AFTER  INITIALIZATION  ---   … then   function set_uuid() {
  actual  : function set_uuid() {
```

10 of the clusters are this one. Reading it with shape 1: reordering `$:` statements from the AST also reorders what the text pipeline had been attaching comments to, and a comment that no longer has its statement in front of it moves.

**3. A `let` loses its declarator to an interposed comment (server).**

```
svelte-table/src/SvelteTable.svelte   expected: let /** @type {TableColumns<any>} */   actual: let colspan;
svelte-ux/…/Icon.svelte               expected: let // Accessibility                   actual: let isLabelled;
```

Worth flagging explicitly because this is the **#2601 family in reverse**. #2642 just landed a fix whose whole point is that `let // c` / newline / `name` hides the declaration from every later pass, so it hoists the comment above the keyword — and official's expected output here is exactly the `let // Accessibility` shape. If this branch is re-ordering declarations before that hoist runs, the two are interacting; if it is doing its own comment attachment, it is re-deriving the same decision in a second place. Either way #2642 is worth reading before fixing this one.

I have not looked at the diff, so this is what the gate observed, not a diagnosis of the cause.

baseballyama:
One correction to narrow shape 3, since I left it ambiguous: **main is green on `svelte-ux/…/Icon.svelte`** — the corpus ratchet is at 0 on all three targets at `1953af66`, which is after #2642 landed. So #2642 alone does not produce `let isLabelled;` where official emits `let // Accessibility`; its hoist only fires for declarators produced by `split_comma_separated_declarations`, and this file does not reach it on main.

That makes the interaction I raised a *possibility to rule out*, not a suspect. Reading #2642 is still worth it — it is the same decision (where a comment sits relative to a declaration keyword) being made in a second place — but the regression is this branch's, not a latent conflict with a merged fix.

baseballyama:
Independent duplicate of this landed in my worktree before I saw the PR — same design
(`analysis.reactive_statement_assignments` alongside `reactive_statement_dependencies`, ordinal
aligned, walking function bodies), same dead-code cascade. Discarding mine; recording the two
things it produced that are not here, so they are not lost:

**1. Discriminating corpus files.** Written and verified green on all three targets against
official, and each checked for whether it actually moves — three of my first four constructions
turned out to be non-discriminating and were rebuilt:

| file | why it discriminates |
|---|---|
| `issues/2646-reactive-order-unspaced-assignment.svelte` | the reported `$: {mid=seed*2}`. Must stay **unformatted** — the formatter spaces the operator, and the spaced form is precisely the one that already worked |
| `matrix/reactive-order/member-target-assignment.svelte` | `obj.p = seed` — `extract_identifiers` drops a bare member target, so upstream does **not** credit `obj`; the scan's member-base extraction did, and the spurious edge reordered the pair |
| `matrix/reactive-order/string-borne-assignment.svelte` | `(mid = 9)` in a template literal, placed so the phantom edge pulls a **later** statement ahead — a phantom that merely restates source order proves nothing, which is what my first version did |
| `matrix/reactive-order/function-body-assignment.svelte` | the negative control: scan and AST agree, order must not move |

I can send these as a follow-up PR once this merges, or you can take them — say which.

**2. The comment-borne twin does not exist for a reason worth writing down.** A phantom only
changes the order when it pulls a later statement ahead, which necessarily *moves* the statement
carrying it — and official's server printer re-homes a moved `$:` statement's interior comment in a
way rsvelte does not match yet (`$: /* mid = 9 */\nmid = seed * 2;` vs `$: mid = seed * 2;`,
neighbour of #2621). So that file cannot land green for a reason unrelated to ordering. Worth a
README line rather than a silently absent case.

Also: `find_colon_at_depth0` dies in this cascade, which closes one of the sites #2378 enumerates.

baseballyama:
Measurement on the 38 new failures, from an independent implementation of the same design
(mine, discarded — see previous comment). Two facts that narrow it:

**1. The regression is comment attachment, not ordering.** On
`svelthree/src/lib/components/AmbientLight.svelte` the diff against official is **89 removed
lines, every one of them a comment**. Filtering the unified diff to lines that are not
comment-only leaves nothing but the two `---`/`+++` headers:

```
$ node scripts/compat-corpus/one.mjs submodules/svelthree/src/lib/components/AmbientLight.svelte --target client \
    | grep -E "^[+-]" | grep -vE "^[+-]\s*(/\*|\*|//|\*/)"
--- …-exp.js
+++ …-act.js
```

A reordering shows up as paired `-`/`+` **code** lines. There are none, so the emitted statements
are byte-identical and in the same order — official parks a block of orphan JSDoc at the end of
the component function and rsvelte drops it. Worth checking `rehome_reactive_statement_comments`
rather than the sort.

**2. Not all 38 share a cause.** With my build, `svelte-fa/src/lib/fa-layers-text.svelte` and
`svelte-toast/src/lib/ToastItem.svelte` — both listed as new failures on this branch — come out
`MATCH`, while `AmbientLight.svelte` reproduces. My tree differs from this one in two ways I can
name: `dep_vars` is filtered by `!assigned_vars.contains(name)` before the sort, and the tree
predates #2655. So the cluster is at least two mechanisms, and treating "38 svelthree-ish files"
as one bug will mislead.

Neither entry is in `known-failures.client.json`, so both do pass on `main` — these are real
regressions, not stale baseline rows.

baseballyama:
`10b5f521` ("sort `$:` statements on upstream's dependency set") moved it, but not to zero. Diffing the two report artifacts so the delta is attributable:

| | before (`31270696397`) | after (`31272790108`) |
|---|---|---|
| `match` | 13168 | 13179 |
| `js-mismatch` | **22** | **19** |
| `js-unparseable` | 0 | 0 |

**Fixed: 3.** `svelte-table/SvelteTable`, `svelte-ux/Icon`, `sveltestrap/Dropdown` — which is exactly **shape 3** from my earlier comment, the `let` that lost its declarator to an interposed comment. That shape is gone.

**New: 0.** So the commit is a strict improvement, not a trade.

**Still failing: 19**, all `svelthree`, and they are shapes 1 and 2 unchanged:

```
6 + 6  client / client-dev   () => ($.deep_read_state(light()), $.get(light_uuid))   →  () => $.deep_read_state(light())
2 + 2  client / client-dev   the same with camera_uuid
  10   server                // ---  AFTER  INITIALIZATION  ---   →  function set_uuid() {
3 + 3  client / client-dev   let on_wheel = $.prop(  →  let on_wheel = $.prop($$props, "…", N, undefined);
```

The dependency-drop (shape 1) is the one to look at first: `$.get(light_uuid)` disappearing from a `$:` dependency list is lost reactivity, and it survived a change that was specifically about which dependency set the sort reads. The `on_wheel` cluster is new to the top of the list only because the other three files left — it was present in the first run too.

The 10 server comment-displacement entries and the 12 dependency-drop entries share `svelthree` and probably a cause; if the sort is what relocates the comment, then shape 1 and shape 2 are one bug seen from two targets rather than two.

Method, so it can be repeated: `gh api repos/baseballyama/rsvelte/actions/runs/<id>/artifacts`, pull `corpus-report`, and compare `report.json` `counts` plus the `failures[].id` sets. `cluster.txt` in the same artifact gives the per-shape counts.

PR #2657, opened 2026-08-08
fix(compiler): stop the client script scan from rewriting comment bodies

baseballyama:
`strip_unnecessary_arrow_body_parens` scanned the client instance script for
`=> (` and dropped the parens. It skipped string and template literals but **not
comments**, so a comment whose text happened to contain an arrow function was
rewritten too:

```js
// values.forEach((v) => (valueFilter[v] = true));   official
// values.forEach((v) => valueFilter[v] = true);     rsvelte
```

That is the same family as #2590 / #2592 / #2596 / #2598 / #2599 / #2603 — a
scanner assuming input it did not get — and it belongs in the table in
`AGENTS.md`.

## The pass is not worth fixing, because it never earns its keep

Measured by building the binding twice and diffing the **raw** corpus trees
(snapshotted after `compile.mjs`, before `verify.mjs` — `verify` rewrites both
trees in place via `oxfmtTree`, so a post-verify tree is not raw output):

| | |
|---|---|
| entries whose output the pass changes | **4 of 14,138** |
| of those, entries where it diverges from official | **4 of 4** |
| entries that become byte-identical to official once removed | **3** |
| regressions | **0** |

The four are `layercake/src/_components/{Sankey,SharedTooltip.html,SharedTooltip.percent-range.html}.svelte`
and `svelte-pivottable/src/lib/UI/FilterBox.svelte` — client and client-dev
only. No `.css`, no `server.js`, no `.svelte.ts`, matching the pass's single
call site exactly.

Everywhere else esrap already prints the parens the way official does. The pass
only ever mattered on inputs where **its own text rewrite** forced the fallback
path, so removing it is a strict improvement rather than a trade. It is deleted
rather than repaired, along with the two helpers it orphaned
(`has_top_level_comma_in_expr`, `find_matching_paren_bytes`) — 265 lines.

## Why every gate was green

The corpus gate falls back to an AST comparison for byte-different outputs, and
`ast_equiv_batch` applies `CommentPolicy::Ignore` unless `--comments` is passed
— which `verify.mjs` does not pass. This is documented at `verify.mjs:470-474`.
A divergence living entirely inside a comment therefore scores `match`.

So the blind spot was already known and written down; what was new is that it
was concealing a live defect. These entries are in **no** ratchet either —
`known-failures.{client,server,client-dev}.json` are all empty, so this was not
a suppressed known failure. It was invisible.

Note also that the *direct* effect of this pass is unobservable to the gate for
a second, independent reason: `oxfmt` collapses `=> (a + b)` to `=> a + b`
during normalization. A green corpus run can never have been evidence that this
pass was safe.

## Test

`crates/rsvelte_core/tests/arrow_parens_in_comment.rs` — tree-stable, so it does
not expire on a submodule bump. Verified discriminating rather than assumed:
with the pass restored, 2 of 3 fail **with the predicted message** ("the comment
body was rewritten"), while the server control stays green — the server never
ran this scan, so a change reaching beyond the client would move it.

## Instrumentation

#2641 added `PN_INV_ARROW` / `PN_CHG_ARROW` around this pass. They are removed
along with it — the counter array, the struct fields and the `compile_profile`
row. Leaving them would have left two instruments that structurally cannot move
off zero, which reads as "this pass changes nothing" rather than "this pass no
longer exists".


baseballyama:
## Side effect: this closes most of the span-validity work the epic was sizing

Measured with `compile_profile --features rsvelte_core/measure-pa-split`,
**both arms built** — the control is this branch with the pass restored, not an
inherited figure:

| corpus | control | after | text still rewritten |
|---|---|---|---|
| bits-ui | 72.40% | **100.00%** | 0 |
| melt-ui | 75.90% | **100.00%** | 0 |
| shadcn-svelte | 99.47% | **100.00%** | 0 |
| flowbite-svelte | 97.36% | **99.76%** | 2 / 832 |
| `submodules/svelte` | 95.80% | 96.05% | 1010 / 25557 |

Three of the four real-world libraries reach 100%, and the fourth drops to 2
files. The premise this work started from — that raising library-side span
validity needed the arrow-paren pass *ported* to the AST — was wrong in the
useful direction: the increment was a deletion.

`submodules/svelte` barely moves, and its residual attributes cleanly
(`IDENTITY … HOLDS`, so the three sets are disjoint):

```
class    587 invoked / 533 changed (90.8%)
split    508 invoked / 396 changed (78.0%)
comments 1703 invoked /  81 changed (4.8%)
```

`class` + `split` account for 929 of the 1010 — but both read `0/0` on all four
shipped libraries, so that residual lives in the fixture population rather than
in shipped code. `comments` misses 95.2% of the time, which is wasted work
rather than lost spans: a pass that does not change the text does not
invalidate a span.

None of this is a reason to merge — the correctness argument above stands on its
own — but it changes what the next increment should be.

baseballyama:
Correcting the scope of the table above: the four libraries I measured are a
subset of the profiler's shipped set. On the full `--shipped` population —
5,879 files / 9.65 MB, adding `layerchart`, `skeleton` and `svelte-ux`, which
skew far more legacy than the libraries — both arms built:

| | control | after |
|---|---|---|
| files whose text a prenormalize pass rewrites | 231 (5.33%) | **22 (0.51%)** |
| spans valid | 94.67% | **99.49%** |

A **90.5% reduction** in rewritten files across shipped code. The per-pass split
shows why:

```
control   comments 89/20   class 11/0   split 13/2   arrow_parens 458/211
after     comments 89/20   class 11/0   split 13/2
```

`arrow_parens` was invoked 458 times and changed 211 files — it was doing
essentially all of the span invalidation on shipped code. What remains is 22
files, 20 of them from `comments`.

Worth noting for whoever picks up the next increment: `class` and `split` read
`11/0` and `13/2` here but account for 929 of the 1010 remaining rewrites in
`submodules/svelte`. That residual is a property of the fixture population, not
of shipped code, so burning it down would not move this table.

baseballyama:
## Correction: what the span-validity numbers above do *not* license

I framed the 94.67% → 99.49% movement as evidence that this "closes most of the
span-validity work the epic was sizing." That over-reads the instrument. Measured
rather than argued, with both arms built:

| | control | after |
|---|---|---|
| `loop_lines` | 86,048 | **86,048** |
| `statements` | 15,454 | **15,454** |
| `processed` | 12,529 | **12,529** |
| spans valid | 94.67% | **99.49%** |

**The text pipeline does byte-for-byte the same work.** Not one line, statement
or processed statement moved. Zero files migrated from the text path to an AST
path.

The code agrees:

- `transform_instance_script_for_visitors` — the text pipeline, `line_loop`
  included — runs **unconditionally** for every file with an instance script
  (`client/mod.rs:599`). Span validity does not gate it.
- What unchanged text actually gates is much narrower: `client/mod.rs:6679`,
  `if original_script.trim() != result.trim() { return None; }`, decides whether
  one later step — the state-variable transform — may **reuse the retained oxc
  program** instead of **re-parsing**. Both branches are AST-based
  (`transform_state_vars_ast_projected_from_program` vs
  `transform_state_vars_ast`). It is reuse-vs-reparse, not AST-vs-text.
- `PN_TEXT_CHANGED` is recorded at `client/mod.rs:4711`, right after
  prenormalize, while the real gate compares after the *whole* text transform.
  So the counter is a **necessary condition measured upstream of the gate**, and
  can hold while the gate still fails.

What the number does license: `arrow_parens` was breaking the retained-reuse
precondition for 211 of 4,332 shipped files and no longer does. What it does not
license: any claim about the text→AST migration, whose surface this change does
not touch.

Still unmeasured, and stated rather than assumed: whether those 211 files now
actually *reach* retained reuse. Only the precondition was measured; the gate at
`:6679` is downstream of it and has no counter.

The correctness argument for this PR is unaffected — it never depended on these
figures.

PR #2658, opened 2026-08-08
fix(analyze): reject an invalid bind: target on components too (#2583)

### LINKED ISSUE #2583: validate: bind_invalid_expression is not raised for a component binding, and the element message text diverges
Script for all three: `import Comp from './Comp.svelte'; let o = $state({}); let obj = {};`

|---|---|---|
| `<Comp bind:value={o.x = obj} />` | error `bind_invalid_expression` at 4:6 — "Can only bind to an Identifier or MemberExpression or a `{get, set}` pair" | **compiles**; emits `return o.x = obj;` as the getter and `o.x = obj = $$value;` as the setter |
| `<Comp bind:this={o.x = obj} />` | error `bind_invalid_expression` at 4:6, same message | **compiles**; emits `$.bind_this(Comp($$anchor, {}), ($$value) => o.x = obj = $$value, () => o.x = obj)` |
| `<input bind:value={o.x = obj} />` | error `bind_invalid_expression` at 4:7, same message | error `bind_invalid_expression`, but the message reads **"Invalid binding expression"** |

Two separate defects, and the split is the useful part:

1. **Component-side, the validation is missing entirely.** rsvelte compiles an assignment expression as a binding target and generates a getter/setter around it. The element form is checked, the component form is not, so the check exists and does not cover both call sites.
2. **Element-side, the check fires but the message is rsvelte's own.** So even where rsvelte rejects, a user comparing output against the official compiler sees different text. This is the population the `validator-message-known-failures.json` ratchet (#2418) covers for warnings and #2558 is extending to errors — worth confirming whether this pair is already listed there or is invisible to it because the *fixture* corpus has no such input.

## The class this belongs to

Fifth instance found today of rsvelte accepting a program upstream rejects:

| | official raises |
|---|---|
| #2547 | `js_parse_error` — `await` in a default parameter value |
| #2561 | `expected_whitespace` |
| #2568 | `derived_invalid_export` |
| #2581 | 15 `expected_whitespace` sites, `require_whitespace` having **0 callers** |
| this | `bind_invalid_expression`, component form |

Each was found while investigating something else, so the count is unknown rather than small. #2568 proposes the way to size it in one pass: diff upstream's `errors.js` code set against the codes rsvelte can raise at all. That is a different question from #2446 / #2558, which gate the errors both compilers already raise — a code rsvelte never raises produces no error to compare, so no parity gate can see it.

Found while reconciling #2569 and #2577; the measurement is in #2577's thread.

baseballyama:
Closes #2583.

## The defect

`<Comp bind:value={o.x = obj} />` compiled. Official rejects it with
`bind_invalid_expression`; rsvelte lowered it into a getter/setter pair wrapped
around the assignment expression.

Upstream runs `object(node.expression)` **once**, before it branches on whether
the binding is on an element or a component, and raises `bind_invalid_expression`
when the expression is not an identifier or a member chain rooted in one. rsvelte
had that check on the element path only: `validate_bind_value_for_component`
early-returned `Ok(())` for every non-identifier expression, so anything that was
not an identifier skipped validation entirely on its way to the getter/setter
lowering.

Two smaller things fell out of the same place:

- `bind:this` on a component took the same unvalidated path.
- The element path raised rsvelte's own wording ("Invalid binding expression")
  rather than upstream's `Can only bind to an Identifier or MemberExpression or
  a \`{get, set}\` pair`.

## The fix

`bind_target_name()` — the port of upstream's `object()` + its diagnostic — is
now a shared helper that **both** paths call. A check that lives at one of two
call sites is what let this drift; there is now one call site's worth of logic
and two callers. The diagnostic moved into `2_analyze/errors.rs` with upstream's
exact message and URL.

## Why this could be reported at all, and what stops the next one

Every gate in this repo compares programs both compilers accept. That is also
all the collected corpus can hold — published code compiles — so
"rsvelte accepts what official rejects" was gated only by the 145
`compiler-errors` fixtures, at **one input per code**. `bind_invalid_expression`
has a passing fixture. It is on an element. The component slot was never asked.

So this PR adds the population that asks it, plus the comparison that can see
the answer:

**1. New matrix family `invalid-bind`** (`matrix/axes.mjs`) — 20 invalid target
expressions × 8 `bind:` slots (element value/group/this/clientWidth, component
value/this/named, `<svelte:window>`) = 160 generated cases × 3 targets. These are
programs official rejects by construction; a collected corpus cannot contain
them.

**2. `error-code-mismatch` in `matrix/run.mjs`** — the family alone would have
measured nothing. `run.mjs` scored any both-reject case as `error-parity` and
`continue`d without looking at the errors, so rsvelte rejecting for an unrelated
reason (or with a wrong code) read as agreement. It now compares the two error
codes. This applies to every family, not just the new one.

The comparison and the population had to land together — either alone is inert.

**Positive control.** The same 480 comparisons, run against a binary built with
the two Rust files reverted to `origin/main` and nothing else changed:

| | `error-parity` | `error-mismatch` |
|---|---|---|
| `origin/main` binary | 300 | **180** |
| this branch | 480 | 0 |

The 180 that move are exactly 20 invalid targets × **3 component slots** × 3
targets. Every element slot and `<svelte:window>` was already rejected before
this PR — which is the shape of the defect, stated as a measurement rather than
as a reading of the code.

## Also updated

- `compatibility/gate-coverage.md` — blind spot 5d rewritten (codes are now
  compared; **message and position still are not**, and that is stated as the
  remaining gap), new blind spot 5e naming what `invalid-bind` does *not* reach:
  every error code other than this one. The same per-call-site drift is possible
  for `{@render}`, `use:`, `{#each … as}` patterns and `<svelte:element>`, and no
  gate generates invalid inputs for them.
- `MIN_MATRIX_CASES` raised 400 → 700 (1017 cases are generated now).
- `errorCode()` extracted to `scripts/compat-corpus/error-code.mjs` so the
  collected-corpus gate and the matrix gate agree on what "the same error" is.

### DIFF
--- a/.changeset/bind-invalid-expression-on-components.md
+++ b/.changeset/bind-invalid-expression-on-components.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Reject `bind:` to an expression that names no binding on a **component**, not only on an element. `<Comp bind:value={o.x = obj} />` compiled and was lowered into a getter/setter around the assignment where the official compiler raises `bind_invalid_expression`. The element path's message is now upstream's text too, so a user comparing diagnostics sees the same string.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -198,9 +198,20 @@ corpus could never have found them — the same lesson as the warning gate, one 
 
 A **generated**, not collected, differential corpus (`pnpm run corpus:matrix`, #2281 Gate 2),
 ratcheted through `compatibility/matrix-known-failures.json` with per-cluster justification in
-the paired `.md`. Two declarative axis families in `matrix/axes.mjs` — binding kind × syntactic
-position, and comment kind × insertion slot — expanded into ~2,000 comparisons that run in
-**~5 s** and need only `submodules/svelte` plus the NAPI binding, so it gates every PR.
+the paired `.md`. Three declarative axis families in `matrix/axes.mjs` — binding kind × syntactic
+position, comment kind × insertion slot, and invalid `bind:` target × directive slot — expanded
+into ~3,000 comparisons that run in **~7 s** and need only `submodules/svelte` plus the NAPI
+binding, so it gates every PR.
+
+The third family is the odd one out and the reason is worth stating: its inputs are programs the
+official compiler **rejects**, which is a population no collected corpus can hold, because
+published code compiles. "rsvelte accepts what official rejects" was otherwise gated only by the
+145 `compiler-errors` fixtures at **one input per code** — and a code with a passing fixture
+reads as covered. #2583 is what that misses: `bind_invalid_expression` had a passing fixture on
+an element while `<Comp bind:value={o.x = obj} />` compiled into a getter/setter around an
+assignment. Adding the family alone would still have measured nothing, because `run.mjs` scored
+any both-reject case as `error-parity` without looking at the codes; **the comparison and the
+population had to land together**.
 
 It exists because the collected corpus samples the **marginal** distribution of published code
 while every bug in the #2253/#2254/#2255/#2256 batch was an **interaction**: #2254's shape occurs
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -213,6 +213,12 @@ assignment. Adding the family alone would still have measured nothing, because `
 any both-reject case as `error-parity` without looking at the codes; **the comparison and the
 population had to land together**.
 
+The family carries **valid** targets against the same slots too, and that half is not
+decoration: the first version had only the invalid rows, and CI then caught an over-rejection
+(a TypeScript assertion, `bind:group={c as T}`) from a corpus file instead of from the gate.
+An over- and an under-rejection are opposite directions of one check, and a population of
+only-invalid inputs is blind to one of them.
+
 It exists because the collected corpus samples the **marginal** distribution of published code
 while every bug in the #2253/#2254/#2255/#2256 batch was an **interaction**: #2254's shape occurs
 **0 times in 14,026 real files**, #2253's likewise, and `client`/`server` were at 0 known failures
--- a/compatibility/gate-coverage.md
+++ b/compatibility/gate-coverage.md
@@ -218,10 +218,10 @@ differ: the prose and span of two unrelated errors say nothing. Those pairs are
 
 ## 5. Generated shape matrix — `scripts/compat-corpus/matrix/`
 
-**Unit.** 1017 generated cases × 3 targets = 3051 comparisons. For the 857 cases both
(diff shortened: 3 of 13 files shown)


baseballyama:
`MSRV, package, and docs` is red on the two pinned-diagnostics tests, and the failure is the guard doing its job — this PR adds a `diagnostics!` constructor, so both pins need updating in `crates/rsvelte_core/src/compiler/phases/2_analyze/diagnostics_test.rs`:

```
line 23:  const ERROR_MACRO_COUNT: usize = 126;                   ->  127
line 37:  const DIAGNOSTICS_DIGEST: u64 = 0x16dd_ae65_8049_6f45;  ->  0x0856_867a_8ed7_4cbb
```

The runner printed the new digest as `0x856867a8ed74cbb`; the constant is written in `_`-grouped form and needs the leading zero to stay 16 digits.

The count assertion is mechanical. **The digest one is not** — its own message says to update it "after confirming the new wording matches official Svelte", and that is the call for whoever wrote the new diagnostic, not something to read off a CI log. I have not checked the new message against `submodules/svelte`, so I am not pushing this.

Diagnosis only, no push: we are both working this queue and a second push to your branch would race yours.

baseballyama:
CI caught a real over-rejection in the first version of this PR, and where it caught it is the interesting part.

`Lint parity` went red on two `svelte/prefer-const` diagnostics vanishing from one flowbite-svelte file. The cause: that file writes `<Radio bind:group={btnColor as ButtonProps["color"]} />`. Upstream analyses the AST with the TypeScript nodes already removed, so `object()` sees the bare identifier; rsvelte's `get_object_node` walks only member chains, so an assertion-wrapped target named no binding and the new check rejected the file outright — which is why its lint diagnostics disappeared rather than changed.

The gap is older than this PR: the element path had the same `get_object_node`, so `<input bind:value={x as T} />` was already over-rejected. Nothing had noticed because no corpus file writes it on an element.

Two things follow, and only one of them is the fix:

1. `get_object_node` and its JSON fallback now unwrap `TSAsExpression` / `TSSatisfiesExpression` / `TSNonNullExpression` / `TSTypeAssertion` / `TSInstantiationExpression`. Confirmed against official: all seven assertion shapes in the new `typescript_assertion_bind_targets_still_compile` test compile there.

2. **The `invalid-bind` family structurally could not have found this.** Every one of its inputs is a program official rejects, so it reports "rsvelte accepts what official rejects" and nothing else. An over-rejection is the opposite direction. The family now also carries 11 **valid** targets against the same 8 slots (TypeScript assertions among them), so both directions are generated: 248 cases, 744 comparisons.

Recorded in `gate-coverage.md` 5e and in AGENTS.md, because the general form is worth keeping: a population of only-invalid inputs measures one direction of a two-directional check, and reads as covered.

PR #2660, opened 2026-08-08
fix(codegen): keep a string literal's source spelling in template expressions (#2653)

### LINKED ISSUE #2653: codegen: a single-quoted string literal loses its `raw`, so `{@const t = 'a\tb'}` emits a real tab
## Summary

`convert_literal` keeps a string literal's `raw` **only when it starts with `"`**
(`3_transform/client/visitors/expression_converter.rs`, both the `JsNode::Literal`
arm and `convert_literal`). A single-quoted literal is converted to its cooked
value and re-serialized by the printer, whose escape set is not esrap's — so an
escape esrap prints verbatim comes back as the raw character.

## Repro

```svelte
{#if true}
  {@const t = 'a\tb'}
  <p>{t}</p>
{/if}
```

```diff
-      const t = $.derived_safe_equal(() => "a\tb");
+      const t = $.derived_safe_equal(() => "a	b");
```

Also reproduces in an event handler, i.e. wherever a template expression is
converted rather than spliced:

```svelte
<button onclick={() => console.log('a\tb')}>go</button>
```

```diff
-	$.delegated('click', button, () => console.log('a\tb'));
+	$.delegated('click', button, () => console.log('a	b'));
```

## What it is and is not

The emitted string's **value is correct** and the output parses — `'a\tb'` and
`'a<TAB>b'` denote the same string. This is a source-text divergence, not a
wrong-output one, which is why it has sat under the parse gate. It is **not**
erased by the corpus normalizer: oxfmt keeps the literal tab, so `verify.mjs`
reports it (checked on the repro above, `--target client`, normalized).

Affected escapes are the ones the printer does not re-emit: `\t`, `\v`, `\b`,
`\f`, and the codepoint escapes (`\x41` / `A` come back as `A`). `\n` is
fine because the printer escapes newline.

## Scope

Found while fixing #2607 and deliberately left out of it: the fold path and the
conversion path are different mechanisms, and #2607's evidence says nothing
about this one. The `\\` case matches today only by coincidence — the printer's
escape for a backslash happens to equal the source's.

Note the asymmetry the current condition encodes: double-quoted literals are
routed through `RawString` **to preserve the quote style** through OXC
normalization. Preserving `raw` unconditionally is what esrap does, so the
`starts_with('"')` test looks like the narrow fix for the quote-style symptom
rather than the general rule.

## Suggested resolution

Preserve `raw` for every source string literal, and land the shape in
`compatibility/pattern-corpus/matrix/string-escape-fold/` (the axis added in
#2607) — its `codepoint-escapes.svelte` / `control-escapes.svelte` files cover
the *fold* site of exactly these escapes, so the conversion site belongs beside
them.

baseballyama:
Closes #2653.

## The defect

esrap writes a literal's `raw` whenever it is set:

```js
const value = node.raw || (typeof node.value === 'string' ? quote(node.value, quote_char) : String(node.value));
```

so official's output carries the source's quote style **and** its escape spelling. rsvelte kept `raw` only when it started with `"` and re-printed everything else from the cooked value, whose escape set is not esrap's.

Measured against official, before this PR:

|---|---|---|
| `{@const t = 'a\tb'}` | `'a\tb'` | `'a<TAB>b'` |
| `console.log('a\tb')` | `'a\tb'` | `'a<TAB>b'` |
| `console.log('\x41')` | `'\x41'` | `'A'` |
| `console.log("a\tb")` | `"a\tb"` | `"a\tb"` |
| `console.log('plain')` | `'plain'` | `'plain'` |

The last two rows are why this survived: double-quoted literals took the `raw` path already, and a single-quoted literal with no escapes round-trips through the printer unchanged.

The emitted string's **value is correct** and the output parses. It is a source-text divergence, so neither the parse gate (#2591) nor any runtime test can observe it — only a byte comparison can.

## The fix

`raw` is preserved whenever it is non-empty, at both conversion sites (the typed `JsNode::Literal` arm and the JSON `convert_literal`). The empty-`raw` guard is what keeps synthesized literals, which carry no source text, on the printer path.

## The gate

**New matrix family `literal-escape`** — 16 escape spellings × 14 template expression slots = 224 cases, 672 comparisons.

It is the first axis in `matrix/axes.mjs` that injects into **markup** rather than into a JS statement inside `<script>`. `gate-coverage.md` names that as this gate's largest blind spot (5c) and as "the highest value-per-cost item in this document"; the slots are `{expr}`, an attribute value, `{@const}`, a handler body, the `{#if}` / `{#each}` / `{#await}` / `{#key}` heads, `{@html}`, `{@render}`, `class:` and `style:` directives, a spread attribute, and an instance declaration.

**Positive control**, same 672 comparisons against a binary built from `origin/main` with nothing else changed:

| | `match` | `js-mismatch` |
|---|---|---|
| `origin/main` binary | 456 | **216** |
| this branch | 672 | 0 |

`newline-escape` and `backslash-escape` are the negative controls — `\n` and `\\` are in the printer's own escape set, so they matched before and must keep matching. They are among the 456.

## Why the repro files are unformatted

`compatibility/pattern-corpus/` normally commits formatter fixed points. These files cannot be: the fmt oracle rewrites every literal to double quotes, and a double-quoted literal is exactly the shape that already worked. The formatted form of every file here reproduces nothing — the formatted shape is not a stricter version of the input, it is a different input.

Checked: `rsvelte-fmt` produces byte-identical output to the oxfmt oracle on all 11 new files, so the hard fmt-parity gate stays green with them committed unformatted. The same precedent is already in the README for `2592-destructure-assignment-line-break.svelte`.

## Verified

- 11 new pattern files + the issue repro: 33 of 33 comparisons byte-identical to official across client / server / client-dev, **without** normalization.
- Full matrix: 3243 comparisons, 206 known divergences, 0 new.
- `compiler_fixtures`, `print`, `compiler_errors` green.


baseballyama:
Handing this branch off — I am **not** pushing to `fix/2653-literal-raw`. (A peer session asked me directly; I could not reach it to reply, so the answer is here where it is discoverable.)

Both conflicting sides are mine (#2658 and this PR both extended the shape matrix), so here is what "merge both" means concretely. Each item is a place where taking either side alone leaves a wrong number in the tree:

- **`matrix/axes.mjs`** — keep BOTH axis blocks. #2658 added `INVALID_BIND_TARGETS` / `VALID_BIND_TARGETS` / `BIND_SLOTS` / `BIND_PREAMBLE` / `BIND_PREAMBLE_TS`; this PR added `LITERAL_ESCAPES` / `EXPRESSION_SLOTS`. Independent exports, no shared identifiers.
- **`matrix/generate.mjs`** — keep both `invalidBindCases` and `literalEscapeCases`, and both entries in `FAMILIES`. Merge the import list rather than picking a side.
- **`matrix/run.mjs`** — the one that needs care. #2658 also changed *behaviour* here: `error-code-mismatch` (comparing the two error codes when both compilers reject, instead of scoring every both-reject case as `error-parity`) and `expectedError` / `actualError` becoming `{message, code}`. **Keep that.** This PR touched only `MIN_MATRIX_CASES` and the case-count comments. Recompute the count after merging rather than copying either side:

  ```
  node -e "import('./scripts/compat-corpus/matrix/generate.mjs').then(m=>console.log(m.generate().length))"
  ```

  Expected: 1105 + 224 = **1329**. Set `MIN_MATRIX_CASES` to 900 and update both prose mentions of the count (the comment above the constant, and the `--update-baseline` refusal message). If the printed number is not 1329, stop — a family was dropped in the merge.
- **`compatibility/gate-coverage.md`** — the row-5 table row, the **Unit** line, blind spots 5c / 5d / 5e and the "Closing 5b/5c" paragraph all carry edits from both sides. Unit and the closing paragraph state totals (1329 × 3 = 3987 comparisons), so neither version is correct verbatim.
- **`AGENTS.md`** and **`pattern-corpus/README.md`** — additive on both sides, keep both paragraphs. AGENTS.md then describes **four** axis families, not three; the sentence reading "Three declarative axis families" needs updating.

After pushing, with a staged NAPI binding: `node scripts/compat-corpus/matrix/run.mjs`. With both families present I measured 0 `error-mismatch`, 0 `error-code-mismatch`, 206 known `js-mismatch`. Anything else means the merge lost a fix rather than a doc line.

Issue #2662, opened 2026-08-08
compiler: a template literal with constant interpolations is not folded (all three targets)


baseballyama:
## Summary

A template literal whose interpolations are all constants is folded by the official compiler and not by rsvelte, on **all three targets**.

## Repro

```svelte
<script>
  const cont = `p${'ab'}q`;
</script>

<p>{cont}</p>
```

client / client-dev:

```diff
-	p.textContent = 'pabq';
+	p.textContent = cont;
```

server:

```diff
-	$$renderer.push(`<p>pabq</p>`);
+	$$renderer.push(`<p>${$.escape(cont)}</p>`);
```

## Cause

`try_insert_constant_value` (`3_transform/server/helpers.rs`) accepts a backtick literal **only** when it contains no `${`, and `try_evaluate_with_constants` has no template-literal branch at all. Upstream's `scope.evaluate` walks the `TemplateLiteral`'s quasis and expressions, so any interpolation whose value is known folds.

## Scope

Found while adding the line-continuation matrix for #2652 and deliberately left out of it — the shape reproduces with no line continuation, so it is a separate gap.

Not wrong output: the value is correct at runtime and only the emitted text differs. It is a corpus-gate divergence, and it is also the reason a constant `{@const}`-style template literal never reaches the static-`textContent` path.

PR #2663, opened 2026-08-08
fix(compiler): treat a string's line continuation as content, not code (#2652)

### LINKED ISSUE #2652: compiler: a string-literal line continuation is re-indented (client) and blocks constant folding (server)
## Summary

A JavaScript **line continuation** inside a string literal (a backslash immediately before the line break) diverges on all three targets, in two different mechanisms:

- **client / client-dev** — the emitted declaration gains a tab, changing the string's value.
- **server** — the constant is not folded at all, so the value is emitted through `$.escape(cont)` where official inlines it.

## Repro

```svelte
<script>
  const cont = 'a\
b';
</script>

<p>{cont}</p>
```

(The `\` is the last character of the line; the string's value is `ab`.)

### client

```diff
 export default function Cont($$anchor) {
 	const cont = 'a\
-b';
+	b';
```

The indenter treats the continued line as ordinary code and indents it. Inside a
string literal the tab is **content**, so the compiled component's `cont` is
`a\tb`, not `ab`. Output is valid JavaScript computing the wrong string — the
parse gate (#2591) cannot see it. Identical on `client-dev`.

### server

```diff
-	$$renderer.push(`<p>ab</p>`);
+	$$renderer.push(`<p>${$.escape(cont)}</p>`);
```

`try_insert_constant_value` (`3_transform/server/helpers.rs`) matches an
initializer against `'…'` on **one line**, so a continued literal never enters
the constants map and the read stays dynamic. Not wrong output, but a
divergence the corpus gate reports.

## Scope

Found while fixing #2607 — the fold now cooks a line continuation to the empty
string correctly (`p.textContent = 'ab'` matches official), which is what left
these two residuals visible. They are separate mechanisms from the cooking, and
from each other:

| | mechanism |
|---|---|
| client indent | the instance-script text pipeline re-indents a line it did not recognize as string-interior |
| server no-fold | the constant extractor's initializer pattern is single-line |

The client half is another instance of the class recorded in `CLAUDE.md` — a
scanner assuming input it did not get — this time assuming a line break ends a
string literal.

## Suggested resolution

Fix each with its own repro, and land the shape in
`compatibility/pattern-corpus/matrix/string-escape-fold/` (the axis added in
#2607), which deliberately excludes it today because the file would not be
green.

### LINKED ISSUE #2661: compiler: a server constant fold turns a string concatenation into its own source text
## Summary

A server-side constant fold treats **any** initializer that starts and ends with the same quote as one string literal, so `'ab' + 'cd'` is folded to the raw text between the outer quotes — quote characters and operator included — and rendered into the page.

## Repro

```svelte
<script>
  const cont = 'ab' + 'cd';
</script>

<p>{cont}</p>
```

```diff
-	$$renderer.push(`<p>abcd</p>`);
+	$$renderer.push(`<p>ab' + 'cd</p>`);
```

The client is unaffected — it does not fold this — so the divergence is server-only, and it is **wrong rendered output**, not a missed fold.

## Cause

`try_insert_constant_value` (`3_transform/server/helpers.rs`) tests

```rust
value.starts_with('\'') && value.ends_with('\'')
```

which is not the question "is the whole expression one string literal". `'a' + 'b'` answers yes to both. `try_evaluate_with_constants` repeats the same test a few lines down, and because it runs *before* its own `+` handling, the concatenation branch is never reached.

Same shape for `"` and for a backtick pair without `${`.

## Scope

Found while fixing #2652 — a two-continuation repro produced `ab' + 'cd`, and the plain form above reproduces without any line continuation, so it is an older defect that the repro merely surfaced.

Both compilers' output parses, and the value is a plausible-looking string, so no parse gate or ratchet flagged it; the corpus simply contains no `'a' + 'b'` initializer that reaches an element's text.

baseballyama:
Closes #2652. Also closes #2661, which this PR's own repro exposed.

## Mechanism 1 — the client re-indenters (a wrong value, not a formatting difference)

```svelte
<script>
  const cont = 'a\
b';
</script>
```

`cont` is `ab`. rsvelte compiled a component whose `cont` was `a\tb`:

```diff
 export default function Cont($$anchor) {
 	const cont = 'a\
-b';
+	b';
```

The output is valid JavaScript computing the wrong string, so the parse gate (#2591) is blind to it and only a byte comparison can see it.

The cause is a shape this repo has hit before: the pipeline re-indents every line of a statement except the interior of a template literal, and a `'…'` carried across a line break by a backslash is interior in exactly the same way. There were **three** implementations of that loop and none of them knew:

| site | reached by |
|---|---|
| `normalize_js_with_oxc` fast path | every instance script |
| `normalize_js_with_oxc` slow path | the comment-bearing fallback |
| `legacy_pre_effect` body indent (`reactive_transforms.rs`) | a `$:` **block** body, and nothing else |

The third one is why this PR does not just add a condition. It ran on a duplicate, bool-returning copy of the scanner — a `bool` cannot represent "inside a quoted string", so it could not have been taught the same lesson. The three sites now share one predicate, `in_string_content`, over one scanner that tracks a `Quoted(quote)` frame alongside `Template` / `Interp`, and the duplicate state machine is deleted.

## Mechanism 2 — the server never folded it

`join_continuation_lines` copies a string literal verbatim, so the "logical line" it produces still held the raw newline and `.lines()` re-split it. The literal never entered the constants map:

```diff
-	$$renderer.push(`<p>ab</p>`);
+	$$renderer.push(`<p>${$.escape(cont)}</p>`);
```

The continuation is dropped while copying a `'…'` / `"…"` (never a backtick, whose newlines are content). `cook_string_literal` already maps `\<line break>` to nothing, so the folded value is byte-identical to what a correct scan would produce.

## #2661, found by the repro

A two-continuation case (`'a\…b' + 'c\…d'`) rendered `ab' + 'cd`. Reduced, it has nothing to do with line continuations:

```svelte
<script>
  const cont = 'ab' + 'cd';
</script>

<p>{cont}</p>
```

```diff
-	$$renderer.push(`<p>abcd</p>`);
+	$$renderer.push(`<p>ab' + 'cd</p>`);
```

`try_insert_constant_value` asked `starts_with('\'') && ends_with('\'')`, which is not the question "is the whole expression one string literal" — `'a' + 'b'` answers yes to both. `try_evaluate_with_constants` repeated the test and, because it ran before its own `+` branch, never reached the concatenation it already knew how to do. `is_whole_string_literal` scans for the quote that actually closes. Both sites use it, and the concatenation now folds correctly.

I filed it separately (#2661) rather than folding it in silently, but it lands here because the repro cannot be green without it.

## Positive control

The 8 repro files, same 24 comparisons, against a binary built with the three source files reverted to `origin/main` and nothing else changed:

| | ok | divergent |
|---|---|---|
| `origin/main` binary | 9 | **15** |
| this branch | 24 | 0 |

`template-literal-newline.svelte` is the **negative control** and is among the 9 on both sides — a backtick really does carry its newline as content and had to keep behaving as it did. `legacy-reactive-block.svelte` moves on client and client-dev only, which is the third re-indenter showing itself.

## Repro files

`pattern-corpus/matrix/string-line-continuation/` crosses the continuation with the path it reaches: the runes instance script, a pre-indented continuation (where the indent is *also* content), a legacy `$:` block body, two continuations in one statement (the state has to survive a line that closes one string and opens the next), and the template-literal control.

Two notes on what is **not** there:

- **The quote character is not an axis in the corpus.** The fmt oracle rewrites every literal to double quotes, so a single-quoted repro stops being one the moment it is committed formatted — a `double-quoted.svelte` written alongside `instance-declaration.svelte` came out byte-identical to it after formatting. That axis lives in `crates/rsvelte_core/tests/string_line_continuation_2652.rs` instead, where no formatter runs.
- **`inside-interpolation.svelte` was removed** before this PR. `` `p${'ab'}q` `` is not folded by rsvelte at all, with or without a line continuation, so it is a separate gap — filed as #2662.

## Verified

- 8 repro files × 3 targets: 24 of 24 byte-identical to official, **without** normalization.
- `string_line_continuation_2652`: 5 tests, including both quote characters and the template-literal control.
- Shape matrix: 3315 comparisons, 206 known, 0 new.
- `compiler_fixtures`, `print`, `ssr`, `legacy_pre_effect_comment_2355`, `legacy_pre_effect_parens_1783`, `reactive_block_comment_delimiter_2351` green.


baseballyama:
`Compiler parity` is red here for a real regression in this PR, not a flake — leaving a note since another session is merging green PRs and this one will come back green shortly.

The fix taught the client re-indenter's scanner to track `'…'` / `"…"` frames, and it pushed a *carried string* frame for any quote it could not close on the line. svelte.dev's `packages/repl/src/lib/Output/Viewer.svelte` has

```js
/** Called everytime a log is pushed. If this is set, the built-in console coming with the Viewer isn't shown */
```

so `isn't` opened a string that never closed, and every line below it stopped being re-indented — 2 new `js-mismatch` (client, client-dev).

That the repros did not catch it is the point worth recording: they cross the continuation with the *path* it reaches, and every one of them contains a quote that is genuinely a string. The population where a quote is **not** a string — an apostrophe in prose, a quote inside `/'/g` — is exactly what a collected corpus has and a hand-written repro does not.

Fixed by making the question narrower instead of the scanner smarter. A `'…'` can only cross a line break through a trailing `\`, so `scan_quoted` now answers `Closed` / `Continued` / `NotAString` and a frame is pushed only for `Continued`. Anything else is an ordinary byte, which covers apostrophes in comments and quotes in regex literals without teaching the scanner to lex either.

Test added: the same component compiled with and without the apostrophe, across all three targets, must differ only in the comment's own text.

baseballyama:
Pushed the fix for the regression above, and a correction to my own diagnosis of it.

## What the fix is

`scan_quoted` returned `Option`, and `None` — "ran to the end of the line" — was read as "carried by a line continuation". It answers `Closed` / `Continued` / `NotAString` now, and a frame is pushed only for `Continued`, because a `'…'` can only cross a line break through a trailing `\`. Anything else is an ordinary byte.

`Viewer.svelte`, normalized exactly as `verify.mjs` does (flatten holes → oxfmt → strip blank lines → `ast_equiv_batch`):

| | client | client-dev | server |
|---|---|---|---|
| this branch before the fix | **code-differs** | **code-differs** | equivalent |
| this branch now | equivalent | equivalent | equivalent |

The five existing repros stay green on the same run.

## Correction: it is not one apostrophe

I said above that `isn't` was the trigger. Replacing only that one leaves the broken binary **still** failing — the file has three stray apostrophes (`isn't` line 33, `can't` 144, `there's` 151) and removing any *one* of the first two is enough to make it pass, while removing only the third is not. Two surviving strays are what corrupts: the first opens a frame, the second closes it, and the region between them is treated as string content. Removing all three makes the broken binary match, which is what establishes the cause.

## Six candidate repros, all dropped

None of them failed on the broken binary, so none of them measure anything:

| candidate | result on the broken binary |
|---|---|
| `comment-apostrophe.svelte` (one apostrophe comment above a function) | passes |
| `regex-quote.svelte` (`/'\|"/g`) | passes |
| reduction: TS interface doc comment + snippet | passes (byte-different, AST-equivalent) |
| reduction: apostrophe comment inside an object literal | passes |
| the same two shapes as a `compile()`-level test | passes |
| two apostrophe comments in one script, four layouts | passes |

They are not in this branch. An entry that cannot fail reads as coverage and is worse than a blank.

What is in the branch instead is `quote_frame_tests` in `3_transform/client/formatting.rs`, which asserts the scanner's own frame state. It fails on the broken scanner with `a quote that closes nothing opened a string: /** the console isn't shown */`, and its second case checks the state **between** the two stray apostrophes — after the fourth line the stack is empty either way, so only the intermediate state can tell them apart.

`pattern-corpus/README.md` records the blind spot: every repro in that directory contains a quote that really is a string, which is exactly why none of them could see a scanner that misclassifies a quote that is not.

PR #2670, opened 2026-08-09
fix(compiler): look past comment text when joining continuation lines (server)

### LINKED ISSUE #2669: Block comment disables const-folding when the value is a string with a line continuation
`main` is red on `Corpus Compat` (run 31283116948, `9da01d5e`). The mutation
ratchet gained one entry, `code-mismatch` 36 → 37:

```
pattern/matrix/string-line-continuation/indented-continuation__m0__block-with-paren.svelte [code-mismatch] (server)
  official: …const cont="a\b";$$renderer.push(`<p>ab0</p>`);
  rsvelte : …const cont="a\b";$$renderer.push(`<p>${$.escape(cont)}0</p>`);
```

Official folds the `const` into the template; rsvelte emits a runtime read. The
seed is the fixture #2663 added, so the shape arrived with that PR — the fix
handled the line continuation itself but not const-folding over one.

## The comment does not touch the declaration

```svelte
<script>
  let n = $state(0);
/* c */
  const cont =
    "a\
		b";
</script>

<p>{cont}{n}</p>
```

The comment sits *between* `let n` and `const cont`. It is not inside the
declaration, not inside the string, and not adjacent to either. Its mere
presence stops `cont` being recognised as a constant.

## Isolated: it is not the delimiter, and it needs both halves

The mutant that caught this is the `block-with-paren` kind (`/* ) c */`), and
`)` looks like the obvious cause. It is not. Removing the delimiter leaves the
divergence intact; removing the *line continuation* removes it:

| comment | string | result |
|---|---|---|
| none | `"a\`⏎`b"` | match |
| `/* c */` (plain) | `"a\`⏎`b"` | **diverge** |
| `/* ) c */` | `"a\`⏎`b"` | **diverge** |
| `// c` | `"a\`⏎`b"` | match |
| `/* ) c */` | `"ab"` | match |
| none | `"ab"` | match |

So the condition is **a block comment AND a string line continuation**, both
required. A `//` line comment does not trigger it, which is the discriminating
pair worth keeping in any fix's test.

Client diverges too, at a different insertion point; server diverges at three of
the seed's line boundaries.

## Not a gate-coverage gap

Unlike #2637's operator class, this one *is* reachable by an existing gate — the
mutation fuzz found it on its first sweep after the fixture landed. What it
shows instead is that a `pattern-corpus` fixture becomes mutation-fuzz seed
material, so adding a fixture can surface defects in code paths the fixture
itself does not exercise.

## Reproduction

Insert `/* c */` at line 2 of
`compatibility/pattern-corpus/matrix/string-line-continuation/indented-continuation.svelte`
and compile for `server`.

Found while investigating why `main` went red; not yet fixed.

baseballyama:
Fixes #2669. **`main` is currently red on `Corpus Compat`** (run 31283116948,
`9da01d5e`) — this closes it.

## What broke

`join_continuation_lines` decides whether a physical line continues onto the
next by reading the last non-whitespace byte it has already emitted:

```rust
let prev = out.as_bytes().iter().rposition(|c| !c.is_ascii_whitespace())…
```

Comment text is emitted into that same buffer, and **a block comment ends in
`/`** — which is on the continuation-operator list as division. So a `/* … */`
on its own line made the *next* line join onto it. A joined `const` declaration
no longer starts with `const`, `extract_constant_vars` stopped recognising it,
and the template read came out as a runtime `$.escape(cont)` where upstream
folds the literal in:

```
official: $$renderer.push(`<p>ab0</p>`);
rsvelte : $$renderer.push(`<p>${$.escape(cont)}0</p>`);
```

The comment does not touch the declaration — it sits between `let n` and
`const cont`.

The fix tracks the length of `out` up to the last byte that was *code*, and the
continuation decision reads from there.

## The delimiter was not the cause

The mutant that found this is the `block-with-paren` kind (`/* ) c */`), so `)`
looks like the obvious culprit — the same shape as #2253. It is not:

| comment | string | before |
|---|---|---|
| none | `"a\`⏎`b"` | match |
| `/* c */` (plain) | `"a\`⏎`b"` | **diverge** |
| `/* ) c */` | `"a\`⏎`b"` | **diverge** |
| `// c` | `"a\`⏎`b"` | match |
| `/* ) c */` | `"ab"` | match |
| none | `"ab"` | match |

Removing the delimiter leaves it; removing the line continuation removes it. A
`//` comment never triggers it because its last byte is its own text rather than
`/`. Both halves are required, and the test pins the pairing — a fix aimed at
the `)` would not pass it.

## Verification

- **Regression test** `crates/rsvelte_core/tests/comment_before_continuation_decl.rs`,
  4 tests including both controls. Demonstrated discriminating: against the
  unfixed tree exactly the two block-comment tests fail and both controls stay
  green.
- **Corpus** 14,138 entries × 3 targets: `js-mismatch 0`, `js-unparseable 0`,
  warning/error ratchets unmoved, 39,569 modules parse.
- **Full mutation sweep** (`--full`, what `main` runs, not the PR sample): back
  to the baselined **36**, with none of the 36 newly passing — so the ratchet
  shrinks on the new entry only and is not falsely shrunk elsewhere.

## Known adjacent defect, not fixed here

A comment placed *inside* the declaration — between `const cont =` and the value
— still diverges, and unlike this one it triggers on `//` too, so it is a
different mechanism (the join swallows the value into the line comment). It is
not what `main` is red on and it is not in the ratchet. Filed separately rather
than folded in, so this PR stays reviewable against one cause.

### DIFF
--- a/.changeset/comment-before-continuation-decl.md
+++ b/.changeset/comment-before-continuation-decl.md
@@ -1,0 +1,16 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Stop a block comment from suppressing constant folding of the declaration below it (server).
+
+`join_continuation_lines` decides whether a physical line continues onto the
+next by reading the last non-whitespace byte it has emitted. Comment text went
+into that same buffer, and a block comment ends in `/` — a division operator —
+so a `/* … */` on its own line joined the next line onto itself. A joined
+`const` declaration no longer starts with `const`, so `extract_constant_vars`
+stopped seeing it and the template read was emitted as a runtime
+`$.escape(...)` where upstream folds the literal in.
+
+The continuation decision now looks past comment text to the last byte that was
+actually code.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/helpers.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/helpers.rs
@@ -908,6 +908,11 @@ fn join_continuation_lines(script: &str) -> String {
     let mut depth_paren: i32 = 0;
     let mut depth_brace: i32 = 0;
     let mut depth_bracket: i32 = 0;
+    // Length of `out` up to the last byte that was *code*. A block comment ends
+    // in `/`, which the newline rule below would otherwise read as a division
+    // operator and use to join the following line onto this one; comment text is
+    // not code, so the continuation decision has to look past it.
+    let mut code_len = 0usize;
     let mut i = 0;
     while i < bytes.len() {
         let b = bytes[i];
@@ -957,6 +962,7 @@ fn join_continuation_lines(script: &str) -> String {
             } else {
                 out.push_str(&strip_line_continuations(&script[s..i]));
             }
+            code_len = out.len();
             continue;
         }
         if b == b'(' {
@@ -981,8 +987,7 @@ fn join_continuation_lines(script: &str) -> String {
             // a backtick, etc. The most common case we care about is `=`,
             // but covering operators avoids surprises with hand-formatted
             // declarations like `const x = a +\n  b`.
-            let prev = out
-                .as_bytes()
+            let prev = out.as_bytes()[..code_len]
                 .iter()
                 .rposition(|c| !c.is_ascii_whitespace())
                 .map(|p| out.as_bytes()[p]);
@@ -1020,6 +1025,7 @@ fn join_continuation_lines(script: &str) -> String {
             next += 1;
         }
         out.push_str(&script[i..next]);
+        code_len = out.len();
         i = next;
     }
     out
--- a/crates/rsvelte_core/tests/comment_before_continuation_decl.rs
+++ b/crates/rsvelte_core/tests/comment_before_continuation_decl.rs
@@ -1,0 +1,83 @@
+//! A block comment ends in `/`, and `/` is a division operator.
+//!
+//! `join_continuation_lines` decides whether a physical line continues onto the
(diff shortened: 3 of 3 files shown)


baseballyama:
## Correction: this PR did not close `main`'s red

The header of this PR says "`main` is currently red on `Corpus Compat` … this
closes it". That was wrong, and the merge commit proves it: run
[31286506148](https://github.com/baseballyama/rsvelte/actions/runs/31286506148)
on `62b43297` — this PR's own merge — reports **the same single new divergence,
unchanged**:

```
pattern/matrix/string-line-continuation/indented-continuation__m0__block-with-paren.svelte [code-mismatch] (server)
```

It is byte-identical to the entry on `9da01d5e`, `374cb1ba` and `c2a8eeb6`.
`main` has been red on this one entry since #2663 and still is.

## What went wrong

I reconstructed the mutant by hand instead of generating it. The generator does
not use slot 0 — it picks `slots[fnv1a("<id>#0") % slots.length]`
(`mutate-corpus.mjs:216-217`), and for this seed that is **line 4**, between `=`
and the value:

```svelte
  const cont =
    /* ) c */
    "a\
		b";
```

My repro put the comment on the line *above* the declaration. That shape is a
real defect and this fix is a real fix for it — the regression tests in this PR
still fail on the pre-fix tree. But it is not the shape the gate reports, so
"this closes main's red" was a claim about my repro, not about the referent.

The fix that actually closes it is #2671, which I had already filed as a
separate defect without noticing it was the red one. Follow-up PR incoming; it
subsumes the `code_len` machinery added here.

PR #2672, opened 2026-08-09
fix(compiler): drop comment text from the server constant-fold line join

baseballyama:
Closes the mutation-fuzz divergence that has `main` red since c2a8eeb6:

```
pattern/matrix/string-line-continuation/indented-continuation__m0__block-with-paren.svelte [code-mismatch] (server)
  official: …n=0;constcont="a\b";$$renderer.push(`<p>ab0</p>`);}
  rsvelte : …n=0;constcont="a\b";$$renderer.push(`<p>${$.escape(cont)}0</p>`);}
```

#2670 addressed the same report and `main` is **still red on the identical entry** — the fix and the reported mutant are in different slots. Both are in `join_continuation_lines`, which exists only to rebuild logical lines for `extract_constant_vars`.

| slot | shape | pre-#2670 | with #2670 | here |
|---|---|---|---|---|
| comment ABOVE the declaration | `/* c */`⏎`const cont =` | broken (block only) | fixed | fixed |
| comment BETWEEN `=` and the value | `const cont =`⏎`/* c */`⏎`"a\`… | broken | **broken** | fixed |

#2670 stops the block comment's trailing `/` from being read as a division operator. That is real, but in the second slot the join is *correct* — `=` already asked for it — and the damage is where the comment lands: in front of the value, where `is_whole_string_literal` tests the first byte. For the `// c` form joining is actively harmful, because the value ends up inside the line comment.

So the change here is one thing instead: **a comment becomes a single space.** That is the whole of what the sole consumer can observe — it reads declarator values, and a comment carries none. `code_len` becomes unnecessary and is removed rather than left describing a rule that no longer applies.

## What the delimiter had to do with it: nothing

The mutant carries `)`, which invites a `#2601`-class reading ("a byte scan reads a delimiter inside a comment as code"). Measured across the full comment-kind × insertion-slot product, that is **wrong**:

```
kind               s1 s2 s3 s4 s5          s4 = inserted INSIDE the string
line                .   .   X   X   .      s5 = before </script>
line-with-brace     .   .   X   X   .
line-with-paren     .   .   X   X   .
line-with-semi      .   .   X   X   .
block               .   X   X   X   .
block-with-brace    .   X   X   X   .
block-with-paren    .   X   X   X   .
svelte-ignore       .   .   X   X   .
```

Every kind fails in the same slots; `)` / `}` / `;` change nothing. The `s2`-only-for-block column is the #2670 mechanism, and it is what makes the delimiter reading superficially plausible.

`s4` is **not** a divergence: both compilers reject it with `js_parse_error` and only the message text differs, which is `error-parity` to the gate. My first probe compared messages and reported it as a failure — recording that here because the number in the table would otherwise look like an unfixed case.

The control that identifies the real trigger: the same declaration **without** the line continuation folds under every kind in every slot. It takes continuation × comment.

## Verification

Run on this change (before the rebase onto #2670; re-running now on the rebased tree, will report):

- **Full mutation sweep** — the gate that is failing: `code-mismatch` 37 → **36**, `✅ no code regressions`, 0 crashes, 0 unparseable.
- **Corpus verify**, all 3 targets: `js-mismatch` **0**, `css-mismatch` 0, `output-unparseable` 0, no warning or error regressions.
- **Shape matrix**: 1749 cases / 5247 comparisons, `✅ no regressions`.
- **Negative control on the new tests**: reverting only `helpers.rs` to `origin/main` makes exactly the two added tests fail (4 passed / 2 failed) and every #2670 test pass. They are discriminating, and they fail *for the predicted reason*.

## Note on the gate, not the PR

A seed added by a PR cannot be mutation-tested by that PR: PRs run `--seeds 1500`, only `main` runs `--full`. #2663 added this seed and could not have seen its mutant. Every future `pattern-corpus` addition carries the same exposure — worth a line in `gate-coverage.md` §20 if someone agrees it belongs there.

### DIFF
--- a/.changeset/comment-in-continuation-fold.md
+++ b/.changeset/comment-in-continuation-fold.md
@@ -1,0 +1,14 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Fold a line-continuation string constant when a comment sits between `=` and the value
+
+`join_continuation_lines` reconstructs logical lines for `extract_constant_vars`,
+and it copied comment text into that reconstruction. A comment then landed in
+front of the declarator's value, where `is_whole_string_literal` tests the first
+byte, so the constant was never recorded and the SSR output read it at runtime
+instead of folding it — output that runs correctly and differs from official's.
+
+Comments now become a single space. That is the whole of the difference the sole
+consumer can observe: it reads values, and a comment carries none.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/server/helpers.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/server/helpers.rs
@@ -908,25 +908,24 @@ fn join_continuation_lines(script: &str) -> String {
     let mut depth_paren: i32 = 0;
     let mut depth_brace: i32 = 0;
     let mut depth_bracket: i32 = 0;
-    // Length of `out` up to the last byte that was *code*. A block comment ends
-    // in `/`, which the newline rule below would otherwise read as a division
-    // operator and use to join the following line onto this one; comment text is
-    // not code, so the continuation decision has to look past it.
-    let mut code_len = 0usize;
     let mut i = 0;
     while i < bytes.len() {
         let b = bytes[i];
-        // Line / block comments — copy as-is.
+        // A comment becomes one space. The sole consumer is the constant
+        // extractor below, which reads declarator values, and a comment carries
+        // none — while keeping the text puts it exactly where that extractor's
+        // `starts_with` tests look, and leaves the continuation rule reading the
+        // comment's last character (`= /* c */` ends in `/` and joins the next
+        // line; `= // c` ends in `c` and does not, and joining anyway would let
+        // the `//` swallow the value).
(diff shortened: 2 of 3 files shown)


baseballyama:
Re-ran the three gates on the rebased tree (this branch, i.e. on top of #2670) rather than leaving the earlier pre-rebase numbers standing:

| gate | result |
|---|---|
| Full mutation sweep (`--full`) | `✅ no code regressions` — **36** known remain (main has 37: the extra one is the entry that is failing) |
| Corpus verify, 3 targets | `✅ all corpus outputs identical after normalization`; `output-unparseable` 0; `✅ no warning regressions` (32 known); `✅ no error regressions` (236 known) |
| Shape matrix | `✅ no regressions` — 206 known, unchanged; `error-code-mismatch` 0 |

Same as before the rebase, which is the expected result but not one I wanted to assume — the rebase replaced #2670's `code_len` with the comment-to-space rule, so the tree that produced the earlier numbers is not this one.

PR #2674, opened 2026-08-09
chore(measure): split the client template-fragment visitor by node kind

baseballyama:
Instrumentation only, off by default (`measure-tf-split`), zero-cost when off.

`Template fragment` was the second-largest Phase 3 bucket (18.2% of compile) and
one timer around one call — `fragment(&ast.fragment, …)`. `visit_node` is the
single dispatch point for every template node kind, so the guard goes there. The
visitor recurses, so an inclusive timer would charge an element for everything
inside it; the guard keeps a child-time accumulator and charges **self time**.

## What it found

Per-corpus, min over 5 runs (the machine was not quiet — see the caveats):

| corpus | `component` | `regular_element` | Σ split / parent | residual |
|---|---|---|---|---|
| bits-ui | **45.2%** (2,788 calls) | 24.0% (1,412) | 93.5% | 6.5% |
| flowbite-svelte | **46.1%** (6,801) | 27.5% (2,802) | 93.9% | 6.1% |
| shadcn-svelte | **67.8%** (16,432) | 15.7% (2,487) | 95.9% | 4.1% |
| layercake | **38.3%** (479) | 37.8% (586) | 93.4% | 6.6% |

Two node kinds carry **62–84%** of the bucket. Everything else — `if_block`,
`each_block`, `render_tag`, `const_tag`, and the 19 other kinds — is under 8%
each. The residual (4–7%) is `fragment()`'s own work plus `Text` and
`ExpressionTag`, which the fragment loop handles inline and which therefore
never pass through `visit_node` (they read as 0 calls, by construction, not
because templates lack them).

Per call, `component` is the **cheaper** of the two (2.8–6.8 µs vs 4.2–5.5 µs).
Its share is call count, and call count is a property of real component
libraries.

## Why the ranking is not the instrument

`component` has the most calls, and the guard adds one `Instant` pair per
visited node, so "the biggest row is the row with the most timer pairs" is the
obvious confound. It is bounded and reported inline:

```
TF-OVERHEAD  20164 pairs x 58.9ns = 1.188ms (0.9% of parent)
```

0.9% cannot produce a 67.8% row. The pair cost also lands in the *parent* frame,
not the child's, because the child's timer starts after its guard's bookkeeping.
A paired feature-on/off A/B is below this machine's noise floor (on < off in 2
of 4 corpora), which agrees with the 0.9% bound rather than adding to it.

## Caveats

- **Do not size this from the fixtures corpus.** There the order inverts —
  `regular_element` 6.9% of compile against `component` 1.0% — one more instance
  of the fixture corpus skewing shares.
- Absolute ms here are not comparable to the clean phase table: the machine had
  other builds running throughout. Shares are min-estimator over 5 runs;
  `component > regular_element` holds in **5/5** runs on bits-ui, flowbite and
  shadcn, and **3/5** on layercake, where the two are within 2% of each other.

### DIFF
--- a/crates/rsvelte_core/Cargo.toml
+++ b/crates/rsvelte_core/Cargo.toml
@@ -58,6 +58,11 @@ measure-pa-split = []
 # its share of `compile()` can be measured. Off by default; the build counts are
 # unconditional, only the `Instant` pair is gated.
 measure-semantic-build = []
+# Repository-local instrumentation: charge self time per template node kind
+# inside the client fragment visitor. The visitor recurses, so the guard also
+# maintains a child-time accumulator; one `Instant` pair per visited node makes
+# it far too heavy to leave on. Off by default and zero-cost when off.
+measure-tf-split = []
 # Select the JavaScript entropy backend when the compiler is embedded in wasm.
 # Binding exports live in `rsvelte_lint_bindings`, not in this library.
 wasm-target = ["dep:getrandom", "getrandom/wasm_js"]
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/types.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/types.rs
@@ -68,7 +68,10 @@ impl<'a> ComponentContext<'a> {
     }
 
     /// Index into `profile::TF_KINDS`. Kept next to `visit_node`'s match so the
-    /// two arms cannot drift apart silently.
+    /// two arms cannot drift apart silently. Gated with its only call site so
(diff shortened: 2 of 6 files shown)


baseballyama:
`skip-changeset` applied. The whole change is behind `measure-tf-split`, which
is off by default, and the lookup that feeds the guard is now gated with its
call site — so a default build of `@rsvelte/compiler` compiles to the same code
it did before this PR. That is what makes the label correct here rather than a
bypass: there is no consumer-visible change to name.

PR #2676, opened 2026-08-09
chore(measure): split build_component into stages

baseballyama:
Stacked on #2674 — review that one first; this adds only the second level.

#2674 found that `component` is the largest row of the template-fragment bucket
on every shipping corpus (38–68%). It is one 675-line function, so the row named
a target without locating it. These guards split it, reusing #2674's accumulator
so the stages are **subtracted from** `component` rather than added beside it and
Σ-against-parent still holds.

Per-corpus, min over 5 runs, share of the `Template fragment` parent:

| corpus | `bc:slot_function` | `bc:regular_attribute` | `component` (rest) | `regular_element` |
|---|---|---|---|---|
| bits-ui | 12.3% (2,135) | 8.2% (3,946) | 9.6% (2,788) | 23.6% (1,412) |
| flowbite-svelte | 17.1% (4,656) | **20.6%** (10,264) | 10.2% (6,801) | 24.8% (2,802) |
| shadcn-svelte | **29.9%** (12,165) | 9.6% (16,915) | 20.3% (16,432) | 14.1% (2,487) |
| layercake | 4.1% (255) | **28.4%** (1,075) | 4.8% (479) | 37.2% (586) |

`component`'s share was **not** one hot spot. It is `build_slot_function` plus
`process_regular_attribute` plus a residual body, and which of the two leads
depends on the corpus — slots on shadcn/bits-ui, attributes on flowbite/
layercake. The other eleven stages are each under 6% everywhere;
`bc:props_expression` runs once per component (16,432 calls on shadcn) for 1.1%.

Per call the two differ in kind, which is the part worth carrying forward:
`bc:slot_function` costs 1.5–1.7 µs and is called once per slot;
`bc:regular_attribute` costs 0.4–2.2 µs — a 5× spread across corpora at the
same call site, which is a question this instrument raises and does not answer.

Same caveats as #2674: the machine was not quiet, absolute ms are not comparable
to the clean phase table, and the fixtures corpus inverts the ordering.

### DIFF
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/component.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/component.rs
@@ -802,6 +802,9 @@ fn process_let_directive(
     lets: &mut Vec<JsStatement>,
     let_names: &mut Vec<(String, Option<String>)>,
 ) {
+    let _tf = crate::compiler::phases::phase3_transform::profile::tf_guard(
+        crate::compiler::phases::phase3_transform::profile::TF_BC_LET,
+    );
     let prop_name = &let_dir.name;
 
     // Check if expression is an Identifier or null (simple case)
@@ -975,6 +978,9 @@ fn process_on_directive(
     context: &mut ComponentContext,
     events: &mut IndexMap<String, Vec<JsExpr>>,
 ) {
+    let _tf = crate::compiler::phases::phase3_transform::profile::tf_guard(
+        crate::compiler::phases::phase3_transform::profile::TF_BC_ON,
+    );
     // If no expression, mark that component needs props for event bubbling
     // This is handled via build_event_handler which sets needs_props_from_events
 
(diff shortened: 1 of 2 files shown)


baseballyama:
Added a third level, because `bc:slot_function` is a wrapper around a single
call and the row could not say whether the wrapper or the call carried it.
Splitting `visit_slot_children` out answers it — the wrapper is nothing:

| corpus | `bc:slot_children` | wrapper (rest of `build_slot_function`) |
|---|---|---|
| shadcn-svelte | 27.1% of parent, 1.66 µs/call | falls out of the top 7 |
| flowbite-svelte | 16.3%, 1.65 µs/call | " |
| bits-ui | 11.6%, 1.41 µs/call | " |

So the save/restore of let-directive transforms costs nothing measurable. One
hypothesis died cheaply on the way: `context.state.transform_deep_read.clone()`
runs unconditionally although only the `let:`-directive branch reads it, which
looks like 12,165 wasted clones on shadcn — but the field is an
`im::HashMap`, whose clone is O(1). Not a defect.

With this level the whole bucket reads, per call:

```
each_block          8.2 µs      bc:snippet_block   2.3-8.2 µs
bc:snippet_block    7.2 µs      regular_element    4.4-4.7 µs
regular_element     4.7 µs      bc:slot_children   1.4-1.7 µs
if_block            6.2 µs      component          0.7-0.9 µs
                                bc:regular_attribute 0.4-1.0 µs
```

The rows that own the bucket are the *cheap* ones. `component` at 0.7–0.9 µs and
`bc:regular_attribute` at 0.4–1.0 µs lead on call count, not unit cost — so an
optimisation aimed at unit cost here has a much smaller ceiling than the share
suggests, and the two expensive kinds (`each_block`, `bc:snippet_block`) are too
rare to matter. That is the finding, not a step towards one.

PR #2699, opened 2026-08-09
fix(compiler): take client statement boundaries from the parser

baseballyama:
First slice of the script-pipeline AST migration. AGENTS.md justifies that work
by correctness — 35 real components where rsvelte emitted output no JS parser
accepts, every one "a scanner assuming input it did not get" — and this replaces
the scanner that decides where a statement ends.

## What it replaces

The line loop settled that question five ways: balanced depths, a trailing comma
in a declaration, **a list of operators a statement cannot end on**, a
brace-less control header, and a lookahead for a continuation token on the next
line. The third is a list, so anything missing from it splits a statement.

A single oxc parse of the script answers it exactly. All five heuristics are
neutralised when the parse succeeds; when it does not, they stay, so nothing
that worked stops working.

## It fixes real shapes, and I measured which

A/B over end-of-line operators in `$: v = a <op>⏎ b;`, `main` vs this branch:

```
main diverges from upstream on:  -  /  ?  .
this branch:                           ?  .
fixed:                           -  /
broken:                          (none)
```

`-` and `/` are the regression tests; `+` and `&&` were already in the list and
are asserted as controls. `?` and `.` still diverge and this change does not
move them — a different cause, not this one.

## The path is actually taken

A byte-identical corpus cannot distinguish "the parser answered" from "we fell
back every time", so the adoption rate is counted:

| corpus | boundaries from the parser |
|---|---|
| bits-ui | 517 / 517 (100%) |
| flowbite-svelte | 743 / 743 (100%) |
| shadcn-svelte | 1050 / 1050 (100%) |
| layercake | 173 / 175 (98.9%) |
| svelte-ux | 166 / 180 (92.2%) |

## Gates

- corpus verify, all three targets: **js-mismatch 0, js-unparseable 0**, all
  14,166 entries identical after normalization; no warning or error regressions
- mutation fuzz, full sweep: **36 known remain, no regressions**
- shape matrix: no regressions (206 known)
- `cargo fmt --check`, `clippy --all-targets --all-features -D warnings` clean

Two wrong versions preceded this one, both caught by the corpus rather than by
reading: cutting at every line no statement covers split multi-line block
comments (419 failures), and then flushing at every comment's end line cut
statements that contain a comment (207). Both are recorded in the code as the
reason those two lines are the way they are.

## Not in this PR

The per-line depth scan still runs, because the scanner is still the fallback.
Removing it on the parser path is the next slice, and it is what pays back this
PR's extra parse; I have not claimed a speedup here and have not measured one.

### DIFF
--- a/.changeset/statement-boundaries-from-parser.md
+++ b/.changeset/statement-boundaries-from-parser.md
@@ -11,5 +11,8 @@ line. Each is an approximation, and the operator list is a list — a line endin
 in `-` or `/` was not on it, so `$: v = a -⏎ b;` split into two statements and
 `b` stopped being a dependency.
 
-The boundaries now come from one oxc parse of the script. A script that does not
-parse at that point keeps the scanner, so nothing that worked stops working.
+The boundaries now come from a parse of the script — the program Phase 1 already
+holds where that text is a verbatim region of it, and a fresh parse otherwise. A
+script that does not parse at that point keeps the scanner, so nothing that
+worked stops working, and the per-line depth scan no longer runs when a parser
+answered.
--- a/.changeset/statement-boundaries-from-parser.md
+++ b/.changeset/statement-boundaries-from-parser.md
@@ -1,0 +1,15 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Take the client instance script's statement boundaries from the parser.
+
+The pipeline decided where a statement ended by scanning characters: balanced
+depths, a trailing comma, a list of operators a statement cannot end on, a
+brace-less control header, and a lookahead for a continuation token on the next
+line. Each is an approximation, and the operator list is a list — a line ending
+in `-` or `/` was not on it, so `$: v = a -⏎ b;` split into two statements and
+`b` stopped being a dependency.
+
+The boundaries now come from one oxc parse of the script. A script that does not
+parse at that point keeps the scanner, so nothing that worked stops working.
--- a/.changeset/ts-boundaries-through-projection.md
+++ b/.changeset/ts-boundaries-through-projection.md
@@ -1,0 +1,7 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Place TypeScript statement boundaries through the strip projection, so the client
+instance-script pipeline reuses the program Phase 1 already parsed instead of
+parsing the script a second time.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -4660,24 +4660,120 @@ fn statement_end_lines_from_spans(
     flush
 }
 
+/// The same spans, for a `script` that TypeScript stripping rewrote, placed
+/// through the projection that records which source bytes survived into it.
+///
+/// Only the first and last byte of each span are mapped, never the whole span:
+/// `let a: number = 1;` loses `: number`, so no annotated statement is one
+/// contiguous copied run, while its two endpoints almost always are — and the
+/// endpoints are all a line boundary needs.
+///
+/// A span whose endpoints both fail to map is a construct stripping deleted
+/// outright (`interface Foo {}`), so it is skipped rather than treated as a
+/// failure: bailing on it would give up on every file that declares a type.
+/// One endpoint mapping without the other is not interpretable, so that bails.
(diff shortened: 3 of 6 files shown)


baseballyama:
## Update: the added parse, and what it cost

CodSpeed priced the first commit honestly: `transform_client[synthetic-legacy-state-heavy]`
8ms → 9.2ms, **−12.91%**, 1 regressed of 149. That is the parse this PR added.

I checked whether dropping the depth scan could pay it back before assuming it would.
On that same input the boundary parse is the file's **only** direct parse
(1 call, 14,320 bytes, 0.07ms), while the depth scan sits inside a 0.06ms residual.
It cannot. So the parse should not have been added at all.

**Phase 1 already parses the instance script and retains the program**, and already
hands it to this very function for the state transform. The boundaries now come from
it when the pipeline's text is a verbatim, uniquely-located, statement-aligned region
of what it parsed; a fresh parse is the fallback, and the scanner is still the
fallback for text that does not parse.

### Measured, on 5,879 shipping files

| | |
|---|---|
| boundaries from a parser | 4,227 / 4,243 (99.62%) — scanner 16 |
| …of those, **reused** Phase 1's parse | 1,231 (**29.1%**) |
| …still **added** a parse | 2,996 (70.9%) |
| oracle (`RSVELTE_BOUNDARY_ORACLE=1`) | **1,232 checks, 0 mismatches** |

On the benchmark that regressed: **1/1 reused, 0 parses added.**

**29% is not most files.** TypeScript components bail — the retained program holds
pre-stripping text, so the match fails — and TS is the bulk of this corpus. This PR
does not make the parse free in general, only for plain-JS components. The mechanism
for the rest already exists (`strip_typescript_from_program_with_projection` maps
stripped text back to original offsets) but wiring it in is a separate change.

### Why there is a hit counter

Reusing a parse answers the same question a fresh parse does, so **bailing on every
file would leave all 14,138 corpus entries byte-identical and the whole path dead**.
No parity gate can see the difference. `the_boundaries_come_from_the_program_phase_1_already_parsed`
counts hits through `compile()`; disabling the reuse leaves the other five tests in
that file green and fails only that one.

It has to run through `compile()`: `transform_component` hard-codes
`retained_scripts: None`, which is what made my first profiler reading say 0%.
That was a defect in the harness reported in the vocabulary of a verdict, and the
harness now measures it through `compile()` with the reason recorded at both sites.

### Gates

- corpus verify ×3 targets: 13,218 match, **0** js-mismatch / js-unparseable / css / error; 39,653 modules parse
- warnings + errors: no regressions (32 / 236 known)
- shape matrix: 3,928 match, 206 known, no regressions
- 6/6 boundary tests, including import-rebasing and the reuse counter

`line_scan` is **5.92ms — 0.4% of compile** on shipped code. Slice 2 is real but small
at corpus scale; the 0.06→0.01ms I measured on one synthetic is not a corpus-wide figure.

baseballyama:
## CodSpeed: the regression is on a path that does not ship

Refreshed on `787a5c4`: **−11.22%**, one benchmark,
`transform_client[synthetic-legacy-state-heavy]`.

That benchmark calls `transform_component`, which hard-codes
`retained_scripts: None` (`3_transform/mod.rs:88`). **The parse reuse cannot fire
there.** `compile()` threads the retained programs; this entry point does not.

CodSpeed measured the control itself, in the same run, on the same input:

| benchmark | entry point | retained | verdict |
|---|---|---|---|
| `transform_client[synthetic-legacy-state-heavy]` | `transform_component` | **None** | ❌ −11.22% |
| `full_compile/client[synthetic-legacy-state-heavy]` | `compile()` | yes | ✅ not in the regression table |

Both draw from the same `workload()`, so this is one input measured two ways,
differing only in whether the retained program is threaded.

The arithmetic agrees: slice 1 cost **+12.91%**, slice 2 (dropping the depth
scan) returns **1.69 points** → **11.22%**, and slice 3 contributes **zero here
by construction** while removing the parse entirely under `compile()`
(1/1 reused, 0 parses added on this exact file).

### This is a gate defect, not a property of the change

`bench_transform` measures a configuration that never ships. Two consequences:

- every optimisation that depends on the retained program is **invisible** to it —
  slice 4 will remove 2,996 real parses and move this benchmark by nothing
- any cost landing on the non-retained path is counted at **full weight**

Worth fixing separately (it shifts the baseline of every transform benchmark, so
it does not belong in this PR).

PR #2748, opened 2026-08-09
fix(check): honour JSDoc types in JavaScript-authored components

### LINKED ISSUE #2730: svelte-check: JSDoc types are ignored in JS-authored components (shadows are always .tsx)
Shadows are emitted as `<name>.svelte.tsx` for every component, including those whose `<script>` is JavaScript. TypeScript honours JSDoc type annotations **only in JS files**, so every `/** @type {...} */` in a JS-authored component is discarded — producing both false positives and false negatives.

Official `svelte-check` does not have this problem: `DocumentSnapshot` gives a JS-authored component `ScriptKind.JSX`, so the same text is parsed as JavaScript and its JSDoc applies.

## The mechanism, isolated

Same file, two extensions, `allowJs: true, checkJs: true`, no `strict`:

```jsx
export function $$render() {
	/** @type {HTMLDivElement} */
	let elem;

	setTimeout(() => { elem.attachShadow({ mode: 'open' }); });
	async () => { elem = document.createElement('div'); };
	return {};
}
```

```
a.tsx(4,6): error TS7034: Variable 'elem' implicitly has type 'any' in some locations ...
a.tsx(7,3): error TS7005: Variable 'elem' implicitly has an 'any' type.
c_nojsdoc.tsx(4,6): error TS7034: ...          <- control: identical without the JSDoc
c_nojsdoc.tsx(7,3): error TS7005: ...
b.jsx: (no errors)
```

The `.tsx` copy and the JSDoc-stripped control produce byte-identical diagnostics — the annotation contributes nothing. `.jsx` is clean.

## Impact, measured

`sveltejs/kit@ef86ba7` `packages/kit/test/apps/basics` (540 components, JS-authored with JSDoc types), `rsvelte-check --tsgo --tsconfig tsconfig.json`:

| | errors |
|---|---|
| `rsvelte-check` | **68** |
| `svelte-check@4` | **0** |

All 68 are `TS7034` / `TS7005` / "implicitly has an 'any' type" on variables that carry a JSDoc `@type`, e.g. `src/routes/routing/shadow-dom/+page.svelte`:

```svelte
<script>
	/** @type {HTMLDivElement} */
	let elem;
	onMount(() => { const shadow = elem.attachShadow({ mode: 'open' }); });
</script>
<div bind:this={elem}>…</div>
```

The false-negative half is the same bug from the other side. A probe component added to that app:

```svelte
<script>
	const zz = 1;
	zz.toUpperCase();

	/** @type {number} */
	const yy = 'a string';
</script>
```

| | `zz.toUpperCase()` | `/** @type {number} */ const yy = 'a string'` |
|---|---|---|
| `rsvelte-check` | reported | **missed** |
| `svelte-check@4` | reported | reported |

So official is genuinely type-checking that project (it reports the plain error too, and 0 elsewhere) — the 68 are ours alone, and the JSDoc violation is invisible to us.

## Fix sketch

Emit `<name>.svelte.jsx` when `looks_like_ts_svelte(&source)` is false. Not a one-line change: the `.tsx` suffix is assumed in the entry/manifest records, in `rewrite_aliased_svelte_imports` (which builds a *target* component's shadow specifier and would need that target's TS-ness), in `emit_svelte_module_bridges` / `shadow_reexport`, and in the mapper's path back to the `.svelte` source. It also needs a decision for a project with `allowJs: false`, where a `.jsx` shadow would leave the program entirely — a silent regression to the current, merely-wrong behaviour.

Found while verifying #2710 / #2714 (fixed in #2729); deliberately left out of that PR.

baseballyama:
Closes #2730.

A `.svelte` file with no `lang="ts"` was shadowed as `.tsx`, and TypeScript ignores JSDoc entirely in a TS file. Every `/** @type {…} */` in such a component was therefore discarded — the annotation neither constrained the value nor reported a violation, while the props it was meant to type produced implicit-`any` errors the author could not act on.

Official svelte-check reaches the right place through `ScriptKind.JS` on its in-memory snapshot (`getScriptKindFromAttributes`); an on-disk overlay has to say it with the file name, so a JS-authored component now gets a `.jsx` shadow. `allowJs` is added to the overlay config when such a shadow exists (it defaults off, and the shadow would otherwise leave the program silently, taking its importers' resolution with it); `checkJs` is deliberately left to the project, exactly as upstream leaves it. The counterpart shadow under the other extension is removed on each run, so a component that changes language does not enter the program twice.

Two svelte2tsx ports were incomplete in the same direction and had to land with it:

- `addComponentExport`'s Svelte 5 isomorphic branch emitted `type X = InstanceType<typeof X>` unconditionally. Upstream picks `export const` plus a `@typedef` when emitting JSDoc, because declaration merging with a `@typedef` needs the const exported. The runes branch already did this; the isomorphic one did not.
- `ExportedNames.handleTypeAssertion` computes its SvelteKit `$types` injection from `tsType || ts.getJSDocType(declaration)`; rsvelte looked only at the TS annotation. `/** @type {any} */ export let form` therefore got `import('./$types.js').ActionData` injected over the author's own type — and on a route with no server actions that reports a member the route's `$types` never had.

### Verification

`packages/kit/test/apps/basics` from https://github.com/sveltejs/kit/pull/16692, against official `svelte-check@4`'s 0 errors:

| | errors |
|---|---|
| before | 68 |
| after | 6 |

A JSDoc violation that used to be accepted silently is now reported, which is the point of the change rather than a side effect.

Of the 6 that remain, 5 are `TS2683 'this' implicitly has type 'any'` and are **not** rsvelte's: TypeScript 7 (`@typescript/native` 7.0.2, which is what `basics` resolves as `typescript`) ignores a JSDoc `@this` when a `@type` is present on the same function, in `.js` and `.jsx` alike, while `tsc` 6.0.3 does not — confirmed on a standalone two-file probe with one tsconfig. The 6th is a `$types` injection whose diagnostic official drops because it lands in generated text; that needs the mapper to distinguish an insertion from an edited chunk and is filed separately.

https://claude.ai/code/session_01CQFGbmKgydpTBfK2T1D7da

### DIFF
--- a/.changeset/svelte-check-jsdoc-shadows.md
+++ b/.changeset/svelte-check-jsdoc-shadows.md
@@ -1,0 +1,6 @@
+---
+'@rsvelte/svelte-check': patch
+'@rsvelte/svelte2tsx': patch
+---
+
+Honour JSDoc types in JavaScript-authored components. A `.svelte` file without `lang="ts"` was shadowed as `.tsx`, where TypeScript ignores JSDoc entirely — so `/** @type {…} */` neither constrained a value nor reported a violation, while the props it left untyped produced implicit-`any` errors the author could not act on. Such a component now gets a `.jsx` shadow (official svelte-check reaches the same place through `ScriptKind.JS`), the overlay adds `allowJs` when one exists, and the Svelte 5 isomorphic component export emits the `export const` + `@typedef` form upstream uses for JSDoc output. `checkJs` is left to the project. A SvelteKit route prop that already carries a JSDoc `@type` also keeps it: `ts.getJSDocType` suppresses the `import('./$types.js')` injection just as a TS annotation does, so `/** @type {any} */ export let form` on a route with no server actions no longer reports a nonexistent `ActionData`
--- a/crates/rsvelte_check/src/svelte_check/overlay.rs
+++ b/crates/rsvelte_check/src/svelte_check/overlay.rs
@@ -585,10 +585,24 @@ pub fn materialize_overlay_with(
         let tsx_path = emit_dir.join(&tsx_rel);
         let dts_path = emit_dir.join(&dts_rel);
         let map_path = append_extension(&tsx_path, ".map");
+        // `X.svelte.d.ts` is TypeScript's declaration counterpart of
+        // `X.svelte.jsx`, so with a JS shadow the bridges' own
+        // `./X.svelte.jsx` resolves to the twin, which re-exports from
+        // `./X.svelte.jsx` — a cycle that types the component `any` and
+        // silently drops every diagnostic about using it. `.tsx` escapes
+        // it through `allowImportingTsExtensions`, which resolves an
+        // explicit TS extension literally. Dropping the twin costs nothing:
+        // `allowJs` is on whenever a `.jsx` shadow exists, so CJS-mode
+        // resolution of `./X.svelte` finds the shadow itself, and ESM mode
+        // still goes through the `.d.svelte.ts` bridge.
+        let emit_dts_twin = ext == ".tsx";
         if let Some(parent) = tsx_path.parent() {
             fs::create_dir_all(parent)?;
         }
         remove_stale_counterpart(&tsx_path, ext);
+        if !emit_dts_twin {
+            let _ = fs::remove_file(&dts_path);
+        }
 
         let stats_match = match (stats, cached_entry) {
             (Some((mtime, size)), Some(entry)) => {
@@ -604,7 +618,10 @@ pub fn materialize_overlay_with(
         // source map; we don't gate cache validity on it, so a workspace
         // that gained / lost source maps still hits the cache. On hit we
         // simply best-effort-read whatever sits at `map_path`.
-        let can_skip = incremental && stats_match && tsx_path.exists() && dts_path.exists();
+        let can_skip = incremental
+            && stats_match
+            && tsx_path.exists()
+            && (!emit_dts_twin || dts_path.exists());
 
         let source_map = if can_skip {
             fs::read_to_string(&map_path).ok()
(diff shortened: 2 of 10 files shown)


baseballyama:
CI found two more defects that only a JS-authored component can reach, both fixed in 9df7b5a9. Each was invisible before this PR because the shadow was `.tsx`, and each showed up as the `test-error` golden fixture reporting **zero** TypeScript diagnostics rather than as a wrong one.

**`Binding.ts`'s `useTypescriptSyntax` was not ported.** A `bind:contentRect` on a component with no `lang="ts"` emitted `null as DOMRectReadOnly`. That is legal in a `.tsx` shadow and is `TS8016` in a `.jsx` one — a *syntax* error, so it suppressed every semantic diagnostic in the program. The svelte2tsx fixture covering this shape passed anyway: the comparison strips `as` assertions and JSDoc casts before the byte match, so `null as T` and `/** @type {T} */ (null)` are indistinguishable to it.

**`<name>.svelte.d.ts` is TypeScript's declaration counterpart of `<name>.svelte.jsx`.** Both bridges next to a JS shadow name it as `./<name>.svelte.jsx`, which therefore resolved to that twin — which re-exports from `./<name>.svelte.jsx`. The cycle types the component `any`, so every diagnostic about *using* it disappears (the fixture lost `TS2741` on a component invoked without its required prop). `.tsx` never had the problem: `allowImportingTsExtensions` resolves an explicit TS extension literally. Two experiments separated it from everything else — the same shadow content renamed `.jsx` → `.tsx` restores `TS2741`, and deleting the twin restores it as `{ b: boolean; }`, which is exactly the type official svelte-check reports.

The twin is now emitted only next to a `.tsx` shadow. `allowJs` is on whenever a `.jsx` shadow exists, so CJS-mode resolution of `./<name>.svelte` finds the shadow itself; ESM mode still goes through the `.d.svelte.ts` bridge. A test pins both directions.

Re-measured on `packages/kit/test/apps/basics` with #2729 and #2751 also applied (cache cleared): **5 errors**, all the TypeScript 7 `@this` + `@type` behaviour described in #2730, none of them rsvelte's.

One caution for anyone re-running this: a measurement on this branch **alone** reports `0 errors`, and that zero is vacuous — without #2729 the `extends: "$app/tsconfig"` chain is not followed, `types: ["$app/types"]` raises `TS2688`, and one config-level diagnostic suppresses every semantic diagnostic program-wide. All three PRs have to be in the tree for the number to mean anything.

Issue #2697, opened 2026-08-09
napi: compile.modernAst and parse.skipCssAst are declared at the boundary, accepted, and have no consumer


baseballyama:
Found while building the napi option-boundary gate (#2539), which crossed 38 of 40
declared option keys. Two of the keys it *did* reach turn out to be declared and inert,
and they are the same defect in two places.

## The two sites

1. **`compile.modernAst` is accepted and ignored.** `crates/rsvelte_core/src/compiler/mod.rs:835`
   hardcodes `ast: None, // TODO`. Setting `modernAst: true` changes nothing. This is
   exactly the #2438 shape and is still live.
2. **`parse()` declares `skipCssAst` and never reads it.** The key is on the shared
   `NapiParseOptions`, so it is declared at the boundary, accepted from JS, and dropped.

## Why these two belong in one issue

They are not "two options to wire up" — they are one *class*: **a key can be declared at
the boundary, accepted without error, and have no consumer.** #2539's gate is keyed on
*declared* fields and asserts that each declared key is crossed by a case, which is what
found these. But "crossed by a case" is not "has an effect": for these two the case
asserts current behaviour, and current behaviour is nothing.

Upstream does not have this failure mode in the same way, because it forwards the parsed
options object into the compiler rather than mapping field by field — the same argument
#2687 makes about `svelte-check` dropping `accessors` / `customElement`. That is the
general form worth fixing; these two are its instances at the napi boundary.

## What "fixed" means

For each key, one of:

- **implement it** — `modernAst` should produce the modern AST that upstream's
  `modernAst: true` produces, and `skipCssAst` should actually skip building the CSS AST;
  or
- **reject it** — if it is not going to be supported, the boundary should say so rather
  than accept and discard, because silent acceptance is what makes a downstream tool
  believe it opted in.

Choosing "reject" for either is a legitimate outcome, but it must be a decision recorded
in the code and covered by a test, not the current default of silence.

## Gate

#2539's `scripts/dev/test-napi-compile-options.mjs` (`pnpm run test:napi-options`) is the
right home: it already enumerates the declared surface both directions. What it needs for
this class is a third state per key beyond covered/uncovered — **declared but inert** —
so that an option with no consumer is an explicit, asserted status rather than a case that
happens to assert nothing. Otherwise a future inert key reads as covered, which is the
same "vacuous measurement" trap `AGENTS.md` warns about: an empty measurement is worse
[...]

PR #2769, opened 2026-08-10
fix(napi): reject skipCssAst outside parseEnvelope (#2697)

baseballyama:
## Summary

- reject `skipCssAst` on the JSON `parse()` endpoint instead of accepting and ignoring it
- retain the supported `parseEnvelope()` path and assert both behaviors in the option-boundary gate

## Validation

- `cargo fmt --check`
- `node --check scripts/dev/test-napi-compile-options.mjs`
- `git diff --check`
- native addon rebuild and `pnpm run test:napi-options` are in progress

Partially addresses #2697; `compile.modernAst` remains tracked there.


baseballyama:
`pnpm run build:vps-native && pnpm run test:napi-options` completed successfully: 95 passed, 0 failed.

PR #2770, opened 2026-08-10
fix(compiler): warn for deprecated options when false (#2542)

### LINKED ISSUE #2542: compiler: options_deprecated_accessors requires a truthy value in runes mode, and options_deprecated_immutable is missing for the compile option
Found while fixing #2486 (PR #2538), which covered the four `options_removed_*` / `options_renamed_*`
codes and deliberately stopped there. Two `options_deprecated_*` codes are also wrong, in
different ways.

## 1. `options_deprecated_accessors` fires under the wrong condition

Upstream warns whenever the key is **present** (`validate-options.js`), which includes
`accessors: false` and non-runes components. rsvelte requires the value to be truthy *and* the
component to be in runes mode:

```rust
options.accessors && runes_mode
```

So `accessors: false` — a user explicitly opting out, who still wants to know the option is going
away — gets nothing, and a legacy-mode component with `accessors: true` gets nothing either.
Both directions are silent where upstream warns.

## 2. `options_deprecated_immutable` is missing for the compile option

Upstream raises it from `validate-options.js:98` for the **compile option**. rsvelte implements
only the `<svelte:options immutable>` attribute site, so the option spelling is unreported.

## Reachability, and why the corpus cannot see either

Both are functions of the *options object*, not of the source. `compile.mjs:99` fixes the corpus
options at `{generate, dev, filename}` (+ `css: 'external'`), so no corpus entry ever passes
`accessors` or `immutable` and the warning ratchets cannot move for either code. That is #2486's
group-B observation, and it is why the acceptance test has to be a direct compile with the option
set, not a corpus run.

The same applies to the fix in PR #2538 — it was verified with a differential run over
`{true, false, null, absent}` × the keys × `compile`/`compileModule` against the official
compiler. The same table is the right shape here, and it is what distinguishes "present" from
"truthy" for `accessors`.

## Note on ordering

Upstream emits these from a validator key table, and #2538 had to order its emissions to match.
Adding two more codes means re-checking that order against upstream rather than appending.

baseballyama:
## Summary

- retain `accessors` and `immutable` option presence through the N-API boundary
- emit the upstream deprecation warnings for explicit `false`, legacy mode, and direct Rust `true` options
- add core and N-API-boundary controls

## Validation

- `cargo fmt --check`
- `cargo test -p rsvelte_core --test legacy_compile_options_warnings -- --nocapture`
- `node --check scripts/dev/test-napi-compile-options.mjs`
- `git diff --check`

Fixes #2542

### DIFF
--- a/.changeset/deprecated-option-presence.md
+++ b/.changeset/deprecated-option-presence.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Warn for deprecated `accessors` and `immutable` options whenever they are supplied, including `false`.
--- a/crates/rsvelte_core/src/compiler/mod.rs
+++ b/crates/rsvelte_core/src/compiler/mod.rs
@@ -115,6 +115,10 @@ pub struct LegacyOptions {
     /// `generate: 'dom' | 'ssr'` — the pre-Svelte-5 spellings of
     /// `'client'` / `'server'`, still honoured but renamed.
     pub generate_dom_ssr: bool,
+    /// `accessors` — deprecated in Svelte 5. Presence matters even for `false`.
+    pub accessors: bool,
+    /// `immutable` — deprecated in Svelte 5. Presence matters even for `false`.
+    pub immutable: bool,
     /// `loopGuardTimeout` — removed in Svelte 5.
     pub loop_guard_timeout: bool,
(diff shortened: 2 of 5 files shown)


baseballyama:
`pnpm run build:vps-native && pnpm run test:napi-options` passed after rebuilding the addon: 96 passed, 0 failed.

PR #2774, opened 2026-08-10
fix(compiler): preserve async derived leading comments (#2755)

### LINKED ISSUE #2755: codegen(client): a block comment before an async $derived emits `$.run([async () => void (/* … */ const a = await …)])` — not JavaScript
Disclosed by the `async-derived` matrix family added in #2540 (PR #2733): **14 ratcheted entries**.
Pre-existing; not what that PR fixed.

## The defect

A block comment before an async `$derived` produces:

```js
$.run([async () => void (/* … */ const a = await …)])
```

A `const` declaration inside a `void` operand is **not JavaScript** — no parser accepts it.

## Family

This is the client instance-script text-splicing class that `AGENTS.md` documents with its
"what the scanner assumed / what broke it" table, and it is the fourth comment-position member found
in this campaign alongside **#2705** (block comment before a next-line operand → `g + )`),
**#2727** (`//` on the last `<script>` line spliced inside `$.set(…)`), and **#2737**. Four sites,
one assumption: that a comment does not sit between the construct being rewritten and the text the
rewriter expects to find next.

## Why no ordinary gate sees it

Two independent reasons, both recorded in `AGENTS.md`:

1. every ratchet scores **match/mismatch**, so it structurally cannot distinguish "wrong text" from
   "text that is not JavaScript" — feeding output to a parser is what found 35 real-world components
   in this class;
2. source **layout** is the fingerprint of the text path, and the collected corpus is normalised by
   formatters, so the comment-before-declaration shape is under-sampled. Absence from 14,026 real
   files is not evidence of rarity.

Note this one *is* currently gated, but only as a ratcheted entry in the `async-derived` family —
i.e. it is recorded, not watched. A ratchet entry suppresses everything about the entry it lists.

## Suggested acceptance

- assert the emitted JS **parses** (acorn / oxc), not merely that it matches
- the 14 entries leave the ratchet in the same PR (every ratchet here is two-sided)
- sweep the neighbours before choosing where the fix goes: `//` in the same slot, a block comment
  before a *non-async* `$derived`, and the same shapes in `<script module>` and a `.svelte.js`
  module — separate parse entry points where a fix in one has repeatedly not reached the others

The durable fix is the AST route rather than teaching the scanner about block comments; `AGENTS.md`
states the justification for the AST pipeline is that this defect class becomes unreachable, not
that it is faster.

Controls that must not move: an async `$derived` with no comment, and a block comment before a
plain `const`.

baseballyama:
Fixes #2755.

Leading line and block comments are now separated before async-body statement classification and retained on the generated hoisted declaration. This prevents a block comment followed by an async `$derived` declaration from being misclassified as an expression and emitted as invalid `void (/* … */ const …)`.

The generated async-derived matrix runs all entry points and targets. This change clears the client and client-dev block-inline rows and re-baselines the shrink-only ratchet. The existing output-parseability gate was also run to verify that output-ratchet entries cannot hide syntactically invalid JavaScript.

Validated with:
- `cargo test -p rsvelte_core --lib async_body::tests::test_leading_comments_on_async_var_declarations -- --nocapture`
- `pnpm run corpus:matrix:update -- --keep-artifacts`
- `node scripts/dev/test-corpus-parse-gate.mjs`

### DIFF
--- a/.changeset/async-derived-leading-comment.md
+++ b/.changeset/async-derived-leading-comment.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve leading comments when lowering asynchronous derived declarations.
--- a/compatibility/matrix-known-failures.json
+++ b/compatibility/matrix-known-failures.json
@@ -32,8 +32,6 @@
 	"async-derived/instance__identifier__block-before.svelte [js-mismatch] (client)",
 	"async-derived/instance__identifier__block-before.svelte [js-mismatch] (client-dev)",
 	"async-derived/instance__identifier__block-before.svelte [js-mismatch] (server)",
-	"async-derived/instance__identifier__block-inline.svelte [js-mismatch] (client)",
-	"async-derived/instance__identifier__block-inline.svelte [js-mismatch] (client-dev)",
 	"async-derived/instance__identifier__block-inline.svelte [js-mismatch] (server)",
 	"async-derived/instance__identifier__line-before.svelte [js-mismatch] (client)",
 	"async-derived/instance__identifier__line-before.svelte [js-mismatch] (client-dev)",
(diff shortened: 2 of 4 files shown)


baseballyama:
Closing this attempt: the proposed leading-comment hoist corrupts an async runtime output into an empty parenthesized expression, caught by the Compatibility Report parser gate. The failure proves this is not a safe local preservation strategy. #2755 remains open; its fix must use the shared comment-cursor path rather than encoding comments into hoisted declaration text.

PR #2776, opened 2026-08-10
test: honor fixture dev options in runtime oracles (#2746)

### LINKED ISSUE #2746: test: do the runtime/SSR oracles compile expectations under the options the harness runs them with? validator's did not (6 uncompared, 2 vacuously green)
Found while fixing #2452/#2579 (PR #2731), which fixed this class **for the validator category
only**. The same mechanism plausibly affects the runtime and SSR oracles, and that has not been
measured.

## The mechanism, as confirmed for validator

`generateValidatorFixture` compiled each sample under **default** options while the harness ran it
under the sample's own `runes` / `customElement`. Official therefore emitted warnings its own
expectations did not contain, and **six fixtures had been silently uncompared** — fixing it moved
the compared population from 159 to **165**.

Worse in kind: `dev: true` was never read, so two samples ran in prod mode expecting `[]`. They
were **vacuously green** — passing a comparison that could not have failed.

## Why this is filed rather than closed with that PR

`parseConfigText` in `scripts/.../generate-fixtures.mjs` deliberately drops `dev` for **all**
categories, with a documented reason about SSR dev-mode divergence. #2731 made the validator
category parse its options locally instead, so validator no longer depends on that path — but
runtime and SSR still do.

So the question is open and specific: **do the runtime and SSR oracles compile their expectations
under the same options the harness later uses to run them?** If not, the same two failure shapes
apply — expectations generated under the wrong options (a comparison that is wrong) and
expectations generated under options that make the assertion trivially true (a comparison that is
vacuous).

The documented reason for dropping `dev` may well be correct for SSR. That is exactly why it needs
measuring rather than assuming: a deliberate exclusion carries its scope, and the scope stated was
SSR dev-mode divergence, not "all categories, all options".

## What would settle it

For each of runtime, SSR, and hydration:

1. enumerate the options a sample's `_config.js` can set that affect compilation
2. compare that set against the options `generate-fixtures.mjs` actually passes when producing the
   expectation
3. report the **count of fixtures where they differ**, not a rate — and per `AGENTS.md`, a zero
   needs a negative control: show a case where a mismatch WOULD be detected, so a zero means "no
   mismatches" rather than "the check cannot see mismatches"

Then, for any mismatch found, classify it as *wrong* (expectations disagree with the run) or
*vacuous* (expectations trivially satisfied). The second is the more dangerous and the harder to
notice; #2731's two `dev: true` samples were green for years.

## Related gate facts from #2731, worth reading first

- `_config.js` is read by **substring match** and only five keys are read; `options.json` is not
  read at all
- the harness passes `generate: client` where upstream passes `generate: false`
- input normalisation differs from upstream's (unmeasured, not a known live defect)

Any of those could hide a mismatch this issue is asking about, so the sweep should account for them
rather than trusting the config reader.

baseballyama:
Fixes #2746.

Runtime, hydration, SSR, skipped-fixture audit, and the compatibility report now compile with each fixture’s configured `compileOptions.dev`, matching the official fixture generator. This exposes option-sensitive divergences instead of comparing production output to a development expectation.

`invalid-nested-svelte-element` is intentionally skipped by upstream because its SSR diagnostic is printed rather than thrown; the shared SSR skip list now mirrors that scope.

Validated with:
- `pnpm run generate-fixtures`
- `cargo test -p rsvelte_core --test runtime -- --nocapture`
- `cargo test -p rsvelte_core --test ssr -- --nocapture`
- `cargo test -p rsvelte_devtools --test compatibility_report --no-run`

### DIFF
--- a/crates/rsvelte_core/tests/audit_skipped.rs
+++ b/crates/rsvelte_core/tests/audit_skipped.rs
@@ -178,6 +178,7 @@ fn audit_runtime(category: &str, name: &str) -> Outcome {
 
     if let Some(expected) = &expected_client {
         let options = CompileOptions {
+            dev: fixture_options.dev,
             generate: GenerateMode::Client,
             filename: Some("main.svelte".to_string()),
             css: CssMode::External,
@@ -203,6 +204,7 @@ fn audit_runtime(category: &str, name: &str) -> Outcome {
 
     if let Some(expected) = &expected_server {
         let options = CompileOptions {
+            dev: fixture_options.dev,
             generate: GenerateMode::Server,
             filename: Some("main.svelte".to_string()),
             css: CssMode::External,
--- a/crates/rsvelte_core/tests/common/mod.rs
+++ b/crates/rsvelte_core/tests/common/mod.rs
@@ -390,6 +390,7 @@ impl FixtureCoverage {
 #[derive(Debug, Clone, Copy, Default)]
 pub struct RuntimeFixtureOptions {
     pub r#async: bool,
+    pub dev: bool,
     pub hmr: bool,
     /// The generator only passes `accessors` to the client compile.
     pub accessors: bool,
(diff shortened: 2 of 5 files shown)


baseballyama:
Closing this attempt: propagating fixture `dev` directly into the compile oracle changes the established runtime execution contract and produces 139 runes plus 27 legacy regressions. The Compatibility Report also catches a reduced SSR fixture population. #2746 remains open; the fix must distinguish fixture runtime mode from the compiler options used to generate its oracle rather than treating them as one boolean.

PR #2816, opened 2026-08-11
fix(napi): reject unsupported modern AST option (#2697)

baseballyama:
## Summary

- reject `modernAst: true` at the N-API compile boundary
- cover the rejection with the native N-API options test
- add a patch changeset

## Root cause

The option was accepted and copied into core compile options, but rsvelte never produces a modern compiler AST, so callers received a successful compile with no effect.

## Validation

- `cargo fmt`
- `cargo clippy -p rsvelte_napi --lib -- -D warnings`
- `pnpm run build:vps-native`
- `pnpm run test:napi-options`

### DIFF
--- a/.changeset/reject-modern-ast.md
+++ b/.changeset/reject-modern-ast.md
@@ -1,0 +1,6 @@
+---
+"@rsvelte/compiler": patch
+---
+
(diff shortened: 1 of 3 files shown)


baseballyama:
Closing per maintainer direction: this is superseded by the current main implementation.

Issue #2756, opened 2026-08-09
codegen: three independent async-derived lowering divergences — $derived.by(async) suspended (13), missing $.save on non-final await (13), $$d temp hoisted (39)


baseballyama:
Disclosed by the `async-derived` matrix family added in #2540 (PR #2733). Three **independent**
clusters, filed together only because they share a family and a burndown; each needs its own fix and
its own control, and a partial fix must not close the others.

Total 65 ratcheted entries. All pre-existing; none is what #2540 fixed.

## Cluster 1 — `$derived.by(async () => …)` is suspended as if it were an async derived (13 entries)

`$derived.by` takes a thunk the user supplies. rsvelte appears to treat an `async` thunk as making
the derived itself async and suspends it. Establish what upstream does before fixing — `$derived.by`
and `$derived` have different lowering in
`submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/VariableDeclaration.js`,
and the answer is a measurement, not an inference.

Control that must not move: `$derived.by(() => …)` with a synchronous thunk.

## Cluster 2 — a multi-`await` derived loses the `$.save(…)` upstream wraps non-final `await`s in (13 entries)

Upstream wraps every **non-final** `await` in `$.save(…)`; rsvelte does not. #2540 fixed the inverse
defect one slot over (rsvelte wrapping `$.save(…)` around destructured async deriveds where upstream
does not), so both directions of this pairing are now known to be wrong — which suggests the
`$.save` placement rule is not modelled at all rather than modelled incorrectly. Worth confirming
before patching a case.

`$.save` exists to preserve reactive context across a suspension point, so a missing one is a
**runtime** divergence, not only a text one. Judge it on that basis, and say in the fix whether a
runtime test is warranted.

Controls: a single-`await` derived (no non-final `await`, so no `$.save`), and the destructured case
#2540 fixed.

## Cluster 3 — the `$$d` destructuring temp is hoisted into the component's top-level `var` list (39 entries)

On **all three targets**. Upstream keeps it scoped. This is the largest of the three and the only one
that is target-independent, which usually means it sits in shared lowering rather than in a
per-target visitor — a useful starting point, not a conclusion.

Control: a non-destructured derived must not gain or lose a hoisted temp.

## Gate

All 65 are the family's live output on `main`, so this is a burndown against an existing,
discriminating gate rather than new gate work. Report the delta **per cluster** so a partial fix
reads as partial, with raw counts rather than a rate.

[...]

PR #2817, opened 2026-08-11
fix(client): preserve async derived context (#2756)

baseballyama:
## Summary

- wrap non-final awaits in async-derived declarations with `$.save`
- keep generated destructuring temporaries inside the async callback
- remove the 17 now-passing async-derived matrix ratchet entries

## Root cause

The instance AST lowering bypassed the existing non-final-await handling, while the shared async-body hoist treated compiler-generated `$$` temporaries as user bindings.

## Validation

- focused `rsvelte_core` regression tests
- `cargo clippy -p rsvelte_core --lib -- -D warnings`
- full matrix: 13,925 comparisons, no regressions (923 known divergences)


baseballyama:
Reopening to re-trigger the missing pull_request CI event for the current head.

PR #2823, opened 2026-08-11
fix client rune trailing comment ownership

### LINKED ISSUE #2718: codegen(client): a trailing comment after $effect(…) changes official's argument-list wrapping; rsvelte keeps the call on one line
Client and client-dev targets. A comment trailing an `$effect(…)` statement makes
official attach it to the effect's **callback argument**, which forces esrap's
wrapped one-argument-per-line call layout. rsvelte attaches it after the whole call
statement and keeps the call on one line. The comment survives on both sides — this
is a placement / layout divergence, not a loss, and it is byte-visible.

Measured on `origin/main` @ `9b1d004c` against the pinned `submodules/svelte`
(Svelte 5.56.8), oxfmt-normalized exactly as `verify.mjs` normalizes.

```js
// a.svelte.js
export function f(a) {
	$effect(() => {
		console.log(a);
	}); // c
}
```

```diff
-  $.user_effect(
-    () => {
-      console.log(a);
-    }, // c
-  );
+  $.user_effect(() => {
+    console.log(a);
+  }); // c
```

## Scope, measured

From the generated `removed-statement-comment` matrix family (comment kind × slot ×
host × successor): **every** `trailing`-slot row diverges on `client` and
`client-dev` — 114 and 144 comparisons respectively — and no `leading` or
`interior` row does, except one `instance-top` pair. It reproduces on all three
hosts the family generates: `compileModule` (`.svelte.js`), the component instance
script's top level, and inside a function in the instance script. It is
target-symmetric between `client` and `client-dev`.

The mechanism is the one `rsvelte_esrap`'s `flush_trailing_comments` implements
(`crates/rsvelte_esrap/src/printer.rs:604-642`, and its call from the call-argument
loop at `:2970`): esrap emits a comment on the same line as an **argument**'s end
and propagates the resulting newline into the surrounding context's `multiline`,
which is what forces the wrapped form. rsvelte's client path never gets the comment
into that position, so the layout decision is never triggered.

## Provenance

Found by the `removed-statement-comment` matrix family, added by the PR that fixes
#2567 — a server-only fix. Pre-existing on `main`: it moved none of these rows in
either direction. The rows are enrolled in
`compatibility/matrix-known-failures.json` with this issue as their justification.

Note the enrolment cost: a ratchet entry suppresses everything about the entry it
lists, so those 258 ids are now blind to any *further* client regression on the same
shapes until this is fixed.

baseballyama:
Closes #2718

Restores upstream comment ownership for rebuilt `$effect`, `$effect.pre`, `$effect.root`, and dev `$inspect` calls by preserving argument spans while clearing synthesized wrapper locations.

Validated with `cargo test -p rsvelte_core --test comment_delimiters_in_initializer`, `cargo clippy --all-targets --all-features -- -D warnings`, and the removed-statement-comment matrix.

### DIFF
--- a/.changeset/client-rune-trailing-comments.md
+++ b/.changeset/client-rune-trailing-comments.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Keep trailing comments attached to transformed client rune arguments.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/to_oxc.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/to_oxc.rs
@@ -188,6 +188,33 @@ struct SingleTargetSequenceRebuilder<'a, 'x> {
     ab: &'x AstBuilder<'a>,
 }
 
+/// Calls rebuilt from user runes retain their argument locations but not a
+/// location for the call itself. Keeping the parsed call span would make a
+/// trailing source comment belong to the generated call rather than its last
+/// user argument.
+struct RebuiltRuneCallLocations;
+
+impl<'a> VisitMut<'a> for RebuiltRuneCallLocations {
+    fn visit_call_expression(&mut self, call: &mut CallExpression<'a>) {
+        oxc_ast_visit::walk_mut::walk_call_expression(self, call);
+
+        if ["user_effect", "user_pre_effect", "effect_root"]
+            .iter()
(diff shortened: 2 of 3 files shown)


baseballyama:
Superseded by the later complete fix for the same issue in #2830, which merged and closed #2718. Closing this stale overlapping PR to avoid a duplicate/conflicting implementation.

Issue #2754, opened 2026-08-09
codegen(dev): <script module> / .svelte.js async-derived lowering is inside out — 154 entries, and it blocks dev instrumentation on that path


baseballyama:
Disclosed by the `async-derived` matrix family added in #2540 (PR #2733): **154 ratcheted
entries**, the largest of five clusters that family exposed. Pre-existing; none of them is what
#2540 fixed.

## The defect

In `<script module>` and `.svelte.js`, the dev lowering is **inside out**:

```js
// rsvelte
(await $.track_reactivity_loss($.async_derived(() => p)))()

// upstream wraps the thunk body, not the call
```

Two further shapes in the same cluster:

- **destructured module declarations are not lowered at all**
- on `server`, a module-level derived is **read without being called**

## Why this blocks work rather than merely diverging

#2540 added dev arguments to `$.async_derived`. Adding them in the module path **would have been
invisible**, because the shape above them is wrong first — the arguments would sit inside a call
that upstream does not emit. So this cluster is upstream of any further dev-instrumentation work on
the module path, in the same way #2644 was upstream of #2626's fixture.

## Entry points

`AGENTS.md` records that the instance script, `compileModule`, and a template expression are three
different parse entry points in rsvelte, and that #2547's fix was incomplete in exactly that way.
This cluster is that hazard realised: the instance-script path is right and the module path is not.
Whoever fixes it should check all three rather than the two named here.

## Gate

Already generated and discriminating — the 154 entries are the family's live output on `main`, so
this is a burndown against an existing gate rather than new gate work. Mirror
`submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/VariableDeclaration.js`
(`:211`, `:245` are the sites #2540 mirrored for the instance path).

Report the delta per shape — call wrapping, destructured declarations, server read-without-call —
so a partial fix reads as partial. Raw counts, not a rate.

## One consequence to fix along the way

A sibling cluster in the same family (the `svelte-ignore` comment not being reproduced on the
hoisted declaration, 18 entries) is also ratcheted. Because a ratchet entry suppresses everything
about the entry it lists, **the ignore axis cannot currently gate what it was added for.** #2540's
two direct Rust tests are what watch it in the meantime. Clearing that cluster restores the axis;
say so explicitly when it is cleared, because nothing else will announce it.

baseballyama:
[...]

PR #2843, opened 2026-08-11
fix(compiler): preserve module derived await patterns

baseballyama:
Fixes the directly-awaited destructured `$derived` lowering in module scripts.

Upstream preserves `const [a, b] = await p`; rsvelte created an async-derived cell in both the client module transform and the SSR module AST transform.

Validation:
- `cargo test -p rsvelte_core --test module_direct_async_derived_2754`
- module-derived AST unit tests
- server AST unit suite
- clippy with warnings denied

This is one validated #2754 subcluster; the remaining module/dev shapes stay tracked by #2754.

### DIFF
--- a/.changeset/module-derived-await-pattern.md
+++ b/.changeset/module-derived-await-pattern.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve directly awaited destructured derived declarations in module scripts.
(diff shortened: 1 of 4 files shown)


baseballyama:
Closing: the full matrix exposed that this direct-await lowering is incorrect under the gated  option. Upstream emits  for destructured module declarations, so this would add 43 new client/server mismatches. The remaining #2754 shapes stay tracked for a semantics-correct fix.

baseballyama:
Correction to the closing note: under the matrix `experimental.async` option, upstream emits `$.async_derived` for destructured module declarations. The proposed direct-`await` lowering would add 43 new client/server mismatches, so it is not valid.

PR #2862, opened 2026-08-12
fix(server): instrument dev option bodies

baseballyama:
## Summary

Match Svelte dev-SSR instrumentation for `<option>` callback bodies by emitting the same `$.push_element` and `$.pop_element` markers around their renderer callback body.

This removes the shared missing-instrumentation cluster while leaving synthetic-value options unchanged, as upstream does.

## Validation

- `ast_matches_oracle_dev_instrumentation` (including standalone and nested select/option cases)
- Raw official-vs-rsvelte dev-SSR comparisons for three corpus option shapes
- Full corpus compile: 14,198 entries × 4 targets; 53,004 parseable output pairs
- Full corpus verifier: client 0, server 0, client-dev 0, server-dev 239 → 99
- Full four-target matrix: 4,711 cases / 18,428 comparisons; baseline unchanged at 872
- `cargo fmt --check`, `git diff --check`


baseballyama:
Superseded by #2860 (`8c35fe482`), which includes the same `emit_option_special` dev instrumentation and completes the full server-dev output ratchet at zero. Closing this duplicate implementation PR; its two direct oracle cases may be considered separately if further coverage is desired.

PR #2864, opened 2026-08-12
fix(server): preserve dev snippet guard order

baseballyama:
## Summary
- emit non-hoistable server-dev snippet stringification guards in template declaration order
- retain the preceding `{@const}` before its snippet guard
- add a focused server-dev regression

## Verification
- `cargo fmt --check`
- `RUST_MIN_STACK=33554432 cargo test -p rsvelte_core --test server_dev_snippet_guards_2610`
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings`
- `pnpm run corpus:compile:ssr-dev && pnpm run corpus:verify:ssr-dev` (14,198 entries; green, artifacts cleaned)
- `pnpm run corpus:matrix` (22,931 comparisons; no regressions; 872 known divergences)

### DIFF
--- a/.changeset/fix-server-dev-snippet-guard-order.md
+++ b/.changeset/fix-server-dev-snippet-guard-order.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve declaration order for dev SSR snippet stringification guards.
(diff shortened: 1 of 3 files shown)


baseballyama:
Superseded by a consolidated PR that will include both this snippet-guard order repair and #2865’s duplicate-target repair. Current main requires both for the corpus gate to pass, so neither standalone PR can attain a meaningful green verdict.

PR #2865, opened 2026-08-12
fix(corpus): remove duplicate server-dev target

baseballyama:
## Summary
- remove the duplicate `server-dev` corpus target descriptor
- add a target-list invariant so mutation accounting cannot enroll a target twice

## Verification
- `pnpm run test:corpus-targets`
- `node --check scripts/compat-corpus/targets.mjs`
- `node --check scripts/compat-corpus/mutate-corpus.mjs`

### DIFF
--- a/.changeset/fix-corpus-server-dev-target.md
+++ b/.changeset/fix-corpus-server-dev-target.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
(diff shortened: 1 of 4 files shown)


baseballyama:
Superseded by a consolidated PR that will include both this duplicate-target gate repair and #2864’s snippet-guard order repair. Current main requires both for the corpus gate to pass, so neither standalone PR can attain a meaningful green verdict.

PR #2881, opened 2026-08-12
fix(svelte2tsx): restore projection registry checks

baseballyama:
Fix current main CI and two exposed projection correctness bugs.

- update projection tests for the current accessor APIs
- mask packed span/VLQ values before narrowing to `u32`
- preserve named `$props()` type references without an unnecessary alias
- synchronize the generated C ABI header wrapping

Validated: projection lib tests (265), projection clippy, C ABI build, registry dependency-surface gate, fmt, diff check.

### DIFF
--- a/.changeset/fix-projection-packed-spans.md
+++ b/.changeset/fix-projection-packed-spans.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/svelte2tsx': patch
+---
+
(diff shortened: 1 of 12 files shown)


baseballyama:
Superseded by #2882: current main already contains the projection accessor updates, packed span/VLQ narrowing fixes, named `$props()` type handling, and generated C header sync. Closing this duplicate branch rather than re-running CI against an obsolete merge base.

PR #3008, opened 2026-08-16
feat(language-server): gate LSP parity against the official server (#1767)

### LINKED ISSUE #1767: Wave 4 M6: LSP parity ratchet, differential harness and benchmarks
Part of the Wave 4 epic. Runs alongside M1–M5.

## Goal

Hold LSP parity to the same standard as the compiler, formatter, svelte2tsx and lint gates: a
shrink-only ratchet with a per-entry justification file.

## Tasks

- [ ] `compatibility/lsp-known-failures.json` + `lsp-known-failures.md`, shrink-only in CI (same shape as the four existing gates)
- [ ] Differential harness: drive the official `svelte-language-server` and `rsvelte-language-server` with the same `initialize` + request stream over stdio and diff the responses. Upstream has **no** end-to-end protocol test, so this has to be built from scratch
- [ ] Reuse upstream fixture suites — `test/plugins/typescript/features/{diagnostics,folding-range,inlayHints}/` are `.svelte` + `expected*.json` pairs and can be diffed directly
- [ ] Port the TS-independent unit tests: `test/plugins/{svelte,html,css}/**`
- [ ] Reuse the 174 files in `test/plugins/typescript/testfiles/` as inputs for the differential harness
- [ ] Run the harness over the real-world corpus repos already pinned in `scripts/compat-corpus/corpus-sources.json` (bits-ui, flowbite-svelte, melt-ui, shadcn-svelte) — request hover/definition/completion at every identifier and diff
- [ ] Benchmarks: cold start, time-to-first-diagnostic, hover/completion latency percentiles, memory on a large SvelteKit project, vs. the official LS
- [ ] Crash/robustness: malformed input, mid-edit parse errors, tsgo child crash and restart, cancellation storms

Tracked by #1768.

baseballyama:
Closes #1767 (Wave 4 M6).

Holds `rsvelte-language-server` to the same standard as the compiler, formatter, svelte2tsx and lint gates: a shrink-only, two-sided ratchet with a per-cluster justification file.

## The harness

Upstream ships **no** end-to-end protocol test — every `language-server` test in `submodules/language-tools` constructs a plugin class in-process — so this had to be built from scratch. `scripts/compat-corpus/lsp-verify.mjs` drives the pinned real `svelte-language-server` and `rsvelte-language-server` over stdio with one identical `initialize`, one identical client-capability set and one identical request stream, and diffs the responses.

| | |
|---|---|
| projects | 11 (2 committed fixture projects, 4 upstream fixture suites, 4 real-world corpus repos, 1 upstream testfiles workspace) |
| units | 227 files |
| response pairs compared | 7231 |
| already agree | 2804 (38.8%) |
| ratchet | 3044 entries, `compatibility/lsp-known-failures.json` |
| wall clock | ~17 min |

Run: `pnpm run test:language-server-parity`. CI: a new `lsp-parity` job in `corpus-compat.yml`.

## The key is a class, not a payload

Entries are `<unit>|<method>|<verdict>` where the verdict is `only-official` / `only-rsvelte` / `count` / `differs:<≤3 fields>` / `error-official:<code>` / `error-rsvelte:<code>` / `error-both:<a>/<b>`.

A payload key would churn on every TypeScript wording change. A `<unit>|<method>` key would let a divergence that *changes kind* reuse an existing entry — the failure mode #2521 recorded for the shape matrix. Diagnostics are compared per `source` (`publishDiagnostics[ts]`, `[svelte]`, `[rsvelte]`, …) for the same reason: rsvelte's server also publishes its native linter's findings, which official has no counterpart for, and one flat key would mask every TypeScript divergence in the same file behind them.

The cost — once an entry exists the payload inside that class is no longer compared — is written down as blind spot 27a rather than left implicit.

## Two defects it found on its first runs

**1. One malformed message killed the server.** `lsp_server::Connection::stdio` ends its reader thread on the first frame whose body will not deserialize, which closes the connection and takes the whole server down — one bad message from any client, extension or proxy in the chain and every open document loses its language features. The body of such a frame has already been consumed in full, so the stream is still framed correctly; `src/transport.rs` drops the message with a warning and keeps reading. A malformed *header* stays fatal, because the reader no longer knows where the next frame begins. (Changeset included.)

**2. The gate's own projection erased semantic tokens.** `data` was dropped everywhere as the opaque `completionItem/resolve` payload — and `data` is also where a semantic-tokens response keeps its entire answer, so every token stream compared equal to every other. The gate reported **zero** semantic-token divergences across all 227 files while rsvelte was returning `{"data": []}` and official a full stream. There are 166 of them. `scripts/dev/test-lsp-normalize.mjs` now fails on a projection that erases the field under comparison, and runs in `ci.yml` without needing either server, the oracle install, or any submodule.

## Positive controls — fatal, not ratcheted

A gate that can go green by measuring nothing is the failure mode this repo has paid for repeatedly, so three things abort the run instead of producing entries:

- **The oracle is calibrated against upstream's own snapshots.** For the 15 `upstream/folding-range` fixtures the official server's stdio response is compared to upstream's `expectedv2.json`. It reproduces **13/15**; the two misses are one extra whole-`<script>` fold each, contributed by the HTML folding provider that upstream's provider-level test does not run. Below 0.8 the run aborts — a run measured against a server not behaving as its own tests say would produce a plausible ratchet full of meaningless entries.
- **A project that agrees on nothing** is one server failing to start, not 800 divergences.
- **`--update`** refuses below 150 compared units, and refuses when `--positions` / `--max-units` / `--corpus-limit` are overridden: those three numbers *are* the population, and a baseline written from a subset deletes every entry it did not measure.

Determinism: two full `--update` runs and one full verify run produced the identical 3044-entry set (0 new, 0 fixed).

## Also in this PR

- **Benchmarks** (`scripts/compat-corpus/lsp-bench.mjs`, `pnpm run lsp-corpus:bench`): cold start, time-to-first-diagnostic, hover/completion p50/p90/p99, and peak RSS **of the whole process tree** (rsvelte's tsgo child is a separate process, so a single-pid reading is not the cost of running the server). Both servers in one invocation. It asserts nothing — it is a benchmark, and there is no performance ratchet for the language server (blind spot 27j).
- **Robustness** (`tests/robustness.rs`, 4 tests): malformed frames and wrong-typed params; a document that is unparseable at nearly every intermediate keystroke; a tsgo child that dies the moment it is spawned; 200 requests each cancelled immediately (every one must still be answered — a client that never hears back leaks the id forever).
- **A case-for-case port of upstream's `html/getFoldingRange.test.ts`** (`tests/upstream_html_folding.rs`). Its expectations come from `vscode-html-languageservice`, not from either server, so they are worth asserting directly. Nine of ten hold as upstream wrote them; the tenth (a stray unmatched close tag, where upstream folds and rsvelte offers nothing) is asserted as today's behaviour with upstream's expectation named beside it.
- `tests/protocol.rs`'s stdio harness moved to `tests/support/protocol_server.rs` and is now shared by all three integration suites.
- **gate-coverage §27** — eleven blind spots, each with the responsible code cited: the payload inside an accepted verdict, sampled positions (identifiers only — never inside a string, comment, CSS selector or attribute value), sampled files, the fields the projection drops (including completion **ordering**), no `*/resolve` anywhere, the methods not in the stream, **no `didChange` at all** (a unit is opened once and never edited, so the incremental path both servers declare is unmeasured), one capability point, one configuration point, the TypeScript-major asymmetry, and no timing assertion.

## What is partial, and why

The issue's *"port the TS-independent unit tests: `test/plugins/{svelte,html,css}/**`"* is done by **input reuse** rather than by transcribing all ~155 inline assertions: those suites' fixture directories are projects in the population, and their expectations come from the live official server, which is both stronger (it stays in sync with upstream) and cheaper to maintain. `html/getFoldingRange.test.ts` is the exception and is ported literally, because its expectations are third-party. Transcribing the other ten files would need its own ratchet — most cases would fail today — and is worth a follow-up issue rather than this PR.

Likewise, the corpus sweep samples (25 components per repo, 4 identifier positions per file) rather than requesting *"hover/definition/completion at every identifier"*: the flags to widen it are there, the numbers are part of the gate's definition, and 27b/27c say so plainly rather than letting the sampling read as coverage.

One thing this PR deliberately does **not** change: a tsgo child that can never start is respawned every 250 ms forever. The robustness test documents that the server survives it; a bounded backoff is a behaviour change that belongs in its own PR.

https://claude.ai/code/session_017LzPenwNb9zAGazC8JK2fH


baseballyama:
Superseded by #2980, which landed on `main` as `a24ad8d2d` and closed #1767 while this branch was in flight. Closing rather than rebasing: the two implementations cover the same issue and #2980 is already merged, so rebasing would mean re-litigating a shipped design.

The one part of this branch that #2980 does **not** contain is a product fix rather than a gate: `lsp_server::Connection::stdio` ends its reader thread on the first frame whose body will not deserialize, which takes the whole server down. That is carried forward on its own in #3009, with a negative control showing the test fails without it.

Issue #3015, opened 2026-08-17
Phase 3 discards source spans, so the source map is rebuilt by text matching


baseballyama:
## The 11 source-map passes are a consequence of a discarded field, not work the map needs

`js_ast/to_oxc.rs` states the assumption that created them:

> Synthesized nodes use the dummy `oxc_span::SPAN`: esrap formats structurally, so their spans do not affect output.

That is true right up until the output *is* a source map. Phase 3 throws the source
position away at IR-construction time, so by the time a map is wanted there is nothing
left to emit from, and #2954 recovers it by matching generated text back against the
source. Each lowering shape needs its own matcher, which is why there are eleven of them
and why every one is a full scan of both texts.

`JsExpr::Raw(CompactString)` is the harder half of the same thing: those fragments have no
structure at all, so there is no node to carry a span even if the IR had the field.

### Measured (58 upstream sourcemap fixtures, 818 official segments)

| pipeline | official segments reproduced | emitted segments |
| --- | --- | --- |
| esrap alone, the eleven passes disabled | **239/818 (29.2%)** — 469 missing, 110 wrong | 863 |
| with the eleven passes (main today) | **815/818 (99.6%)** | 1,589 |

So the text matchers are producing **576 of the 818 segments — 70% of the map**, not a
correction at the margin.

### Cost

Interleaved A/B against the merge base, minimum of three rounds each, `crates/rsvelte_bench`:

| benchmark | base | with #2954 |
| --- | ---: | ---: |
| `compile/client/runes-counter` | 221.7 µs | +44.9% |
| `compile/client-dev/typescript-generics` | 378.1 µs | +39.3% |
| `compile_both/data-table` | 1003 µs | +30.3% |
| `compile/server/legacy-reactive` | 397.8 µs | +18.1% |
| `compile/client/css-heavy` | 2033 µs | +8.9% |
| `printer/decoded-map/rsvelte-esrap` | 240.5 µs | +118.7% |

The printer row is **not** waste and must not be optimized away: rsvelte now emits 13,000
mappings for the 50 KB printer corpus against official esrap's 12,464, where the previous
implementation emitted 1,082. Same density, and still 6.4× faster than official esrap
(550 µs against 3,538 µs). Only the eleven passes are removable, and they are 10–18% of
compile after #2954's scan fusion (down from up to 30%; `append_generated_lines` alone went
135 µs → 1.9 µs by replacing a per-line `contains` with one SIMD scan).

### What replaces them

[...]

PR #3028, opened 2026-08-17
perf(client): emit the source map from spans, not from text matching

baseballyama:
Closes part of #3015 — the span-carrying architecture it asks for, plus the falsifiable step 3.

## What the issue asked for

> Stamp the source span on the IR node where the visitor builds it … have `to_oxc` pass that through instead of `SPAN`, and let esrap emit the map from spans … All eleven passes then delete.

Two of the eleven delete. The other nine are still load-bearing, and this PR measures exactly what each one still carries and why — which is what step 3 said the useful finding would be either way.

## The denominator, first

`sourcemap_gate_measure` scores **818** official segments across 29 samples, and **488 of them are client**; the other 330 belong to the server path, which has its own passes and is untouched. The issue's "239/818" is really **239/488** — quoting the whole 818 understates a client-only change by ~1.7×.

| configuration | `main` | this PR |
| --- | ---: | ---: |
| all enrichment passes | 815/818 | 815/818 |
| every client pass disabled | 567/818 (**client 239/488**, 49.0%) | 749/818 (**client 421/488**, 86.3%) |

The ratchet (`compatibility/sourcemap-known-failures.json`) is unchanged: 815/818 either way, the same 3 entries.

## Where the 182 segments came from

| change | client segments |
| --- | ---: |
| identity chunk projection for non-TypeScript scripts | +57 |
| `JsBlockStatement::brace_span` — the component function's braces | +76 |
| `Synth::reserve_anchor` — those braces under split coordinates | +8 |
| esrap `Printer::map_position` — real spans map under split coordinates | +10 |
| the identifier's span travels *into* the read transform | +6 |
| longest-run resync in the chunk projection | +25 |

Two of these are worth calling out because they are bugs rather than gaps:

**`has_loc` answers a comment question, and the printer was reading it as a mapping question.** Under split coordinates `loc_base` is set *above* every real source offset, so `has_loc` is false for all of them. That is right for "may this node carry comments" and exactly backwards for "is this a source position worth a segment" — every `JsExpr::Spanned` in a component whose script has a comment was silently dropped from the map. `map_position` is the same lookup without the gate; formatting decisions keep `offset_to_line_col`, because a comment-space line and a source-space line are not comparable.

**A span belongs on the identifier, not on the wrapper the lowering builds around it.** Upstream writes `b.call(b.id(name, loc))`; spanning the call makes the segment cover `foo()` where official covers `foo`. The read transform now receives the spanned identifier.

**A third thing this PR learned the expensive way: an in-band wrapper variant is only safe where every downstream matcher on that position has been enumerated.** A nineteenth revision of this branch also kept the span on a member expression's *object*, worth **+18 segments**, and it passed the entire source-map gate. It also broke **49 runtime fixtures** — `component-binding-*`, `binding-input-group-each-*`, `binding-this-each-*` — because the client lowering walks a member chain with `while let JsExpr::Member(m) = root` and then asks `if let JsExpr::Identifier(name) = root`. A `Spanned` answers neither, so `shared/component.rs`'s `member_root_info` came back `None` and a `bind:` setter fell through to `bar.baz = $$value` instead of `bar(bar().baz = $$value, true)` — output that parses, computes a value, and drops the parent notification. **This gate cannot see that class at all**: its unit is a segment, not the generated statement. `without_outer_source_span` therefore stays on the member object, and the case is recorded as gate-coverage 14g.

## Deleting the passes (step 3)

Segments lost when exactly one client pass is disabled, everything else on:

| pass | `main` | this PR |
| --- | ---: | ---: |
| `default_function_wrapper` | 84 | **0 — deleted** |
| `effect_callback` | 8 | **0 — deleted** |
| `token` | 80 | 23 |
| `template_element_runtime` | 25 | 21 |
| `legacy_prop_read` | 16 | 16 |
| `inline_script` | 7 | 4 |
| `bind_value` | 5 | 5 |
| `component_bind` | 5 | 4 |
| `verbatim_import` | 4 | 4 |
| `collapsed_declaration` | 0 | 0 |
| `rune` | 0 | 0 |

The before column is the attribution: the two deleted passes carry a *movement*, not just a zero.

**`collapsed_declaration` and `rune` cost 0 on both trees, and that is not evidence they are redundant.** They contributed nothing before this change either, so deleting them cannot be attributed to it, and 29 samples cannot tell "no lowering here fires this pass" apart from "a span supplies it now". Both stay; the distinction is recorded as gate-coverage 14f, together with the fact that there is no corpus-wide source-map gate to fall back on.

What the remaining nine still carry is a named lowering, not a mystery — and every row is a `Raw` fragment or a builder call that *had* the span and dropped it, so #3015's step 1 really is the prerequisite it claims to be:

| still carried by a pass | why the position is lost |
| --- | --- |
| `let x = $.prop($$props, …)` and its default | written by the script *text* rewriter, which records nothing; the projection has to re-derive it by alignment |
| hoisted verbatim `import` lines | `extract_imports` returns text with no offset, so they are emitted as `JsStatement::Raw` |
| element/component identifier *uses* (`pre.textContent`, `$.sibling(div, 2)`) | `flush_node` stamps the declaration; `SiblingPrev::Reuse` carries a bare `b::id` |
| component `bind:` accessor pairs (`get potato()`) | built from `JsExpr::Raw` text |
| the `(deps, $.untrack(…))` sequence tail | builder-made, with no source anchor |

## Two hazards this created and paid for

**The projection stopped being TypeScript-only, so its consumers stopped being cheap.** `RestoreRawMappedSpans::source_offset` was a linear scan of `copied_spans` per AST node, and `take_chunk_region` pushed one `loc_map` entry per mapped *byte* — both fine when only TypeScript scripts had a projection, both quadratic once every script does. They are now a binary search and one `LocRange { linear: true }` per copied run, and the projection is computed only when `enable_sourcemap` is set.

**A resync window that is too wide is worse than the nearest-byte rule it replaces.** Scoring resync candidates by longest common run over 256 bytes fixed `export let x = …` → `let x = $.prop(…)` and *broke* the TypeScript sample by 4 segments the passes had been covering: a slightly longer run a hundred bytes away beat the right one next door. The rule that holds is *nearest candidate starting at least a token's worth of agreement, within 32 bytes, and only when the nearest single byte buys less than that* — no regression with the passes, and better than the wide window without them (measured 767 vs 756 on the tree that still carried the member-object span; both shift down by the 18 that span held, and the ordering is what the rule is chosen on).

## Not done here

Step 1 (`Raw` / `RawMapped` eradication in the client path) is untouched; the table above is the priority order for it, sized by segments held. #3015 stays open for that and for the nine remaining passes.


baseballyama:
## Green-ing this branch: one parse failure, one silent mis-compile, and a representation regression

Four commits on top of `1180cf5ce`. Everything below is measured against `main` (`ade62185f`).

### 1. The `Compiler parity` failure — `loc_base` is a boundary, not a high-water mark

`esrap_corpus` rejected 1 of 13,284 files, `svelte.dev/.../llms-small.txt/content-sveltekit.md/10.svelte`:

```js
$.template_effect(() => $.set_text(text, // "data" prop contains property "result" $$props.data.result));
```

The `//` swallows the rest of the line, so the argument list never closes.

`block_body` called `note_span(end)` with the **instance script's end offset**. `loc_base` is
computed as `max_span + 2`, and it is the boundary between the comment buffer and source
coordinates — so a source offset raising it puts the boundary *inside* the source range. Here
`loc_base` landed at 134 while `{data.result}` starts at 133: the template's own offsets then read
as comment-space positions, `has_loc` said yes, and the printer flushed a `<script>` comment at
one of them. `has_newline_between` compared a comment-space end (177) with a source-space start
(139) and answered "same line", so no newline was emitted.

Dropping that one `note_span` restores the documented invariant and reproduces upstream's output
byte for byte. Regression test in `client/tests.rs`; it fails on the old tree.

### 2. A second defect the same corpus job would have caught, if it had got that far

`verify.mjs` is skipped when the parse gate fails, so nothing compared outputs. Compiling all
**14,393** `.svelte` files under `submodules/` on both trees and diffing the generated JS found
**8 divergences** — 7 in layerchart, 1 in svar-core, both corpus sources:

```js
$$props[children]?.(node_4, props)   // this branch
$$props.children?.(node_4, props)    // main, and upstream
```

`respan` hands the read transform the identifier *inside* its `Spanned` wrapper, and
`non_source_prop_read` picks the member property by variant — the unrecognised wrapper fell to
`JsMemberProperty::Expression`, which prints computed. With a `{#snippet children}` shadowing the
prop, `children` is the local snippet, so the emitted program renders the snippet instead of the
prop. Same class as the member-object case already in the PR description; the span now travels
onto the property as `SpannedIdentifier`.

After the fix, generated JS is byte-identical to `main` on **all 14,393 components across all three
targets** (client / server / client-dev) and on the 452 `.svelte.js|ts` modules. Source maps differ
on 4,875 client entries, which is the point of the PR. `sourcemap_gate_measure` still reports
**815/818**.

### 3. Performance — the regression is in the representation, and CodSpeed cannot see it

`benchmark_runner` hard-codes `enable_sourcemap: false` (two sites) while `CompileOptions::default()`
is `true`, so the benchmarks compile a configuration in which the projection never runs. CodSpeed's
"will not alter performance" is correct about the workload it ran.

Measured with `alloc_count` on flowbite-svelte (1296 files) — a deterministic counter, because wall
clock on this box moved **31% from run order alone** at load average 160:

| tree | allocations | requested bytes |
| --- | ---: | ---: |
| `main` | 2,120,446 | 578,635,400 |
| this branch as pushed | 2,124,328 | 592,917,695 (**+2.47%**) |
| now | 2,125,314 | 581,451,429 (**+0.49%**) |

77% of that came from one field. `brace_span: Option<(u32,u32)>` on `JsBlockStatement` grew
`JsStatement` 192 → 208 bytes and `JsExpr` 184 → 200 — an 8.3% / 8.7% tax on the whole IR — because
that struct is inside every statement and expression, for a range exactly **one** block per program
carries. There is no cheaper field either: `JsBlockStatement` is a bare `Vec` at 24 bytes with
8-byte alignment, so any field rounds it to 32. It now lives on `JsProgram`, keyed by the component
function's name.

The other two are the same shape — per-byte work that stopped being TypeScript-only.
`copied_spans_for_normalized_code` built a `Vec<Option<u32>>` the length of the script and walked
it byte by byte; with no erasure to project through that table is `Some(original_offset + i)` at
every `i`, so the split loop can never break a run and the whole pass reduces to one span per
matched run. And an identifier no transform rewrote was rebuilt as a *second* `Spanned` arena node
holding the same span over a clone of the same identifier.

What remains (+0.49%) is the `Spanned` node for identifiers a transform *does* rewrite — the
feature, not overhead around it.

Also parenthesised the `rune` kill switch, which bound to the first `contains` only.

### What I did not measure

A time figure. Wall clock is unusable on this machine right now (the min for one arm moved
1599 ms → 2092 ms between two runs that differed only in arm order), so every number above is an
allocation count, and an allocation-byte share is not a CPU share.

Written up in `docs/phase3-ast-refactor-plan.md` and `compatibility/gate-coverage.md`
(*the perf gates also compile with a different option set*).

baseballyama:
### The time figure, now that the box is quiet

My previous comment said wall clock was unusable here — at load average 160 one arm's minimum
moved 1599 ms → 2092 ms between two runs that differed only in arm order. Load has since dropped
to ~10 on a 10-core machine, so I re-ran it properly: three release binaries, rotated through all
6 orders, 18 samples each, compiling the same 3,637 `.svelte` files (bits-ui / flowbite-svelte /
melt-ui / shadcn-svelte) as CSR with `enable_sourcemap: true`.

| arm | min | mean |
| --- | ---: | ---: |
| `main` (ade62185f) | 947.6 ms | 957.7 ms |
| this branch **before** the perf commit | 928.2 ms | 941.9 ms |
| this branch now | 925.9 ms | 938.3 ms |

Two things follow, and the second one matters more than the first.

**There is no wall-clock regression against `main` on this workload, and there was not one before
my perf commit either.** The branch is ~2% *faster*, which is itself below the ~5% code-layout
floor this repo has measured for separate-binary A/B, so the honest reading is *parity*.

**So the +2.47% allocation-byte figure was real but did not convert into measurable time.** I found
it with a deterministic counter and fixed it because an 8.3% growth of `JsStatement` and 8.7% of
`JsExpr` is a representation regression that compounds with everything built on those types — not
because it showed up on a timer. The remaining 938.3 → 941.9 ms difference between the two branch
arms is 0.4% and I am not claiming it.

Both populations are library corpora, which gate-coverage records as the runes-heavy end; I did not
measure application source (huly / carbon / open-webui), where the instance-script text machinery
carries a much larger share.

baseballyama:
## The mutation gate cannot move, measured rather than argued

GitHub Actions has had this branch's `CI` and `Corpus Compat` runs queued with no runner for
hours, so the remaining risk was checked locally. The one gate whose *input* this PR could
plausibly disturb is Gate 3 (`mutate-corpus.mjs`) — it inserts a semantics-preserving comment at
a line boundary inside `<script>`, which is exactly the region whose coordinate space the
`loc_base` fix touches. The corpus gates' other units are covered by the byte-identity already
reported; a mutant is a *different input*, so it is not.

Reproducing the mutation differentially: every `.svelte` in the corpus (including
markdown ` ```svelte ` blocks) with `// mutant ) } ; ' " \`` inserted after the first line of its
first `<script>` — **12,141** mutants, of the 14,393 sources, the rest having no script — compiled
on `main` (`ade62185f`) and on this branch, and the generated JS hashed.

| target | mutants | code hashes differing |
| --- | ---: | ---: |
| client | 12,141 | **0** |
| client-dev | 12,141 | **0** |
| server | 12,141 | **0** |

(583 client / 582 server mutants fail to compile on **both** trees — mutants of sources that do
not compile unmutated either. Source maps differ, which is the PR.)

A delimiter-carrying comment is the shape that found #2351 and #2347, and it is deliberately the
comparison the gate itself does not make: `mutate-corpus.mjs` ratchets `code-mismatch` after
normalizing comments and whitespace away, so this raw-byte comparison against `main` is strictly
stronger than the verdict CI will compute.

`cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` are
both clean on `117500f82`, and `sourcemaps_gate` is 815/818 unchanged.

PR #3313, opened 2026-08-21
fix(compiler): run the host-independent bind: rules for every host, and read a duplicate this= like upstream

### LINKED ISSUE #3242: await in a bind: get/set pair is accepted; official rejects it
`await` inside a `bind:` get/set pair is rejected by official with
`experimental_async` and accepted by rsvelte.

```svelte
<script>
	let s = $state('v');
	const p = Promise.resolve(1);
</script>
<input bind:value={() => s, async (v) => { await p; }} />
```

| | verdict |
|---|---|
| official | `experimental_async` at 5:43 |
| rsvelte | compiles |

Both halves of the pair are affected — the getter too:

```svelte
<input bind:value={async () => await p, (v) => { s = v; }} />
```

official `experimental_async` at 5:31, rsvelte compiles.

The neighbouring positions are handled correctly, which localises it to the
`bind:` get/set slot rather than to the async check in general:

| shape | official | rsvelte |
|---|---|---|
| `{#snippet w()}<b>{await p}</b>{/snippet}` | `experimental_async` | same |
| `onclick={async () => { await p; }}` | compiles | same |
| `use:act={async () => await p}` | compiles | same |
| `bind:value={() => s, (v) => { s = v; }}` | compiles | same |

This is an **over-acceptance**, the direction a collected corpus can never
find, because published code compiles. The `bind-setter` matrix family
generates setter shapes but no async one.

Found by a 28-shape `await`-in-template grid (84 cells); it is the only real
divergence in it — the other row the grid flagged was my own fixture's name
clash, not a defect.

### LINKED ISSUE #3264: Over-acceptance: three `bind:` validations skip a host (`<svelte:element>` size bindings, component `bind:group`) — one emits unparseable output
Three `bind:` validations are implemented for some hosts and not others, so
rsvelte compiles programs official rejects. One of the three emits output no JS
parser accepts.

## 1. `bind_invalid_expression` is not checked on `<svelte:element>`

```svelte
<svelte:element this={tag} bind:clientWidth={o?.k}>x</svelte:element>
```

official: `bind_invalid_expression` at 1:27. rsvelte compiles, and the client
output is **not JavaScript**:

```js
$.bind_element_size($$element, 'clientWidth', ($$value) => o?.k = $$value);
```

An optional chain is not a valid assignment target. Reproduces for
`clientWidth`, `clientHeight`, `offsetWidth`, `offsetHeight`, `contentRect` and
`this`, on client, server and client-dev.

Controls: `<div bind:clientWidth={o?.k}>`, `<div bind:this={o?.k}>`,
`<input bind:value={o?.k}>` and `<C bind:x={o?.k} />` all raise
`bind_invalid_expression` on both sides. The check exists — it just does not run
for `SvelteElement`.

## 2. `bind_invalid_value` is not checked on `<svelte:element>`

```svelte
<svelte:element this={tag} bind:clientWidth>x</svelte:element>
```

official: `bind_invalid_value` at 1:32 (there is no `clientWidth` variable to
shorthand-bind to). rsvelte compiles and emits a write to an undeclared name:

```js
$.bind_element_size($$element, 'clientWidth', ($$value) => clientWidth = $$value);
```

Reproduces for the same five size bindings. Controls: `<div bind:clientWidth>`
and `<div bind:offsetWidth>` are rejected by both, and
`<svelte:element bind:this>` matches — so it is the element-size family on this
one host.

## 3. `bind_group_invalid_expression` is not checked on a component

```svelte
<C bind:group={() => v, (nv) => (v = nv)} />
```

official: `bind_group_invalid_expression` at 1:3 — `bind:group` takes no
get/set pair. rsvelte compiles it into a getter/setter prop pair. All three
targets.

Controls: `<input bind:group={() => v, (nv) => (v = nv)} />` is rejected by
both, and `<C bind:value={() => v, (nv) => (v = nv)} />` is accepted by both.

## Why the fixtures read as covered

All three codes have a passing fixture in the 145 `compiler-errors` set, at one
input each — and a code with a passing fixture reads as covered. This is the
`directive-element` drift the shape matrix was built for, one host axis over:
the rule is one `parent_type` test upstream and one arm per visitor here.

## Found by

Adversarial `bind:` grid: 37 binding properties × 9 target-expression shapes ×
15 hosts × 3 targets = 15,795 cells, 3,444 accepted by official, 41 divergences
over 15 keys — of which 12 cells are unparseable, all in family 1.

### LINKED ISSUE #3265: Server: `<select bind:value={getter, setter}>` renders the setter function instead of calling the getter
On the server, `<select bind:value={getter, setter}>` renders the **setter
function** as the select's value. Upstream calls the getter.

```svelte
<script>
	let v = $state('a');
</script>
<select bind:value={() => v, (nv) => (v = nv)}>
	<option value="a">a</option>
</select>
```

`generate: 'server'`:

```js
// official
$$renderer.select({ value: (() => v)() }, …)

// rsvelte
$$renderer.select({ value: (() => v, (nv) => v = nv) }, …)
```

rsvelte emits the whole get/set sequence expression as the value. A sequence
expression evaluates to its **last** operand, so `value` is the setter arrow —
the rendered `<option selected>` decision is made against a function instead of
against `v`, and nothing is ever selected. Both outputs parse.

Reproduces for `<select multiple bind:value={…}>` too. Controls: the same
get/set pair on `<input>`, `<textarea>` and a component all match, and the
client target matches for `<select>` — so this is the server `select` builder
alone, which is the one place that has to *read* the bound value while
rendering rather than only wire it up.

## Found by

Adversarial `bind:` grid (37 properties × 9 expression shapes × 15 hosts × 3
targets); `value / getset / select` and `value / getset-nl / select` diverged on
server only.

### LINKED ISSUE #3266: Over-rejection: `<C bind:this={x} bind:this={x} />` is refused with `attribute_duplicate`; official compiles it
**Over-rejection**: rsvelte refuses a component that official compiles.

```svelte
<script>
	import C from './C.svelte';
	let el = $state(null);
</script>
<C bind:this={el} bind:this={el} />
```

official: compiles (client and server). rsvelte: `attribute_duplicate` at 5:18.

Upstream exempts `bind:this` on a component from the duplicate-attribute check;
rsvelte applies it.

## Controls

Every neighbouring cell matches, which is what narrows this to one arm:

| source | both |
|---|---|
| `<div bind:this={el} bind:this={el}>` | accepted |
| `<svelte:element this={tag} bind:this={el} bind:this={el}>` | accepted |
| `<svelte:component this={C} bind:this={el} bind:this={el} />` | accepted |
| `<C bind:value={v} bind:value={v} />` | accepted |
| `<C title="a" title="b" />` | accepted |

So it is not the duplicate check in general, not `bind:this` in general, and not
components in general — it is the product of the three.

## Found by

Adversarial construct × host validation grid (27 constructs × 15 hosts × 2
targets = 810 cells): the only over-rejection in the sweep.

### LINKED ISSUE #3267: A second `this=` on `<svelte:element>` / `<svelte:component>` is dropped instead of passed through as an attribute/prop
A second `this=` on `<svelte:element>` or `<svelte:component>` is silently
dropped. Upstream consumes the first as the tag / component and passes the
second through as an ordinary attribute or prop.

```svelte
<script>
	import C from './C.svelte';
	let tag = $state('div');
	let tag2 = $state('span');
</script>
<svelte:element this={tag} this={tag2}>x</svelte:element>
```

`generate: 'client'`:

```js
// official
$.element(node, () => tag, false, ($$element, $$anchor) => {
	$.attribute_effect($$element, () => ({ this: tag2 }));
	…
});

// rsvelte — the second `this` is gone
$.element(node, () => tag, false, ($$element, $$anchor) => {
	…
});
```

Same for a component:

```svelte
<svelte:component this={C} this={C} />
```

```js
// official
$$component($$anchor, { get this() { return C; } });
// rsvelte
$$component($$anchor, {});
```

Both targets, both hosts. Neither compiler errors, so the attribute simply
disappears from rsvelte's output — a rendered `this="span"` attribute (or a
`this` prop the child reads) that is present with the official compiler and
absent here.

## Found by

Adversarial construct × host validation grid (27 constructs × 15 hosts × 2
targets = 810 cells).

baseballyama:
Five `bind:` / `this=` defects, all of the same shape: a rule that upstream
applies once, ported here as one copy per host visitor, with some copies short.

Closes #3242
Closes #3264
Closes #3265
Closes #3266
Closes #3267

## What was wrong

Upstream's `BindDirective` visitor gates **only** the `binding_properties` block
on the parent's type. Everything below it — the `{get, set}` pair checks,
`object()` → `bind_invalid_expression`, the `bind_invalid_value` identifier
check, and the `await` gate — runs for every host. rsvelte had a hand-written
copy per visitor and three of them were short:

| host | what it did not reach |
|---|---|
| `<svelte:element>` | every target-shape rule below the pair branch |
| component | the `{get, set}` pair rules |
| `<svelte:component>` | the pair rules *and* the target-shape rules |
| every host | the `await` gate — it had no port at all |

Consequences, all confirmed against the official compiler:

- `<svelte:element this={tag} bind:clientWidth={o?.k}>` compiled into
  `($$value) => o?.k = $$value` — **not JavaScript**. Same for `offsetWidth`,
  `clientHeight`, `offsetHeight`, `contentRect` and `this`.
- `<svelte:element this={tag} bind:clientWidth>` emitted a write to an
  undeclared name.
- `<C bind:group={() => v, (nv) => (v = nv)} />` was lowered into a
  getter/setter prop pair; so were a parenthesised pair and a three-element
  pair on a component / `<svelte:component>`.
- `await` anywhere a bind expression can reach it compiled.

Two more, adjacent:

- On the server, `<select bind:value={get, set}>` rendered the whole sequence
  expression as `value`. A sequence evaluates to its **last** operand, so the
  rendered value was the setter function and no `<option>` was ever selected.
  Upstream emits `b.call(expression.expressions[0])`.
- Upstream applies the duplicate-attribute rule **once**, while reading the
  attributes, and exempts every attribute named `this`; the tag definition it
  then consumes is `attributes.splice(index, 1)` — the *first* `this` only.
  rsvelte had a second copy of the duplicate check in the phase-2 component
  visitor with no exemption (so `<C bind:this={x} bind:this={x} />` was
  **rejected** where official compiles it), and its tag-definition extraction
  filtered out *every* `this` attribute rather than the first (so a second
  `this=` disappeared from the output instead of being passed through).

## How it is fixed

- `is_get_set_pair` / `validate_get_set_pair` / `walk_get_set_pair` /
  `walk_bind_expression` are one shared implementation in
  `2_analyze/visitors/bind_directive.rs`; `<svelte:element>`, a component,
  `<svelte:component>` and `<svelte:self>` now call it instead of carrying a
  partial copy. `validate_bind_value_for_component` is renamed
  `validate_bind_value_target` — it was never component-specific.
- The `await` gate is `VisitorContext::bind_await_depth`: the `function_depth`
  at which an `await` below a bind expression suspends. A function-like half of
  a `{get, set}` pair records `depth + 1`, which is exactly upstream's "visit
  `get.body` with `state.expression` installed" — it jumps across the function
  that would otherwise reset it, while a function *below* that still resets.
- The phase-2 duplicate-attribute check is deleted rather than patched: it was
  a second port of a rule the parser already applies to every element, and the
  parser's port already had the `this` exemption.
- The tag-definition extraction removes the first `this` by index.
- The `bind:group` collector now reads `<svelte:self>`'s own attributes; it
  visited every other host's and only recursed past this one, so
  `<svelte:self bind:group={x} />` never declared its group array.

## Evidence

A host × bind-expression-shape grid (12 hosts × 28 shapes × 2 targets = 672
cells) run against the official compiler:

| | before | after |
|---|---|---|
| match | 645 / 672 | **672 / 672** |

The 27 pre-fix divergences were, in order: 12 `<svelte:component>` cells that
reached none of the shape rules (a host the issues did not name — found by
crossing the axis), 12 `await` cells, 2 `<svelte:element>` cells, and the
`<svelte:self bind:group>` output.

Both directions are in the regression tests, on the same host axis: the illegal
shapes must be rejected **and** ~24 legal shapes per host must still compile on
both targets. `bind_host_axis_3264.rs`'s legal-shape list is what caught the
one over-rejection I introduced while writing it.

Every markup asserted by the new tests was first run through the official
compiler (152 cells, 152/152 agreement) before the assertion was written, so no
assertion encodes rsvelte's behaviour instead of upstream's.

## Ratchets

None touched. No `compatibility/*known-failures*` file is modified by this PR.

## Test plan

- `cargo test -p rsvelte_core` — green (whole suite)
- `cargo test -p rsvelte_projection` — green
- `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — clean
- new: `bind_await_3242.rs`, `bind_host_axis_3264.rs`,
  `ssr_select_bind_getset_3265.rs`, `this_attribute_3266_3267.rs`

### DIFF
--- a/.changeset/bind-host-axis-and-this-attribute.md
+++ b/.changeset/bind-host-axis-and-this-attribute.md
@@ -1,0 +1,12 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Run the host-independent half of the `bind:` rules for every host, and read `this=` the way the official compiler does.
+
+- `await` anywhere a `bind:` expression can reach it — including both halves of a `{get, set}` pair — is now rejected with `experimental_async`, as the official compiler does; an `await` inside a function *below* the pair still compiles.
+- `<svelte:element>` reached none of the target-shape rules, so `bind:clientWidth={o?.k}` compiled into `($$value) => o?.k = $$value`, which no JS parser accepts, and a shorthand `bind:clientWidth` emitted a write to an undeclared name. A component and `<svelte:component>` never reached the `{get, set}` pair rules, so `bind:group={get, set}`, a parenthesised pair and a three-element pair were all accepted.
+- On the server, `<select bind:value={get, set}>` rendered the sequence expression — whose value is the *setter* — instead of calling the getter, so no `<option>` was ever selected.
+- `<C bind:this={x} bind:this={x} />` was rejected with `attribute_duplicate`; the official compiler exempts every attribute named `this` from that rule.
+- A second `this=` on `<svelte:element>` / `<svelte:component>` was dropped instead of being passed through as an attribute / prop.
+- `<svelte:self bind:group={x} />` did not declare its binding group array.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/state/element.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/state/element.rs
@@ -685,17 +685,9 @@ impl<'a> Parser<'a> {
                 // Extract the "this" attribute to get the expression
                 let expression = self.extract_this_attribute(&attributes);
 
-                // Filter out the "this" attribute from the list
-                let filtered_attrs: Vec<_> = attributes
-                    .into_iter()
-                    .filter(|attr| {
-                        if let crate::ast::Attribute::Attribute(node) = attr {
-                            node.name.as_str() != "this"
-                        } else {
-                            true
-                        }
-                    })
-                    .collect();
+                // Upstream splices out only the *first* `this` (element.js L266-280);
+                // a second one stays in the list and is passed through as a prop.
+                let filtered_attrs = remove_first_this_attribute(attributes);
 
                 TemplateNode::SvelteComponent(Box::new(SvelteComponentElement {
                     start: start as u32,
@@ -739,43 +731,31 @@ impl<'a> Parser<'a> {
                 // Check if the "this" attribute is a string value (not an expression)
                 // and emit svelte_element_invalid_this warning if so.
                 // Corresponds to element.js L288-289: if (!is_expression_attribute(definition)) { w.svelte_element_invalid_this(definition); }
-                for attr in &attributes {
-                    if let crate::ast::Attribute::Attribute(node) = attr
-                        && node.name.as_str() == "this"
-                    {
-                        let is_expression_attribute = match &node.value {
-                            AttributeValue::Expression(_) => true,
-                            AttributeValue::Sequence(parts) => {
-                                parts.len() == 1
-                                    && matches!(&parts[0], AttributeValuePart::ExpressionTag(_))
-                            }
-                            _ => false,
-                        };
-                        if !is_expression_attribute {
-                            self.parse_warnings.push(crate::ast::template::ParseWarning {
-                                code: "svelte_element_invalid_this".to_string(),
-                                message: "`this` should be an `{expression}`. Using a string attribute value will cause an error in future versions of Svelte\nhttps://svelte.dev/e/svelte_element_invalid_this".to_string(),
-                                start: node.start,
-                                end: node.end,
-                            });
+                if let Some(node) = definition {
+                    let is_expression_attribute = match &node.value {
+                        AttributeValue::Expression(_) => true,
+                        AttributeValue::Sequence(parts) => {
+                            parts.len() == 1
+                                && matches!(&parts[0], AttributeValuePart::ExpressionTag(_))
                         }
+                        _ => false,
+                    };
+                    if !is_expression_attribute {
+                        self.parse_warnings.push(crate::ast::template::ParseWarning {
+                            code: "svelte_element_invalid_this".to_string(),
+                            message: "`this` should be an `{expression}`. Using a string attribute value will cause an error in future versions of Svelte\nhttps://svelte.dev/e/svelte_element_invalid_this".to_string(),
+                            start: node.start,
+                            end: node.end,
+                        });
                     }
                 }
 
                 // Extract the "this" attribute to get the tag expression
                 let tag = self.extract_this_attribute(&attributes);
 
-                // Filter out the "this" attribute from the list
-                let filtered_attrs: Vec<_> = attributes
-                    .into_iter()
-                    .filter(|attr| {
-                        if let crate::ast::Attribute::Attribute(node) = attr {
-                            node.name.as_str() != "this"
-                        } else {
-                            true
-                        }
-                    })
-                    .collect();
+                // Upstream splices out only the *first* `this` (element.js L282-296);
+                // a second one stays in the list and is rendered as an attribute.
+                let filtered_attrs = remove_first_this_attribute(attributes);
 
                 TemplateNode::SvelteElement(Box::new(SvelteDynamicElement {
                     start: start as u32,
@@ -2856,4 +2836,19 @@ fn shorthand_first_invalid_offset(name: &str) -> Option<usize> {
             .find(|(_, c)| !is_identifier_continue(*c))
             .map(|(i, _)| i),
     }
+}
+
+/// Upstream consumes a `<svelte:element>` / `<svelte:component>` tag definition
+/// with `attributes.splice(index, 1)` on the *first* `this` attribute, so a
+/// second one survives as an ordinary attribute/prop.
+fn remove_first_this_attribute<'a>(
+    attributes: Vec<crate::ast::Attribute<'a>>,
+) -> Vec<crate::ast::Attribute<'a>> {
+    let mut attributes = attributes;
+    if let Some(index) = attributes.iter().position(|attr| {
+        matches!(attr, crate::ast::Attribute::Attribute(node) if node.name.as_str() == "this")
+    }) {
+        attributes.remove(index);
+    }
+    attributes
 }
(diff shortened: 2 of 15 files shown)


baseballyama:
Cross-check from a parallel salvage effort, in case it is useful here.

A killed agent's branch `fix/3267-second-this-attribute` (commit `5a94666e4`, now on `origin`) implements the same fix as this PR's `remove_first_this_attribute`. The two were diffed and are **semantically identical — the difference is decoration only** (free function vs associated function). This PR already declares `Closes #3267`, so **that branch is not being turned into a competing PR.**

One caveat on that equivalence claim, stated because it bounds it: it comes from a **source diff**, not from running this PR's tests against that branch. If you want it stronger, that is the measurement to take.

What the rescued branch carries that this PR appears not to, and which is worth lifting rather than dropping:

- 3 `compatibility/pattern-corpus` repro files
- `crates/rsvelte_core/tests/second_this_attribute_3267.rs`
- the matching `compatibility/pattern-corpus/README.md` line
- a changeset

The repros matter beyond bookkeeping here: the four corpora that produce this class of defect are not corpus sources, so the collected gate baselines at 0 while the instances live outside the population it inspects — which is why the convention in this repo is that each such fix lands a `pattern-corpus` repro.

Nothing needs to be done with the branch itself; it stays on `origin` as a rescue artifact.

Unrelated context for anyone reading this today: GitHub Actions is not dispatching account-wide (653 queued runs, 0 executing jobs; a control repository shows the same from ~09:59Z), so a green/absent check state on this PR currently means "never ran", not "passed".

Issue #3057, opened 2026-08-18
parser: comment between an {#each} item pattern and } is accepted — official rejects with expected_token


baseballyama:
### Summary

The official parser rejects a comment between an `{#each}` item pattern and the closing `}` with
`expected_token — Expected token }`; rsvelte accepts every form and compiles:

```svelte
{#each [1, 2] as x /* c */}<i>{x}</i>{/each}
{#each [1, 2] as [x] /* c */}<i>{x}</i>{/each}
{#each [1, 2] as [,] /* c */}{/each}
```

official: `expected_token` on all three. rsvelte: compiles all three. "rsvelte accepts what
official rejects" is the class the matrix `bind:` / `param-default` families exist for — no
collected corpus can hold these programs because published code compiles.

### Where

The each-context pattern reader needs to stop consuming a trailing comment before `}` (or the
`}` scan must not skip comments there), matching upstream's `read_pattern` + explicit `}`
expectation.

Issue #3284, opened 2026-08-21
A duplicate `{:else}` is rejected with three different codes depending on the host; official compiles it


baseballyama:
Official compiles a block with two `{:else}` branches. rsvelte rejects it — and
rejects it with **three different error codes depending on where the block
sits**, which is what makes this a parser falling over rather than a deliberate
check.

```svelte
{#each arr as v}{v}{:else}a{:else}b{/each}
```

| host | rsvelte | official |
|---|---|---|
| root | `block_unclosed` | compiles |
| inside `<div>` | `block_invalid_continuation_placement` | compiles |
| inside `{#if}` | `expected_token` | compiles |
| inside `{#each}` | `block_unclosed` | compiles |
| inside `{#snippet}` | `block_unclosed` | compiles |

`{#if}` reproduces it too, so it is not `{#each}`-specific:

```svelte
{#if x}a{:else}b{:else}c{/if}
```
`rsvelte: block_unclosed` · `official: compiles`

Whether accepting a second `{:else}` is right is arguable — but byte parity is
the goal, and the host-dependent code is a defect on its own terms regardless of
which way the acceptance question is settled: one input should not produce three
diagnostics.

10 cells (5 hosts × 2 targets) for `{#each}`, found while crossing 18 extra
`{#each}` header shapes with 5 hosts (612 cells, 382 compiled on both sides).

PR #3321, opened 2026-08-21
fix(parser): validate {#each} and {#await} block-header patterns

### LINKED ISSUE #3245: {#each} item and index patterns are unvalidated: unparseable output, or a dropped index
`{#each}`'s item pattern and index are not validated. A non-pattern item is
spliced verbatim into the generated arrow's parameter list, producing text no
JS parser accepts; a non-identifier index is silently dropped.

### Item — unparseable output

```svelte
<script>
	let arr = $state([1]);
</script>
{#each arr as 1}<b>x</b>{/each}
```

official: `expected_pattern`. rsvelte compiles to

```js
$.each(node, 17, () => arr, $.index, ($$anchor, 1) => {
```

`($$anchor, 1) =>` is not a parameter list. Same for a string literal and for
a member expression:

| source | rsvelte's parameter list | official |
|---|---|---|
| `{#each arr as 1}` | `($$anchor, 1)` | `expected_pattern` |
| `{#each arr as "a"}` | `($$anchor, "a")` | `expected_pattern` |
| `{#each arr as o.k}` | `($$anchor, o.k)` | `expected_token` |

### Index — silently dropped

```svelte
{#each arr as v, 1}<b>{v}</b>{/each}
```

official: `expected_identifier`. rsvelte compiles to
`($$anchor, v) => {` — the index is simply gone. The output **parses**, so
only output equality reports it. Same for `, o.k` and `, "i"`.

A call expression as the item (`{#each arr as f()}`) is rejected by both with
`js_parse_error`, so this is a specific missing pattern check rather than the
each-head being unvalidated.

Found by writing a **second** input for Svelte error codes that already have a
passing `compiler-errors` fixture — the gap AGENTS.md names: 145 fixtures at
one input per code, and a code with a passing fixture reads as covered.
46 shapes, 90 cells, and these plus #3246 are the only divergences.

### LINKED ISSUE #3283: Nine `{#each}` / `{#await}` header validations are missing: illegal index forms, member targets, rest and default patterns, duplicate `{:then}`/`{:catch}`
Nine block-header shapes the official compiler rejects compile here. Every one is
uniform across all five hosts (root, inside an element, `{#if}`, `{#each}`,
`{#snippet}`) and both targets, so these are absent checks rather than host-arm
gaps.

## `{#each}`

|---|---|---|
| `{#each arr as , i}{i}{/each}` | `expected_pattern` | **compiles** |
| `{#each arr as ...r}{r}{/each}` | `expected_pattern` | **compiles** |
| `{#each arr as v = 1}{v}{/each}` | `expected_token` | **compiles** |
| `{#each arr as o.b}x{/each}` | `expected_token` | **compiles** |
| `{#each arr as v, i = 0}{i}{/each}` | `expected_token` | **compiles** |
| `{#each arr as v, [i]}{v}{/each}` | `expected_identifier` | **compiles** |
| `{#each arr as v,}{v}{/each}` | `expected_identifier` | **compiles** |

The index must be a plain identifier and the item must be a binding pattern;
rsvelte enforces neither. `{#each arr as o.b}` is the sharpest of these — it
compiles into an assignment to a member expression as if it were a binding.

## `{#await}`

|---|---|---|
| `{#await p}w{:then v}{v}{:then u}{u}{/await}` | `block_duplicate_clause` | **compiles** |
| `{#await p}w{:catch e}{e}{:catch f}{f}{/await}` | `block_duplicate_clause` | **compiles** |
| `{#await }w{/await}` | `js_parse_error` | **compiles** |
| `{#await p}w{:then ...r}{r}{/await}` | `expected_pattern` | **compiles** |
| `{#await p}w{:then v = 1}{v}{/await}` | `expected_token` | **compiles** |

`{:then}` / `{:catch}` take a binding pattern, not a parameter list, so a rest
element and a default are both illegal — and a second `{:then}` is a duplicate
clause. None of the four is checked.

## Denominators

| grid | cells | compiled both | errored both | divergent |
|---|---|---|---|---|
| `{#await}` header shapes × 5 hosts | 310 | 220 | 40 | 60 |
| `{#key}` header shapes × 5 hosts | 150 | 110 | 40 | 20 |
| `{#each}` pattern × key-expression shape, + 18 extra headers × 5 hosts | 612 | 382 | 100 | 195 |

### Measured clean in the same grids

`{#key}` has **no** acceptance divergence over 150 cells — `{#key x}`,
`{#key o.b}`, `{#key f(x)}`, a template literal, a parenthesised sequence, a bare
comma, an assignment, a literal, an object literal, an arrow, `{#key await x}`
and nested `{#key}` all agree, and its 20 divergent cells are message/position
only (filed separately).

On the `{#each}` side the 18 pattern shapes (object, renamed, defaulted, rest,
nested, array, holes, double holes, trailing comma, nested array) crossed with 12
key-expression shapes agree on acceptance everywhere except the seven rows above.

### LINKED ISSUE #3286: Block-header error fields: six message / code / position divergences in `{#each}`, `{#await}` and `{#key}`
Block headers that both compilers reject, where the reported fields differ. Split
out from #3283 (which is about headers rsvelte does not reject at all) because
these need field parity rather than a new check.

## Message only

| source | rsvelte | official |
|---|---|---|
| `{#key }v{/key}` | `Empty parenthesized expression` | `Unexpected token` |
| `{#each arr as v, i ()}{i}{/each}` | `Empty parenthesized expression` | `Unexpected token` |

10 and 28 cells respectively — the `{#each}` one across all 14 pattern shapes, so
it is the empty key parens rather than any pattern.

## Message and position

`{#key ...x}v{/key}`

| | message | start | end |
|---|---|---|---|
| rsvelte | ``Expected `=>` but found `EOF` `` | 6:10 | 6:10 |
| official | `Unexpected token` | 6:6 | 6:6 |

rsvelte scans past the spread looking for an arrow. 10 cells, and the 4-column
offset is constant across all five hosts.

## Code and message

`{#each arr as}x{/each}`

| | code | message |
|---|---|---|
| rsvelte | `js_parse_error` | `Unexpected token` |
| official | `expected_whitespace` | `Expected whitespace` |

10 cells.

## Code, message and position

`{#each arr as , i (i = 1)}{i}{/each}`

| | code | message | start | end |
|---|---|---|---|---|
| rsvelte | `constant_assignment` | `Cannot assign to constant` | 6:19 | 6:24 |
| official | `expected_pattern` | `Expected identifier or destructure pattern` | 6:14 | 6:14 |

16 cells. rsvelte gets past the malformed pattern and then complains about the
key expression; official stops at the pattern. Conditional on #3283 — once the
missing `expected_pattern` check exists, these should agree by construction.

## `end` only

`{#each  as v}{v}{/each}` — rsvelte `end` 6:12, official 6:11. 10 cells.

## Position only

`{#await p}w{:else}z{/await}` — rsvelte reports one column earlier than official
at every host (6:11 vs 6:12 at root, 6:16 vs 6:17 in an element, 6:18 vs 6:19 in
`{#if}`, 6:27 vs 6:28 in `{#each}`, 6:25 vs 6:26 in a snippet). 10 cells, constant
one-column offset, so it is one rule.

## Denominators

| grid | cells | compiled both | errored both | error-field divergent |
|---|---|---|---|---|
| `{#await}` | 310 | 220 | 40 | 10 |
| `{#key}` | 150 | 110 | 40 | 20 |
| `{#each}` | 612 | 382 | 100 | 64 |

baseballyama:
Block headers were the one place the parser handed a scanned region straight to
the pattern parser instead of reading a pattern. Upstream reads **one** binding
pattern for the `{#each}` item and the `{#await}` value
(`1-parse/read/context.js`), and **one** identifier for the `{#each}` index
(`Parser#read_identifier`); anything after that has to be the header's next
token. rsvelte scanned to the next `,` / `(` / `}` and accepted whatever fell
inside.

Two symptoms, and only one of them is loud:

- `{#each arr as 1}` compiled to `$.each(node, 17, () => arr, $.index, ($$anchor, 1) => {` — **a parameter list no JS parser accepts**, from a call that returned successfully.
- `{#each arr as v, 1}` compiled to `($$anchor, v) => {` — **the index is simply gone**, and the output parses, so only output equality could ever report it.

## What changed

`state/tag.rs` now mirrors `read_pattern` and `read_identifier` structurally:
the pattern is an identifier or a `{` / `[` destructuring, everything past it
must be whitespace or a `:` continuing it as a TypeScript type annotation, and
the index is an identifier or nothing. That **subsumes** the `#3057` comment
special case — a comment before the pattern is not an identifier and not a
bracket, so `each_segment_comment` and its `code_bytes` scan are deleted rather
than kept alongside.

Also from the three issues:

- a second `{:then}` / `{:catch}` is `block_duplicate_clause` (the opening tag's own `then` / `catch` counts, as upstream's `block.then` does);
- `{#await }` with no expression is `js_parse_error`;
- `{#each xs as}` reaches `require_whitespace` and is `expected_whitespace` instead of a `js_parse_error` about `arr as`;
- a misplaced block continuation is located on the `:`, which is upstream's `start = parser.index - 1` in `next()` — the fixture `compiler-errors/samples/catch-before-closing` pins that at `[35, 35]`, which is exactly the `:`;
- `{#key }` and `{#key ...x}` report acorn's `Unexpected token` at acorn's position. Both messages were **the `(…)` wrapper `check_js_parse_error_with_pos` adds describing itself** — `Empty parenthesized expression` for `()`, `Expected `=>` but found `EOF`` for `(...x)` read as arrow params, four columns to the right of where acorn stops. Acorn parses in place, so those two shapes are answered before wrapping;
- `expected_token` is a point span. Upstream's `e.expected_token(index, token)` takes a bare index; rsvelte reported `end = start + 1` at **every** raising site.

`{#each arr as , i (i = 1)}` resolves by construction, as #3286 predicted: the
missing `expected_pattern` now fires before the key expression is ever reached.

## Over-rejection control

The new test carries 28 legal neighbours alongside the 40 rejected ones —
object / array / nested / defaulted / rest / hole patterns, `as const as item`,
newline-split headers, comments *inside* a pattern, `{#await p then }` (whose
`\s*}` escape exists in the opening tag and, deliberately, **not** in
`{:then }`), and `{:catch}` before `{:then}`. Every code, message and span in
the file was read off the official compiler at the pinned version rather than
from the issue text.

`compiler-errors` (145), `parser_fixtures`, `compiler_fixtures`, `validator`,
`ssr`, `css`, `print`, all of `rsvelte_projection` (svelte2tsx, the loose-parse
consumer), `rsvelte_lint`, `rsvelte_formatter` and `rsvelte_fmt` are green
locally.

## Ratchets

Per the ordering constraint on `compatibility/error-*-known-failures.*.json`:
**no baseline-update flag was used here.** The `expected_token` point span and
the continuation `:` position are both parity *fixes*, so they may retire
entries a two-sided ratchet then reports as passing. If CI names any, they will
be removed by hand — only the ones CI reproduces as PASS — after
`origin/main` carries #3176's re-measured baseline.

Closes #3245
Closes #3283
Closes #3286


baseballyama:
I was assigned #3283 / #3286 / #3349 and wrote a competing implementation before
reading this PR. **Yours is strictly better and mine is discarded** — this is the
measurement that decided it, and it also says this PR closes more issues than its
body claims.

## Measurement

I rebuilt the population rather than trusting the issue text: **21 invalid + 31
valid** block-header shapes × **5 hosts** (root, element, `{#if}`, `{#each}`,
`{#snippet}`) = **260 cells**, each compiled by the official compiler and by
rsvelte, classified as `MISSING-CHECK` (official rejects, rsvelte accepts) /
`OVER-REJECT` (the reverse) / `code|message|start|end-mismatch` /
`error-parity` / `both-accept`. The valid half is there because a population of
only-invalid inputs is blind to over-rejection.

| tree | MISSING-CHECK | OVER-REJECT | code | message | start | end | error-parity | both-accept |
|---|---|---|---|---|---|---|---|---|
| `origin/main` | 70 | 10 | 10 | 15 | 5 | 5 | 0 | 145 |
| **this PR** | **0** | 10 | **0** | **0** | **0** | **0** | **105** | 145 |
| #3348 | 70 | **0** | 10 | 15 | 5 | 5 | 0 | 155 |
| my own impl (discarded) | 10 | **0** | 5 | 15 | 0 | 5 | 80 | 155 |

Fixed / regressed stated separately: this PR is **105 fixed / 0 regressed**
against `origin/main`. Mine was 80 fixed / 0 regressed and *left* 25 cells —
it never got the acorn-wrapper messages, `expected_whitespace`, or the
`expected_token` point span.

## This closes #3286 as well

All six of #3286's families reach `error-parity` here, in every host:

| #3286 family | source | verdict on this PR |
|---|---|---|
| message only | `{#key }v{/key}` | parity |
| message only | `{#each arr as v, i ()}{i}{/each}` | parity |
| message + position | `{#key ...x}v{/key}` | parity |
| code + message | `{#each arr as}x{/each}` | parity |
| code + message + position | `{#each arr as , i (i = 1)}{i}{/each}` | parity |
| `end` only | `{#each  as v}{v}{/each}` | parity |
| position only | `{#await p}w{:else}z{/await}` | parity |

No rows left, so **`Closes #3286` belongs in the body** — could you add it? (I
can't; it's your branch.) errH was independently about to work #3286 and stood
down on this measurement.

#3283's twelve families are likewise all at parity, and #3349's two duplicate
clauses plus two ordering variants I added
(`{:then}` after a `{:catch}`, `{:catch}` after a `{:then}`) are too.

## The one cell left, and where it lives

`OVER-REJECT: 10` is `{#if c}a{:else}b{:else}d{/if}` and
`{#each arr as v}{v}{:else}b{:else}d{/each}` — rsvelte raises `block_unclosed`
where **official accepts**. That is #3284, and this PR only changes those two
sites' error *position*, so it stays. #3348 closes it. **This PR + #3348 =
260/260.** The orchestrator has recorded the order as
`3478 → 3504 → 3321 → 3348 → 3331 → …`.

## One thing worth a follow-up (not a change request)

The duplicate-clause *rejection* for `{#await}` is inline in your await loop, and
the duplicate-clause *acceptance* for `{:else}` is inline in #3348's if/each loop
— two copies of one decision ("does a repeated clause error, per block type"),
which is exactly how #3284 and #3349 came to be reported as separate issues
pointing opposite ways. I'm landing a small `Clause::duplicate_is_error()` table
on top of #3348 so both sides read one exhaustive `match`; the `{#await}` arms are
already in it and the call sites here can read them once this merges. Nothing for
you to do — flagging it so the follow-up isn't a surprise.

Issue #3198, opened 2026-08-21
`import … assert { … }` on a new line: rsvelte over-rejects in TS and diverges on message/position in JS


baseballyama:
`assert` on the line after the module specifier is treated differently by the two
parsers. OXC's `parse_import_attributes` only accepts the keyword when
`!self.cur_token().is_on_new_line()` (`oxc_parser/src/js/module.rs:369`), so it
applies ASI and errors on the next token; acorn has no such guard and reads the
clause, then rejects the deprecated `assert` spelling at the keyword.

```svelte
<script>
	import d from "./d.json"
		assert { type: "json" };
	let z = d;
</script>
```

| wrapper | official | rsvelte |
|---|---|---|
| instance | `js_parse_error@44` "Unexpected token" | `js_parse_error@43` "Expected a semicolon or an implicit semicolon…" |
| `<script module>` | `js_parse_error@51` "Unexpected token" | `js_parse_error@50` same |
| `lang="ts"` | **OK** | `js_parse_error@53` same |

The TS row is the one that matters: acorn-typescript keeps the deprecated clause,
so official accepts the file and rsvelte **over-rejects** it.

The same-line spelling agrees on both message and position after
`strict_mode::check_with_clause`; this row is untouched by that scan, because OXC
never produces a `WithClause` for it and its own error fires first.

baseballyama:
Re-measured against `svelte@5.56.9` on the current `main`. **The two JavaScript
rows in the table no longer diverge** — official and rsvelte both raise
`js_parse_error` "Unexpected token" at the same byte, and the byte is the `{`
after `assert`, not the keyword:

| wrapper | official | rsvelte (today) |
|---|---|---|
| instance | `js_parse_error` "Unexpected token" @43 | identical |
| `<script module>` | `js_parse_error` "Unexpected token" @50 | identical |
| `lang="ts"` | **OK** | `js_parse_error` @53 |
| `module lang="ts"` | **OK** | `js_parse_error` @60 |

`realign_missing_semicolon` already moves OXC's insertion-point report ("Expected
a semicolon or an implicit semicolon…", raised at the end of the `assert`
identifier) to the token acorn stops on, which lands exactly where acorn does:
`assert` becomes an expression statement under ASI and the `{` on the same line
cannot continue it. #3351 pins those rows, plus the same-line spelling and the
`with` clause, which neither parser restricts on a new line.

**Only the TypeScript row is still open**, and it is worth recording why it is
not a small fix.

[...]

PR #3351, opened 2026-08-21
fix(parse): template-expression language mode, failure classification, and trailing line comments

### LINKED ISSUE #3261: Template expressions are parsed as TypeScript with no `lang="ts"` anywhere: 330 over-acceptances across 11 constructs and 15 slots
Template expressions are parsed as TypeScript whatever the component declares.
Official turns TypeScript on for the whole component only when some script has
`lang="ts"`; rsvelte accepts TS-only syntax in a template even when no script
does.

```svelte
{y as string}
```

official `expected_token` · rsvelte **compiles**

### Grid

TS-only construct × template slot × which script (if any) declares `lang="ts"` —
**768 cells, 438 agree, 330 divergent, every one an over-acceptance.**

| construct | official | cells |
|---|---|---|
| `y as string` | `expected_token` | 30 |
| `y as const` | `expected_token` | 30 |
| `y as unknown as string` | `expected_token` | 30 |
| `y satisfies string` | `expected_token` | 30 |
| `y!` | `expected_token` | 30 |
| `y!.k` | `expected_token` | 30 |
| `<string>y` | `js_parse_error` | 30 |
| `f<string>()` | `js_parse_error` | 30 |
| `((a: string) => a)("")` | `js_parse_error` | 30 |
| `((): string => "")()` | `js_parse_error` | 30 |
| `(<T,>(a: T) => a)(1)` | `js_parse_error` | 30 |

Uniform across all eleven, so the construct is not the discriminating axis — the
**host** is:

| host | divergent |
|---|---|
| no script | 165 |
| plain `<script>` | 165 |
| `<script lang="ts">` | **0** |
| `<script module lang="ts">` only | **0** |

Official turns TS on for the whole component if *any* script declares it,
including a module-only one, and rsvelte agrees there. The divergence is exactly
"no script declares `lang="ts"`".

15 of the 16 slots reproduce it: `{…}`, an attribute value, a spread, an `{#each}`
collection, an `{#each}` body, an `{#if}` test, `{#key}`, `{#await}`, `{@const}`,
`{@html}`, a `{@render}` argument, an event handler, `style:`, `class:`, and a
transition/animate parameter object.

### Controls

- **Negative control on the construct axis**: a twelfth construct that is plain
  JavaScript (`y`) is 0 divergent over its 64 cells. It is TypeScript-ness, not
  the grid.
- **The one clean slot is reachable and discriminating**: `{#snippet s(a = …)}`
  never diverged, and it is not a hole — both sides reject `y as string` there
  with `js_parse_error` under `no_script`, and both accept it under
  `instance_ts`.

### Relation to #3203

#3203 is the same question asked of **script bodies** — 30 TypeScript constructs
× 5 script wrappers — and it records `as`, `as unknown as`, `satisfies` and `!` as
*rejected on both sides at a different position*. In a template expression they
are not rejected at all, so this is a second site with a worse outcome, not a
restatement.

The instance script, `compileModule` and a template expression are three separate
parse entry points in rsvelte, and this is the same shape as #2547, whose fix
covered the script paths while `{(async (p = await x) => p)}` still compiled. A
fix here should be checked against all three entry points, and against
`{#snippet}` type parameters, which are the same defect in the block-header
parser:

| | official | rsvelte |
|---|---|---|
| `{#snippet s<T>(a)}` with no `lang="ts"` | `expected_token` | **compiles** |
| `{#snippet s<T>()}` with no `lang="ts"` | `expected_token` | **compiles** |
| `{#snippet s<T>(a)}` with `lang="ts"` | accepts | accepts |
| `{#snippet s(a: string)}` with no `lang="ts"` | `js_parse_error` | `js_parse_error` |

### LINKED ISSUE #3287: Over-rejection: a trailing `//` comment before a template expression's closing `}` raises `js_parse_error`
**Over-rejection**: a `//` comment whose newline is the last thing before a
template expression's closing `}` makes rsvelte raise `js_parse_error`. Official
compiles every case below. All targets.

```svelte
<div class={f
	? "a"
	: "b" // no
}>x</div>
```

official: compiles. rsvelte: `js_parse_error` at 3:12.

That shape is ordinary formatted code. So is this one:

```svelte
<button onclick={() => sink(1) // do
}>x</button>
```

## Slots

Every one of these is accepted by official and rejected by rsvelte, on client
and server alike:

| slot | source |
|---|---|
| mustache | `<p>{1 // c⏎}</p>` |
| mustache, two comments | `<p>{1 // a⏎+ 2 // b⏎}</p>` |
| attribute | `<p title={1 // c⏎}>x</p>` |
| event handler | `<button onclick={() => sink(1) // c⏎}>x</button>` |
| `{#if}` header | `{#if true // c⏎}<p>x</p>{/if}` |
| `{#each}` key | `{#each [1] as q (q // c⏎)}…{/each}` |
| `{@html}` | `{@html String(1) // c⏎}` |
| spread | `<p {...{ a: 1 } // c⏎}>x</p>` |
| ternary, comment on the last arm | the first repro above |

## What does NOT trigger it

These are the controls, and they place the defect precisely:

| shape | verdict |
|---|---|
| the comment is **leading** (`{// c⏎1}`) | accepted (comment placement differs — see #3070) |
| the comment is in the **middle** (`{1 + // c⏎2}`) | accepted |
| a `)` closes before the `}` (`{cls("a" // c⏎)}`) | accepted |
| a `}` of an inner object closes first (`{st({ a: 1 // c⏎})}`) | accepted |
| an arrow **body** brace closes first (`{() => {⏎sink(1); // c⏎}}`) | accepted |
| a block comment instead (`{1 /* c */}`) | accepted |
| `{#each [1] as q // c⏎}` (the header, not the key) | rejected by **both** |

So it is specifically: the comment's newline, then whitespace, then the
expression's **own** terminator. Any intervening closer moves it back onto the
accepted path — which is why the collected corpus, whose expressions are mostly
single-line or end in `)`, supplies none.

## Found by

Adversarial markup-lexer grid (49 terminator-lookalike inputs × 2 targets = 98
cells, 15 diverging rows), then two follow-up probes: 19 comment slots and 11
realistically-formatted shapes.

baseballyama:
Two parse-phase fixes and one pin, all in the same place: how a template
expression is handed to OXC.

Closes #3287
Closes #3261

`#3198` is **not** closed — see the last section.

## 1. A trailing `//` comment before the expression's own terminator (#3287)

rsvelte slices an expression out of the template, whitespace-trims it and
re-parses it wrapped in `(` … `)`. The trim deletes the newline that terminated
a trailing line comment, so the synthetic closing paren landed *inside* the
comment:

```svelte
<div class={f
	? "a"
	: "b" // no
}>x</div>
```

Official compiles that; rsvelte raised `js_parse_error`. The wrap now emits a
newline before the closing paren, which is what upstream has for free because it
parses in place. Nine slots from the issue plus `{#snippet}` params, `{@const}`
and `{@attach}` are covered, with the issue's six accepted controls and the
`{#each}` *header* (rejected by both) as negative controls.

## 2. A template expression is parsed with the component's one language mode (#3261)

Upstream picks the acorn variant once per component from `parser.ts` — set when
*any* script declares `lang="ts"`, module-only included. rsvelte parsed the
component's mode and then **retried in the other mode on failure**, so
TypeScript-only syntax compiled in a component with no `lang="ts"` anywhere:
11 constructs × 15 slots × 2 hosts, every cell an over-acceptance. `{#snippet
s<T>(a)}`'s generic clause had the same shape in the block-header parser and is
now guarded on `parser.ts` exactly as upstream guards it.

**Removing the retry made a second defect observable, and it is not about
TypeScript.** rsvelte decided `expected_token` vs `js_parse_error` from OXC's
first error label alone, which is a different question from the one upstream
answers: acorn parses ONE maximal expression and only *then* expects the close
token, so leftover input is leftover only when what precedes it is itself a
complete expression. `trailing_token_offset` now checks that, which fixes six
slots on inputs carrying no TypeScript at all — `{@html String(a b)}`,
`<p {...{ a: a b }}>`, `{#each [a b] as q}` and friends were reported as a
missing `}` where official reports a broken expression. An attribute value gets
the classification too (`read_attribute_value` is `read_expression` +
`eat('}', true)`), and a TypeScript-only postfix operator that OXC *recovers* in
a JavaScript file (`as`, `satisfies`, `!` — it reports them as a diagnostic over
the whole expression rather than failing at the token) is located from the
recovered node instead.

Measured against `svelte@5.56.9` on the issue's grid, plus a plain-JavaScript
grid built the same way:

| grid | before | after |
|---|---|---|
| 11 TS constructs × 15 slots × 2 plain hosts | 330 divergent (all over-acceptances) | 66 |
| 3 plain-JS invalid expressions × 15 slots × 2 hosts | 22 divergent | 9 |

Every remaining cell is one of three slots — `{#await}`, `{@const}` and a
`{@render}` argument — which swallow a template-expression parse error outright
and accept plain-JS `{#await y +}` too. That is a separate pre-existing defect,
filed as a follow-up; the test file says so in its header rather than leaving the
slots silently absent.

## 3. `import … assert { … }` on a new line (#3198) — pinned, not closed

The two JavaScript rows and the same-line spelling **already agree** with
official on `main` (`realign_missing_semicolon` moves OXC's insertion-point
report to the token acorn stops on), so this PR adds the regression test that
pins them, plus the `with` spelling, which neither parser restricts on a new
line.

The TypeScript row is still an over-rejection and I am deliberately not claiming
it: OXC's guard is `Kind::Assert if !self.cur_token().is_on_new_line()` in
`parse_import_attributes`, so OXC applies ASI and never builds a `WithClause`,
while acorn-typescript keeps the clause and official compiles the file. Fixing
it needs a repaired re-parse, and the repaired text cannot be threaded through
`RetainedProgram`, whose owner borrows the component source for `'source` — and
even with the parse fixed the client's line-based `extract_imports` drops a
clause that starts on the next line (see the follow-ups). #3198 stays open with
that analysis recorded on the issue.

## Testing

- `crates/rsvelte_core/tests/template_expr_trailing_line_comment_3287.rs`
- `crates/rsvelte_core/tests/template_typescript_mode_3261.rs` (six tests: both
  directions of the mode check, a plain-JS negative control on the construct
  axis, the classification half with no TypeScript in it, snippet generics, and
  the `{#snippet}` default-parameter slot the issue records as already clean)
- `crates/rsvelte_core/tests/import_assert_newline_3198.rs`
- Every expectation was read off `svelte@5.56.9` cell by cell, not guessed, and
  each test was observed failing before its fix.
- Green locally: `parser_fixtures`, `compiler_errors`, `compiler_fixtures`,
  `validator`, `ssr`, `print`, `css`, `block_head_parse_errors`,
  `parser_hardening`, `parser_scanner_lexical`, the eleven `ts_*` suites, and
  `rsvelte_projection`'s svelte2tsx suites.


baseballyama:
Measured this branch against the 472-row gate that ships with #3551 (`crates/rsvelte_core/tests/template_expression_ts_gate.rs`), so this is a cross-execution, not a comparison of the two PRs' self-reported numbers. Oracle is the **pinned submodule** (`submodules/svelte`, 5.56.9), not `node_modules/svelte`.

**38 of 472 rows diverge from official on this branch**, and the direction matters more than the count:

| direction | count |
|---|---|
| official rejects, rsvelte accepts (under-reject) | 16 |
| **rsvelte rejects, official accepts (over-reject)** | **18** |
| both reject, different code | 4 |

The 18 over-rejects are code that official compiles and this branch refuses, which is the direction a user notices. They are **not scattered** — every one is `js_parse_error`, and they collapse onto a single axis:

| | values |
|---|---|
| slot | `each-pattern-default` 10, `await-then-pattern-default` 4, `await-catch-pattern-default` 4 |
| construct | `as-string` 6, `angle-cast` 6, `satisfies` 2, `non-null` 2, `arrow-param-type` 2 |
| host | `instance-ts` 9, `module-ts-only` 9 |

All three slots are the **default value inside a destructuring pattern**, and all 5 TS constructs × both TS hosts fail in each. Two examples:

```svelte
<script lang="ts">let y: any = "s";</script>
{#each [{}] as { a = y as string }}<p>{a}</p>{/each}
{#await Promise.resolve({})}<p>x</p>{:then { a = <string>y }}<p>{a}</p>{/await}
```

Official accepts both; this branch reports `js_parse_error`. One of the 18 carries no TS type syntax in the pattern at all — `{#each [{}] as { a = ((a: string) => a)("") }}` — so it is the pattern path, not the construct.

That points at one site: **`parse_destructuring_pattern` in `phases/1_parse/read/expression.rs`**, which is the same function #3551 edits. #3551 measures 472/472 on this gate.

### Suggested reduction

Given the above and the merge order the team settled on (**#3551 first, then this**), the cheapest way to land the parts of this PR that #3551 does not cover is to reduce it to:

1. the **#3287** hunk (trailing line comments) and its test,
2. the **#3198** pin,

and drop the #3261 half — #3551 closes #3261 with the gate attached, so keeping both means resolving a conflict in `read/expression.rs` for work that would then be measured twice.

### One caveat on the number

The 38 above is scored on **accept vs reject** plus the error **code**. It is not a claim about message text or position — the gate drops both, which is recorded as blind spot 39a in `compatibility/gate-coverage.md`. So a row scored as agreeing here can still differ in what the user reads.

Issue #3295, opened 2026-08-21
Three result-object fields diverge: `ast` is null without `modernAst`, `js.map.file` is a constant, and `outputFilename` adds a `./` to `js.map.sources`


baseballyama:
Three fields on the object `compile()` returns diverge. The corpus comparator
reads `js.code` and `css.code`, so none of them is under any gate today.

## 1. `ast` is `null` unless `modernAst: true`

```js
compile(src, { filename: 'Probe.svelte', generate: 'client' }).ast
```

| | |
|---|---|
| rsvelte | `null` |
| official | `{ html, instance, module, css, _comments }` (the legacy AST) |

| `modernAst` | rsvelte | official |
|---|---|---|
| omitted | **`null`** | `_comments,css,html,instance,module` |
| `false` | **`null`** | `_comments,css,html,instance,module` |
| `true` | `comments,css,end,fragment,js,options,start,type` | identical |

So the modern AST is right and the legacy one is simply not produced. The key is
present on the result object, which means a caller doing
`if ('ast' in result)` sees no difference — it has to read the value.
Svelte-4-era tooling that consumes `result.ast` gets `null` with no error.

## 2. `js.map.file` is a constant that has nothing to do with the input

| | `js.map.file` |
|---|---|
| rsvelte | `"input.svelte.js"` |
| official | absent |

The filename passed was `Probe.svelte`. `"input.svelte.js"` appears to be a
baked-in default. Present on every one of the 2,136 compiled cells in the
matrix, both targets, every option combination.

## 3. `js.map.sources` gains a `./` prefix when `outputFilename` is set

| options | rsvelte | official |
|---|---|---|
| default | `["Probe.svelte"]` | `["Probe.svelte"]` |
| `outputFilename: 'out.js'` | `["./Probe.svelte"]` | `["Probe.svelte"]` |

38 cells in the single-axis sweep, 12 more in the pairs. With `outputFilename`
set rsvelte also reports `file: "out.js"` where official still omits `file`
entirely — so setting that option changes two map fields, neither the way
official does. A bundler resolving `sources` relative to the map's location
gets a different path.

## Population and one correction

Compile-options matrix: 12 axes × 19 bodies × 2 targets = 1,444 cells (1,440
compiled on both sides), plus 13 axis pairs = 720 cells (696 compiled on both
sides).

**A warning divergence this matrix reported is not real and I want it on the
record rather than in the count.** The sweep showed 150 cells where rsvelte
[...]

Issue #3226, opened 2026-08-21
Injected CSS is minified wrong: a stray `;` and unbalanced braces on nesting


baseballyama:
The CSS that gets **injected into the JS bundle** (`css: "injected"`, and every
`customElement` component, which injects unconditionally) is minified wrong: a
declaration that has no trailing `;` in the source emits a stray ` ;`, and a
nested rule emits an extra `{` and an extra `;` — leaving the stylesheet with
**unbalanced braces**.

```svelte
<svelte:options css="injected" />
<script>
	const n = 1;
</script>

<b class="a">{n}</b>
<style>
	.a { color: red; &:hover { color: blue } }
</style>
```

```
official : .a.svelte-1lj1c2h {color:red;&:hover {color:blue}}
rsvelte  : .a.svelte-1lj1c2h {{color:red;;&:hover {color:blue ;}}
```

Three opening braces against two closing ones — the injected stylesheet is not
parseable CSS, so everything after it in the same `<style>` is dropped by the
browser.

| source declarations | official | rsvelte |
|---|---|---|
| `color: red` | `{color:red}` | `{color:red ;}` |
| `color: red;` | `{color:red}` | same |
| `color: red; margin: 0` | `{color:red;margin:0}` | `{color:red;margin:0 ;}` |
| `margin: 0 1px 2px 3px` | `{margin:0 1px 2px 3px}` | `{margin:0 1px 2px 3px ;}` |
| `color: red; &:hover { color: blue }` | `{color:red;&:hover {color:blue}}` | `{{color:red;;&:hover {color:blue ;}}` |

The **external** `css.code` — what the default `css: "external"` returns and what
the corpus gate compares — is byte-identical on every one of these, so no
existing gate sees any of it: the injected string lives inside `js.code` as a
quoted literal, and the only headers that produce it are `css="injected"` and
`customElement`.

Whether a declaration ends in `;` is the discriminator for the first four rows,
and published code almost always writes the `;` — which is why a collected
corpus of real components scores this clean.

Found by a generated grid over 20 `<svelte:options>` headers × 5 bodies × 3
script forms (900 cells, 26 diverging).

baseballyama:
A 34-shape at-rule grid (`@media`, `@supports`, `@container`, `@layer`,
`@scope`, `@starting-style`, `@keyframes`, `@font-face`, `@import`,
`@charset`, `@namespace`, `@page`, `@property`, `@counter-style`,
[...]

PR #3362, opened 2026-08-21
fix: hmr output parity, cssHash at the NAPI boundary, and two fabricated js.map fields

### LINKED ISSUE #3240: hmr: true diverges in four places, and no gate compiles with it
`hmr: true` is the Vite plugin's compile path and no gate compiles with it —
the corpus targets are `client`, `server` and `client-dev`. A 12-component x
16-option grid finds four divergences, and each one is attributable to `hmr`
alone: the same component with `dev: true` and `hmr` unset matches, so
neither is a dev-mode difference.

### 1. The accept hook does not clean up injected styles

```svelte
<script>
	let n = $state(1);
</script>
<div class="a">{n}</div>
<style>
	.a { color: red }
</style>
```

```js
// official
import.meta.hot.accept((module) => {
	$.cleanup_styles('svelte-1lj1c2h');
	A[$.HMR].update(module.default);
});

// rsvelte
import.meta.hot.accept((module) => {
	A[$.HMR].update(module.default);
});
```

Every hot update leaves the previous version's injected stylesheet in the
document, so rules accumulate across edits.

### 2. A custom element is redefined on every hot update

```svelte
<svelte:options customElement="my-el" />
<script>
	let { a = 1 } = $props();
</script>
<div>{a}</div>
```

| | tail of the module |
|---|---|
| official | `if (customElements.get('my-el') == null) customElements.define('my-el', …)` |
| rsvelte | `customElements.define('my-el', …)` |

`customElements.define` throws on a name that is already defined, so the
second hot update of a custom-element component throws.

### 3. A root `{@render}` gets an anchor comment it does not need

```svelte
{#snippet s(x)}<b>{x}</b>{/snippet}{@render s(1)}
```

```js
// official
$.add_svelte_meta(() => s($$anchor, () => 1), 'render', A, 1, 35);

// rsvelte
var fragment = $.comment();
var node = $.first_child(fragment);

$.add_svelte_meta(() => s(node, () => 1), 'render', A, 1, 35);
$.append($$anchor, fragment);
```

### 4. SSR drops the trailing anchor after a component

```svelte
<script>
	import C from "./C.svelte";
</script>
<C a={1}>x</C>
```

official closes the `$$renderer.component(…)` callback with

```js
$$renderer.push(`<!---->`);
```

which rsvelte omits, so the server HTML is missing the closing hydration
anchor and hydration of that subtree mismatches.

Only `hmr` + `server`; the same component under plain `server` and
`server+dev` matches.

### LINKED ISSUE #3294: `cssHash` is unusable: `compile()` ignores it silently and `compileWithCssHash()` passes the wrong argument shape, then kills the process
`cssHash` is a documented public compile option. Neither of the two entry points
that should honour it works.

## `compile()` ignores it silently

```js
compile(src, { filename: 'Probe.svelte', cssHash: () => 'fixed-hash' })
```

```
rsvelte  css: .k.svelte-4psua6 { color: red }
official css: .k.fixed-hash { color: red }
```

No error, no warning — the component just gets a different class name than the
caller asked for, in both `css.code` and every `class` attribute in `js.code`.
This is deliberate at the NAPI layer (`crates/rsvelte_napi/src/lib.rs:620`:
"Unknown JS fields (e.g. the `cssHash` / `warningFilter` callbacks Vite passes)
are silently ignored"), with `compileWithCssHash` documented as the route for a
real callback. The silence is the problem: a drop-in caller gets a wrong answer
rather than an unsupported-option error.

## `compileWithCssHash()` calls the callback with the wrong signature, then kills the process

Official invokes the callback with **one object**:

```
OFFICIAL argc=1 arg0 type=object value={"css":"\n.k { color: red }\n","filename":"Probe.svelte","name":"Probe","hash":"[fn]"}
RSVELTE  argc=3 arg0 type=string value="Probe"
```

Three positional arguments, the first being the component **name**. So:

- `({ hash, css }) => \`zz-${hash(css)}\`` — the documented idiom — throws
  `TypeError: hash is not a function`; `hash` is not passed at all.
- `({ css }) => …` throws on `css.length`.
- `({ name }) => …`, `({ filename }) => …` and `() => 'fixed'` all fail with
  `[Error: Call the PromiseRaw::then failed] { code: 'InvalidArg' }`.

Every one of those is an **uncaught exception** — the `try`/`catch` around
`await compileWithCssHash(...)` never runs and the process exits. Same failure
mode as #3292 on the preprocess side, so the two may share a bridge.

| callback | result |
|---|---|
| `(a) => { inspect(a); return 'x'; }` | arg is the string `"Probe"`, then uncaught `PromiseRaw::then failed` |
| `() => 'fixed-hash'` | uncaught `PromiseRaw::then failed` |
| `({ name }) => \`n-${name}\`` | uncaught `PromiseRaw::then failed` |
| `({ filename }) => …` | uncaught `PromiseRaw::then failed` |
| `({ css }) => \`c-${css.length}\`` | uncaught `TypeError` |
| `({ hash, css }) => \`zz-${hash(css)}\`` | uncaught `TypeError: hash is not a function` |

## Population

Found on a compile-options matrix: 12 option axes × 19 component bodies × 2
targets (1,444 cells, 1,440 compiled on both sides) plus 13 axis pairs on the
bodies that exercise them (720 cells, 696 compiled on both sides). All four
`cssHash` values diverge on `css.code` and on `js.code`, on both targets, for
every body with a scoped selector — 34 `js` and 24 `css` divergences in the
single-axis sweep alone.

### Measured clean in the same matrix

`css: 'injected' | 'external'`, `discloseVersion`, `preserveComments`,
`preserveWhitespace`, `runes: true|false`, `dev`, `namespace`, `name`,
`enableSourcemap` and `modernAst` produce **no** `js.code` or `css.code`
divergence of their own on any of the 19 bodies. The only other output
divergence in the matrix is the stray `;` of #3226 (and the `$$css` object
line-break that follows from it — both sides break that object past a width
threshold, and rsvelte crosses it two characters earlier because of the `;`, so
that is one defect and not two).

baseballyama:
Three compile-option boundary defects, each in a place no gate looks at.

Closes #3240
Closes #3294

## 1. `hmr: true` — four divergences, none of them dev-mode (#3240)

`hmr` is the Vite plugin's compile path and none of the three corpus targets
(`client`, `server`, `client-dev`) sets it, so the whole option went ungated.
Each fix is upstream's own condition:

| | upstream | rsvelte was |
|---|---|---|
| accept hook | `if (analysis.css.hash)` unshifts `$.cleanup_styles(hash)` (`transform-client.js`) | a hard-coded `Raw` block with no CSS branch — every hot update left the previous stylesheet in the document |
| custom element | `b.if(customElements.get(tag) == null, define)` when `options.hmr` | a bare `customElements.define`, which throws on the second hot update |
| `is_standalone` | gated on `hmr` for the **`Component` arm only** (`3-transform/utils.js`) | a blanket `!options.hmr` in the client fragment visitor (so a root `{@render}` also lost standalone), no `hmr` term in the component visitor, and none at all on the server |

The last row is one defect with two ports: the client emits a `$.comment()` /
`$.first_child` / `$.append` wrapper it should not for a `{@render}`, and the
server drops the trailing `$$renderer.push(``)` it should keep after a
component — which is a hydration mismatch. `hmr` now threads into
`clean_nodes` / `clean_nodes_refs` / `clean_node_list` and into the server's
`is_standalone_fragment`, so the two ports share one rule.

`crates/rsvelte_core/tests/hmr_option_parity.rs` pins all four against the
official compiler's verbatim output. A repro must carry **both** a root
`{@render}` and a root `<Component>`: one alone cannot tell upstream's rule
from a blanket one. Negative control run per port — removing the `!hmr` term
from the client predicate fails only
`hmr_suppresses_standalone_for_components_but_not_render_tags`, removing it
from the server predicate fails only
`server_keeps_the_trailing_anchor_after_a_component_under_hmr`.

## 2. `cssHash` (#3294)

Neither entry point that should honour it worked.

**The synchronous entries dropped it silently.** The options decoder ignored
unknown JS fields, so `compile(src, { cssHash: () => 'x' })` returned a
component with a *different scope class* than the caller asked for — in
`css.code` and in every `class` attribute of `js.code` — with no error. They
now reject a function-valued `cssHash` naming `compileWithCssHash`, and report
a non-function with upstream's `validate-options` wording. `LenientScalar`
grows a `Function` variant so the two diagnoses stay apart.

**`compileWithCssHash` called the callback with the wrong shape.** Upstream
passes ONE object, `{ hash, css, name, filename }`; rsvelte passed three
positional arguments starting with the component name, so the documented
`({ hash, css }) => \`x-${hash(css)}\`` idiom threw `hash is not a function`.
The bridge now builds that object, with `hash` a real JS function wrapping the
compiler's own digest, and reads a plain string back. The JS-side adapter in
`index.cjs` is gone and the entry is exported from the package.

**And a throwing callback took the process with it.** napi's
`ThreadsafeFunction::call_async` documents that a JS throw is routed through
`napi_fatal_exception`, which terminates the host — so the `try`/`catch` around
the `await` never ran (the awaited promise settled as `oneshot canceled` while
Node reported an uncaught exception). `call_async_catch` hands the throw back
as `Err`, which becomes a compile failure the way upstream's exception does.
That is a **discriminating** control: built with `call_async`, the new gate's
throwing-callback case kills the runner.

New gate **38**, `scripts/dev/test-napi-css-hash.mjs` (`pnpm run
test:napi-csshash`, wired into the `vps-shim` CI job in its own process, since
two rsvelte addons in one process have been observed to SIGSEGV). It compares
the produced scope class to the **official** compiler across five callback
shapes × two targets, and inspects the callback's own argument list — comparing
only the resulting class would pass against a callback invoked with the right
values in the wrong shape. Its blind spots are written up in
`compatibility/gate-coverage.md` § 38, including 38c: `cssHash` has three ports
(napi, wasm, the `rsvelte` facade) and no gate compares them to each other,
which is exactly how the wasm port ended up with the official shape while the
napi port did not.

Gate 22's `outputFilename` case had to move (see below): its marker was
`js.map.file === 'out.js'`, a field that should not exist. It now discriminates
on `sources`, with a nested `filename`/`outputFilename` pair — a
same-directory output leaves the bare basename, which is what the baseline
already is.

## 3. Two of the three result-object fields of #3295

* `js.map.file` was a baked-in constant `"input.svelte.js"` on every compile,
  unrelated to the input. Upstream builds the JS map with esrap's `print()`,
  which sets no `file` key at all. `generate_sourcemap_json{,_multi}` now take
  `file: Option<&str>`; the **CSS** map keeps its `file`, which upstream does
  set, and that is the control in the new test.
* `outputFilename` prefixed `js.map.sources` with `./`. Upstream's
  `get_relative_path` (`utils/mapped_code.js`) joins the parts verbatim, so the
  result is a bare basename or a `../`-prefixed path, never `./`. A bundler
  resolving `sources` relative to the map's location got a different path.
  `get_basename` also now splits on `\` as upstream does.

`crates/rsvelte_core/tests/js_sourcemap_result_fields.rs` pins both against
values read off the official compiler; three of its five cases failed before
the fix and the two CSS-map cases passed throughout.

**#3295 is not closed by this PR.** Its first item — `result.ast` is `null`
unless `modernAst: true`, where official returns the Svelte-4 legacy AST —
needs a port of `submodules/svelte/packages/svelte/src/compiler/legacy.js`
(641 lines of `convert()`), which is its own piece of work and does not belong
in this change.

## Verification

* `cargo test -p rsvelte_core` — lib (1732), `hmr_option_parity`,
  `js_sourcemap_result_fields`, `compiler_fixtures`, `ssr`, `sourcemaps`,
  `sourcemaps_gate`, and `--release --test runtime` (1207 legacy + 1007 runes +
  hydration + browser) all green.
* `pnpm run test:napi-options` — 104 passed, 0 failed; the denominator moves
  39/41 (`cssHash` is now a declared key, justified as `UNCOVERED` there with a
  pointer to gate 38).
* `pnpm run test:napi-csshash` — 29 passed, 0 failed.
* `pnpm run test:vps-shim` — 62 passed, 0 failed.

### DIFF
--- a/.changeset/hmr-map-fields-and-css-hash.md
+++ b/.changeset/hmr-map-fields-and-css-hash.md
@@ -1,0 +1,10 @@
+---
+"@rsvelte/compiler": patch
+"@rsvelte/vite-plugin-svelte-native": patch
+---
+
+`hmr: true` now matches the official compiler in four places it diverged in: the `import.meta.hot.accept` hook calls `$.cleanup_styles(<hash>)` before the update when the component has CSS (so injected stylesheets no longer accumulate across hot updates), `customElements.define` is guarded by `customElements.get(tag) == null` (a second hot update of a custom-element component used to throw), and `is_standalone` is suppressed for a root component but *not* for a root `{@render}` — which restores the anchor comment on the client and the trailing `<!---->` hydration anchor on the server.
+
+`js.map` no longer carries a `file` key (upstream's esrap-produced map has none; rsvelte emitted a constant `"input.svelte.js"`), and `outputFilename` no longer prefixes `js.map.sources` with `./` — the relative path is joined verbatim, as `get_relative_path` does upstream. The CSS map keeps its `file` key, which upstream does set.
+
+`cssHash` works at the NAPI boundary. `compileWithCssHash` now invokes the callback the way the official compiler does — one `{ hash, css, name, filename }` argument, returning the scope class — so `({ hash, css }) => \`x-${hash(css)}\`` works verbatim, and a throwing callback rejects the returned promise instead of terminating the process. The synchronous entries (`compile`, `compileEnvelope`, …) now *reject* a function-valued `cssHash` naming `compileWithCssHash`, rather than dropping it and silently returning a different scope class.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1091,6 +1091,11 @@ jobs:
       - name: Run NAPI option-boundary gate
         run: pnpm run test:napi-options
 
+      # Again a separate process, for the same reason: this one is the only gate
+      # that drives the `cssHash` callback across the boundary.
+      - name: Run NAPI cssHash gate
+        run: pnpm run test:napi-csshash
+
   vps-sourcemaps:
     name: VPS postprocessing source-map parity
     runs-on: ubuntu-latest
--- a/apps/npm/vite-plugin-svelte-native/index.cjs
+++ b/apps/npm/vite-plugin-svelte-native/index.cjs
@@ -54,32 +54,6 @@ function prepareCompileOptions(options) {
 	return { options: resolved, warningFilter: hasWarningFilter ? warningFilter : undefined };
 }
 
-// Port of Svelte's `hash()` (submodules/svelte/packages/svelte/src/utils.js),
-// handed to a user `cssHash` callback so custom scope classes match upstream's digest.
-const regexReturnCharacters = /\r/g;
-function hash(str) {
-	str = str.replace(regexReturnCharacters, '');
-	let h = 5381;
-	let i = str.length;
-	while (i--) h = ((h << 5) - h) ^ str.charCodeAt(i);
-	return (h >>> 0).toString(36);
-}
-
-// Adapt a user `cssHash` to the bridge shape. It must never reject: a rejected
-// Promise crossing the NAPI boundary can crash V8 during threadsafe-function
-// teardown — so a throw becomes `{ error }` (Rust turns it into a compile failure)
-// and a non-string return becomes `{ value: null }` (Rust falls back to the default hash).
-function makeCssHashCallback(userCssHash) {
-	return async (name, filename, css) => {
-		try {
-			const result = await userCssHash({ hash, css, name, filename });
-			return { value: typeof result === 'string' ? result : null };
-		} catch (err) {
-			return { error: err instanceof Error ? err.message : String(err) };
-		}
-	};
-}
-
 function applyWarningFilter(result, warningFilter) {
 	if (!warningFilter || result == null) return result;
 	const warnings = result.warnings;
@@ -183,11 +157,9 @@ async function compileAsync(source, options) {
 	const { options: resolved, warningFilter } = prepareCompileOptions(options);
 	if (typeof options?.cssHash === 'function') {
 		// The bridge entry returns a plain (JSON) CompileResult, not an envelope.
-		const result = await binding.compileWithCssHash(
-			source,
-			resolved,
-			makeCssHashCallback(options.cssHash),
-		);
+		// The callback is handed straight through: Rust calls it with upstream's
+		// own `{ hash, css, name, filename }` object and reads the string back.
+		const result = await binding.compileWithCssHash(source, resolved, options.cssHash);
 		return applyWarningFilter(result, warningFilter);
 	}
 	if (resolved?.modernAst) {
@@ -254,6 +226,11 @@ module.exports.compileModuleBuffers = binding.compileModuleBuffers;
 module.exports.compileBatch = compileBatch;
 module.exports.compileBatchRaw = binding.compileBatch;
 module.exports.compileAsync = compileAsync;
+// The only entry that honours a function-valued `cssHash`. The callback takes
+// upstream's single `{ hash, css, name, filename }` argument and returns the
+// scope class; the synchronous entries reject a function `cssHash` rather than
+// dropping it. Returns a plain CompileResult (no envelope).
+module.exports.compileWithCssHash = binding.compileWithCssHash;
 module.exports.compileBatchAsync = compileBatchAsync;
 module.exports.compileEnvelopeAsync = binding.compileEnvelopeAsync;
 module.exports.compileBatchAsyncRaw = binding.compileBatchAsync;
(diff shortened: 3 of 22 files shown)


baseballyama:
Extra check while CI queues: a differential sweep of `hmr: true` vs the official compiler over **18 template shapes × client/server × hmr on/off** (72 pairs), driven through the freshly built NAPI addon.

**0 hmr-attributable divergences.** The 8 pairs that do diverge appear identically at `hmr: false`, so none of them is this option: they are all `css: 'injected'` — rsvelte's minifier emitting `.a.svelte-1lj1c2h {color:red;}` where official emits `.a.svelte-1lj1c2h{color:red}`, and the `$$css` object then breaking across lines because it crosses the width threshold two characters earlier. That is #3226 and the line-break that follows from it, already recorded as one defect in #3294's own "measured clean" table.

Shapes covered beyond the four in #3240: a component in a slot, in `{#if}`, in `{#each}`, in `{#snippet}`, inside `<svelte:boundary>`, nested each-in-if, a component carrying a `--x` custom property, a dynamic component, and a render tag inside `{#if}` — i.e. the fragment kinds that reach `clean_nodes` / `is_standalone_fragment` through a path other than the root.

Issue #3342, opened 2026-08-21
A nested `$inspect(…)` statement in a module drops the empty statements official keeps


baseballyama:
Splitting two module-only divergences out of the #3315 paren grid. Both are on the
`bare` row, so they are **not** paren defects — they reproduce with no parentheses at
all, and the paren rows only inherit them.

## `$inspect` in a module function body — client

```js
// m.svelte.js
let base = $state(1);
export function go() { $inspect(base); }
```

```js
// official                       // rsvelte
export function go() {            export function go() {}
	;;
}
```

Upstream's non-dev `CallExpression` visitor turns the `$inspect` call into `b.empty`,
and the enclosing `ExpressionStatement` is kept — so the pair of `;` is what esrap
prints. rsvelte removes the whole statement. The output parses and runs identically,
so only output equality can see it; it is nonetheless a byte divergence on every
`.svelte.js` that inspects inside a function.

The component instance script already gets this right (`inspect_hole_placeholder` in
`3_transform/server/ast/script.rs` exists precisely to emit `;;`), so this is the
module path missing the same distinction.

Found while fixing #3303 / #3315 / #3336; deliberately left out of that PR because it
is a different axis (the statement's *residue*, not its bracketing).

baseballyama:
Scope correction: this is not limited to a module **function body**. Measured over 8 hosts × dev/non-dev:

| host (non-dev) | official | rsvelte |
|---|---|---|
| `$inspect(n);` at module top level | `;;` | (nothing) |
| inside an exported `function` | `;;` | (nothing) |
| inside a bare block in a function | `;;` | (nothing) |
| `$inspect(n).with(console.log);` at top level | `;;` | (nothing) |
| two `$inspect` statements in a row | `;; ;;` | (nothing) |
| inside a `static {}` block | `;;` | (nothing) |
| inside an arrow body | `;;` | (nothing) |
| a module with no `$inspect` (control) | — | — |

**7 of 7 hosts diverge in non-dev; all 8 agree in dev** (where the call is kept, so there is no
hole to place). The two controls are the last row and the whole dev column.

So the fix is at the statement level in the module pipeline, not at any particular host — the
same distinction `inspect_hole_placeholder` already makes for the component instance script.

Found by an option × module-source grid over `compileModule` (220 cells); the `$inspect` row
diverged under every one of the 10 option variants, which is what says the host is not the axis.

baseballyama:
[...]

Issue #2986, opened 2026-08-15
SSR .svelte.ts: a `class ` occurrence in a comment or string makes the class-field scanner rewrite a factory function into a class body (invalid JS)


baseballyama:
## Summary

In the SSR `.svelte.js` / `.svelte.ts` path, the class-field transform locates the class with a raw
byte search for `class ` over the whole script — comments and string literals included. Any
occurrence outside a class makes everything from there to the next `{` a "class header" and the
following function body a "class body", so a local `const x = $derived(…)` is lowered to a private
class field and **the emitted module is not JavaScript**.

## Repro (current `main`, `generate: 'server'`)

```js
// we avoid class here
export const make = () => {
	const flag = $derived(a !== b);
	return { read() { return flag; } };
};
```

`compileModule(source, { generate: 'server' })` emits:

```js
export const make = () => {
		#const_flag = $.derived(() => a !== b);

		get const flag() {
			return this.#const_flag();
		}

		set const flag($$value) {
			return this.#const_flag($$value);
		}

	return { read() { return flag; } };
	};
```

`#const_flag = …` outside a class body and `get const flag()` are both syntax errors — a bundler
rejects the module (`Expected 'in' but found '='`).

**Negative control:** delete the comment line and the same input compiles correctly. Removing only
the trailing word (`// we avoid class`) is also correct — the search needs the trailing space.

A string literal is enough as well:

```js
const label = 'class name';
export const make = () => {
	const flag = $derived(a !== b);
	return { read() { return flag; } };
};
```

`class C { flag = $derived(1); }` (a real class) stays correct, and the `client` target is
unaffected for all of the above.

## Mechanism

`crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_script.rs:4918` and `:4927`:

```rust
if memmem::find(script_bytes, b"class ").is_none() … { return script.to_string(); }
let Some(class_pos) = memmem::find(script_bytes, b"class ") else { … };
let after_class = &script[class_pos..];
let Some(brace_pos) = after_class.find('{') else { … };
```

Both the keyword search and the `find('{')` that follows are raw byte scans. The step immediately
after them — `find_matching_bracket` — *is* comment/string aware, and its comment says so (#2253);
the entry point that decides *where the class starts* never got the same treatment. It also takes
the **first** `class ` in the file, so a file with a comment mentioning `class` before a real class
mis-locates the body too.

## Real-world impact

[...]

Issue #3271, opened 2026-08-21
Prod: erasing `$inspect` leaves a different number of empty statements than upstream (`;;` vs `;` / nothing)


baseballyama:
In **prod** mode `$inspect(x)` is erased, and upstream leaves **two** empty
statements where rsvelte leaves one, none, or a different grouping. Client and
server; dev mode matches everywhere (the call survives there).

```svelte
<script>
	let base = $state(1);
	function f() { $inspect(base); }
	f();
</script>
<p>{base}</p>
```

`generate: 'client'`:

```js
// official
function f() {
		;;
}
// rsvelte
function f() {
		;
}
```

`generate: 'server'` for the same source: official emits `;;`, rsvelte emits
nothing at all.

Two `$inspect` calls in the instance script diverge in grouping rather than
count — official emits `;;` on each of two lines, rsvelte emits `;;;;` on one.

## Scope

| context | client | server | dev (both) |
|---|---|---|---|
| instance top level, one `$inspect` | match | match | match |
| `$inspect(x).with(console.log)` | match | match | match |
| `$inspect.trace()` in a function | match | match | match |
| a statement after the `$inspect` | match | match | match |
| inside a function declaration | `;;` vs `;` | `;;` vs nothing | match |
| inside an arrow body | `;;` vs `;` | `;;` vs nothing | match |
| inside a bare block / `if` / `try` | `;;` vs `;` | `;;` vs nothing | match |
| inside a class method | `;;` vs `;` | `;;` vs nothing | match |
| `<script module>` | `;;` vs nothing | match | match |
| a `.svelte.js` module | `;;` vs nothing | — | match |
| two `$inspect` in the instance script | grouping | grouping | match |

The four matching rows are the controls: the erasure itself is right, and it is
only the residue that differs — so this is where the removed statement's slot
is filled, not whether the call is removed.

## Note on visibility

Empty statements are exactly what the corpus comparison's oxfmt normalization
removes, so this class cannot fail `verify.mjs` — the same reason #3045's
"empty `$:` statement" row is byte-parity-only. Filing for byte equality, not
because anything misbehaves at runtime.

## Found by

Adversarial rune × context grid (17 rune forms × 10 contexts × 4 targets = 646
cells), then the 13-context follow-up above with hand-written sources (the
grid's `class-field` context mangles multi-statement bodies and its rows are
discounted).

baseballyama:
[...]

PR #3364, opened 2026-08-21
fix: keep the `;;` slots of a removed $inspect and normalise a template CR in the SSR fold

### LINKED ISSUE #3231: `$inspect` in a `.svelte.js` module drops the statement official keeps as `;;`
`$inspect` in a `.svelte.js` / `.svelte.ts` module is removed entirely here;
official leaves the statement slots behind.

```js
// m.svelte.js
let a = $state(1);
$inspect(a);
```

```js
// official
let a = 1;

;;
// rsvelte
let a = 1;
```

Both are correct JavaScript and both do nothing at runtime — `$inspect` is a
no-op outside dev — so this is a byte divergence in generated code, which the
corpus gate compares directly. Reproduces on `.svelte.js` and `.svelte.ts`
alike.

Official's `;;` is what its statement-level rune removal leaves when it replaces
the expression with nothing rather than dropping the statement; matching it means
reproducing the empty statements, not tidying them away.

Found by a generated `compileModule` grid over 40 module bodies × 2 languages
(76 cells, 6 diverging).

### LINKED ISSUE #3282: Server: a lone CR in a template literal is deleted from the folded value instead of normalised to LF
On the **server**, a lone CR inside a template literal is dropped from the
folded value instead of being normalised to LF.

```svelte
<script>
	const s = `a<CR>b`;
</script>
<p>{s}</p>
```

(`<CR>` is a literal U+000D, no LF after it.)

`generate: 'server'`:

```js
// official
export default function A($$renderer) {
	const s = `a
b`;
	$$renderer.push(`<p>a
b</p>`);
}

// rsvelte — the line break is gone from the rendered text
export default function A($$renderer) {
	const s = `a
b`;
	$$renderer.push(`<p>ab</p>`);
}
```

Both sides print the *declaration* the same way; they disagree only on the
constant-folded value that goes into the rendered output, so the SSR HTML is
missing a newline the client render has.

JS template-literal semantics normalise a lone CR in the cooked value to LF —
official does that, rsvelte deletes the character.

## Shapes

| source | client | server | client-dev |
|---|---|---|---|
| `` `a<CR>b` `` | match | **DIFF** | match |
| `` `a<CR>b<CR>c` `` | match | **DIFF** | match |
| `` `a<CR><LF>b` `` (CRLF) | match | match | match |
| `` `a<LF>b` `` | match | match | match |
| `` `a\rb` `` (escape, not a raw CR) | match | match | match |
| `'a\rb'` (string literal) | match | match | match |
| `` `a<CR>${n}b` `` (with an interpolation) | match | match | match |
| a raw CR in `<pre>` text | match | match | match |
| a raw CR in an attribute value | match | match | match |

So it needs all three of: server, a raw CR (not `\r`, not CRLF), and a template
literal with **no** interpolation — the interpolated form takes a different
fold path and is correct.

## Found by

Adversarial escaping grid (29 payloads × 15 slots × 3 targets = 1,305 cells),
then the 9-shape follow-up above.

baseballyama:
Two SSR/module output divergences, both verified byte-for-byte against the official
compiler (`compileModule` / `compile` from `submodules/svelte`).

Closes #3231
Closes #3282

## #3231 — a removed `$inspect(...)` must leave its statement slots

Upstream's server/client `CallExpression` visitor returns `b.empty` for the **call**
(`if (!dev) return b.empty`) while the `ExpressionStatement` around it survives, so the
statement prints as `;;`. rsvelte's `.svelte.(js|ts)` module pipeline deleted the whole
statement instead — with a raw `memmem::find(b"$inspect(")` scan.

Three things fall out of one fix:

* the statement slots are kept, so a module's output matches official again;
* the scan is now AST-driven (`inspect_rune_ast::transform_module_inspect_removal_ast`),
  so `$inspect(` spelled inside a **string, a template literal or a comment** is no longer
  rewritten — the `opaque-keyword` class, three more instances of #2986/#2987/#2988;
* consecutive holes no longer merge. `rsvelte_esrap`'s `joined` rule made *every* adjacent
  kept-`EmptyStatement` join, so three `$inspect` statements printed `;;;;;;` instead of
  `;;\n;;\n;;`. Joining now alternates, which also fixes the **component instance script**
  path (`server/ast/script.rs`'s `empty_kept` pair) — that port had the same defect and no
  gate compared the two ports to each other.

Measured over a 28-case × 2-target grid against the official compiler: **38 → 8** diverging
cells. The remaining 8 are inputs where **official itself emits text no JS parser accepts**
(`export const x = ;;`, `return [;];`, `(;, console.log(a));` — a `$inspect` call in
expression rather than statement position, which the validator does not reject) plus
`if (a) $inspect(a);`, where rsvelte prints the semantically identical `if (a) ;\n\n;`.
Reproducing unparseable output is deliberately not attempted.

## #3282 — a lone CR in a template literal is not a deleted character

ECMA-262 normalises `<CR><LF>` and a lone `<CR>` to `<LF>` in a template literal's cooked
value. The SSR constant fold (`server/helpers.rs::try_insert_constant_value`) harvests the
literal from **raw source text**, so the carriage return went into the rendered HTML and the
SSR output disagreed with the client render — which folds through the AST's `cooked` value
and was already correct. Normalisation now happens before the cook, for the backtick form
only (a string literal cannot carry a raw line terminator).

Measured over a 15-shape × 3-target grid: **7 → 0** diverging cells.

## Tests

`crates/rsvelte_core/tests/module_inspect_removal_3231.rs` and
`crates/rsvelte_core/tests/template_literal_cr_fold_3282.rs`. Every expectation is the
official compiler's recorded output for the same input, so the assertions are parity
assertions rather than snapshots of current behaviour.

Negative controls (each change reverted in isolation, rebuilt, re-run):

* dropping the CR normalisation → the 7 server cells fail, each keeping `\r` where official
  has `\n`;
* dropping the esrap `joined` alternation → `two` / `three` fail in **both** the module and
  the component grid, printing one merged `;`-run;
* the whole `$inspect` change → 38 module cells fail.


baseballyama:
This PR fixes two open issues that its body does not cite, so they will not auto-close when it lands. **Please add `Closes #3342` and `Closes #3562` to the description.**

I measured a 176-cell `$inspect` / empty-statement grid against the pinned oracle (`submodules/svelte` @ `20b341f10048`, `VERSION` checked at runtime as `5.56.9`) while working on #3271, and both issues land on decision points this PR owns:

* **#3342** — "a nested `$inspect(…)` statement in a module drops the empty statements official keeps". That is the module removal loop this PR rewrites AST-driven (`client/inspect_rune_ast.rs` / the `memmem::find(b"$inspect(")` loop in `client/mod.rs`). #3342's own body already names the same cause ("rsvelte removes the whole statement"), and its scope comment adds the top-level, `.with()`, `static {}` and arrow hosts.
* **#3562** — "consecutive removed `$inspect` statements collapse to `;;;;` on one line". That is the `joined` rule in `rsvelte_esrap/src/printer.rs` this PR alternates. #3562 explicitly records that the **component instance** path has the same defect on `origin/main` with no patch applied, and this PR's body says it fixes that port too (`server/ast/script.rs`'s `empty_kept` pair).

Neither issue number appears in any open PR body, so a search by number reads them as unclaimed — I only found the overlap by searching for the *mechanism* (`printer.rs`'s `joined`, the module `$inspect` loop) rather than the number.

**Coordination**, so we do not collide: I have #3604 open for #3271 and it deliberately does **not** touch `rsvelte_esrap/src/printer.rs` or `client/inspect_rune_ast.rs`. The one file we share is `client/rune_transforms.rs`, and only in the non-dev `$inspect` removal branch.

Two things worth checking on your side, both measured rather than guessed:

1. **This branch is behind `origin/main`** (`git merge-base --is-ancestor origin/main <head>` fails), and `main` has since moved again to `613fa1eb9`.
2. I could not build this branch to verify the above end-to-end. My first attempt failed with three errors in `client/visitors/expression_converter.rs` (`ComponentAnalysis` not in scope, `evaluate_unknown` module missing, an arity mismatch) — but **that was a contaminated working tree, not this branch**: another session was editing the same checkout, and `origin/fix/3228-3213-derived-known-inspect-operand` failed with the identical three errors. Two branches failing the same way means the common factor is the environment. I am flagging it rather than asserting this branch is broken; a clean build here is worth one run.

PR #3374, opened 2026-08-22
fix(compiler): strip a leading UTF-8 BOM before compiling

### LINKED ISSUE #3277: A leading UTF-8 BOM is emitted into the template as document content
A UTF-8 BOM at the start of a `.svelte` file is emitted into the component's
template as document content. Official strips it.

```svelte
<!-- file begins with U+FEFF -->
<p>x</p>
```

```js
// rsvelte (client)
var root = $.from_html(`﻿<p>x</p>`, 1);   // BOM inside the template, and the
                                               // extra-node flag is now set

// official
var root = $.from_html(`<p>x</p>`);
```

Server target likewise: `$$renderer.push(\`﻿<p>x</p>\`)` against
`$$renderer.push(\`<p>x</p>\`)`.

Consequences beyond byte parity: the rendered DOM gains a leading zero-width
no-break space, and the template gains a node so the `, 1` flag and the
`$.next()` / `var fragment = root()` shape change with it — hydration walks a
different node count than the server produced.

With a `<script>` present the injected text is `﻿ ` (BOM plus the space that
followed the script tag), so the divergence is not confined to a
markup-only file.

### Controls

| source | result |
|---|---|
| no BOM | **identical** |
| BOM immediately after `<script>` | **identical** |
| BOM at file start, markup only | diverges |
| BOM at file start, with a `<script>` | diverges |

So it is specifically a BOM in leading position, not U+FEFF anywhere in the file
— which is the case a Windows editor produces by default and every corpus
repository happens not to contain.

baseballyama:
## What

A leading UTF-8 BOM was compiled as document content.

Upstream `submodules/svelte/packages/svelte/src/compiler/index.js` calls
`remove_bom(source)` at **every** public entry — `compile`, `compileModule`,
`parse`, `parseCss` — so `U+FEFF` at index 0 never reaches the parser and every
position downstream is relative to the trimmed source. rsvelte kept it.

## Measured, before

```
$ printf '\xef\xbb\xbf<p>x</p>\n' > t.svelte     # BOM, then markup
```

```diff
-var root = $.from_html(`<p>x</p>`);
+var root = $.from_html(`﻿<p>x</p>`, 1);

 export default function T($$anchor) {
-	var p = root();
+	$.next();
+	var fragment = root();
+	$.next();
-	$.append($$anchor, p);
+	$.append($$anchor, fragment);
 }
```

So it is not only a byte divergence: the template gains a node, so the
extra-node flag and the `$.next()` / `var fragment = root()` shape change with
it, and a hydrating client walks a different node count than the server
produced. The server pushed the BOM into its markup for the same reason.

## Fix

`compiler::remove_bom` (one `strip_prefix('\u{feff}')`), applied at the two
funnels the public API goes through:

* `toolchain::PreparedComponent::new` — covers `compile`, `compile_both`,
  `compile_client_with_program_sink`, `compile_with_external_sourcemap_content`,
  `compile_batch` and `Toolchain::prepare`, and stores the trimmed slice, so
  transform, warnings and `sourcesContent` all see one source;
* `compiler::compile_module`.

It is a strict no-op for a source that does not start with a BOM.

## Verification

Every expectation was read off the official compiler
(`submodules/svelte/packages/svelte/src/compiler/index.js`), one input per
process, against `compile_one`:

| source | before | after |
|---|---|---|
| BOM + `<p>x</p>`, client | DIVERGE | MATCH |
| BOM + `<p>x</p>`, server | DIVERGE | MATCH |
| BOM + `<script>` + markup, client / server | DIVERGE | MATCH |
| `<p>a␎FEFF␏b</p>` (BOM **not** leading), client | MATCH | MATCH |
| BOM + `<p>a␎FEFF␏b</p>` | — | MATCH (only index 0 is stripped) |

`compileModule('﻿' + src)` is byte-identical to `compileModule(src)`
upstream; `crates/rsvelte_core/tests/leading_bom_3277.rs` pins that too.

Regression test: `crates/rsvelte_core/tests/leading_bom_3277.rs` (5 tests,
covering client, server, `compile_both`, `compile_module`, and the
non-leading-BOM control).

Not in scope: the standalone `parse()` / `parseEnvelope` NAPI surface, which
upstream also strips. Stripping inside `phase1_parse::parse` would shift every
returned offset by three bytes relative to a caller that still holds the
untrimmed string (svelte2tsx, the formatter, the language server all call it
directly), so it needs its own boundary rather than a change to the shared
parser.

Closes #3277

### DIFF
--- a/.changeset/quiet-jars-smile.md
+++ b/.changeset/quiet-jars-smile.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Strip a leading UTF-8 BOM before compiling, as `svelte/compiler` does. A BOM at the start of a `.svelte` or `.svelte.js` file was treated as template content, which added a text node to the client template (changing the extra-node flag and the fragment shape) and a stray zero-width no-break space to the server output.
--- a/crates/rsvelte_core/src/compiler/mod.rs
+++ b/crates/rsvelte_core/src/compiler/mod.rs
@@ -533,6 +533,13 @@ fn tabs_to_spaces_column(line: &str, column: usize) -> usize {
     taken + leading_tabs.min(taken)
 }
 
+/// Drop a leading byte order mark, which would otherwise be template content.
+/// Upstream `compiler/index.js` does this at every public entry point, so every
+/// position downstream is relative to the trimmed source.
+pub(crate) fn remove_bom(source: &str) -> &str {
+    source.strip_prefix('\u{feff}').unwrap_or(source)
+}
+
 /// Parse phase shared by [`compile`] and [`compile_both`]: the fixed component
 /// parse options plus [`phase1_parse::parse`](phases::phase1_parse::parse).
 ///
@@ -931,6 +938,7 @@ pub fn compile_module(
     source: &str,
     options: ModuleCompileOptions,
 ) -> Result<CompileResult, CompileError> {
+    let source = remove_bom(source);
     // Parse JS source into an AST using the same infrastructure as component scripts.
     // Upstream `compileModule` → `analyze_module` always parses with
     // `typescript: false` (2-analyze/index.js `parse(source, comments, false,
--- a/crates/rsvelte_core/src/toolchain.rs
+++ b/crates/rsvelte_core/src/toolchain.rs
@@ -340,6 +340,7 @@ impl std::fmt::Debug for PreparedComponent<'_> {
 
 impl<'source> PreparedComponent<'source> {
     pub(crate) fn new(source: &'source str, options: CompileOptions) -> Result<Self, CompileError> {
+        let source = crate::compiler::remove_bom(source);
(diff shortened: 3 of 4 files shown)


baseballyama:
PR #3531 covers the parse half of the same upstream call — `parse`, `parseEnvelope` and the wasm `parse_svelte` — and touches none of this PR's files (`compiler/mod.rs`, `toolchain.rs`, `tests/leading_bom_3277.rs`). Measured, not assumed: with a BOM, `parse()` diverged from official on every source I tried in both AST modes (extra `Text` node, `.end` off by one, `.css` spans shifted), and the same sources without a BOM were clean.

The two are complementary and conflict-free. Once both land, this PR's `pub(crate) compiler::remove_bom` and #3531's `pub phase1_parse::remove_bom` should collapse into one — I left them separate rather than pre-empting a PR I do not own.

PR #3392, opened 2026-08-22
fix(parse): decode textarea content with the attribute-value entity rule

### LINKED ISSUE #3205: `<textarea>` decodes a semicolon-less entity prefix the official compiler leaves literal
`<textarea>` content that contains a **semicolon-less legacy named reference as a
prefix** is decoded here and left literal by official, so the rendered text
differs.

```svelte
<textarea>a&notreal;b</textarea>
```

```js
// official server output
$$renderer.push(`<textarea>a&amp;notreal;b</textarea>`);
// rsvelte
$$renderer.push(`<textarea>a¬real;b</textarea>`);
```

`&notreal;` is not an entity, so official keeps the `&` literal and escapes it;
rsvelte matches the `&not` prefix from HTML's legacy list and decodes it to `¬`,
leaving `real;` behind. Same with `&ampx;` → official `&amp;ampx;`, rsvelte
`&amp;x;`.

The scope is narrow and that is what makes it a decoder bug rather than a policy:

| input | verdict |
|---|---|
| `<textarea>a&notreal;b</textarea>` | **diverges** (server only) |
| `<textarea>a&ampx;b</textarea>` | **diverges** (server only) |
| `<div>a&notreal;b</div>` | agrees |
| `<pre>a&notreal;b</pre>` | agrees |
| `<title>a&notreal;b</title>` | agrees |
| `<textarea>a&not;b</textarea>` | agrees (a real entity, with its semicolon) |
| `<textarea>a&amp;b</textarea>`, `&nbsp;`, `&lt;`, `a & b`, `&copy b` | agree |

Found by a generated grid over **29 text payloads × 12 hosts × 3 targets** (1044
cells): 4 diverging, 2 of them this. The payload axis carries entity forms
(named, numeric, hex, unknown, bare `&`), whitespace runs and newlines, astral
and combining characters, a ZWJ, CRLF, adjacent mustache tags and a comment; the
host axis carries the raw-text and whitespace-preserving elements alongside
ordinary ones, which is what localises this to `<textarea>`.

baseballyama:
## What

Two defects in one function, coupled: fixing the first makes `<textarea>` reach
the second.

### 1. `<textarea>` content was decoded in the wrong mode

Upstream reads a `<textarea>`'s children with `read_sequence`
(`1-parse/state/element.js:401`), whose `flush` calls
`decode_character_references(current_chunk.raw, **true**)` — the
**attribute-value** rule. Under it a semicolon-less legacy name is written
`${name}\b(?!=)`, so `&not` in `&notreal;` has no word boundary after it and
stays literal. Ordinary text goes through `state/text.js`, which passes `false`
and does decode the prefix. rsvelte passed `false` in both places.

```
<textarea>a&notreal;b</textarea>
  official  $$renderer.push(`<textarea>a&amp;notreal;b</textarea>`);
  rsvelte   $$renderer.push(`<textarea>a¬real;b</textarea>`);
```

Client-invisible: the client template carries the text node's `raw`, so only the
server target — and any `<textarea>` with an expression tag, which turns the
content into a `value` attribute — reports it.

### 2. `_` did not close the word boundary

`\b` is JavaScript's, so `\w` is `[A-Za-z0-9_]`. rsvelte's guard tested
`b == b'=' || b.is_ascii_alphanumeric()`, one character short, so `&amp_b`
decoded in **every** attribute value:

```
<div title="a&amp_b">x</div>
  official  `<div title="a&amp;amp_b">x</div>`
  rsvelte   `<div title="a&amp;_b">x</div>`
```

This is not a `<textarea>` case — the plain `title` attribute above reproduces it
on its own — but it is why the two land together: with only fix 1, a `<textarea>`
switches to attribute mode and immediately hits it.

## Verification

Generated grid, 32 entity payloads × 10 hosts × {client, server}, both compilers,
official as oracle (`submodules/svelte/.../compiler/index.js`):

| | match | diverge |
|---|---|---|
| before | 582 | **54** |
| after | 636 | **0** |

The 54 split as 42 `<textarea>` cells (defect 1) and 12 `_` cells across
`attr-quoted` / `attr-unquoted` / `textarea-value` (defect 2).

Element-kind axis, run separately because "which parents does this rule apply
to" is one `parent_type` test upstream and one visitor arm here — 6 payloads ×
24 hosts covering `RegularElement` (text, quoted attr, unquoted attr, `style`
value), `<textarea>` (content, uppercase tag, own attribute, `value` attribute,
inside `<svelte:head>`), `SvelteElement` (`this="textarea"` and an attribute),
`Component`, `SvelteSelf`, `SvelteWindow`, `SvelteBody`, `SvelteDocument`,
`SvelteHead`, `SvelteFragment`, `SlotElement`, `TitleElement`, `<option>`,
`<pre>`, and an each-block child: **228 compared, 0 divergences** (60 cells
official rejects for unrelated placement/attribute rules, e.g. `title` on
`<svelte:window>`). Raw-text handling is keyed on the tag NAME in both compilers
(`name == "textarea"` here, `tag.name === 'textarea'` upstream), so
`<svelte:element this="textarea">` correctly gets the ordinary path on both
sides.

The unit assertions added to `entities.rs` were each checked against
`decode_character_references` directly:

```
"&amp_b" attr => "&amp_b"     "&amp_b" content => "&_b"
"&not_x" attr => "&not_x"     "&not real;" attr => "¬ real;"
"&notreal;" attr => "&notreal;"   "&ampx;" attr => "&ampx;"
```

## Tests

* `crates/rsvelte_core/tests/textarea_legacy_entity_3205.rs` — 6 tests, both
  targets, with the three controls that separate "changed mode" from "stopped
  decoding" (`&not;`, `&not b`, `&#65b`) and the content-mode hosts that must
  not move.
* `compatibility/pattern-corpus/issues/3205-textarea-legacy-entity.svelte` and
  `3205-entity-word-boundary-underscore.svelte`, with their README rows. Both
  verified byte-identical to official on client and server, and both already
  prettier-clean.
* Existing `entities` / `html` unit tests extended rather than replaced (27
  pass).

Closes #3205


baseballyama:

Issue #3323, opened 2026-08-21
Client: a sequence expression in a concatenated interpolation is missing its `?? ''` guard


baseballyama:
When several interpolations are concatenated into one template literal, official
guards each with `?? ''`. rsvelte omits the guard for a `SequenceExpression`.

```svelte
<script>
	let s = "x"; let n = 1; let arr = [1]; void s; void n; void arr;
</script>
{#each arr as q}{(n, s)}{q}{/each}
```

```js
rs: $.template_effect(() => $.set_text(text, `${(n, s)}${$.get(q) ?? ''}`));
sv: $.template_effect(() => $.set_text(text, `${(n, s) ?? ''}${$.get(q) ?? ''}`));
```

The `{q}` slot right next to it gets the guard on both sides, so the guard is
being applied per interpolation and skipped for this one node kind.

Runtime effect: if the sequence evaluates to `null` or `undefined`, rsvelte
renders the string `"undefined"` where official renders nothing.

## Scope

**4 cells**: `in_each` and `in_await`, on `client` and `client-dev`. Those are
the two hosts in the grid that put a second interpolation next to the first —
the concatenation is the precondition, not the block.

Controls, same source with one thing varied:

| source | both sides |
|---|---|
| `{#each arr as q}{s}{q}{/each}` | `` `x${$.get(q) ?? ''}` `` — identical |
| `{#each arr as q}{(n, s)}{/each}` | `text.nodeValue = (n, s);` — identical |
| `{(n, s)}` | `text.nodeValue = (n, s);` — identical |

A sequence alone is a one-shot assignment on both sides and needs no guard;
another expression kind in the same concatenation slot gets it on both sides. It
is specifically a sequence *in* a concatenation.

## Negative control

The tags grid's plain-mustache control family — 19 expression kinds x 10 hosts x
3 targets = 570 cells — diverges on only two kinds, this one and `new_expr`
(separate issue). The other 17, including `s || o.k`, `(s = s)`, `n ? s : o.k`,
`o?.k` and `(() => s)()`, are byte-identical in the same `in_each` slot.

Svelte 5.56.8.

Issue #3027, opened 2026-08-17
codegen (client): `$derived(cond ? undefined : null)` is constant-folded, silently dropping reactivity


MathiasWP:
### Summary

`undefined` and `null` appear to be treated as the same value when constant-folding a ternary. A `$derived` whose two branches are *different* nullish literals is therefore judged constant, and the value is emitted non-reactively — the attribute or prop never updates when the condition changes.

Versions: `@rsvelte/vite-plugin-svelte@0.5.2`, `@rsvelte/vite-plugin-svelte-native@0.3.7`, svelte 5.56.8, vite 8.2.0.

### Repro

```svelte
<script>
	const { n } = $props();
	const c = $derived(n > 3 ? undefined : null);
</script>

<div title={c}></div>
```

`compile(source, { generate: 'client', dev: false })`:

```js
// upstream svelte 5.56.8
$.template_effect(() => $.set_attribute(div, 'title', $.get(c)));

// @rsvelte/vite-plugin-svelte-native 0.3.7 — hoisted out of the effect, never re-runs
$.set_attribute(div, 'title', $.get(c));
```

Same defect for a component prop — upstream emits a reactive getter, rsvelte a one-shot value:

```js
// upstream                       // rsvelte
C($$anchor, {                     C($$anchor, { to: $.get(c) });
	get to() { return $.get(c); }
});
```

### Truth table

| expression | upstream reactive | rsvelte reactive | correct? |
| --- | --- | --- | --- |
| `cond ? undefined : null` | yes | **no** | ✗ |
| `cond ? null : undefined` | yes | **no** | ✗ |
| `cond ? undefined : undefined` | no | no | ✓ genuinely constant |
| `cond ? null : null` | no | no | ✓ |
| `cond ? undefined : 1` | yes | yes | ✓ |
| `cond ? null : 1` | yes | yes | ✓ |
| `cond ? void 0 : null` | yes | yes | ✓ |

Note the last row: `void 0` vs `null` is *not* folded, so the fold appears keyed on the `undefined` identifier specifically.

Non-nullish literal pairs (`cond ? 1 : 2`, `cond ? 'a' : 'b'`, `cond ? true : false`) are all handled correctly.

### Why it matters

`undefined` and `null` are distinguishable, and code does distinguish them. Our real hit is a feed row that renders a "field was cleared" transition:

```svelte
const cleared = $derived(previousValue === undefined ? undefined : null);
<FieldChangeRow ... to={cleared} />
```

[...]

PR #3400, opened 2026-08-22
fix(compiler): walk a NewExpression's children for client template metadata

### LINKED ISSUE #3322: Client: `{new X()}` in a text interpolation is wrapped in `template_effect` where official assigns `nodeValue` once
A `new` expression in a text interpolation is classified as reactive by rsvelte
and as one-shot by official.

```svelte
<script>
	let s = "x"; void s;
</script>
{new String(s)}
```

```js
rs: $.template_effect(() => $.set_text(text, new String(s)));
sv: text.nodeValue = new String(s);
```

`new Date()` reproduces it identically, so it is the `NewExpression` node and not
what it constructs or what it reads.

## rsvelte disagrees with itself, not only with official

The same grid, one node kind changed:

| expression | rsvelte | official |
|---|---|---|
| `{String(s)}` | `$.template_effect(($0) => $.set_text(text, $0), [() => String(s)])` | identical |
| `{new String(s)}` | `$.template_effect(() => $.set_text(text, new String(s)))` | `text.nodeValue = new String(s)` |

So a `CallExpression` gets the **dependency-array** form and matches official
exactly, while a `NewExpression` gets the **bare-closure** form — a third
answer that neither the call path nor official produces. Whatever decides
"reactive, and in which form" is reached for `NewExpression` through a different
branch.

## Hosts

7 of 10 hosts x 2 targets = **14 cells**: `root`, `in_if`, `in_snippet`,
`component_child`, `in_key`, `in_boundary`, `in_fragment`, on `client` and
`client-dev`. `element_child`, `in_each` and `in_await` do not diverge, and
`server` does not.

## Negative control

Found by the negative-control family of the tags grid — 19 expression kinds x 10
hosts x 3 targets = 570 cells in a plain mustache, of which **only** `new_expr`
and `sequence` (the separate issue) diverge. The other 17 kinds, including
`f()`, `o.f()`, `o?.k`, `(() => s)()`, `` `a${s}b` ``, `s || o.k`, `(s = s)`,
`n ? s : o.k`, `[...arr].join("")`, `await …`, and the four literal forms, are
byte-identical in every host and every target. So the interpolation lowering is
right for everything else in the same slot.

Svelte 5.56.8.

baseballyama:
Closes #3322

## What

`NewExpression` reached no arm of the client's template-expression metadata walk, so it was answered by four different defaults at once:

| predicate | old answer for a `new` | upstream |
|---|---|---|
| `has_state` | `_ => props.has_state = true` (the catch-all) | from the callee and the arguments |
| `has_call` | `_ => false` | from the callee and the arguments |
| `has_member` | `_ => false` | ditto |
| `has_await` | `_ => false` | ditto |

Upstream's `2-analyze/visitors/NewExpression.js` sets no metadata flag of its own — it ends in `context.next()`, so every flag comes from the children. The fix adds that arm to `analyze_props_json` and folds `NewExpression` into the `CallExpression` arms of `has_call_json` / `has_member_json` / `has_await_json`, which already walk `callee` + `arguments`.

Measured consequences, all against the official compiler as the oracle:

```svelte
<script>let s = "x"; void s;</script>
{new String(s)}
```
```js
// before      $.template_effect(() => $.set_text(text, new String(s)));
// official    text.nodeValue = new String(s);
```

```svelte
<script>function f() { return "f"; } void f;</script>
{new String(f())}
```
```js
// before      $.template_effect(() => $.set_text(text, new String(f())));
// official    $.template_effect(($0) => $.set_text(text, $0), [() => new String(f())]);
```

The second row is the opposite direction of the first, and a fix that only stops over-reporting `has_state` leaves it wrong — which is why both are pinned.

`has_reactive_state_json` and the typed `typed_has_reactive_state` already had correct `NewExpression` arms; only the combined single-pass walk and the three single-flag walks were missing one.

## Server

Checked per the #3027 rule. The server transform has **no port of this function** — `crates/rsvelte_core/src/compiler/phases/3_transform/server/` contains no `has_call` / `has_reactive_state` / `analyze_props` and no `NewExpression` reference at all, because SSR does not classify reactivity. Confirmed empirically as well: a 140-case × 3-target grid (7 hosts × 10 expression kinds × legacy/runes) diverged in **0** server cells before and after the change.

## Measurement

Grid of 7 hosts (`root`, `element_child`, `in_if`, `in_each`, `in_key`, `in_snippet`, `in_await`) × 10 expression kinds × legacy/runes × {client, client-dev, server} = 420 cells, compared byte-for-byte against `submodules/svelte`'s compiler:

- before: **96** divergent cells
- after: **8** — all of them `{(n, s)}` in a concatenated interpolation, which is #3323 and lands separately

## Regression tests

- `crates/rsvelte_core/tests/new_expression_metadata_3322.rs` — 4 tests. Negative control run on the reverted tree: 3 fail, and the fourth (a reassigned `$state` argument, which must **stay** reactive) passes in both trees, which is what keeps the fix from reading as "a `new` is never reactive".
- `compatibility/pattern-corpus/issues/3322-*.svelte` — the same four points as corpus entries, all `0 divergent` on client / client-dev / server.


baseballyama:

Issue #3411, opened 2026-08-22
Legacy `$:` compound assignment: the dependency thunk loses its parentheses


baseballyama:
The dependency thunk of a `legacy_pre_effect` loses a pair of parentheses when
the `$:` body is a compound assignment.

```svelte
<script>
	let q;
	$: q += 1;
</script>
<p>{q}</p>
```

```js
// rsvelte
$.legacy_pre_effect(() => $.get(q), () => {
	$.set(q, $.get(q) + 1);
});

// official
$.legacy_pre_effect(() => ($.get(q)), () => {
	$.set(q, $.get(q) + 1);
});
```

Semantically identical; byte-different, so only output equality sees it.

## Population

Found while measuring #3272 on a generated grid. After #3272's boundary fix,
**14 of the 32 residual divergent cells are this** — the `$: q += 1;` row, for
every one of the 14 trailing forms including `nothing` and `next_line`.

That is the evidence it is unrelated to #3272: it reproduces with nothing else on
the line, and reproduced identically on the pre-#3272 binary. The other bodies in
the grid (`$: q = 1`, `$: q++`, `$: f()`, `$: { … }`, `$: if (…) { … }`,
`$: ({a: q} = o)`, `$: [q] = arr`, `$: o.a = 1`, `$: q = c ? 1 : 2`) all match, so
the divergence is specific to how the compound-assignment case builds its
dependency list.

Issue #3410, opened 2026-08-22
Legacy `$:`: an empty statement (`;;`) survives into the client output


baseballyama:
A `;` that is not part of any statement — an `EmptyStatement` — survives into the
client output, where official drops it.

```svelte
<script>
	let q;
	$: q = 1;
	;;
</script>
<p>{q}</p>
```

```js
// rsvelte                              // official
$.legacy_pre_effect(() => {}, () => {   $.legacy_pre_effect(() => {}, () => {
	$.set(q, 1);                            $.set(q, 1);
});                                     });

;;                    // <- kept
$.legacy_pre_effect_reset();            $.legacy_pre_effect_reset();
```

The output parses and runs; only the text differs, so no gate but output equality
can see it.

## Population

Found while measuring #3272 on a generated grid (`$:` body shape × what shares
its line). After #3272's boundary fix, **21 of the 32 residual divergent cells are
this**: one per `$:` body shape, plus one per leading form — every cell whose
trailing form is `;;`.

It is **not** a boundary defect and is independent of #3272: it reproduces with
the `;;` on its own line, and it reproduced identically on the pre-#3272 binary.
The `$:` is not needed either in principle; it is simply what the grid varied
around.

## Where to look

The client instance-script pipeline emits each accumulated line; an empty
statement is a line whose only content is `;`. Upstream never emits one because
it rebuilds the body from the AST, where an `EmptyStatement` has no printer
output.

PR #3412, opened 2026-08-22
fix(compiler): take a legacy `$:` boundary from the parser, and count every ECMAScript line terminator

### LINKED ISSUE #3272: Legacy `$:`: a statement on the same line breaks the boundary scan — 81 unparseable outputs, plus silently dropped and silently added reactivity
A legacy `$:` statement's boundary is found by a scan that assumes the `$:`
occupies its own line. Put anything else on that line — before it or after it —
and the boundary lands in the wrong place. Three different wrong outputs come out
depending on which side, and **one of the three is not JavaScript**.

## 1. `$: <assignment>` followed by a statement → unparseable output

```svelte
<script>
	let q;
	$: q = 1; void q;
</script>
<p>{q}</p>
```

```js
// rsvelte
	$.set(q, 1; void $.get(q));      // <- `;` inside an argument list
});

// official
void $.get(q);
	$.set(q, 1);
});
```

The following statement is spliced **inside the `$.set(…)` call**. Confirmed
against acorn: **81 distinct cases produce output no JS parser accepts, and 0 do
on official's side.** Nothing downstream can consume the file.

## 2. `$:` with a non-assignment body → the next statement is pulled into the effect

```svelte
<script>
	let q; function f() {}
	$: f(); q = 2;
</script>
```

```js
// rsvelte                            // official
$.legacy_pre_effect(() => {}, () => { $.legacy_pre_effect(() => {}, () => {
	f();                                  f();
	$.set(q, 2);   // <- swallowed      });
});                                   $.set(q, 2);
```

`q = 2` was a one-time initialisation and is now re-run on every dependency
change. **The output parses and the component runs**; only the behaviour is
different. Reproduced for every non-assignment body: bare expression, `$: { … }`,
`$: if (…) { … }`, `$: ({a: q} = …)`, `$: [q] = …`, `$: q++`.

## 3. A statement *before* the `$:` on the same line → the reactivity is dropped

```svelte
<script>
	let q;
	let z = 1; void z; $: q = 1;
</script>
```

```js
// rsvelte                          // official
let z = 1;                          let z = 1;
void z;                             void z;
$: $.set(q, 1);   // <- a bare JS   $.legacy_pre_effect(() => {}, () => {
                  //    label           $.set(q, 1);
                                    });
```

There is no `legacy_pre_effect` at all — the assignment runs once at init and is
never reactive again. The output parses, because a labelled statement is legal
JavaScript. This is the shape CLAUDE.md records for #2598 ("emitted a bare `$:`
labelled statement that every parser accepts"), reproduced from a different
input.

## Population

`$:` body shape × what follows on the same line, and what precedes × what
follows — **602 cells, 586 compiled on both sides, 231 divergent**, 0 error
divergences.

| | divergent |
|---|---|
| `body_*` (13 body shapes × 14 trailing forms) | 122 |
| `before_*` (7 leading forms × 14 trailing forms) | 105 |
| `control_plain` | 4 |

Trailing forms that are clean for every body: `nothing`, `next_line`,
`semi_only`, `empty_semis`. So one statement per line — which is what every real
source does, and why the corpus is at zero here — is the only layout that works.

### Negative control

The same 21 layouts with a plain `q = 1;` in place of the `$:` are **0 divergent
over the 34 cells that contain neither a `$:` nor a trailing comment**. The
control is not stuck at zero: 4 of its cells do diverge, and both causes are
identifiable and separate from this one —

- `after_another_reactive` / `before_another_reactive` still contain a `$:`
  (they were built to place a *second* one), so they are not controls;
- `after_line_comment` / `after_block_comment` diverge with **no `$:` anywhere**:
  `q = 1; // trailing` emits `$.set(q, 1 // trailing\n);`. That is a separate
  defect, adjacent to #3251 but with a wider population — #3251 is a comment on
  its own line after a `$:`, and this one needs neither. Commented there rather
  than filed twice.

The discriminating pair is `$: q = 1; void q;` (unparseable) against
`q = 1; void q;` (byte-identical) — same layout, same trailing statement, and the
only difference is the two characters `$:`.

## Why no gate sees it

Both #2 and #3 **parse**, so the parse gate is blind to them; only output
equality reports them. And the input shape — more than one statement on a line
containing a `$:` — does not occur in formatted source, which is what every
corpus repository ships. This is the `script_text` scanner class CLAUDE.md
already describes ("an RHS ends at `;`, `,` or an unbalanced closer"), reached
from the leading side as well as the trailing one.

### LINKED ISSUE #3278: U+2028 / U+2029 do not terminate a `//` comment, and CR / U+2028 / U+2029 break the legacy `$:` line scan
U+2028 (LINE SEPARATOR) and U+2029 (PARAGRAPH SEPARATOR) end a line in
ECMAScript, so they terminate a `//` comment. rsvelte's script handling does not
treat them as terminators, and swallows the code that follows into the comment.

```svelte
<script>
// c<U+2028>const a = 1;
</script>
<p>{a}</p>
```

```js
// rsvelte                    // official
// c const a = 1;        // c
                              const a = 1;
```

The declaration disappears from the emitted script. U+2029 reproduces it
identically; a plain `\n` in the same position is **byte-identical**, which is
the control.

## The same terminator gap reaches the legacy `$:` scanner

```svelte
<script>\rlet a = 1;\r$: b = a + 1;\r</script>\r<p>{b}</p>
```
(CR-only line endings — classic-Mac style, still produced by some tooling.)

```js
// rsvelte                       // official
$: $.set(b, a + 1);              $.legacy_pre_effect(() => {}, () => {
                                     $.set(b, a + 1);
                                 });
```

A bare `$:` label reaches the output and the statement is never reactive — the
same wrong output as #3272, reached through a different trigger: there the `$:`
shared a **line** with another statement, here it is alone on its line but the
line is separated by CR / U+2028 / U+2029 rather than LF. U+2028 and U+2029
reproduce it; LF is identical. Fixing #3272's boundary arithmetic will not cover
this unless the scanner's notion of "line terminator" is widened at the same
time, so the two are noted together.

## Measured clean beside it

From the same grid (94 line-terminator cells, 88 compiled on both sides): CRLF
throughout, LF throughout, a terminator inside a string literal, inside a
template literal, inside markup text, splitting an attribute list, splitting a
`<style>` block, a file with no trailing newline, a script with no trailing
newline, an empty file, a file that is only a newline, a file that is only a CR,
and a file mixing CRLF/CR/LF are all byte-identical.

baseballyama:
Closes #3272
Closes #3278

Both issues are the same question asked twice — **where does a statement end** — so
they land together: #3278's `$:` half is the same cut as #3272's, and its comment
half is the same "what ends a line" answer one layer down in the printer.

## What was wrong

**1. A legacy `$:` sharing its physical line (#3272).** The client instance-script
pipeline reads one physical line as one statement. Anything on the `$:`'s line
therefore landed on the wrong side of the reactive boundary, three different ways:

| input | rsvelte before | official |
|---|---|---|
| `$: q = 1; void q;` | `$.set(q, 1; void $.get(q))` — **not JavaScript** | `void $.get(q);` then the effect |
| `$: f(); q = 2;` | `q = 2` swallowed into the effect body — **reactivity added** | `$.set(q, 2)` before the effect |
| `let z = 1; void z; $: q = 1;` | a bare `$: $.set(q, 1)` label — **reactivity dropped** | `$.legacy_pre_effect(...)` |

**2. CR / U+2028 / U+2029 (#3278).** The same split reads a line as `\n`-delimited
text, so a `$:` alone on a CR- or separator-terminated line arrived glued to its
neighbour and produced the third row above. And the printer answers "do this
comment and the next node share a line?" the same `\n`-only way, while esrap reads
that off acorn's `loc.line`, which advances on CR and U+2028 / U+2029 too — so a
`//` comment terminated by one of those absorbed the statement after it, which
then **disappeared from the output** on client and server alike.

## The fix

- `separate_same_line_legacy_export_declarations` already cut a same-line pair
  using the parser's top-level statement spans; it becomes
  `separate_same_line_top_level_statements` and also cuts on either side of a `$`
  label. A gap holding no `\n` is what makes a pair same-line, so CR / U+2028 /
  U+2029 cut as well. Formatted source has a `\n` in every gap, so the pass
  returns `Cow::Borrowed` and changes nothing there. When it does cut, the
  retained program and the source projection built for the caller's text no
  longer describe this one, so both are dropped for the rest of the pass.
- `rsvelte_esrap`'s two "is there a line break between these offsets" predicates
  (`line_starts` and the `has_newline` text scan) now count every ECMAScript
  `LineTerminator`.

Boundaries come from the parser, never from a semicolon scan.

## Measurement

A generated grid — 13 `$:` body shapes × 14 same-line trailing forms, plus 7
leading forms × the same 14 — compiled with both compilers, plus an acorn parse
oracle on both sides and a 15-cell line-terminator axis:

| | before | after |
|---|---|---|
| divergent (of 270 compared) | 218 | **32** |
| rsvelte output no JS parser accepts | 54 | **0** |
| official output no JS parser accepts | 0 | 0 |
| line-terminator axis divergent (of 15) | 9 | **0** |
| negative control (same layouts, plain `q = 1;` instead of `$:`) | 0 of 78 | 0 of 78 |

The six repros from the two issues are now **byte-identical to official** on the
client; the server was already identical and is the control.

**The 32 residual cells are two pre-existing defects that have nothing to do with
layout**, each confirmed against the pre-fix binary with a one-statement-per-line
repro:

- 21 cells: a trailing `;;` leaves stray empty statements official drops —
  reproduces with the `;;` on its own line, filed as #3410.
- 14 cells (`$: q += 1`): the dependency thunk prints
  `() => $.get(q)` where official prints `() => ($.get(q))` — reproduces with
  nothing else on the line, filed as #3411.

## Tests

`crates/rsvelte_core/tests/legacy_reactive_same_line_3272.rs` and
`crates/rsvelte_core/tests/script_line_terminators_3278.rs`, every expectation
measured against the official compiler on the same source.

Negative controls, taken separately for each of the two fixes:

- reverting the statement separator fails exactly the three #3272 tests and
  #3278's reactive test, each with its predicted message (`spliced into the
  argument list` / `pulled into the effect body` / `the reactive wrapper was
  dropped`); the plain-assignment control and the server control stay green.
- reverting the esrap change fails exactly `a_line_comment_ends_at_a_separator`
  with the predicted `// c const a = 1;`; the LF control and every #3272 test
  stay green.

Suites run locally: `compiler_fixtures`, `sourcemaps_gate`, `runtime`, `ssr`,
`validator`, `print`, `client_transforms_pin`, `client_script_comments_pin`, and
the whole `rsvelte_esrap` suite — all green.

---

## Revision (3rd commit): the esrap fix is contained to comment placement

The second commit widened `printer::line_starts` itself to count every ECMAScript
`LineTerminator`. **That was wrong, and @legD caught it: the same table is the
source-map coordinate table.** The third commit replaces it with a contained fix.

### Call-site audit of `printer::line_starts`

Two disjoint sets of readers:

| reader | decides |
|---|---|
| `line_of` → `has_newline_between`, `comment_starts_on_earlier_line` | comment **placement** |
| `source_map_line_starts()` → `command::flatten_with_map`; `offset_to_line_col` (`printer.rs:757`); the two column computations (`printer.rs:810`, `889`) | source-map **coordinates** |

The coordinate readers take `map_line_starts.as_deref().unwrap_or(&self.line_starts)`
and `print_with_map` never sets `map_line_starts`, so there it *is* `line_starts`.
`print_split` does set it — but builds it with the same `printer::line_starts`
(`lib.rs:191`) from `map_source`, which for a component is the Svelte source text
(`3_transform/client/mod.rs:2554`). Both map paths were in scope of the 2nd commit.

### The coordinate path

Official counts original positions with **LF alone**. @legD decoded `mappings` and
counted which original lines are referenced, for
`<script>\nlet aaa = 1;{T}let bbb = 2;\n</script>\n<p>{aaa}{bbb}</p>`:

| `{T}` | official | rsvelte on `main` | rsvelte, 2nd commit | rsvelte, this PR now |
|---|---|---|---|---|
| `\n`, `\r\n` | 1,2,3,4,5 | 1,2,3,4,5 | 1,2,3,4,5 | 1,2,3,4,5 |
| `\r`, U+2028, U+2029 | **1,2,3,4** | 1,2,3,4 | **1,2,3,4,5** | **1,2,3,4** |

The last column holds **by construction, not by measurement**: the body of
`line_starts` in this branch is byte-identical to `main` (`diff` of the function
between `origin/main` and `HEAD` is empty — only its doc comment changed, to record
why it counts LF alone), and nothing else the coordinate readers touch was
modified. There is no input on which the map can move.

`sourcemaps_gate` was **green through the 2nd commit as well** (3 passed, 1 ignored).
It asserts structural invariants rather than equality with official, and a uniform
line shift satisfies every one of them — **the gate is not a negative control for
this**, which is why the table above is in the PR body.

### What replaced it

`Printer` gains `placement_source: Option<&'opt str>`, threaded from the
`comment_source` / `source` the callers already hold, and **only** the two placement
predicates consult it (`placement_text()` = `comment_source.or(placement_source)`),
scanning for any `LineTerminator`. @legD reached the same design independently.

### Still fixed, and what is not

The same six line-terminator cells pass, the LF control stays green, and the three
`#3272` tests are unaffected (suites re-run on the contained version:
`rsvelte_esrap` 55, `sourcemaps_gate`, `print`, `compiler_fixtures`,
`compile_both_parity`, plus both new suites).

@legD's grid separately measured **8 residual cells** of the shape
`let a = 1; // c{T}$: b = a + 1;`. They diverge on `main` too and the server half
diverges under LF as well, so they are comment *attachment*, not line termination —
filed as #3469, out of scope here.

### DIFF
--- a/.changeset/legacy-reactive-boundary-line-terminators.md
+++ b/.changeset/legacy-reactive-boundary-line-terminators.md
@@ -1,0 +1,17 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Fix the legacy `$:` boundary when a statement shares its line, and treat CR / U+2028 / U+2029 as line terminators
+
+The client instance-script pipeline read one physical line as one statement, so a
+`$:` sharing its line with another statement put the boundary in the wrong place —
+splicing the next statement into the `$.set(...)` call (output no JS parser
+accepts), swallowing it into the effect body, or dropping the `legacy_pre_effect`
+wrapper and emitting a bare `$:` label. Top-level statement boundaries now come
+from the parser.
+
+The same "line" notion was `\n`-only in two places: that split, and the printer's
+decision about whether a comment and the node after it share a line. A `//`
+comment terminated by CR / U+2028 / U+2029 therefore absorbed the statement that
+followed it, which disappeared from the output.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2642,7 +2642,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_esrap"
-version = "0.10.32"
+version = "0.10.33"
 dependencies = [
  "compact_str 0.10.0",
  "memchr",
--- a/crates/rsvelte_core/Cargo.toml
+++ b/crates/rsvelte_core/Cargo.toml
@@ -94,7 +94,7 @@ oxc_ast = { version = "0.145", features = ["serialize"] }
 oxc_span = "0.145"
 oxc_diagnostics = "0.145"
 oxc_codegen = "0.145"
-rsvelte_esrap = { version = "=0.10.32", path = "../rsvelte_esrap" }
+rsvelte_esrap = { version = "=0.10.33", path = "../rsvelte_esrap" }
 oxc_syntax = "0.145"
 oxc_semantic = "0.145"
 
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -5097,13 +5097,23 @@ fn starts_with_else_keyword(line: &str) -> bool {
     })
 }
 
+/// A top-level `$:` label, which is what the legacy pipeline rewrites.
+fn is_legacy_reactive_label(statement: &oxc_ast::ast::Statement<'_>) -> bool {
+    matches!(
+        statement,
+        oxc_ast::ast::Statement::LabeledStatement(labeled) if labeled.label.name == "$"
+    )
+}
+
 /// Makes an AST statement boundary visible to the legacy per-line pipeline.
 ///
-/// The pipeline historically treated each physical line as one statement. That
-/// lets an `export let` rewrite consume a following declaration on the same
-/// line. We use the parser's top-level statement spans rather than scanning
-/// for semicolons, whose grammar is precisely what this path must not infer.
-fn separate_same_line_legacy_export_declarations<'a>(
+/// The pipeline treats each physical line as one statement, and it reads a line
+/// as `\n`-delimited text. So two top-level statements reach it as one whenever
+/// they share a line — and also when the only thing between them is a CR or a
+/// U+2028 / U+2029, which end a line in ECMAScript but not in `str::lines`. We
+/// use the parser's top-level statement spans rather than scanning for
+/// semicolons, whose grammar is precisely what this path must not infer.
+fn separate_same_line_top_level_statements<'a>(
     script: &'a str,
     is_typescript: bool,
 ) -> Cow<'a, str> {
@@ -5111,7 +5121,9 @@ fn separate_same_line_legacy_export_declarations<'a>(
     use oxc_parser::Parser;
     use oxc_span::{GetSpan as _, SourceType};
 
-    if !script.contains("export let ") && !script.contains("export var ") {
+    let has_export = script.contains("export let ") || script.contains("export var ");
+    let has_label = memmem::find(script.as_bytes(), b"$:").is_some();
+    if !has_export && !has_label {
         return Cow::Borrowed(script);
     }
 
@@ -5132,8 +5144,11 @@ fn separate_same_line_legacy_export_declarations<'a>(
         let next = pair[1].span();
         let previous_text = &script[previous.start as usize..previous.end as usize];
         let gap = &script[previous.end as usize..next.start as usize];
-        if (previous_text.trim_start().starts_with("export let ")
-            || previous_text.trim_start().starts_with("export var "))
+        let export_declaration = previous_text.trim_start().starts_with("export let ")
+            || previous_text.trim_start().starts_with("export var ");
+        if (export_declaration
+            || is_legacy_reactive_label(&pair[0])
+            || is_legacy_reactive_label(&pair[1]))
             && !gap.contains('\n')
             && gap.chars().all(char::is_whitespace)
         {
@@ -5639,8 +5654,12 @@ fn transform_instance_script_for_visitors(
     if script.is_empty() {
         return String::new();
     }
-    let separated_script =
-        separate_same_line_legacy_export_declarations(script, analysis.is_typescript);
+    let separated_script = separate_same_line_top_level_statements(script, analysis.is_typescript);
+    // A cut moves every byte after it, so the spans and the projection that were
+    // built for the caller's text no longer describe this one.
+    let separated = matches!(separated_script, Cow::Owned(_));
+    let retained_program = retained_program.filter(|_| !separated);
+    let source_projection = source_projection.filter(|_| !separated);
     let script = separated_script.as_ref();
     let original_script = script;
 
--- a/crates/rsvelte_core/tests/legacy_reactive_same_line_3272.rs
+++ b/crates/rsvelte_core/tests/legacy_reactive_same_line_3272.rs
@@ -1,0 +1,115 @@
+//! Issue #3272 — a legacy `$:` statement that shares its physical line.
+//!
+//! The client instance-script pipeline reads one physical line as one
+//! statement, so anything written before or after a `$:` on the same line
+//! landed on the wrong side of the reactive boundary: the following statement
+//! was spliced *inside* the `$.set(...)` call (output no JS parser accepts),
+//! or swallowed into the effect body (reactivity silently added), or the
+//! reactive wrapper was dropped and a bare `$:` label reached the output
+//! (reactivity silently removed).
+//!
+//! Every expectation below was measured against the official compiler
+//! (`submodules/svelte/.../compiler/index.js`) on the same source. The server
+//! target is pure AST and was already byte-identical on all of these, so it is
+//! the control.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+fn compile_to(source: &str, generate: GenerateMode) -> String {
+    compile(
+        source,
+        CompileOptions {
+            filename: Some("A.svelte".to_string()),
+            generate,
+            ..Default::default()
+        },
+    )
+    .expect("compile failed")
+    .js
+    .code
+}
+
+/// A statement after a `$: <assignment>` on the same line.
+const TRAILING_AFTER_ASSIGNMENT: &str =
+    "<script>\n\tlet q;\n\t$: q = 1; void q;\n</script>\n<p>{q}</p>\n";
+
+/// A statement after a `$:` whose body is not an assignment.
+const TRAILING_AFTER_CALL: &str =
+    "<script>\n\tlet q; function f() {}\n\t$: f(); q = 2;\n</script>\n<p>{q}</p>\n";
+
+/// A statement before the `$:` on the same line.
+const LEADING_BEFORE_REACTIVE: &str =
+    "<script>\n\tlet q;\n\tlet z = 1; void z; $: q = 1;\n</script>\n<p>{q}</p>\n";
+
+#[test]
+fn a_statement_after_a_reactive_assignment_is_not_spliced_into_the_set_call() {
+    let out = compile_to(TRAILING_AFTER_ASSIGNMENT, GenerateMode::Client);
+    // Official: `void $.get(q);` stands on its own, before the effect.
+    assert!(
+        out.contains("$.set(q, 1);"),
+        "the reactive assignment lost its own boundary:\n{out}"
+    );
+    assert!(
+        out.contains("void $.get(q);"),
+        "the trailing statement did not survive as a statement:\n{out}"
+    );
+    assert!(
+        !out.contains("$.set(q, 1; void"),
+        "the trailing statement was spliced into the argument list:\n{out}"
+    );
+}
+
+#[test]
+fn a_statement_after_a_reactive_call_is_not_swallowed_by_the_effect() {
+    let out = compile_to(TRAILING_AFTER_CALL, GenerateMode::Client);
+    let set = out.find("$.set(q, 2)").expect("the assignment is missing");
+    let effect = out
+        .find("$.legacy_pre_effect(")
+        .expect("the reactive effect is missing");
+    // Official emits the one-time assignment before the effect; inside it, the
+    // assignment would re-run on every dependency change.
+    assert!(
+        set < effect,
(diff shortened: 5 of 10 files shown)


baseballyama:
I worked #3272 / #3278 in parallel by a coordination mistake on our side — this PR
is the one that ships. Our diagnoses agree (two sites: the client
instance-script line scan, and `rsvelte_esrap`'s comment placement), so this is
only the part where the two fixes differ, plus the measurements I collected that
may be useful here.

## One thing to check before merge: `line_starts` is also the source-map table

This PR widens `printer::line_starts` itself to the full LineTerminator set. That
function is not only the comment-placement table — it is the table original
source-map positions are resolved against:

- `lib.rs:191` — `let map_line_starts = map_source.map(printer::line_starts)`
- `printer.rs:757 / 810 / 889` — `offset_to_line_col` and the two column
  computations read `map_line_starts`, falling back to `line_starts`
- `3_transform/client/mod.rs:2554` calls `print_split(..., map_source, ...)` with
  `map_source = source`, i.e. **the Svelte component source**

So after this change, a component containing a lone `\r`, U+2028 or U+2029
resolves its original positions against a table with more lines in it.

**Measured against the official compiler** (decode `compile(src, {generate:
'client'}).js.map.mappings`, collect the distinct original lines any segment
points at), for
`<script>\nlet aaa = 1;{T}let bbb = 2;\n</script>\n<p>{aaa}{bbb}</p>`:

| `{T}` | original lines official's map references |
|---|---|
| `\n` | 1,2,3,4,5 |
| `\r\n` | 1,2,3,4,5 |
| `\r` | 1,2,3,4 |
| U+2028 | 1,2,3,4 |
| U+2029 | 1,2,3,4 |

Official counts **only `\n`** for original positions — the two statements stay on
one original line for `\r` / U+2028 / U+2029. esrap's *comment placement* reads
acorn `loc`, which does advance on all four (your comment is right about that),
but the map's original side does not follow it. `sourcemaps_gate.rs` /
`compatibility/sourcemap-known-failures.json` is a shrink-only ratchet over this
area, so a change here wants to be deliberate.

**Confirmed on the rsvelte side too**, same probe against
`rsvelte_core::compile(...).js.map`:

| `{T}` | rsvelte today / with `placement_source` | rsvelte with this PR's `line_starts` |
|---|---|---|
| `\n` | 1,2,3,4,5 | 1,2,3,4,5 |
| `\r\n` | 1,2,3,4,5 | 1,2,3,4,5 |
| `\r` | 1,2,3,4 | **1,2,3,4,5** |
| U+2028 | 1,2,3,4 | **1,2,3,4,5** |
| U+2029 | 1,2,3,4 | **1,2,3,4,5** |

I produced the right-hand column by applying only this PR's `line_starts` body to
an otherwise unchanged tree, so nothing else is confounded. Today rsvelte agrees
with official on all five; after this change it disagrees on three. And the map
gate asserts that rsvelte's map is *structurally well-formed*, calibrated against
official rather than compared to it, so a uniform line shift is exactly the kind
of divergence it can stay green through.

If that turns out to matter, the containment I used was to leave `line_starts`
alone and add a `placement_source: Option<&'opt str>` to `Printer`, set from
`print_split`'s existing `comment_source` argument (it is already in scope, just
not retained) and from `print_with_map`'s `source`. Then only
`has_newline_between` / `comment_starts_on_earlier_line` scan the text for the
full terminator set; every coordinate path keeps its table. It is a smaller blast
radius for the same six cells. Happy to send it as a follow-up if you'd rather
not respin this PR.

## Measurements that may be worth reusing

A 26-cell grid — 5 line terminators × 5 comment/statement layouts + a control —
compared official vs rsvelte on **both** targets:

- before: `match=32 diff=20`
- after the two fixes: `match=44 diff=8`

The 15-case #3272/#3278 repro set × 2 targets: `match=18 diff=12 unparseable=1`
→ **`match=30 diff=0 unparseable=0`**.

All 8 remaining cells are one shape, `let a = 1; // c{T}$: b = a + 1;`, and they
are **not** a residue of this work — they diverge identically on `main` before
either fix, and the server half diverges on plain LF too, so the terminator axis
only carries it. It is a comment-*attachment* difference (upstream hoists the
`let b;` declaration and flushes the pending comment at it; rsvelte emits the
declaration first and the comment is still pending at the `$:`). Filed as #3469
so it is not rediscovered as a regression of this PR.

## The negative control split three ways, which is the reusable part

Reverting **only** the client-scan half of the fix failed 3 of my 6 regression
tests, and each failed through a *different* assertion:

| assertion | what it caught |
|---|---|
| `assert_parses` (oxc) | output no JS parser accepts |
| `contains("$.legacy_pre_effect(")` | reactivity silently dropped — output parses and runs |
| `assert_eq!` vs official's bytes | the terminator half — output parses, runs, and is reactive |

Reverting only the esrap half failed the other 2, and 4 passed. Worth stating
because it is the concrete case for why a parse oracle and byte equality are both
required: neither one alone sees more than a third of the defect, and the middle
row is invisible to every check except comparison with official.

baseballyama:
Follow-up datapoint on the map question above: I applied only this PR's `line_starts` body to an otherwise unchanged tree and ran the map gate — `cargo test -p rsvelte_core --test sourcemaps_gate` is **green** (3 passed, 1 ignored) with the shifted table in place. So the gate does not discriminate here, as expected: it calibrates structural invariants against official rather than comparing the two maps, and a uniform line shift on `\r` / U+2028 / U+2029 sources satisfies every invariant it checks. The decode probe in my previous comment is the discriminating measurement.

baseballyama:
@legD you were right, and the audit backs it: `printer::line_starts` is read both by
`line_of` (placement) and, through `map_line_starts.unwrap_or(&line_starts)`, by
`offset_to_line_col` and the two column computations — and `print_with_map` never
sets `map_line_starts`, so there it *is* the coordinate table.

Replaced in f459cdf: `line_starts` is back to LF only, byte-identical to `main`, and
`Printer` carries a `placement_source` that only `has_newline_between` /
`comment_starts_on_earlier_line` read. I had converged on the same containment from
the call-site audit before your patch arrived — same field name, same two predicates —
so this is two independent derivations of the same fix rather than a hand-off.

The PR body now carries your mappings table and states why the last column holds by
construction rather than by measurement, plus the note that `sourcemaps_gate` was
green through the bad version and so is not a negative control here.

The 8 residual cells are #3469, out of scope.

Issue #3098, opened 2026-08-19
Comments in a template expression tag and at the end of the instance script are dropped


baseballyama:
A comment inside a template expression tag is dropped by rsvelte and kept by official.

```svelte
<script>
	let n = $state(0);
</script>

<p>{/* inline */ n}</p>
```

client:

```js
// rsvelte
$.append($$anchor, p);
// official
$.append($$anchor, p /* inline */);
```

A trailing comment at the end of the instance script is the server-side twin: official flushes it at the next located node, rsvelte drops it.

```svelte
<script>
	let n = $state(0);
	// trailing script comment
</script>

<p>{n}</p>
```

```js
// official, server
	// trailing script comment
}
// rsvelte, server
}
```

Neither is visible to the corpus output gate: byte-different pairs fall through to `ast_equiv_batch`, which is invoked with an empty argv and therefore ignores comments, so both entries score `match`. They show up only under a comment-aware comparison (the matrix and mutation gates), and pattern-corpus is in neither.

baseballyama:
The contribution here is **scope**, not a new case: this issue records the `expr-tag` position,
and measured across template positions the loss is uniform.

Method: **one comment at a time**. With several comments in one file official's cursor places
them differently, so a multi-comment probe cannot attribute a divergence to a position — my
first probe put four comments in one component and came back with `{@const}` and the event
arrow apparently clean on both sides, which is an artifact of the siblings, not a fact about
those positions.

Oracle: `submodules/svelte` @ `20b341f10048` = **5.56.9** with **esrap 2.2.12** (resolved
through the submodule's own `node_modules`, not the repository root's). rsvelte: `main`
c8a67c5aa. Raw byte comparison, no normalization.

Each row is `<script>let v = $state(1);</script>` plus one template line carrying a single
`/* c */`:

| position | target | official | rsvelte | where official puts it |
|---|---|---|---|---|
| `{/* c */ v}` | client | 1 | **0** | `$.append($$anchor, b /* c */);` |
| `{#if /* c */ v}` | server | 1 | **0** | `if (/* c */ v) {` |
| `{#each [/* c */ v] as i}` | client | 1 | **0** | `$.each(node, 17, () => [/* c */ v], $.index, …` |
| `{#each [/* c */ v] as i}` | server | 1 | **0** | `$.ensure_array_like([/* c */ v]);` |
| `{@const c = /* c */ v * i}` | client | 1 | **0** | `($$anchor, i /* c */) => {` |
[...]

Issue #3070, opened 2026-08-19
Comment placement: five classes still diverge in the JS regions (server interior slot, own-line spill, dangling tail, dev tag, template leaks)


baseballyama:
A comment-placement sweep over the JS regions found five classes still
diverging after #3059, #3068 and the follow-ups on this branch. Each is
`comment-mismatch` only — the code is identical — but byte parity is the goal.

## 1. Server: an interior comment before the binding name of a rune-lowered declaration

```svelte
<script>
	let /* c */ x = $state(2);
</script>

<p>{x}</p>
```

official (server): `let /* c */ x = 2;`
rsvelte:           `/* c */ let x = 2;`

Same for `$derived`. A plain `let /* c */ x = 2` (no rune) is correct, and so is
`let x = /* c */ $state(2)` — only the slot *before the name* moves, and only
when the declaration is lowered. Mechanism: `place_on_region` stamps the
re-parsed statement with `Place::Shift(base + reparse_origin - prev_end)`, a
single constant. The lowered text is shorter than the source region (the rune
call collapsed, the comment is carried separately), so an interior offset does
not line up and the comment sorts ahead of the whole statement instead of ahead
of the identifier. A constant shift cannot express a two-edit region; the map
has to be piecewise.

## 2. Server: an own-line comment after a rune argument

```svelte
<script>
	let d = $state.raw(
		[1, 2]
		// trailing
	);
	$effect(() => {});
</script>
```

The client half is fixed (the comment becomes the next statement's leading
comment). The server still glues it to the rewritten statement, and the
`$effect(/* fx */ …)` pre-comment moves out on the server too.

## 3. Server: a dangling comment at the end of an instance script

```svelte
<script module>
	export const K = 1;
</script>

<script>
	let v = $state(K);
	// dangling at the end of instance
</script>

<p>{v}</p>
```

official keeps `// dangling at the end of instance` in the component function's
tail; rsvelte drops it. Server and server-dev only.

## 4. dev: the pre-comments of a tagged `$state` go outside `$.tag(...)`

`let a = /* c */ $state(0)` in dev:
official `let a = $.tag(/* c */ $.state(0), 'a');`, rsvelte
`let a = /* c */ $.tag($.state(0), 'a');`.

## 5. Comments official leaks into generated statements

Upstream's comment cursor puts a mustache's comment into unrelated generated
code — `{/* c */ x}` becomes `var text = $.child(p, /* c */ true);`, a dangling
[...]

PR #3415, opened 2026-08-22
fix(client): keep a rune declarator's leading comment inside the lowered call

baseballyama:
A comment written between a declarator's `=` and its rune call was emitted ahead of the
whole declaration. Upstream never leaves it there — and the four rune forms put it in four
*different* slots, which is why one rule ("hoist it into the wrapper") would have been
wrong for three of them.

Addresses #3070 class 4, plus three sibling classes the same grid found.

## The rule, and where it comes from

The comment prints just before the first node of the **output** that still carries a
source range at or after it. `VariableDeclaration.js` decides which node that is:

| rune | how upstream builds it | where the comment lands |
|---|---|---|
| `$state` / `$state.raw` | `b.id('$.state', init.callee.loc)` — the callee **inherits the source `loc`** (`VariableDeclaration.js:149`) | before `$.state`, i.e. just inside `$.tag(` |
| `$state({…})`, non-reactive | `b.call('$.proxy', value)` — plain string callee, no `loc` | before the **argument**, inside `$.proxy(` |
| `$derived(expr)` | `b.call('$.derived', b.thunk(expr))` — the arrow has no `loc`, its body does | inside the synthesized thunk's **empty parameter parens** |
| `$derived.by(fn)` | `b.call('$.derived', fn)` — the argument is the user's own located function | before that argument |

rsvelte emitted `/* c */ $.tag(…)` / `/* c */ $.derived(…)` for all four.

```svelte
<script>
	let a = /* c */ $state(0);
	let d = /* c */ $derived(a * 2);
</script>
```

```js
// official (dev)                            // rsvelte, before
let a = $.tag(/* c */ $.state(0), 'a');      let a = /* c */ $.tag($.state(0), 'a');
let d = $.tag($.derived((/* c */) => …), 'd');  let d = /* c */ $.tag($.derived(() => …), 'd');
```

## The fix

All in `3_transform/client/ast_state_transform.rs`, using the comment helpers that were
already there (`rune_call_comment_slots`, `flush_trivia_comments`, `maybe_tag_declarator`).
No coordinate-space change, no new scan of the script:

* `declarator_lead_comment_spans` returns the comment run **adjacent** to the call. Any
  code byte restarts the run, so a comment inside a type annotation
  (`let d: Record<string, "/*"> = /* c */ $derived(…)`) is not one of them; string literals
  are skipped so a `/*` inside one cannot open a comment.
* `maybe_tag_declarator_with_lead` places the run inside `$.tag(` and reports whether it
  wrapped, so the caller knows whether its replacement span has to swallow those bytes.
  `$.tag_proxy` deliberately does **not** take it — `$.proxy` has no `loc`, so its caller
  folds the run into the argument text instead.
* The two `$derived` handlers prepend the run to the `pre_spans` they already collect from
  inside the parens, which is the same slot; `$derived(await …)` is left alone, because the
  async form drops the thunk parens the comments would have gone into.

## Measurement

24 declarator shapes × 4 targets = 96 comparisons against the pinned official compiler
(`submodules/svelte`, 5.56.10):

| | diverging |
|---|---|
| merge base | 29 |
| this branch | **4** |

25 fixed, **0 regressions** — checked as a set difference: all 4 remaining are in the
merge base's 29. They are two classes this PR does not claim: a **trailing** comment after
the rune call (`$.state(0 /* c */)` vs `$.state(0); /* c */`, client), and the server's
interior slot for a comma declarator, which is #3070 class 1.

A second grid printed the declarator line for 13 shapes × dev on/off: **26/26 match**,
including the `//` spellings, where esrap breaks `$.tag(` across lines and puts the closing
paren of the thunk on its own line.

`crates/rsvelte_core/tests/rune_declarator_leading_comment_3070.rs` pins all four slots
with every expectation taken from the official compiler's own output, plus four negative
controls that are the reason the four slots had to be separated:

* `let /* c */ a = $state(0)` — a comment before the **name** is not adjacent to the call
  and must not move.
* `let a = $state(/* c */ 0)` — a comment already in the right slot must not move twice or
  duplicate.
* `let o = /* c */ $state(1)` folded to `let o = /* c */ 1` — no wrapper to move into.
* the non-reactive proxy row, which is what falsifies "always put it after the tag's `(`":
  that rule would print `$.tag_proxy(/* c */ $.proxy({ n: 0 }), 'o')` and official prints
  `$.tag_proxy($.proxy(/* c */ { n: 0 }), 'o')`.

Suites run green: `--lib` (1732), `compiler_fixtures`, `ssr`, `runtime`, and the ten
comment/rune-adjacent suites (`client_script_comments_pin`, `client_transforms_pin`,
`class_field_leading_comment_2437`, `class_field_tag_label_2021`,
`async_derived_comment_2755`, `comment_delimiters_in_initializer`,
`comment_before_continuation_decl`, `comma_declaration_split`,
`client_function_body_comment_2527`, `client_effect_trailing_comment_2718`).

## #3070 class 5 — upstream is the one that is wrong, and this records the decision

Class 5 is the row where upstream leaks a template-expression comment into unrelated
generated code. `upstream_issues/3070-svelte-template-comment-leaks-into-generated-code.md`
carries the report. The mechanism is three upstream decisions composing, and I traced all
three rather than describing the symptom:

1. `transform-client.js:394` `component_block.loc = instance.loc` — literally commented
   *"trick esrap into including comments"* — keeps one cursor live over the **file's**
   comment list for the whole component body, template comments included.
2. `shared/fragment.js:47-52,113-117` builds each generated element binding as
   `b.id(scope.generate(name), node.name_loc)`, so `button` reports itself as living at the
   `<button` **tag name**.
3. esrap's `CallExpression` writes the argument separator *before*
   `flush_trailing_comments(arg.loc?.end, next)` (`esrap@2.2.12`, `ts/index.js:611` then
   `:617`), and that flush keys on the comment's **line** matching `arg.loc.end.line`.

So `{/* c */ n}` becomes `$.child(button, /* c */ true)`; move the `<p>` to the next line
and the same comment is **dropped instead**, because it no longer shares a line with a tag
name. That line-move is the discriminating case, and the server rows are the control — the
same comments land beside their own expression there, so this is a client-codegen artifact
rather than how Svelte treats comments in expressions.

Per the standing rule (reproduce an upstream defect when the output runs; do not when the
output is not JavaScript) the default here is **reproduce** — the output is valid JS. The
write-up records that intent and splits the work by what is actually reachable:

* **Server rows: reproduce, and they are a plain rsvelte defect.** That is #3098's template
  half; cmtA's #3326 already built the `ChunkRegistry` machinery it needs.
* **Client rows: blocked on a prerequisite, not declined.** rsvelte's client comment space
  is the generated script text —
  `js_ast/to_oxc.rs::into_coordinate_free_program` clears `program.comments` outright
  because "a comment's position is a coordinate too", and
  `client/visitors/expression_converter.rs` rebuilds template expressions with no comment
  channel at all. There is nothing to place yet. So the write-up states the four-step rule
  to implement once the comment stream exists, so it is a lookup rather than a
  re-derivation.

## Not fixed

* #3070 classes 1-3 (server interior slot, own-line spill, dangling instance tail) — cmtA's
  #3326 territory; this branch touches none of its files.
* #3070 class 5 client rows — above.
* A **trailing** comment after a rune call (`$.state(0 /* c */)`). Diverges on the merge
  base too; this change does not widen it.

## Ratchets

None touched. Comment-only divergences are invisible to the corpus output gate
(`ast_equiv_batch` runs with an empty argv and ignores comments), so `--update-baseline`
was not run. If the matrix gate's `comment-mismatch` verdict moves, I will re-baseline off
a fresh `origin/main`.

Refs #3070 — this closes class 4 and three siblings; classes 1-3 and 5 stay open (see
*Not fixed*), so the issue is deliberately not auto-closed.

### DIFF
--- a/.changeset/rune-declarator-lead-comment.md
+++ b/.changeset/rune-declarator-lead-comment.md
@@ -1,0 +1,8 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Keep a comment written between a declarator's `=` and its rune call inside the lowered
+call, the way the official compiler places it: inside `$.tag(...)` for `$state`, inside
+`$.proxy(...)` for a non-reactive proxied initializer, inside the synthesized thunk's
+parameter parens for `$derived(expr)`, and before the argument for `$derived.by(fn)`.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
@@ -905,6 +905,52 @@ impl<'a, 's> StateVarCollector<'a, 's> {
         None
     }
 
+    /// The comment run written between a declarator's binding and its rune
+    /// call. Any code byte restarts the run, so a comment inside a type
+    /// annotation — which is not adjacent to the call — is not one of them.
+    ///
+    /// Upstream never leaves these ahead of the declarator: the lowered call
+    /// either inherits the source callee's `loc` (`$state`) or hands the
+    /// argument to a node esrap flushes them inside, so they end up within the
+    /// wrapper rather than before it.
+    fn declarator_lead_comment_spans(&self, from: u32, to: u32) -> Vec<(u32, u32)> {
+        let bytes = self.source.as_bytes();
+        let mut i = from as usize;
+        let end = (to as usize).min(bytes.len());
+        let mut run: Vec<(u32, u32)> = Vec::new();
+        while i < end {
+            match bytes[i] {
+                b if b.is_ascii_whitespace() => i += 1,
+                b'/' if bytes.get(i + 1) == Some(&b'/') && i + 1 < end => {
+                    let stop = memchr::memchr(b'\n', &bytes[i..end]).map_or(end, |p| i + p);
+                    run.push((i as u32, stop as u32));
+                    i = stop;
+                }
+                b'/' if bytes.get(i + 1) == Some(&b'*') && i + 1 < end => {
+                    let Some(close) = memchr::memmem::find(&bytes[i + 2..end], b"*/") else {
+                        return Vec::new();
+                    };
+                    let stop = i + 2 + close + 2;
+                    run.push((i as u32, stop as u32));
+                    i = stop;
+                }
+                quote @ (b'\'' | b'"' | b'`') => {
+                    run.clear();
+                    i += 1;
+                    while i < end && bytes[i] != quote {
+                        i += if bytes[i] == b'\\' { 2 } else { 1 };
+                    }
+                    i += 1;
+                }
+                _ => {
+                    run.clear();
+                    i += 1;
+                }
+            }
+        }
+        run
+    }
+
     /// Same-line trailing comments (esrap's `flush_trailing_comments`): each is
     /// emitted after a space; a `//` comment forces a newline so it cannot
     /// swallow what the caller appends. Comments past the first newline are not
@@ -1020,16 +1066,37 @@ impl<'a, 's> StateVarCollector<'a, 's> {
     /// no longer has to walk the script in dev mode, eliminating one
     /// O(text_len) buffer pass per component.
     fn maybe_tag_declarator(&self, var_name: &str, replacement: String) -> String {
+        self.maybe_tag_declarator_with_lead(var_name, replacement, "")
+            .0
+    }
+
+    /// [`Self::maybe_tag_declarator`], with the declarator's leading comment run
+    /// placed just inside the tag call and a flag saying whether a wrap
+    /// happened — the caller needs it to decide whether its replacement span
+    /// has to swallow those comments.
+    ///
+    /// `lead` only belongs inside the wrap for `$.state(`: upstream builds that
+    /// callee with the source `$state` callee's `loc`, so esrap flushes the
+    /// comments right before it. `$.proxy(` is built with a plain string callee
+    /// and carries no `loc`, which puts the comments before its argument
+    /// instead — its caller folds them into the argument text rather than
+    /// passing them here.
+    fn maybe_tag_declarator_with_lead(
+        &self,
+        var_name: &str,
+        replacement: String,
+        lead: &str,
+    ) -> (String, bool) {
         if !self.dev {
-            return replacement;
+            return (replacement, false);
         }
         let head = replacement.as_str();
         if head.starts_with("$.state(") || head.starts_with("$.derived(") {
-            format!("$.tag({}, '{}')", replacement, var_name)
+            (format!("$.tag({lead}{replacement}, '{var_name}')"), true)
         } else if head.starts_with("$.proxy(") {
-            format!("$.tag_proxy({}, '{}')", replacement, var_name)
+            (format!("$.tag_proxy({replacement}, '{var_name}')"), true)
         } else {
-            replacement
+            (replacement, false)
         }
     }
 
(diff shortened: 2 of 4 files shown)


baseballyama:
Measurement handed over from the #3426 / #3428 work (PR #3591, server-only), because this PR owns the client comment cluster and no issue should be filed for it.

The grid there is 10 script shapes × 7 tail-comment shapes × 21 template shapes = 1,470 cases, compiled for `client` and `server` = **2,940 comparisons** against the pinned oracle (`submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'`, asserted at run time). #3591 takes the **server** half to 0. The **client** half stands at **121** and #3591 does not move it: all 1,470 client outputs are byte-identical between the #3326 tree and the #3591 tree (0 cells differ), so this is untouched territory rather than fallout. For context, #3326 itself already took the client from 289 to 121 — 168 cells — so the residue below is what its rule does not reach.

The shape is the one this PR is already about, one host further out: **a comment left pending at the end of an instance script, with no statement after it to flush it.** Only the two script shapes that end in a comment reach it (`plain-tail`, `store-tail`); every `$:`-bearing shape is clean on the client.

## The 121, split by (does official print it / does rsvelte print it / same place)

| cells | official | rsvelte | template columns |
|---|---|---|---|
| 60 | inside the parameter list of the thunk it builds — `$.html(node, (/* tail */⏎) => b)` | **dropped** | `{#each}`, `{#key}`, `{#await}`, `{@const}`, `<svelte:element>` |
| 12 | same | printed one statement too late, at the fragment tail | `{@html}` |
| 24 | **leading** the first template statement | **dropped** | component prop, `<svelte:head>` |
| 24 | same | **trailing** that same statement | component spread, `{@render}` |
| 1 | *(no comment in the source — a different family, see below)* | | `<svelte:head>` |

Each of the ten comment-carrying columns fails all 12 of its cells (2 script shapes × 6 comment shapes), so the comment *spelling* — line, block, same-line, own-line, two-line, JSDoc — separates nothing. What separates is which node the client builder emits first and whether upstream gives it a `loc`.

Two worked examples:

```svelte
<script>
	export let a = 1;
	let b = a;
	/* tail */
</script>
{@html b}
```

```js
// official                          // rsvelte
$.html(node, (/* tail */             $.html(node, () => b);
) => b);                             $.append($$anchor, fragment);
                                     /* tail */
$.append($$anchor, fragment);
```

```svelte
<Comp {...b} />
```

```js
// official                                  // rsvelte
/* tail */                                   Comp($$anchor, $.spread_props(() => b));
Comp($$anchor, $.spread_props(() => b));     /* tail */
```

So the 60 + 12 rows are one mechanism (upstream parks the run in a **builder-made arrow's parameter parens** — the same slot this PR's `$derived` row already describes, reached from the template side instead of the declarator side), and the 24 + 24 rows are the other (the run flushes **before** the first located template statement, and rsvelte either loses it or attaches it after).

## The one cell that is not about comments

`store-tail__none__svelte-head` diverges with no comment anywhere in the source:

```svelte
<script>export let a = 1; let b = a + 1;</script>
```

```js
// official                       // rsvelte
$.document.title = b;             $.document.title = b ?? '';
```

Probed against the oracle directly: `let b = a` → both emit `b ?? ''`; `let b = a + 1` → official drops the coalesce and rsvelte does not; a bare store read `let b = $s` → both emit `b ?? ''`. The trigger is the `+`, not the store — upstream's `scope.evaluate` knows a `+` cannot yield nullish. That belongs to the `fold-value-type` / "is this value known" family, not here; flagging it so nobody folds it into the 120.

## Reproducing

Harness (oracle generator, dump test, comparator, decomposer) is at `…/scratchpad/salvA-3426-{grid,compare,decompose,show}.mjs`; the dump side is a throwaway `#[test]` over `salvA-3426-cases/`. Nothing is committed — the committed regression tests in #3591 cover the server half only.

baseballyama:
Correction to the measurement above, one line: the cell I flagged as "not about comments" now has its own issue — **#3599**. It is a `fold-value-type` defect (`<svelte:head><title>` emits a `?? ''` that upstream drops once `scope.evaluate` can prove the value is non-nullish; the trigger is a `+` in the initializer, isolated against three negative controls).

So the split is **120 comment-attribution cells + 1 unrelated**, not 121 of one kind. The four rows of the table above account for all 120.

PR #3417, opened 2026-08-22
fix(parse): reject a TypeScript namespace with non-type nodes through every modifier

### LINKED ISSUE #3244: typescript_invalid_feature misses export namespace, declare module and declare global
`typescript_invalid_feature` is not raised for a `TSModuleDeclaration` that
carries a modifier. Svelte erases types rather than compiling TypeScript, so
a non-erasable construct has to be rejected; rsvelte compiles three of them.

```svelte
<script lang="ts">
	export namespace N { export const a = 1; }
</script>
<div>x</div>
```

official: `typescript_invalid_feature`. rsvelte: compiles.

Same for `declare module "x" { export const a: number; }` and
`declare global { const g: number }`.

The discriminator is the modifier, not the construct:

|---|---|---|
| `namespace N { export const a = 1; }` | `typescript_invalid_feature` | same |
| `export namespace N { export const a = 1; }` | `typescript_invalid_feature` | **compiles** |
| `declare module "x" { … }` | `typescript_invalid_feature` | **compiles** |
| `declare global { … }` | `typescript_invalid_feature` | **compiles** |

A bare `namespace` is caught and the `export`-wrapped and `declare`-wrapped
forms are not, so the check is looking at a bare `TSModuleDeclaration`
statement and never unwrapping `ExportNamedDeclaration` / the `declare`
modifier. This is the same "the exported form is a second code path" shape
already recorded for `params.rest` in the serialized program.

Everything else in the non-erasable set is handled correctly on both entry
points — `enum`, `const enum`, a non-exported `namespace`, a parameter
property (`constructor(private x: number)`) and a class decorator all raise
`typescript_invalid_feature` in a component script and `js_parse_error` in a
`.svelte.ts` module, matching official exactly. Twelve erasable controls
(`type`, `interface`, `satisfies`, `as`, `!`, generics, `import type`,
inline `type` specifiers, `export type`, `declare` field, `y!:`, `as const`)
all match too.

Found by a 30-shape TypeScript grid x 3 entry points (90 cells); these three
rows and the already-known `readonly` field in `compileModule` (#3203) are
the only divergences.

baseballyama:
A TypeScript namespace carrying non-type nodes is rejected through every
modifier, and the decision about its body is now upstream's rather than a
hand-written predicate.

Closes #3244.

## What was wrong

Two independent causes, both measured against `submodules/svelte` on the same
input.

**1. The module node never reached the stripper.** Upstream's
`TSModuleDeclaration` visitor keys on one thing — `if (!node.body) return
b.empty` — and never looks at `declare`; `ExportNamedDeclaration` walks
`context.next()` into its declaration. rsvelte's converter answered both
differently:

| shape | rsvelte before |
|---|---|
| `declare module "x" { … }` / `declare namespace N { … }` | an early `if (module_decl.declare)` returned an `EmptyStatement` |
| `declare global { … }` | oxc models it as a third variant (`TSGlobalDeclaration`); there was no arm at all, so `filter_map` dropped it |
| `export namespace N { … }` | the `Declaration` (export) path returned an `EmptyStatement` unconditionally |

`export declare namespace N { … }` stays legal, and the reason is not the
`declare`: acorn-typescript gives that export `exportKind: 'type'`, so upstream
returns `b.empty` before the namespace is ever visited. oxc's
`ExportDeclaration::export_kind()` already derives the same thing from
`declaration.declare()`, so that row needed no code.

**2. The body test was a predicate, not a visit.** Upstream maps its *whole*
visitor over `node.body.body` and errors when any result is not the `b.empty`
singleton. rsvelte had a hand-written `is_type_only_namespace_member`, and it
diverges on every shape its author did not enumerate — in both directions:

| body | upstream | predicate |
|---|---|---|
| `enum E { A }`, `const enum E { A }` | `typescript_invalid_feature` **'enums'**, reported on the *enum* | accepted |
| `namespace M { const a = 1 }` | rejected, reported on the *inner* namespace | accepted |
| `class C {}`, `function f() {}`, `const a = 1` | rejected | accepted |
| `;` | rejected — a source `EmptyStatement` is not the `b.empty` singleton | accepted |
| `declare const a: number` | accepted (`VariableDeclaration` → `b.empty`) | **rejected** |
| `namespace M { }` | accepted | **rejected** |

Recursing with `recurse_range` reproduces the decision exactly, including which
node and which *feature* a nested raise reports.

## Measurement

A 37-construct × 4-modifier grid (`''` / `export ` / `declare ` /
`export declare `) × 2 entry points = 296 cells, verdicts read off the official
compiler and diffed against rsvelte before and after:

```
unchanged 244, fixed 38, regressed 0
```

The regression column is the point of the exercise: the first version of this
fix — modifiers only, predicate untouched — scored **fixed 18, regressed 8**,
and every one of the 8 was the predicate meeting a body shape it did not
enumerate. The number came from the grid, not from review.

## What still diverges, and why it is not this PR

80 of the 290 comparable cells still differ, all pre-existing and all one cause:
**oxc enforces TypeScript's ambient-context rules and acorn-typescript does
not.** `declare class D { constructor(private x: number) {} }` and
`declare function f() {}` are `js_parse_error` here and legal upstream;
`declare namespace N { const a = 1 }` is `js_parse_error` upstream (an
initializer in an ambient context) and reaches `typescript_invalid_feature`
here. Every one of these was already divergent on `main` — the delta above
moved none of them — and closing them means reconciling two JS parsers, not
this visitor. `namespace N { export default 1; }` (4 cells) is the one residue
with a different cause: the statement never reaches the module body.

## Tests

`crates/rsvelte_core/tests/ts_namespace_3244.rs`, four cases over both
`GenerateMode`s and all three entry points rsvelte has:

- the seven modifier shapes, each asserted on code, message **and** span;
- the body-content axis asserted on the node *and* the feature, so
  `namespace N { enum E { A } }` pins `'enums'` at the enum's own span rather
  than `'namespaces…'` at the module's;
- 20 legal neighbours as the over-rejection control;
- `compileModule`, which parses as JavaScript whatever the filename says and so
  answers every one of these with acorn's `js_parse_error` — including the
  shapes a component script accepts.

Every code, message and span in the file was read off the official compiler at
the pinned revision.

`ts_namespace_3244`, `ts_decorators_2597`, `compiler_errors` (145),
`parser_fixtures`, `compiler_fixtures`, `validator` (333) and `ssr` (99) pass.

### DIFF
--- a/.changeset/ts-namespace-modifiers-3244.md
+++ b/.changeset/ts-namespace-modifiers-3244.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Reject a TypeScript namespace with non-type nodes through every modifier: `export namespace N { … }`, `declare module "x" { … }` and `declare global { … }` now raise `typescript_invalid_feature` at upstream's span instead of compiling.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -8253,84 +8253,82 @@ fn convert_statement_for_program(
 
         // TypeScript module/namespace declarations - emit so remove_typescript_nodes can detect them
         oxc_ast::ast::Statement::TSExternalModuleDeclaration(module_decl) => {
-            // `declare module`, `declare global`, `declare namespace` etc. are
-            // type-only and must be stripped.
-            if module_decl.declare {
-                let start = offset + module_decl.span.start as usize;
-                let end = offset + module_decl.span.end as usize;
-                return Some(JsNode::EmptyStatement {
-                    start: start as u32,
-                    end: end as u32,
-                    loc: create_typed_loc(start, end, line_offsets),
-                });
-            }
-            let start = offset + module_decl.span.start as usize;
-            let end = offset + module_decl.span.end as usize;
-            let loc = create_typed_loc(start, end, line_offsets);
-
-            // Include body so remove_typescript_nodes can check for non-type nodes
-            let body = module_decl.body.as_ref().map(|block| {
-                let block_body: Vec<JsNode> = block
-                    .body
-                    .iter()
-                    .filter_map(|stmt| {
-                        convert_statement_for_program(arena, stmt, offset, line_offsets)
-                    })
-                    .collect();
-                arena.alloc_js_node(JsNode::BlockStatement {
-                    start: start as u32,
-                    end: end as u32,
-                    loc: loc.clone(),
-                    body: arena.alloc_js_children(block_body),
-                })
-            });
-
-            Some(JsNode::TSModuleDeclaration {
-                start: start as u32,
-                end: end as u32,
-                loc,
-                body,
-            })
+            Some(convert_ts_module_declaration_as_node(
+                arena,
+                module_decl.span,
+                module_decl.body.as_deref(),
+                offset,
+                line_offsets,
+            ))
         }
         oxc_ast::ast::Statement::TSNamespaceDeclaration(module_decl) => {
-            let start = offset + module_decl.span.start as usize;
-            let end = offset + module_decl.span.end as usize;
-            let loc = create_typed_loc(start, end, line_offsets);
-            if module_decl.declare {
-                return Some(JsNode::EmptyStatement {
-                    start: start as u32,
-                    end: end as u32,
-                    loc,
-                });
-            }
-            let body = match &module_decl.body {
-                oxc_ast::ast::TSNamespaceDeclarationBody::TSModuleBlock(block) => {
-                    let block_body = block
-                        .body
-                        .iter()
-                        .filter_map(|stmt| {
-                            convert_statement_for_program(arena, stmt, offset, line_offsets)
-                        })
-                        .collect();
-                    Some(arena.alloc_js_node(JsNode::BlockStatement {
-                        start: start as u32,
-                        end: end as u32,
-                        loc: loc.clone(),
-                        body: arena.alloc_js_children(block_body),
-                    }))
-                }
-                oxc_ast::ast::TSNamespaceDeclarationBody::TSNamespaceDeclaration(_) => None,
-            };
-            Some(JsNode::TSModuleDeclaration {
-                start: start as u32,
-                end: end as u32,
-                loc,
-                body,
-            })
+            Some(convert_ts_module_declaration_as_node(
+                arena,
+                module_decl.span,
+                ts_namespace_block(&module_decl.body),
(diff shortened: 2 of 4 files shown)


baseballyama:
salvD です。殺されたセッションが残した `fix/3217-3243-3244-js-early-errors` ブランチが #3244 を同じ方向で直していたので、**この PR に包含されるかを両方ビルドして測りました。結論: 包含されます。この PR で進めてください。** 私の側からの PR はありません。

## 差集合は空です（両方向）

| 測定 | 結果 |
|---|---|
| 私のテスト `ts_module_declaration_3244.rs`（2 本）を **この PR のツリー**で実行 | **2/2 PASS** |
| この PR のテスト `ts_namespace_3244.rs`（4 本）を **私のブランチ**で実行 | **4/4 PASS** |
| 36 本体 × 2 入口（instance script / `compileModule`）= 72 セルの出力を両ツリーでダンプして `diff` | **72 行とも完全一致** |

3 行目が「テストが表現していない差」まで潰す部分です。実装上の唯一の相違は
`strip_ts_module_declaration_typed` の本体走査（私は各エントリを個別に visit して `?`、
この PR は `recurse_range` で範囲ごと）でしたが、それを分けようとして書いた本体
（`namespace N { let x = 1; namespace B { let y = 1; } }` のような、後ろのエントリが
エラーを上げる形／先頭で既に非型ノードが確定する形を含む）でも**出力はバイト一致**でした。

オラクルは `submodules/svelte` @ `20b341f10048`、`VERSION === '5.56.9'`
（`git ls-tree origin/main submodules/svelte` と一致 = CI が読むツリー。
`node_modules/svelte` が解決する 5.56.10 ではありません）。

## 副産物: 72 セルのうち 4 セルは**両ツリーとも**公式と違います

**この PR の退行ではありません**（`origin/main` でも同じ）。別 issue に分けました:

- **#3566** — `export namespace N { type T = 1; }` / `{ interface I {} }` が instance script で
  公式に無い `$$props` 仮引数を生みます。`export type` / `export interface` /
  `export declare const` / `export declare namespace` は全て正しいので、**`export namespace`
  だけが「型だけの宣言」判定から漏れています**。
- **#3568** — ドット形 `namespace N.M { … }` が instance script にあると**公式が生の
  `TypeError`（`node.body.body.map is not a function`、code も position も無し）で落ちます**。
  rsvelte は通します。`AGENTS.md` の規則どおり再現しませんが、`upstream_issues/` の
  write-up と rsvelte 側挙動のピン留めが**まだ書かれていません**。

この PR のスコープを広げる必要はないと思います（どちらも `remove_typescript_nodes` の
別の判定です）が、#3566 は 1 行で足りる可能性があるので、着手されるならお任せします。

Issue #3247, opened 2026-08-21
An unterminated { is blamed on the enclosing element, and at the root is not reported at all


baseballyama:
An unterminated `{` in the template is reported against the **enclosing
element or block** instead of against the expression, so both the error code
and the position are wrong — and at the root of the template it is not
reported at all.

```svelte
<div>{1</div>
```

| | error |
|---|---|
| official | `js_parse_error` at 1:9 |
| rsvelte | `element_unclosed` at 1:0 |

All twelve shapes I tried diverge. The pattern is that rsvelte unwinds to the
construct that *contains* the expression and blames that:

| source | official | rsvelte |
|---|---|---|
| `<div>{1</div>` | `js_parse_error` 1:9 | `element_unclosed` 1:0 |
| `<div>{x</div>` | `js_parse_error` 1:9 | `element_unclosed` 1:0 |
| `<div>{</div>` | `js_parse_error` 1:6 | `element_unclosed` 1:0 |
| `<div><span>{1</span></div>` | `js_parse_error` 1:26 | `element_unclosed` 1:5 |
| `{#each [1] as v}{1{/each}` | `expected_token` 1:18 | `block_unclosed` 1:0 |
| `{#if true}{@const c = 1{/if}` | `expected_token` 1:23 | `block_unclosed` 1:0 |
| `<div title={1>x</div>` | `js_parse_error` 1:17 | `expected_token` 2:0 |
| `<C a={1 />` | `expected_token` 1:9 | `expected_token` 2:0 |
| `{1<div>x</div>` | `js_parse_error` 1:10 | `js_parse_error` 1:14 |

**Two of them are not reported at all:**

```svelte
{1
```

official: `expected_token` at 1:2. rsvelte compiles it, treating the
unterminated tag as if it had closed:

```js
export default function A($$anchor) {
	$.next();

	var text = $.text();

	text.nodeValue = '1';
	$.append($$anchor, text);
}
```

`{x` behaves the same way (`text.nodeValue = x`). The `{@html z` spelling of
this is already #3219.

Worth separating from the position-only rows: the **code** diverges on six of
these, and `code` is the one error field the corpus gate reports as saturated
— 0 divergences over the 2,843 `(id, target)` pairs both compilers reject. A
population of deliberately unterminated templates is not in that corpus,
which is why a saturated field still had this in it.

Found by writing a second input for error codes whose `compiler-errors`
fixture already passes (`unbalanced-curly-element`,
[...]

Issue #3246, opened 2026-08-21
{@const c} with no value compiles and references an undeclared binding


baseballyama:
`{@const c}` — a const tag with no `=` — is accepted, the declaration is
dropped, and the body keeps referring to the name that was never declared.

```svelte
<script>
	let flag = $state(true);
</script>
{#if flag}{@const c}<b>{c}</b>{/if}
```

official: `expected_token`.

rsvelte:

```js
var consequent = ($$anchor) => {
	var b = root();

	b.textContent = c;
	$.append($$anchor, b);
};
```

There is no `const c` anywhere in the module, so this throws
`ReferenceError: c is not defined` when the branch renders. The output
**parses**, so the parse oracle cannot see it — only output equality can.

The two neighbouring shapes are handled correctly, which localises it to the
"identifier but no initialiser" case rather than to const-tag parsing:

| source | official | rsvelte |
|---|---|---|
| `{@const}` | `expected_whitespace` | same |
| `{@const a = 1, b = 2}` | `const_tag_invalid_expression` | same |
| `{@const c}` | `expected_token` | **compiles** |

Found the same way as #3245 — a second input for an error code whose
`compiler-errors` fixture already passes.

Issue #3202, opened 2026-08-21
`{@const}`, `{#await}` and `{@render}` swallow every JS parse error


baseballyama:
`{@const}`, the `{#await}` head and `{@render}` route their expression through
`parse_js_expression_internal`, whose contract is
`.unwrap_or_else(|(_, pos)| create_empty_identifier(...))` — it turns **any**
parse error into an empty identifier and compiles on. The failure is not
strict-mode-specific: ordinary broken JavaScript compiles successfully.

```svelte
{#if true}{@const c = 1 +}<b>{c}</b>{/if}
{#if true}{@const c = 42 = nope}<b>{c}</b>{/if}
{#await 1 + then v}<b>{v}</b>{/await}
{#await 42 = nope then v}<b>{v}</b>{/await}
```

| input | official | rsvelte |
|---|---|---|
| `{@const c = 1 +}` | `js_parse_error` "Unexpected token" | **compiles** |
| `{@const c = 42 = nope}` | `js_parse_error` "Assigning to rvalue" | **compiles** |
| `{#await 1 + then v}` | `expected_token` "Expected token }" | **compiles** |
| `{#await 42 = nope then v}` | `js_parse_error` "Assigning to rvalue" | **compiles** |
| `{@render s(42 = nope)}` | `js_parse_error` "Assigning to rvalue" | `render_tag_invalid_expression` |

The `{@render}` row shows the shape clearly: the expression is swallowed into an
empty identifier, which is then not a call expression, so the *downstream*
validation fires with a different code and no position — a second error standing
in for the one that was dropped.

Every other template slot reports these correctly (`{#if}`, `{#each}`, `{#key}`,
a text tag, an attribute, `{@html}`, a spread, a directive, an event handler, a
component prop), which is what makes this three entry points rather than a
policy.

Found while closing #3201: these are the only three of that grid's 15 slots where
the fix had no effect, because the error never reaches the reporter.

baseballyama:
Measured while sweeping #3539, and one half of this is louder than the issue records: on the
**client** the swallowed parse error produces output **no JS parser accepts**.

Oracle `submodules/svelte`, `VERSION` read off the compiler's own export: **5.56.9**. Two inputs
official rejects (`-9 ** 4` → `js_parse_error: Unexpected token`; `--3` → `js_parse_error:
Assigning to rvalue`), placed in seven hosts × three targets:

| host | rsvelte |
|---|---|
[...]

PR #3429, opened 2026-08-22
fix(parse): raise the JS early errors that need the surrounding context

### LINKED ISSUE #3243: Eight JS early-error classes are accepted, and all of them emit unparseable output
Eight classes of JavaScript **early error** — syntactically shaped but
illegal — are accepted by rsvelte and rejected by official. In **every** case
rsvelte's output is text no JS parser accepts, because the illegal construct
is simply copied through.

```js
// m.svelte.js
export function f() { super(); }
```

| | |
|---|---|
| official | `js_parse_error` — `'super' keyword outside a method` |
| rsvelte | `export function f() { super(); }` — emitted verbatim |

The full set, each on all three entry points (`.svelte.js`, `.svelte.ts`,
component instance script) unless noted, and each with unparseable output:

| shape | official's message |
|---|---|
| `class K { constructor() {} constructor() {} }` | duplicate constructor |
| `function f() { super(); }` | `'super' keyword outside a method` |
| `function f() { break nope; }` | `Unsyntactic break` |
| `function f() { continue; }` | unsyntactic continue |
| `function f() { a: a: for (;;) break a; }` | duplicate label |
| `class K { m() { return this.#nope; } }` | `Private field '#nope' must be declared in an enclosing class` |
| `class K { #a = 1; #a = 2; }` (instance script only) | duplicate private name |
| `function f() { import "./x.js"; }` and the `export` form | `'import' and 'export' may only appear at the top level` |

25 of 111 cells in a 37-shape grid; **all 25 accept-divergences are
unparseable**, and there are no over-rejections in the grid. Twenty-nine
other early errors are handled correctly — `let a; let a;`, `const a;`,
`function f(a, a)`, `delete a` in strict mode, `with`, legacy octal,
`yield` outside a generator, `await` in a parameter default, assignment to a
call, `__proto__` twice, a bad regex flag, `...a` not last, and others — so
this is a specific set of missing checks rather than "rsvelte does not
validate JS".

What the eight have in common is that none is decidable by the tokenizer
alone: each needs the surrounding scope or class context. That points at the
checks oxc performs in `SemanticBuilder` rather than in `Parser`, which
rsvelte does not run.

**Do not size this against the per-pass `SemanticBuilder` result.** That
measurement (#2602) rejected constructing a `SemanticBuilder` **11–21 times
per file**, at ~2% of compile with a 3.3% ceiling. Running the early-error
checks **once** per script is a different quantity and has not been
measured; the earlier negative result closed a means, not the territory.
Whether oxc can surface these without a full semantic build is the first
thing to check.

None of this is reachable from the collected corpus — published code
compiles — which is why the direction is over-acceptance in all 25.

baseballyama:
Closes #3243

## What

acorn raises a family of JavaScript **early errors** that OXC does not, because none of them is decidable from the token stream alone — each needs the class, function or label context the construct sits in, which OXC leaves to `SemanticBuilder` and this pipeline never runs it. Every one was accepted and copied verbatim into output no JS parser reads.

The check is a single visitor (`1_parse/read/early_errors.rs`) reached from the existing `acorn_only_violation` seam, so it competes with every other acorn-only restriction by source position — acorn is single-pass and non-recovering, so the earliest violation is the one it would have thrown.

`expression.rs` gains exactly one line (a new entry in that seam's array); the logic lives in its own module.

## Rules implemented

Each was ported from acorn's own site and each expectation below — message **and** byte offset — was measured against `svelte.compile` / `svelte.compileModule`, never written by hand.

| shape | acorn's message |
|---|---|
| `class K { constructor() {} constructor() {} }` | `Duplicate constructor in the same class` |
| `function f() { super(); }` | `'super' keyword outside a method` |
| `class K { constructor() { super(); } }` | `super() call outside constructor of a subclass` |
| `function f() { break nope; }` | `Unsyntactic break` |
| `function f() { continue; }` | `Unsyntactic continue` |
| `function f() { a: a: for (;;) break a; }` | `Label 'a' is already declared` |
| `class K { m() { return this.#nope; } }` | `Private field '#nope' must be declared in an enclosing class` |
| `class K { #a = 1; #a = 2; }` | `Identifier '#a' has already been declared` |
| `class K { #a = 1; m() { delete this.#a; } }` | `Private fields can not be deleted` |
| `function f() { import "./x.js"; }` | `'import' and 'export' may only appear at the top level` |
| `function f(a = 1) { 'use strict'; }` | `Illegal 'use strict' directive in function with non-simple parameter list` |

All three entry points — `.svelte.js` / `.svelte.ts` via `compileModule`, and a component's instance script including `lang="ts"`.

## How the set was closed

The issue names eight classes. Rather than trust that list, I enumerated **every `raise` / `raiseRecoverable` message literal in acorn 8.17.0** (90 distinct) and wrote an input for each context-dependent one, then diffed rsvelte against the official compiler on all of them. That found a ninth class the issue does not list — the `'use strict'` directive with a non-simple parameter list, which reproduces on 13 of 20 shapes — and `super()` outside a derived constructor, and `delete` on a private field.

After the fix that independent sweep reports **zero** accept-divergences over 50 inputs. The 17 rows that still differ are all cases where **both** compilers reject and only the message wording or the byte offset diverges (`Missing catch or finally clause` at a different offset, `Invalid regular expression flag` vs OXC's wording, and so on) — a separate defect family, and not one this PR touches.

## Over-rejection is the other half

Adding a rejection is the easy direction to get wrong, so 55 of the 141 measured inputs are shapes acorn **accepts** and each sits one step from a rejection above: `super` in a method / field initialiser / static block / arrow-inside-a-method / object method, a getter-setter private pair, a private name declared in an enclosing class, a label reused across a function boundary, a chained label on a loop, `break` in a `switch`, `import()` and `import.meta` nested, a simple-parameter `'use strict'`, and a `'use strict'` that is not in the directive prologue.

Two TypeScript shapes needed their own handling and are pinned as tests, because a naive port rejects them:

- an **overload signature** repeats the member's name without defining it, so `constructor(a: string); constructor(a: any) {}` must not count as two constructors and `#m(a: string): void; #m(a: any) {}` must not collide;
- a statement inside a `TSModuleBlock` is not measured for the top-level `import`/`export` rule — upstream rejects the whole namespace as `typescript_invalid_feature` first, and firing here would replace that diagnostic with the wrong one.

## Negative control

Removing the one wiring line in `expression.rs` and re-running the suite: **12 of 14 tests fail, every one of them at `must not compile` / `must reject`** — the predicted symptom, not a different assertion — while both control tests (`legal_shapes_still_compile`, `legal_typescript_overloads`) stay **green**, so the legal half was never being carried by the new code.

## Tests

`crates/rsvelte_core/tests/js_early_errors_3243.rs`, 14 tests. Positions are asserted, not just codes: a `@` in each source marks the byte the official compiler reports and is stripped before compiling, so a fix that rejects at the wrong place fails.

Existing suites re-run green on this tree: `compiler_errors`, `parser_fixtures`, `compiler_fixtures`, `validator`, `ssr`, `runtime`.

## Scope note

`{@const}` / `{#await}` / `{@render}` swallowing JS parse errors (#3202), `{@const c}` with no value (#3246) and the unterminated-`{` attribution (#3247) are separate defects in `1_parse/state/tag.rs` and are **not** in this PR.

### DIFF
--- a/.changeset/js-early-errors-needing-context.md
+++ b/.changeset/js-early-errors-needing-context.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Reject the JavaScript early errors that need the surrounding class, function or label context to decide — a duplicate constructor, `super` outside a method, `super()` outside a derived constructor, an unsyntactic `break` / `continue`, a duplicate label, an undeclared or duplicated private name, `delete` on a private field, a nested `import` / `export`, and a `'use strict'` directive in a function with a non-simple parameter list — which OXC leaves to a semantic pass this pipeline never runs, so each was copied into output no JS parser accepts
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/early_errors.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/early_errors.rs
@@ -1,0 +1,516 @@
+//! JavaScript early errors acorn raises and OXC does not.
+//!
+//! Every one of these is syntactically shaped but illegal, and none is decidable
+//! from the token stream alone: each needs the class, function or label context
+//! the construct sits in. OXC leaves them to `SemanticBuilder`, which this
+//! pipeline never runs, so without this pass they are copied into the output and
+//! produce text no JS engine accepts.
+//!
+//! acorn is single-pass and non-recovering, so it throws on the first violation
+//! it reaches and never sees any that follow — the caller takes the earliest by
+//! position.
+
+use oxc_ast::ast::{
+    AccessorProperty, ArrowFunctionExpression, BindingPattern, CallExpression, Class, ClassElement,
+    Expression, FormalParameters, Function, MethodDefinition, MethodDefinitionKind, ObjectProperty,
+    Program, PropertyDefinition, PropertyKey, PropertyKind, Statement, StaticBlock, Super,
+    TSModuleBlock,
+};
+use oxc_ast_visit::{Visit, walk};
+use oxc_span::GetSpan;
+use rustc_hash::{FxHashMap, FxHashSet};
+
+/// The earliest early error in `program`, as `(offset, message)`.
+pub fn find_violation(program: &Program<'_>, source: &str) -> Option<(u32, String)> {
+    // Every check below needs one of these spellings in the source. `class`
+    // covers the private-name and constructor families, `:` the labels.
+    if !(source.contains("super")
+        || source.contains("break")
+        || source.contains("continue")
+        || source.contains("import")
+        || source.contains("export")
+        || source.contains("class")
+        || source.contains('#')
+        || source.contains(':')
+        || source.contains("use strict"))
+    {
+        return None;
+    }
+    let mut scan = Scan::default();
+    scan.visit_program(program);
+    scan.hits.into_iter().min_by_key(|(at, _)| *at)
+}
+
+/// What a label can be a destination for. acorn's `loopLabel` / `switchLabel`
+/// carry no name; a labelled statement carries a name and takes its kind from
+/// the statement it ultimately labels.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum LabelKind {
+    Loop,
+    Switch,
+}
+
+struct Label {
+    name: Option<String>,
+    kind: Option<LabelKind>,
+}
+
+impl Label {
+    fn loop_label() -> Self {
+        Label {
+            name: None,
+            kind: Some(LabelKind::Loop),
+        }
+    }
+
+    fn switch_label() -> Self {
+        Label {
+            name: None,
+            kind: Some(LabelKind::Switch),
+        }
+    }
+}
+
+/// One class body's private names: what it declares, and every `#name` read
+/// inside it that has not been resolved yet.
+#[derive(Default)]
+struct PrivateFrame {
+    declared: FxHashSet<String>,
+    used: Vec<(String, u32)>,
+}
+
+#[derive(Default)]
+struct Scan {
+    hits: Vec<(u32, String)>,
+    /// Set for the direct children of `Program` only, and consumed by the first
+    /// `visit_statement` that follows — acorn's `topLevel` argument.
+    top_level: bool,
+    /// acorn-typescript parses a namespace body with its own statement rules,
+    /// and upstream rejects the whole construct as `typescript_invalid_feature`
+    /// before any of this can apply.
+    in_ts_module: bool,
+    super_allowed: bool,
+    direct_super_allowed: bool,
+    /// Handed to the next `Function` visited, which is how a method's body gets
+    /// `super` while a plain function nested in the same place does not.
+    pending_super: bool,
+    pending_direct_super: bool,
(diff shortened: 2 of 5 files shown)


baseballyama:
Withdrawing this in favour of tsE's implementation of #3243, which was assigned in parallel by mistake and is further along on the axis that matters most (it also closes the **redeclaration** classes, which this PR does not touch).

Handing over the measurements so none of it has to be re-derived:

**The set was closed from acorn's own message list, not from the issue's table.** Enumerating every `raise` / `raiseRecoverable` literal in acorn 8.17.0 (90 distinct), writing one input per context-dependent rule and diffing against `svelte.compileModule` found **three classes the issue does not list**:

| shape | acorn's message | official position |
|---|---|---|
| `function f(a = 1) { 'use strict'; }` | `Illegal 'use strict' directive in function with non-simple parameter list` | the function node's start |
| `class K { constructor() { super(); } }` | `super() call outside constructor of a subclass` | the `super` token |
| `class K { #a = 1; m() { delete this.#a; } }` | `Private fields can not be deleted` | the `delete` keyword |

The `'use strict'` one reproduces on **13 of 20** shapes (default / rest / destructured / array parameter, generator, async, method, block-body arrow, a later directive in the same prologue) and is inert on the other 7 (simple parameters, no parameters, a concise-body arrow, a `'use strict'` that is not in the directive prologue) — the rule is `!isSimpleParamList(params) && the body has a use-strict directive`, reported at the function node's start.

**Two TypeScript shapes reject a naive port and are worth keeping as tests:**

- an **overload signature** repeats the member's name without defining it, so `constructor(a: string); constructor(a: any) {}` must not count as two constructors and `#m(a: string): void; #m(a: any) {}` must not collide. In OXC terms: `value.body.is_none()`.
- a statement inside a `TSModuleBlock` must be exempt from the top-level `import`/`export` rule — upstream rejects the whole namespace as `typescript_invalid_feature` first, and firing here replaces that diagnostic with the wrong one.

**Positions, all measured against `svelte.compile` / `svelte.compileModule`:** duplicate constructor at the second element's start (a `"constructor"` string key counts, a `static constructor` and a computed `["constructor"]` do not); `super` at the keyword; `break`/`continue` at the keyword; a duplicate label at the second label identifier; a private-name conflict at the second `#`; an undeclared private name at its `#`; a nested `import`/`export` at the keyword.

**The residue of that sweep is now #3432** — 17 inputs where **both** compilers reject and the message, the byte offset or the error `code` diverges. Its first section is the redeclaration family (`let a; let a;` → rsvelte answers `declaration_duplicate` from analysis where upstream answers `js_parse_error` from acorn), which is exactly the ground tsE's version covers.

**One correction to the issue body, from tsE's measurement:** "all 25 accept-divergences are unparseable" does not hold — `function f() { break; }` is accepted by OXC and its output is *different from official*, not *invalid JavaScript*. The parse oracle does not discriminate that class, so output equality is the only gate that sees it.

The test file here (`js_early_errors_3243.rs`, 14 tests, 141 measured inputs of which 55 are shapes acorn **accepts**) is the part most worth salvaging if any of it is useful — every position is asserted with a `@` marker stripped before compiling, so a fix that rejects at the wrong byte fails rather than passes.

Issue #3214, opened 2026-08-21
A comment inside an attribute expression is dropped


baseballyama:
A comment written inside an attribute's expression is dropped.

```svelte
<div title={/* c */ s}>x</div>
```

```js
// official
$.set_attribute(div, /* c */ 'title', s);
// rsvelte
$.set_attribute(div, 'title', s);
```

Official carries the comment into the generated call — at an odd place, because
the expression node it was attached to was replaced — and rsvelte loses it. The
loss is total: no `/* c */` appears anywhere in rsvelte's output.

Reproduces on **all 11 attribute/element pairs the grid carries** — `title`,
`class`, `style`, `value`, `placeholder`, `href`, `src`, `data-x`, `aria-label`,
a component prop, and an SVG `fill` — and on `client`, `server` and `client-dev`
alike. `value` on an `<input>` lands the comment in a different call
(`$.remove_input_defaults(input /* c */)`), which is the same loss seen through a
different rewrite.

Found by a generated grid over **36 attribute-value spellings × 12
attribute/element pairs × 3 targets** (1296 cells): **33 diverging, all of them
this one value**. The other 35 spellings — every quoting form, entity, escape,
astral character, concatenation, template literal, a `}` inside a string and a
`}` inside a regex — agree byte for byte.

baseballyama:
The same loss reaches a **block head**, which is a different parse function from
an attribute value:

```svelte
{#if /* c */ flag}<b>a</b>{/if}
```

| host | official | rsvelte |
|---|---|---|
| root, server | `if (/* c */ flag) {` | `if (flag) {` |
| inside a `<div>` | `var node = $.child(div /* c */);` | `var node = $.child(div);` |
| inside `{#each}` | `($$anchor, q /* c */) => {` | `($$anchor, q) => {` |

Official's placement is nonsense in two of the three — the comment lands on the
`$.child` argument and on the each callback's parameter — which is the same
cursor artifact this issue already describes, seen from a slot where the
replaced node is further from the comment.

Notably the root/`client` cell agrees while root/`server` diverges, so the loss
is not uniform across targets for one host either.

Found by a generated grid over 28 `{#if}` / `{@debug}` / `{@html}` shapes × 5
hosts (405 cells, 7 diverging — all of them this one shape).

Issue #3112, opened 2026-08-19
A comment on a constant-folded template expression is dropped instead of flushed onto the next node


baseballyama:
## Summary

A comment on a template expression that gets constant-folded is dropped. Official carries it to the next expression emitted in the same run, the way esrap's comment cursor does everywhere else.

## Repro

```svelte
<script>
	let n = $state(0);
	const f = (x) => x + 1;
</script>

<div>
	{n /* b */}
	<span>{f(n)}</span>
</div>
```

`generate: 'server'`:

```js
// official
	$$renderer.push(`<div>0 <span>${$.escape(
		/* b */
		f(n)
	)}</span></div>`);

// rsvelte
	$$renderer.push(`<div>0 <span>${$.escape(f(n))}</span></div>`);
```

`{n}` folds to the literal `0`, so its node is gone by print time; the comment then flushes onto `f(n)`, the next located node. Both targets diverge, client and server.

## Why the gates do not see it

`verify.mjs` falls back to `ast_equiv_batch` with an empty argv, which ignores comments, so the corpus scores this `match`. The matrix gate has a `comment-mismatch` verdict but its comment axis injects at line boundaries inside `<script>`, never into a template expression that folds. This was found by hand-written adversarial input, and it is the same cursor family as #2990 and #3005 — with the sign reversed: here rsvelte is the one losing the comment.

## Not reproduced by

- `{n /* b */}` alone, or followed by `<span>{n + 1}</span>` — the follower folds too, so neither compiler has a node to flush onto and both drop it.
- `{/* a */ ''}` followed by a non-folding expression behaves the same way as the above, i.e. it is the same defect, not a second one.

PR #3430, opened 2026-08-22
fix(server): carry a comment written inside a template expression

baseballyama:
## What

A JS comment written inside a template expression's `{ … }` was dropped by the
**server** target. Upstream keeps ONE esrap comment cursor over the whole file,
so such a comment is flushed before the next node the printer reaches that
carries a location — including when the expression it was written in
constant-folds away and the flush lands on the following one.

```svelte
<script>
	let s = 'a';
	let n = 0;
	const f = (x) => x + 1;
</script>

<div title={/* c */ s}>x</div>
<div>
	{n /* b */}
	<span>{f(n)}</span>
</div>
```

```js
// official                                     // rsvelte (before)
$.attr('title', /* c */ s)                      $.attr('title', s)

$.escape(                                       $.escape(f(n))
	/* b */
	f(n)
)
```

This is the **server half** of #3112 and #3214. The client half is a separate
problem and is *not* in this PR — see *Not fixed* below.

## Why it was dropped

rsvelte rebuilds a template expression from the structured `JsNode` AST
(`jsnode_to_oxc`), which stamps every node with the placeholder `SPAN` — so the
comment had no located node to flush against and never entered the registry.

## How

Three pieces, all under `3_transform/server/ast/`:

* `ServerTransformState::template_region_comments` harvests the comments of the
  `{ … }` region by re-parsing that source slice, and
  `place_template_expression_comments` registers the region on the existing
  `ChunkRegistry` and stamps the converted expression at the offset the *source*
  puts it at. The buffer is a verbatim source slice, so the line and column
  distances esrap measures between a comment and its anchor are the source's own
  — which is what reproduces upstream's multi-line flush above.
* A folded expression calls `defer_template_expression_comments` instead; the
  next surviving expression widens its region back to the deferred one's start,
  so the carried comment keeps its real line distance. That is the single cursor,
  not a special case for folding.
* `place_on_region` (script) now falls back to a position chunk when a statement
  owns no comment. Upstream keeps every source node's `loc`, so a location-less
  body it prints re-syncs the cursor; rsvelte's had none, and printing
  `const h = () => {};` **killed the whole comment list** before the template was
  reached. Measured: with that one statement in the script the comment vanished,
  without it the comment survived.

## Two negative controls that moved the design

Both were found by the divergence-set sweep below, not by review.

* Upstream's `component_block.loc = instance.loc` (`transform-server.js`,
  *"trick esrap into including comments"*) is what gets template comments printed
  at all. A component with **no `<script>`** has no `loc` to copy, so esrap's
  `reset_comment_index` discards the whole list — official drops the comment.
  A `<script module>` is not the one it copies. Gated on the instance script.
* The converted expression is stamped at ONE address, so a comment written
  **inside** it cannot be placed where upstream puts it: `{f(n) /* c */ + 1}`
  came out as `f(n) + 1 /* c */` where upstream prints `f(n) + /* c */ 1`. Those
  are now skipped — dropped, as before this PR — rather than emitted in the wrong
  place. `an_interior_comment_is_still_dropped` records that boundary.

## Evidence

Divergence **set** against the pinned official compiler over every `.svelte` in
`submodules/svelte/packages/svelte/tests`, client + server, raw bytes, both
binaries built from this tree with only the changed files reverted:

| | compared | diverging |
|---|---|---|
| merge base | 7,695 | 171 |
| this branch | 7,695 | **170** |

Compared as a set, not a count: `runtime-runes/samples/state-snapshot-uncloneable-ignored`
(server) newly matches, and **nothing new diverges**.

A generated grid of 25 comment-host shapes × 3 targets (75 comparisons, raw
bytes) goes from 59 diverging to 48; the whole 11-cell drop is the server column
(14 → 6).

Suites green: `ssr`, `compiler_fixtures`, `print`, `validator`,
`module_server_comment_ownership_2307`, `server_effect_strip_comment_delimiter`,
`legacy_derived_server_fold`, `reactive_block_comment_delimiter_2351`,
`reactive_label_comment_ownership`, `reactive_statement_comment_rehome_2368`,
`module_comment_ownership`, `module_toplevel_comments_2399`,
`each_header_comments`, `await_comment_run_relocation`, plus the new suite.

The regression test asserts the exact bytes in a Rust suite rather than a
`pattern-corpus` file, because the corpus verdict falls back to
`ast_equiv_batch`, which ignores comments — a repro filed there would score
`match` whether the comment is placed correctly, misplaced, or dropped. Every
expected string in it was read off the official compiler, not written by hand.

## Not fixed (both issues stay open)

* **The client and client-dev halves.** rsvelte's client comment coordinate space
  is the GENERATED script text (`js_ast/to_oxc.rs::parse_chunk`), so a template
  expression has no channel into it at all. Upstream's placement there also
  depends on the *element's* source line, not only the expression's: with the
  start tag and the comment on one line it prints
  `$.set_attribute(div, /* c */ 'title', s)`, and with the tag on the line above,
  `$.set_attribute(div, 'title', /* c */ s)` — the first is
  `flush_trailing_comments` firing on the `node_id` identifier, which upstream
  stamps with the element's `loc`. Reproducing that needs the element's
  coordinates in the same space as the expression's, which is the unification
  #3326 named.
* **Server hosts that are not an `ExpressionTag`**: a spread attribute,
  an `{#if}` / `{:else if}` test, an `{#each}` collection, `{@html}`,
  `{@render}` arguments and `{#await}`. Each needs its own `{ … }` region and
  diverges on the merge base in the same rows; this change does not widen it.

## Ratchets

None touched, and `--update-baseline` was NOT run. Comment-only divergences are
invisible to the corpus output gate (`ast_equiv_batch`, no `--comments`); if the
matrix gate's `comment-mismatch` verdict moves, CI will say so.

Contributes to #3112 and #3214.


baseballyama:
Three things from measuring the client half against this PR's scope statement. The third one
bears on **"the client and client-dev halves … a template expression has no channel into it at
all"** — I think that is stronger than the code supports.

Oracle throughout: `submodules/svelte` @ `20b341f10048` = **5.56.9**, with **esrap 2.2.12**
resolved through the submodule's own `node_modules`. Raw byte equality, no normalization, no
parser oracle.

## 1. How big the client half is: official emits in 14 cells, rsvelte in 0

One comment at a time (with several comments in one file official's cursor places them
differently, so a multi-comment probe cannot attribute a divergence to a position), across nine
template positions × two targets, on `main` c8a67c5aa:

| position | client | server |
|---|---|---|
| `{/* c */ v}` | official 1 / rsvelte 0 | both 0 |
| `{#if /* c */ v}` | both 0 | official 1 / rsvelte 0 |
| `{#each [/* c */ v] as i}` | official 1 / rsvelte 0 | official 1 / rsvelte 0 |
| `{@const c = /* c */ v * i}` | official 1 / rsvelte 0 | official 1 / rsvelte 0 |
| `title={/* c */ v}` | official 1 / rsvelte 0 | official 1 / rsvelte 0 |
| `onclick={() => /* c */ v++}` | official 1 / rsvelte 0 | both 0 |
| `{@html /* c */ v}` | official 1 / rsvelte 0 | official 1 / rsvelte 0 |
| `{#key /* c */ v}` | official 1 / rsvelte 0 | both 0 |
| `{@render s(/* c */ v)}` | official 1 / rsvelte 0 | official 1 / rsvelte 0 |

**Official emits in 14 of 18; rsvelte in 0** — client 8, server 6. So the client half is 8 of
the 14, not a tail.

## 2. There is a rescued commit that moves one of those client cells, in `parse_chunk`

`cac630f20` (a WIP rescue commit on `fix/3112-3214-client-template-expression-comments`) is 6
lines in `js_ast/to_oxc.rs` plus a test:

```rust
- if ret.program.comments.is_empty() && !self.synth.borrow().enabled {
+ if ret.program.comments.is_empty() {
      self.note_span(text.len() as u32);
      return Some(…);
  }
- if !ret.program.comments.is_empty() {
-     self.synth.borrow_mut().saw_comments = true;
- }
+ self.synth.borrow_mut().saw_comments = true;
```

The last two hunks are semantically equivalent — reaching that line means the comment list is
non-empty, because the early return above fires when it is empty. So the whole change is the
**first** condition: in the synth (second) pass, a comment-free chunk now takes the
"no location" fast path instead of the located one.

**That is not a new channel. It is a branch condition on the channel that already exists** —
the synth two-pass machinery in `parse_chunk` is the route. Its test asserts

```
<div title={/* c */ s}>x</div>   ->   $.set_attribute(div, /* c */ 'title', s);
```

which is exactly official's placement for that cell as I measured it independently.

So **"the client … has no channel into it at all" reads as *cannot*, and the code suggests
*already half-does*.** That matters because it is the justification for this PR's scope: "we
are not doing the client half" is a decision, "the client half is not reachable" is a claim
about the code. I have not built that commit yet — rescued branches have failed to compile
three times today — so treat this as "the scope statement needs re-checking", not as a finished
counterexample.

## 3. If you adopt that commit's expectations, note where they came from — but they check out

Its test header says the expected strings were read off **Svelte 5.56.10**, which is
`node_modules/svelte`, not the pinned submodule (5.56.9). Since the quantity being asserted is
*comment placement*, and the component that decides comment placement is esrap, the two
oracles differ on both axes:

    pinned submodule    svelte 5.56.9    esrap 2.2.12
    node_modules        svelte 5.56.10   esrap 2.3.5

I measured whether that matters: 13 cases × 3 targets = **39 pairs, byte-identical, 0
divergences**, including the three cells that test asserts. So the expectations are correct
despite the provenance — but the provenance is worth fixing, and the 39 pairs are a population
chosen for comment placement, not a general licence to read expectations off 5.56.10.

Related measurements, in case they are useful for the client half: of the 150 comment-bearing
cells in a 10-host grid, official emits in 120 — rsvelte emits byte-equal in 46, emits in the
**wrong slot** in 59, and emits **nothing** in 15 (all of them `{@const}`). In the 30 cells
where official emits nothing, rsvelte emits in **0**, so any regression from a fix here must
show up in those 30 and nowhere else. Details in #3515.

PR #3431, opened 2026-08-22
fix(svelte2tsx): read the generics attribute with a TypeScript parse, not a comma scan

### LINKED ISSUE #3252: svelte2tsx: a comma in a `generics=` constraint splits the type parameter, producing unparseable TSX
`generics="…"` is split into type parameters by a textual comma scan that knows
only about `<…>`. A comma at top level of any **other** bracket kind — an object
type, a tuple, a parameter list, or a string literal — splits the constraint in
the middle, and the fragments are then emitted as *type arguments*.

```svelte
<script lang="ts" generics="T extends { a: string, b: number }">
	let { v }: { v: T } = $props();
</script>
```

| | |
|---|---|
| rsvelte | `props(): ReturnType<typeof $$render<T,b:>>['props']` |
| official | `props(): ReturnType<typeof $$render<T>>['props']` |

`<T,b:>` is not a type argument list. Run through the TypeScript parser
(`ts.createSourceFile`, `ScriptKind.TSX`), rsvelte's output for that component
yields **45 parse errors** (`'>' expected.`) and official's parses clean. The
whole component's type surface is gone in the editor, from source that is
ordinary TypeScript.

The definition string is corrupted too, more quietly — rsvelte rejoins the
fragments with a bare `,`, so `{ a: string, b: number }` comes back as
`{ a: string,b: number }` everywhere the definition is printed
(`class __sveltets_Render<…>`, the `$$IsomorphicComponent` members, the exported
`type` alias).

### Which bracket kinds separate the two

Same script body, only the attribute value varies:

| `generics=` | identical? |
|---|---|
| `T extends Record<string, number>` | yes |
| `T extends Array<[1, 2]>` | yes |
| `T extends { a: string; b: number }` (semicolons) | yes |
| `T extends A, U extends B` | yes |
| `T extends { a: string, b: number }` | **no** |
| `T extends [a, b]` | **no** |
| `T extends [a: 1, b: 2]` | **no** |
| `T extends (a: 1, b: 2) => void` | **no** |
| `T = { a: 1, b: 2 }` | **no** |
| `T extends 'a,b'` | **no** |

Angle brackets are handled and nothing else is. `Array<[1, 2]>` passes only
because the outer `<…>` already covers the inner comma — which is why this never
showed up in the corpus: the shapes that reach it need a comma that is *not*
inside angle brackets.

### Cause

Upstream does not scan. `svelte2tsx/nodes/Generics.ts` builds the string
`` `<${raw}>() => {}` ``, hands it to `ts.createSourceFile`, and reads
`arrowFunction.typeParameters` — `definitions` are `param.getText()` and
`references` are `param.name.getText()`. Both lists come from the TypeScript
parser, so no bracket kind is special.

Splitting on commas is the same shape as the `opaque-keyword` family one level
up: a raw scan deciding structure. The fix is to parse the attribute rather than
to add `{`/`[`/`(`/quote awareness to the scan — a scan will keep being one
bracket kind behind.

### Method

`generics="…"` value × where `T` is used in the script × whether the file is TS —
108 cells, 108 compiled on both sides, 48 divergent. Found by crossing the
attribute **value** with the script body; every existing exercise of this
attribute holds the value near-fixed and varies the body, which is why the
splitter was never separated from its input.

Related but a different predicate, filed separately: the *presence* test.

### LINKED ISSUE #3253: svelte2tsx: `generics=` presence is decided from the raw attribute, where upstream decides it from a TypeScript parse
Upstream answers "does this component have generics?" **twice, from two different
sources**, and rsvelte answers it once.

`svelte2tsx/createRenderFunction.ts:79` keys on `generics.genericsAttr` — the raw
attribute, truthy whenever it is non-empty — and splices the raw text into
`function $$render<…>`. `svelte2tsx/addComponentExport.ts:43` keys on
`generics.has()`, i.e. `definitions.length > 0`, and `definitions` is populated
only if `ts.createSourceFile("<" + raw + ">() => {}")` actually yields an arrow
function with type parameters.

So when the attribute is present but does not parse as a type parameter list,
upstream writes the raw generics onto `$$render` **and** emits the *non-generic*
component export. rsvelte treats "attribute present" as one decision and emits
the generic form throughout.

```svelte
<script lang="ts" generics="T extends string ? 1 : 2">
	let x: T = null as any; void x;
</script>
```

Both sides emit `;function $$render<T extends string ? 1 : 2>() {`. After that:

```
rsvelte  class __sveltets_Render<T extends string ? 1 : 2> { … }
         interface $$IsomorphicComponent { new <T extends string ? 1 : 2>(…) … }
         const Probe__SvelteComponent_: $$IsomorphicComponent = null as any;
         type Probe__SvelteComponent_<T extends string ? 1 : 2> = InstanceType<typeof Probe__SvelteComponent_<T>>;

official const Probe__SvelteComponent_ = __sveltets_2_isomorphic_component(__sveltets_2_with_any_event($$render()));
         type Probe__SvelteComponent_ = InstanceType<typeof Probe__SvelteComponent_>;
```

### Attribute values that separate the two

16 of 36 values tested, each across all three hosts (TS instance, TS instance
with `$props()`, JS file) — 48 divergent cells of 108, all 108 compiled on both
sides:

`T extends string ? 1 : 2` · `T extends Array<infer U> ? U : never` ·
`T extends string ? U extends number ? 1 : 2 : 3` · `,T` · `in T` · `out T` ·
`in out T` · `T extends [a: 1, b: 2]` · `T extends Array<string>>` ·
`T extends Array<<string` · `" "` · `extends string` · `1` · `class` ·
`T> = (x: T) => x; type Q = <U` · `T = ({ a: 1 } satisfies object)`

Two of these are worth separating from the rest, because rsvelte's answer is not
merely "generic where upstream said non-generic" — it invents a type parameter
name out of the leading token:

| attribute | rsvelte references | official |
|---|---|---|
| `in T` | `<in>` | bails to non-generic |
| `extends string` | `<extends>` | bails to non-generic |
| `T extends Array<infer U> ? U : never` | `<T>` | `<T,U,never>` (TS error recovery yields three parameters) |

Neither side produces good output for a malformed attribute — upstream's own two
answers contradict each other, and `function $$render<T extends string ? 1 : 2>()`
is invalid TypeScript on both sides. Byte equality is the goal, so this is
reported rather than assumed benign; if the decision is to match upstream, the
predicate to port is `Generics.has()`, not "the attribute string is non-empty".

Same family, different mechanism, filed separately: #3252 (the comma splitter),
which is reachable from **valid** TypeScript and produces unparseable output.

baseballyama:
Closes #3252
Closes #3253

## What

`generics="…"` was split into type parameters by a textual comma scan that knew only about `<…>`, and "does this component have generics?" was answered from the raw attribute being non-empty. Both now come from the same place upstream reads them from: `Generics.getGenericTypeParameters` parses `` `<${raw}>() => {}` `` and reads `arrowFunction.typeParameters`, with `definitions = param.getText()` and `references = param.name.getText()`.

`parse_generics_attr` does the same with oxc (`SourceType::ts()`), so no bracket kind is special. `Generics.has()` becomes `!definitions.is_empty()`, which is what `add_component_export` now keys on; the raw attribute still reaches `function $$render<…>` verbatim, exactly as upstream's `createRenderFunction` keys that on `genericsAttr` rather than on `has()`.

## Measurement

Oracle: the official `svelte2tsx` built from `submodules/language-tools` at the pinned svelte (5.56.9). Grid: 28 attribute values × 2 script bodies (a plain `T` annotation, and a `$props()` destructuring) = 56 cells, all 56 compiled on both sides, compared byte-for-byte.

| | identical |
|---|---|
| before | **10 / 56** |
| after | **52 / 56** |

The 4 that remain are two attribute values — `T extends Array<infer U> ? U : never` and `T = ({ a: 1 } satisfies object)` — where TypeScript's parser *error-recovers* into three bogus type parameters and official then emits text that is itself not valid TSX (`<T = ({ a: 1 },satisfies,object>`). oxc bails to non-generic instead. Both inputs are invalid TypeScript and neither side produces usable output; matching them would mean porting TS's recovery, so they are left as they are rather than chased.

Two measurements worth recording because the issue text predicts otherwise, and the oracle disagrees with it:

- `out T` **is** generic on both sides (TS parses `<out T>() => {}` as an arrow with one type parameter). `in T` and `in out T` are not, on both sides. The regression test uses the two that bail.
- `T,` and `  T  ` were divergent before and are fixed too — they are not in the issue's table.

## Which gate sees this

The svelte2tsx **TSX-text** gate (`compatibility/svelte2tsx-known-failures.json`). The source-map gate is untouched by this change: nothing about the MagicString layout of the instance script moves — the corrected text is all in the appended component export, which is generated content with no source mapping.

Neither gate had a chance of catching it: the attribute values that reach the splitter need a comma that is *not* inside angle brackets, and the collected corpus holds none.

## Tests

`crates/rsvelte_projection/tests/svelte2tsx_generics_attribute_parse.rs` — 5 tests, expectations measured against the oracle: the six bracket kinds that used to split, multi-parameter joining, the non-generic export for an attribute that does not parse, the raw attribute still reaching the render header, and `const` / trailing-comma handling.

### DIFF
--- a/.changeset/s2t-generics-attribute-parse.md
+++ b/.changeset/s2t-generics-attribute-parse.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/svelte2tsx": patch
+---
+
+Read the `<script generics="…">` attribute with a TypeScript parse instead of a comma scan. The scan knew only about `<…>`, so a comma at the top level of an object type, a tuple, a parameter list or a string literal split the type parameter and the fragments were emitted as *type arguments* — `<T,b:>` is not a type argument list, so the whole component's TSX stopped parsing. Whether a component has generics at all now comes from the same parse (upstream's `Generics.has()`), so an attribute that is not a type parameter list keeps the non-generic component export while still reaching `function $$render<…>` verbatim.
--- a/crates/rsvelte_projection/src/svelte2tsx/add_component_export.rs
+++ b/crates/rsvelte_projection/src/svelte2tsx/add_component_export.rs
@@ -9,7 +9,7 @@ use super::interfaces::{Svelte2TsxOptions, SvelteVersion};
 use super::magic_string::MagicString;
 use super::nodes::component_documentation::extract_component_documentation;
 use super::nodes::component_events::build_events_str;
-use super::nodes::generics::{compact_generic_params, split_generic_param_names};
+use super::nodes::generics::parse_generics_attr;
 use super::nodes::slot::build_slots_str;
 use super::script::{ComponentEvents, ExportedNames};
 use super::template;
@@ -199,8 +199,12 @@ pub fn add_component_export(
         ""
     };
 
-    // Determine if this component has generics (either from generics= attribute or $$Generic)
-    let has_generics = !exported_names.dollar_generics.is_empty() || generics_attribute.is_some();
+    // Upstream keys the component export on `Generics.has()`, i.e. the attribute
+    // parsed as a type parameter list — not on the attribute merely being there.
+    let attribute_generics = generics_attribute
+        .map(parse_generics_attr)
+        .unwrap_or_default();
+    let has_generics = !exported_names.dollar_generics.is_empty() || attribute_generics.has();
 
     // Build generics strings for component export
     let (generics_params, generics_names) = if !exported_names.dollar_generics.is_empty() {
@@ -220,12 +224,11 @@ pub fn add_component_export(
             .map(|(name, _)| name.clone())
             .collect();
         (params.join(","), names.join(","))
-    } else if let Some(g) = generics_attribute {
-        // Create compact params string (strip leading spaces from each param)
-        let params_str = compact_generic_params(g);
-        // Split generic params at top-level commas (not inside angle brackets)
-        let names = split_generic_param_names(g);
-        (params_str, names.join(","))
+    } else if attribute_generics.has() {
+        (
+            attribute_generics.definitions_str(),
+            attribute_generics.references_str(),
+        )
     } else {
         (String::new(), String::new())
     };
--- a/crates/rsvelte_projection/src/svelte2tsx/nodes/generics.rs
+++ b/crates/rsvelte_projection/src/svelte2tsx/nodes/generics.rs
@@ -112,81 +112,73 @@ pub fn type_text_typeof_references_local_value(
     false
 }
 
-/// Split a generics string like "T extends Record<string, any>, U" into
-/// just the type parameter names: `["T", "U"]`.
-/// Handles nested angle brackets and commas inside constraints.
+/// The parsed `generics="…"` attribute: upstream's `Generics` lists.
+///
+/// `definitions` are `param.getText()` and `references` are
+/// `param.name.getText()` over the type parameters of
+/// `` `<${raw}>() => {}` ``, so no bracket kind is special-cased — a comma
+/// inside an object type, tuple, parameter list or string literal belongs to
+/// the parameter it sits in.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct ParsedGenerics {
+    pub definitions: Vec<String>,
+    pub references: Vec<String>,
+}
+
+impl ParsedGenerics {
+    /// Upstream `Generics.has()` — the attribute being present is not enough;
+    /// it must parse as a type parameter list.
+    pub fn has(&self) -> bool {
+        !self.definitions.is_empty()
+    }
+
+    /// Upstream `toDefinitionString()` without the surrounding `<`/`>`.
+    pub fn definitions_str(&self) -> String {
+        self.definitions.join(",")
+    }
+
+    /// Upstream `toReferencesString()` without the surrounding `<`/`>`.
+    pub fn references_str(&self) -> String {
+        self.references.join(",")
+    }
+}
+
+/// Parse a `generics="…"` attribute value the way upstream does: wrap it in an
+/// arrow function's type parameter list and read the parameters back off the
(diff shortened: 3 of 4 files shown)


baseballyama:
salvB より。#3252 について**別の（後発の）実装が救出ブランチ `fix/3252-3255-s2tx-generics-snippet` にありますが、この PR が先行しており同じ上流忠実アプローチ（`<${raw}>() => {}` をパースして `typeParameters` を読む）なので、そちらは出しません。**

ひとつだけ、この PR に**無い**もので使えそうなものがあります: `compatibility/pattern-corpus` の repro です。AGENTS.md の運用では「収集コーパスが届かない欠陥には pattern-corpus の repro を置く」ことになっていますが、この PR の変更ファイル一覧には pattern-corpus のエントリがありません。

該当ブランチの `e73c8601f` に入っています:

- `compatibility/pattern-corpus/issues/3252-generics-comma-outside-angle-brackets.svelte`

```svelte
<script lang="ts" generics="T extends { a: 1, b: 2 }, U extends [1, 2]">
	let { value, pair }: { value: T; pair: U } = $props();
</script>

<p>{value.a}{pair[0]}</p>
```

  トップレベルのカンマがオブジェクト型とタプルの中に 1 つずつ入っており、`<…>` だけを追うスキャンでは両方で割れます。

- `compatibility/pattern-corpus/README.md` に対応する 1 行（この repro が届くのは共有マニフェストの svelte2tsx ゲートだけで、コンパイラは属性を消すため client/server の行は対照になる、という理由つき）。

必要なら `git cherry-pick` するか、そのまま写してください。取り込むかどうかはこの PR のオーナー判断で構いません。

PR #3445, opened 2026-08-22
fix(esrap): print `#x in o` instead of an `/*unsupported:*/` comment

### LINKED ISSUE #3413: The `#x in o` brand check prints as `/*unsupported:PrivateInExpression*/` — unparseable output in 4 of 6 host positions
## Summary

`#x in o` — the ES2022 ergonomic brand check — is printed as the literal comment
`/*unsupported:PrivateInExpression*/`. In four of six host positions the result is **output no JS
parser accepts**; in the other two it parses and computes the wrong value.

```svelte
<script>
	class K { #x = 1; has(o) { return #x in o; } }
	void K;
</script>
<b>x</b>
```

```js
/* official */  return #x in o;
/* rsvelte  */  return /*unsupported:PrivateInExpression*/;      // returns undefined
```

## Matrix — the host decides how loudly it fails

| host | rsvelte output | verdict |
|---|---|---|
| `if (#x in o) …` | `if (/*unsupported:PrivateInExpression*/) return 1;` | **unparseable** |
| `#x in o ? 1 : 0` | `return /*unsupported:…*/ ? 1 : 0;` | **unparseable** |
| `#x in o && 1` | `return /*unsupported:…*/ && 1;` | **unparseable** |
| `const b = #x in o` | `const b = /*unsupported:…*/;` | **unparseable** |
| `return #x in o` | `return /*unsupported:…*/;` | parses, returns `undefined` |
| `String(#x in o)` | `return String(/*unsupported:…*/);` | parses, calls `String()` with no argument |

Reproduces on `client`, `server` and `client-dev`, and through `compileModule` on a
`.svelte.(js|ts)` module, in a method body, an arrow class field, and a `static` context — 4 host
shapes × 4 entry points, all leaking.

## The marker is a safety net with no production consumer

`printer.rs:1520`:

```rust
fn unsupported(&mut self, kind: &'static str, ctx: &mut Context<DIRECT>) {
    if self.missing.is_none() { self.missing = Some(Unsupported(kind)); }
    // Emit a marker so output is obviously wrong if a miss slips through a
    // test that forgot to check `missing`.
    ctx.write(format_compact!("/*unsupported:{kind}*/"));
}
```

`self.missing` is the intended bail-out signal, and **the only reader is
`crates/rsvelte_esrap/src/internal_tests/golden.rs`**. No production caller checks it, so the
marker is not a debugging aid that a fallback catches — it is the shipped output. Any future
printer gap fails exactly the same way.

`printer.rs:3442` (`other => self.unsupported(expression_kind(other), ctx)`) is the arm reached
here; there are 12 `unsupported(…)` call sites in total.

## Scope: this is the only live gap found

21 other constructs were checked and none reaches a marker — private field reads, private methods,
`static {}`, tagged templates, `yield`, `super()` and `super.m()`, `new.target`, `import.meta`,
dynamic `import()`, regexp literals, `for await…of`, optional-chained calls, destructuring holes
with defaults, `export default class`/`function`, labelled `continue`, spread in `new`, a TS
non-null in a chain. So the immediate fix is one expression kind; the structural fix is giving
`missing` a production reader.

baseballyama:
`#x in o` — the ES2022 ergonomic brand check — was printed as the literal comment
`/*unsupported:PrivateInExpression*/`.

ESTree has no node for it: it is a `BinaryExpression` whose `left` is a `PrivateIdentifier`, which
esrap's `operand_needs_wrap` special-cases to "never parenthesize". oxc gives the form its own
`Expression::PrivateInExpression`, so the port had no arm and fell to the catch-all at
`printer.rs:3442`. `ChainElement::PrivateFieldExpression` (`o?.#x`) reached the same catch-all.

**How loudly it failed depended on the host, not on the defect** — four of six syntactic positions
produced text no JS parser accepts, and the other two produced text that parses and computes
`undefined`:

| host | before |
|---|---|
| `if (#x in o)` | unparseable |
| `#x in o ? 1 : 0` | unparseable |
| `#x in o && 1` | unparseable |
| `const b = #x in o` | unparseable |
| `return #x in o` | parses, returns `undefined` |
| `String(#x in o)` | parses, calls `String()` with no argument |

All three targets and `compileModule`.

## Verification

39 cells — 13 shapes × 3 targets — now match the official compiler byte for byte, including all
six hosts above, the `static`/arrow/class-expression forms, a parenthesized right operand, and two
controls (a plain `"a" in o` and a plain `this.#x` read, both of which were already correct).

Two `pattern-corpus` repros land with the fix, formatted through the gate's own oxfmt oracle and
verified fmt-stable.

## Note

`printer.missing` — the bail-out signal this marker was meant to back up — is read only by
`crates/rsvelte_esrap/src/internal_tests/golden.rs`. No production caller checks it, so any future
printer gap fails the same way. Left as-is here; it is a structural change, not part of this fix.

Closes #3413


baseballyama:

PR #3454, opened 2026-08-22
fix(compiler): a SequenceExpression is never statically defined

### LINKED ISSUE #3323: Client: a sequence expression in a concatenated interpolation is missing its `?? ''` guard

baseballyama:
Closes #3323

## What

The `?? ''` guard on a concatenated interpolation is upstream's

```js
if (!evaluated.is_defined) value = b.logical('??', value, b.literal(''));
```

and `evaluated` comes from `scope.evaluate`, which has **no `SequenceExpression` case at all** — it falls through to `default: this.values.add(UNKNOWN)`, and UNKNOWN clears `is_defined`. rsvelte's built-value path (`is_js_expr_defined`) instead answered from the sequence's last element:

```rust
JsExpr::Sequence(seq) => seq.expressions.last().is_some_and(|e| is_js_expr_defined(e, arena, context)),
```

so `{#each arr as q}{(n, s)}{q}{/each}` emitted

```js
// before     $.template_effect(() => $.set_text(text, `${(n, s)}${$.get(q) ?? ''}`));
// official   $.template_effect(() => $.set_text(text, `${(n, s) ?? ''}${$.get(q) ?? ''}`));
```

A sequence evaluating to `null` / `undefined` therefore rendered the string `"undefined"` where official renders nothing. The arm is deleted, so a sequence falls to the same `_ => false` that stands for upstream's `default`.

## Why "evaluate the last element" is not the rule

`{(n, "lit")}` is the discriminator: its last element is a string literal, so a last-element rule judges it defined and drops the guard, while official writes `?? ''` there too. That case is a pattern file and a test.

## Scope of the predicate

`is_js_expr_defined` has three callers — `build_template_chunk` in `shared/utils.rs` plus two more copies of the same chunk logic in `shared/element.rs` (attribute values, element children). All three make the identical `is_defined → ?? ''` decision, so the fix applies uniformly; the attribute slot is covered by its own pattern file and test rather than assumed.

## Server

No port to check. `crates/rsvelte_core/src/compiler/phases/3_transform/server/` contains no `is_js_expr_defined` / `is_defined` guard logic — SSR interpolates through `$.escape(...)`, which handles nullish itself. Confirmed empirically: **0** server divergences in both grids below, before and after.

## Measurement

Both grids compare byte-for-byte against `submodules/svelte`'s compiler on client / client-dev / server.

1. The tags grid from the issue (7 hosts × 10 expression kinds × legacy/runes × 3 targets = 420 cells): **96 → 88** divergent cells, and the 8 that closed are exactly the `seq` cells. Nothing else moved — the remaining 88 are #3322, which lands in #3400.
2. A dedicated grid built for this predicate — 8 slots (`{#each}` text, element children, attribute concatenation, two-interpolation attribute, `style:` directive, class attribute, `<title>`, root text) × 8 expression kinds (5 sequence shapes + 3 controls) × legacy/runes × 3 targets = 384 cells: **16** divergent, all in the `style:` slot **and including its non-sequence control**, so they are pre-existing and unrelated — filed separately (see below).

## Regression tests

`crates/rsvelte_core/tests/sequence_nullish_guard_3323.rs` — 4 tests. Negative control on the reverted tree: 3 fail, and the fourth (a lone `{(n, s)}`, which must **not** gain a guard) passes in both trees.
`compatibility/pattern-corpus/issues/3323-*.svelte` — the same four points, `0 divergent` on client / client-dev / server.

`ssr`, `css`, `validator`, `print` and the full `runtime` suite pass.

## Out of scope, found by grid 2

A `style:` directive whose value has ≥2 interpolations and whose **first** one folds to a known literal loses its `$.template_effect`:

```svelte
<script>let s = "x"; let o = { k: "k" }; void s; void o;</script>
<div style:color="{s}{o.k}"></div>
```
```js
// rsvelte    $.set_style(div, '', {}, { color: `x${o.k ?? ''}` });
// official   let styles;
//            $.template_effect(() => styles = $.set_style(div, '', styles, { color: `x${o.k ?? ''}` }));
```

`style:color="x{o.k}"`, `style:color="{o.k}{s}"` and `title="{s}{o.k}"` are all byte-identical, so it is the style-directive chunk builder specifically. Not touched here.


baseballyama:

Issue #3401, opened 2026-08-22
CSS: type selectors are matched case-sensitively, so `DIV` (and every wrong-case SVG name) is pruned as unused


baseballyama:
## Summary

rsvelte matches a CSS **type selector** against the markup case-sensitively; official matches
case-insensitively. Every wrong-case type selector is therefore pruned as unused, the rule is
replaced by an `/* (unused) … */` comment, and a `css_unused_selector` warning is raised that
official does not raise.

```svelte
<div class="a"><p>x</p></div>

<style>
	DIV { color: red }
</style>
```

```css
/* official */  DIV.svelte-1lj1c2h { color: red }
/* rsvelte  */  /* (unused) DIV { color: red }*/      + css_unused_selector
```

The component silently loses the rule.

## Matrix

20 cases × 2 targets (`client`, `server`; both behave identically). **12 diverge.**

| case | selector | official | rsvelte |
|---|---|---|---|
| uppercase | `DIV` | scoped | pruned |
| mixed case | `Div` | scoped | pruned |
| uppercase compound | `DIV.a` | scoped | pruned |
| uppercase descendant | `DIV P` | scoped | pruned |
| uppercase nested | `DIV { P { … } }` | scoped | pruned (2 warnings) |
| uppercase inside `:is()` | `:is(DIV)` | scoped, no warning | **same CSS text**, one extra warning |
| SVG camelCase, lowercased | `clippath` | scoped | pruned |
| SVG camelCase, uppercased | `CLIPPATH` | scoped | pruned |
| `lineargradient` | (matches `<linearGradient>`) | scoped | pruned |
| `fegaussianblur` | (matches `<feGaussianBlur>`) | scoped | pruned |
| `RECT` | (matches `<rect>`) | scoped | pruned |
| uppercase `DIV` wrapping SVG | `DIV` | scoped | pruned |

**Controls that match (8):** `div`, `clipPath`, `linearGradient`, `feGaussianBlur`, `rect` — every
exact-case selector; `SECTION` — an uppercase selector that genuinely matches nothing, so both
prune it; and `div:not(SPAN)`, where the uppercase name sits inside `:not()` and neither side
prunes.

The `:is(DIV)` row is the sharpest: the emitted CSS is byte-identical and only the **warning** set
differs, so a gate comparing `css.code` alone would score it a match. The `SECTION` control is what
separates "case handling" from "pruning is broken" — pruning works; the comparison is the defect.

## Note

[...]

Issue #3204, opened 2026-08-21
`:is()` / `:where()` / `:has()` argument lists are not pruned


baseballyama:
Official prunes each unused compound out of an `:is()` / `:where()` / `:has()`
argument list, replacing it with an `/* (unused) … */` comment. rsvelte keeps
every argument and scopes them all, so the emitted CSS differs even though the
warning is raised.

```svelte
<b>x</b>
<style>
	:is(b, i) { color: red }
</style>
```

```css
/* official */  :is(b.svelte-1lj1c2h /* (unused) i*/) { color: red }
/* rsvelte  */  :is(b.svelte-1lj1c2h, i.svelte-1lj1c2h) { color: red }
```

It is the argument list that is not pruned, not the rule: with **every** argument
unused, official comments out the whole rule and raises one warning at the
selector's start, while rsvelte keeps the rule and raises one warning **per
argument**.

| markup | official | rsvelte |
|---|---|---|
| `<span>x</span>` + `:is(b, i)` | `/* (unused) :is(b, i) { … }*/`, `css_unused_selector@6:1` | rule kept, `css_unused_selector@6:5` **and** `@6:8` |

Same for `:has()` — `div:has(b, i)` against `<div><b>x</b></div>` keeps `i` —
and for a compound argument (`:is(b.a, i.a)`) and a nested one
(`div :is(b, i)`). `:not()` is correctly left alone by both: its arguments are
not a match list that can be pruned.

Measured with a generated grid over **40 CSS rules × 14 markups × 3 targets**
(1638 cells): **66 diverging, and every one of them is `:is` or `:where`** —
every other rule form, including `:global`, nesting, `@media` / `@supports` /
`@container` / `@layer`, `@keyframes`, an escaped class and a non-ASCII class,
agrees byte for byte.

Issue #3130, opened 2026-08-19
CSS: pseudo-class arguments — An+B gated on the name, and comments dropped when the list is rebuilt


baseballyama:
## Summary

Two independent divergences share one site — `SelectorParser`'s handling of a pseudo-class
argument list. Measured on a 10 pseudo-class names × 7 argument grid: **42 of 70 cells
diverge**.

## (a) The `An+B` branch is gated on the pseudo-class NAME

rsvelte enters its `Nth` path only for `nth-child` / `nth-last-child` / `nth-of-type` /
`nth-last-of-type`. Upstream gates the same branch on `inside_pseudo_class` alone
(`read_selector`, `1-parse/read/style.js` L299), so an `An+B` token is legal in the argument
list of *any* pseudo-class:

```css
.a:is(2n) { }        /* official: OK    rsvelte: css_expected_identifier */
.a:not(2n of .a) { } /* official: OK    rsvelte: css_expected_identifier */
.a:where(5) { }      /* official: OK    rsvelte: css_expected_identifier */
.a:global(2n) { }    /* official: OK    rsvelte: css_expected_identifier */
```

Conversely, inside the four nth names rsvelte's acceptance test is a heuristic —
`trimmed == "even" || trimmed == "odd" || trimmed.contains('n') || any digit || starts with
+/-` — where upstream is a regex with a lookahead:

```
/(even|odd|\+?(\d+|\d*n(\s*[+-]\s*\d+)?)|-\d*n(\s*\+\s*\d+))((?=\s*[,)])|\s+of\s+)/y
```

Nine of 34 `An+B` spellings are over-accepted as a result, every one at the same position
official reports (the token start):

```
-2n-1    -1    2foo    2n /* t */    n+    2n+    2N    2e    3 n
```

Note the fallback that makes some near-misses legal: when the regex does **not** match,
upstream falls through to `read_identifier`, so `:nth-child(-n-1)` is a `TypeSelector` named
`-n-1` and compiles, while `:nth-child(-1)` is rejected by
`REGEX_LEADING_HYPHEN_OR_DIGIT`. Any fix has to reproduce the fallback, not only the regex.

## (b) Comments inside `:is/:not/:has/:where` arguments are dropped

`transform_is_not_args` (`3-transform/css.rs`) rebuilds the argument list from the
transformed children and joins them with `", "`, so every byte the source had between them
is lost. Official's printer edits a copy of the source, so it keeps them:

```
[...]

Issue #2535, opened 2026-08-07
css: 76 css_unused_selector divergences remain after #2474, in five families no fixture can see


baseballyama:
Found while fixing #2474 (PR #2534). That PR built a differential sweep — 15 component templates
× 40 nested stylesheets = 600 generated cases, comparing both `css_unused_selector` warnings and
`css.code` against official 5.56.8 — and used it to show its own fix was sound:

```
main:          250 mismatches
with the fix:   76 mismatches, 0 new
```

The 76 that remain are **not** part of #2474 and have no tracker. They are pre-existing
divergences in distinct families:

| family | shape |
|---|---|
| explicit `&` under a non-ancestor parent | `.a { & .b { … } }` where `.a` is not an ancestor |
| pseudo-class arguments | `:is()` / `:where()` / `:not()` argument selectors |
| `:root` | |
| trailing `:global()` | |
| dynamic attributes | `class={…}`, `{...spread}`, `class:` |

## Why this is worth its own issue

The CSS suite is 181/181 and stays there — PR #2534 also ran a probe comparing rsvelte's
`css_unused_selector` set (code + line + column + text) against `warnings.json` for **all 181
official samples** and found **0 mismatches before and after** its change. So the fixture suite
is saturated: it cannot see any of these 76, and it could not see #2474 either. The generated
sweep is what made them visible, which is #2281's argument (generated shapes, not collected
ones) applied to CSS.

The sweep harness was a throwaway in that PR. The first task here is to decide whether it should
be a committed gate — a small generated CSS matrix alongside `scripts/compat-corpus/matrix/`,
ratcheted the same way — or whether these families should just be fixed one by one against
hand-written fixtures. Landing the count without landing the harness means the next person
re-derives all 600 cases.

## Numbers to re-measure, not inherit

The 250 / 76 figures come from one harness on one day against 5.56.8. Whoever picks this up
should regenerate before splitting the work: family sizes decide the order, and a family that
looks large here may be one predicate.

PR #3475, opened 2026-08-22
fix(compiler): resolve `&` inside `:is()` / `:where()` / `:has()`

### LINKED ISSUE #3133: CSS: a rule whose subject is reached through `:has()` is kept where official prunes it
## Summary

rsvelte keeps rules the official compiler prunes when the subject is reached through `:has()`.
Eight cells of a 516-cell selector grid, all one direction: **rsvelte under-prunes**.

```svelte
<div class="a"><b class="b">x</b></div>

<style>
	.a :has(.b) { color: red }
</style>
```

```css
/* official */  /* (unused) .a :has(.b) { color: red }*/
/* rsvelte  */  .a.svelte-xxx :where(.svelte-xxx):has(.b:where(.svelte-xxx)) { color: red }
```

The same happens with `.a > :has(.b)`, `.a { :has(.b) { … } }`, `.a { & :has(.b) { … } }`,
and with an attribute subject inside `:has()` — `.a:has([x])`, `[x=y]`, `[x="y z"]`,
`[x^="y"]`, `[x|="y"]`, `[x="y" i]`, `[x="y" s]` — plus `.a:has(:has(.b))`.

Note the two shapes differ in what official reports: a **bare** `:has()` compound
(`.a :has(.b)`, no other simple selector beside it) comes out `(unused)` / `(empty)`, while
`.a:has([x])` also comes out `(unused)`. rsvelte emits a scoped rule in every one.

## Why it matters

Under-pruning is the quiet direction: the stylesheet still *works*, so no test that renders
the component can see it. Only output equality does. It also means rsvelte ships CSS bytes
official drops, which is the opposite of the over-prune #2535 was about — so a fix has to
carry both directions, not just make `:has()` prune more.

Found by a generated grid rather than by the collected corpus: the shape needs a `:has()`
whose argument matches nothing in the component, which published code does not contain
(published CSS is written against the markup it ships with).

### LINKED ISSUE #3371: CSS: `&` inside `:is()`/`:where()`/`:has()` is not resolved to the parent, so elements are scoped in both directions
## Summary

`&` inside a **functional pseudo-class** (`:is()`, `:where()`, `:has()`) is not resolved to the
parent selector, so the set of elements that receive the scoping class is wrong — in **both
directions**, depending on where the pseudo-class sits in the complex selector.

The emitted `css.code` is **byte-identical** in 10 of the 12 diverging cases: only `js.code`
(which elements carry `svelte-<hash>`) diverges. So this is not a printing difference — it is
`apply_selector` deciding which nodes a rule matches.

```svelte
<div class="card wide"><p class="a">x</p><span class="b">y</span></div>

<style>
	.card { :is(&) { color: red } }
</style>
```

| | scoped elements |
|---|---|
| official 5.56.9 | `div.card` |
| rsvelte | `div.card`, **`p.a`**, **`span.b`** |

Both emit the same CSS: `.card.svelte-1lj1c2h { :is(&) { color: red } }`. The rule therefore
applies `color: red` to elements the author never selected — `:is(&)` behaves as "matches any
element".

The opposite direction, same construct one combinator to the left:

```svelte
<style>
	.card { :is(&) .a { color: red } }
</style>
```

| | scoped elements |
|---|---|
| official 5.56.9 | `div.card`, `p.a` |
| rsvelte | `div.card` |

Again the CSS agrees — `.card.svelte-1lj1c2h { :is(&) .a:where(.svelte-1lj1c2h) { color: red } }` —
so rsvelte emits a rule whose `.a:where(.svelte-…)` **can never match**, because it never put the
hash on `<p class="a">`. The style silently does not apply.

## Matrix

22 cases × 4 targets (`client`, `server`, `client-dev`, `server-dev`) = 88 cells.
**48 diverging cells / 12 diverging cases**; every diverging case diverges on all four targets.

**Over-scoping** — rsvelte adds the hash to elements official leaves alone:

|---|---|---|
| `.card { :is(&) { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card { :where(&) { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card { :has(&) { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card { :is(&, .b) { … } }` | `.card`, `.b` | `.card`, `.a`, `.b` |
| `.card { :is(&, .nope) { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card.wide { :is(&) { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card { :has(&) .a { … } }` | `.card` | `.card`, `.a`, `.b` |
| `.card { :is(&) { … } }` with a `.card` nested in a `.card` | both `.card`s | both `.card`s, `.a` |

**Under-scoping** — rsvelte omits the hash official adds, so the emitted rule cannot match:

|---|---|---|
| `.card { :is(&) .a { … } }` | `.card`, `.a` | `.card` |
| `.card { :is(&) .b { … } }` | `.card`, `.b` | `.card` |
| `.card { :is(&) > .a { … } }` | `.card`, `.a` | `.card` |
| `.card { :where(&) .a { … } }` | `.card`, `.a` | `.card` |

**Controls that match** (10, so the axis is `&` *inside a functional pseudo-class*, not nesting
and not functional pseudo-classes):

`.card { & { … } }` · `.card { &.wide { … } }` · `.card { & .a { … } }` ·
`.card { :is(.wide) { … } }` · `.card { :is(.nope) { … } }` · `.card { :not(&) { … } }` ·
`.card { .wide { :is(&) { … } } }` · `.card { .a { :is(&) { … } } }` ·
`:is(&) { … }` at top level (no parent) · `.nope { :is(&) { … } }` (parent matches nothing)

`:not(&)` matching is the sharpest control: both compilers scope all three elements there, so
the divergence is specific to `:is` / `:where` / `:has`.

## The `:has(&)` rows also diverge in `css.code` and in warnings

```svelte
<style>
	.card { :has(&) { color: red } }
</style>
```

```css
/* official */  /* (empty) .card { :has(&) { color: red } }*/     + css_unused_selector@3:8
/* rsvelte  */  .card.svelte-1lj1c2h { :where(.svelte-1lj1c2h):has(&) { color: red } }   + no warning
```

That half is the same under-prune as #3204 / #3133 (a subject reached through `:has()` is kept
where official prunes it), and is expected to be fixed with them. The **element-scoping**
divergence above is separate: it is present on `:is` and `:where`, where nothing is pruned by
either side and the CSS text agrees exactly.

## Not covered by an existing issue

- #3204 is the **argument list** not being pruned inside `:is()`/`:where()`/`:has()`. Here the
  argument list agrees (`:is(& /* (unused) .nope*/)` is byte-identical on both sides) and the
  markup does not.
- #3133 is the `:has()` under-prune; it does not reach `:is`/`:where`, and the eight
  over-scoping rows include four with no `:has()` at all.
- #3166 (closed) was `propagate_ancestor_scoping` scoping an ancestor whose combinator chain
  does not match — the trigger there was *another* rule scoping the subject; here there is one
  rule and no other.

## Reproduction

`p-ampis4.mjs`-style probe: for each case compile the markup + `<style>` on all four targets and
compare `js.code`, `css.code` and the warning list. Against `submodules/svelte` at 5.56.9 and the
NAPI binding at `3516c2f`.

baseballyama:
Closes #3371. Closes #3133.

> **Stacked on #3424** (`fix/3130-3133-3204-css-pseudo-class-args`), which this branch was cut
> from — its three commits are the first three here. Merge #3424 first. If it lands as a **squash**,
> rebase this branch before merging it (`git rebase --onto origin/main 896d61eaf`), otherwise the
> shared commits come back as duplicates.

## What was wrong

`&` inside a **functional pseudo-class** was never resolved. Both phases have a port of
upstream's `get_relative_selectors`, and both looked for `&` **one level deep** — over the
compound's simple selectors, never into a pseudo-class's argument list. Upstream finds it with a
zimmerframe `walk`, which descends into `args`. So for `.card { :is(&) { … } }` rsvelte
concluded "this rule has no `&`", and two things followed from that one wrong answer:

1. it prepended the implicit parent, making the rule `.card :is(&)` instead of `:is(&)`;
2. it left the `&` in place, where the element matcher reads a `NestingSelector` it cannot
   resolve as *matches anything*.

The result is wrong in **both directions**, and #3371's measurement is the part worth keeping:
the emitted CSS was **byte-identical to official in ten of the twelve diverging cases** — only
the set of elements carrying `svelte-<hash>` moved. Over-scoping when the pseudo-class is the
subject (`:is(&)` scoped every element in the component), under-scoping when it has a descendant
(`:is(&) .a` scoped none of them, so the emitted `.a:where(.svelte-…)` could never match and the
style silently did not apply). Nothing that *renders* the component can see either one; only
`js.code` equality can.

## The fix

**Phase 2** (`css_scoping.rs`): `substitute_nesting` now asks `complex_has_nesting`, which
descends into `PseudoClass(_, Some(args))`, and substitutes through the argument list
(`substitute_nesting_in_args`) before the compound itself. That is one predicate deciding both
"is the parent prepended" and "what does `&` become", which is upstream's shape.

**Phase 3** (`3_transform/css.rs`): the `:has()` subject is now taken **under every host chain**
the enclosing rules resolve to (`has_pseudo_unused_under_every_host` / `effective_host_chains`),
and an `&` inside an argument list is substituted before the branch is judged
(`branch_alternatives`, `replace_nesting_in_complex`).

**The host chains come from `build_parent_chains`, which was already there** — this file's port
of `get_relative_selectors`. Routing `:has()` through it is a reuse, not a second
implementation; writing the implicit-`&` prepend a second time is exactly the shape #3204 was
about, one directory over. Ditto the branch verdicts, which go through the `unused_branches`
marking set #3204 introduced rather than being recomputed.

Because a substituted branch already carries the parent, it must not *also* be required to sit
below it — `without_parent_preludes` states that, and getting it wrong is measurable rather than
theoretical: the first version of this patch scored **11/23** on the grid where the previous
commit scored 19/23.

## #3133's last cell: a `:has()` inside a `:has()` argument

`.a:has(:has(.b))` was kept where official prunes it, for an unrelated reason that only became
reachable once the rest of #3133 was fixed. `is_has_argument_unused` bails out with "this
argument has no tag/class/id, so it can match anything" — true of `:focus-visible`, false of
`:has(.b)`, which constrains the element as much as a class does. The argument now gets its own
subject set (the elements the outer argument can reach, via `elements_matching_relative`) and
the inner `:has()` is resolved against *those*.

## Measurements

| gate | before | after |
|---|---|---|
| #3371's own 22 cases × 4 targets (`js` + `css` + warnings) | 48 diverging cells | **88/88 cells identical** |
| #3133's 19 shapes × 3 targets | 3 diverging | **57/57 identical** |
| #3130 / #3204 / #3133 grid (148 cells) | 147/148 | **148/148** |
| `css-prune-sweep --check` (1969 components) | 1969 matched | **1969 matched, 0 diverged** |
| attractions + layercake + sveltestrap + `svelte/tests` (534 styled files) | — | **0 diverging** |
| CSS fixtures + `css_pseudo_class_args` + validator | pass | pass |

The over-prune direction is the one that matters here — #2535 shipped an over-prune that three
real `svelte.dev` components reproduced — so every sweep above is a two-sided check, and each
new test carries its positive control: `:not(&)` scopes all three elements on both sides,
`.a:has(:has(.b))` is kept when a descendant *does* have a `.b`, and `.card { :has(&) }` is kept
when a `.card` really is nested in a `.card`.

The `scoped()` test helper asserts it found a `data-n` before reporting an empty set, because
"nothing was scoped" is also what a probe that cannot read the template says.

## Not closed by this PR

Two shapes outside both issues' matrices still diverge, both filed nowhere yet:

- `.card { .a:is(&) { … } }` — official prunes the rule, rsvelte keeps it. The verdict needs
  `is_simple_selector_unused` to read the `unused_branches` marking set instead of its own
  isolated check, and that function is @cmtA's in #3401; the switch has an ordering hazard
  (the marking walk itself calls it), so it belongs in that PR, not this one.
- `.card { :is(& .a) { … } }` — official additionally scopes `.b`; rsvelte does not. The CSS
  text agrees.


baseballyama:
Follow-up commit `0fd637a` — **the first version of the nested-`:has()` half over-pruned**, and the
input that separates it from the correct behaviour is one level of nesting.

`selector_matches_element` answers "matches nothing" for a compound with no tag/class/id, so the
candidate set built for a `:has()` argument that is itself a `:has()` came out **empty**, and an
empty candidate set reads as "unreachable". On the flat markup the fix was written against
(`<div class="a"><b class="b">`) that is the right answer by coincidence — official prunes there
too. On `<div class="a"><div class="mid"><b class="b">` official **keeps** the rule and this
pruned it.

No gate saw it: `css-prune-sweep` stayed 1969/1969, the matrix stayed green, and the 148-cell
grid stayed 148/148. What caught it was the negative control written into the test in the same
commit as the positive one — `.a:has(:has(.b))` asserted **kept** when a descendant really does
have a `.b`. That is the #2535 direction, and it is only visible if the test carries both
directions of the same rule.

Re-measured after the fix, on the rebuilt binary:

| gate | result |
|---|---|
| #3371's 22 cases x 4 targets | 88/88 identical |
| #3133's 19 shapes x 3 targets | 57/57 identical |
| #3130/#3204/#3133 grid | 148/148 |
| `css-prune-sweep --check` | 1969 matched / 0 diverged |
| `submodules/` + `compatibility/` sweep | 686 compared / 0 diverging |
| `cargo test -p rsvelte_core --test css --test validator --test css_pseudo_class_args` | 2 + 3 + 25 pass, EXIT=0 |

baseballyama:
Two independent findings from verifying this locally. Neither says the PR is wrong about what it set out to fix — `:has(&)` on a multi-compound parent genuinely improves — but one of them blocks the merge and the other is a hard CI failure.

## 1. The suite holds the parent's compound count fixed at one

`.card.wide` is a **single compound carrying two classes**, not a chain, so every nesting row in `tests/css_pseudo_class_args.rs` has a single-compound parent. `nesting_substitute_alternatives` documents keeping only the chain's last compound and dropping the ancestor requirement — which is only observable when the parent *is* a chain.

That is the #2535 shape recorded in `CLAUDE.md`: a 1,955-row generated grid stayed green while an over-prune shipped, because the shape needed a two-compound parent **and** a subject `&`, and every row its author wrote had a single-compound parent.

Varying only that axis:

```svelte
<style>
  .card .inner { :is(&) { color: red } }
</style>
```

with markup carrying an `.inner` **inside** `.card` (`in`) and one **outside** it (`out`):

| tree | scoped set |
|---|---|
| official 5.56.9 | `["card", "in"]` |
| `main` `c75956b20` | `["a", "card", "in"]` — no `out` |
| **this PR** `a6240ac7a` | `["card", "in", "out"]` — **`out` added** |

The same for `:where(&)`, for `.card > .inner`, and for the two-level `.card { .inner { :is(&) … } }`.

**It cuts both ways, and that is the honest summary.** Divergence count is 9 of 10 on both trees, and each tree gets exactly **one** row right — a different one:

- this PR **fixes** `:has(&)` on a multi-compound parent — `main` scoped all six elements (`["a","a2","card","in","other","out"]`), this PR gives `["card","in"]`, matching official exactly
- this PR **breaks** `:is(&)` / `:where(&)` in the other direction by adding `out`

So the failure mode moved rather than closed: `main` under-scopes and misattributes `a`, this PR over-scopes `out`.

### ⚠ Why "main also diverges 9/10" cannot be used as a defence

The probe carries a `:not(&)` control, because `:not` is the one functional pseudo where `out` *should* be scoped. It is **live on this PR and dead on `main`**:

```
.card .inner { :not(&) { color: red } }
  official  : ["a","a2","card","in","other","out"]
  this PR   : ["a","a2","card","in","other","out"]   <- control passes
  main      : ["a","card","in"]                      <- `out` absent, control FAILS
```

`main`'s `:not(&)` is independently broken, so that tree fails the very check that proves the probe is reading this defect. **The two 9s are not commensurable** and the `main` column should not be cited as "no worse than before".

## 2. `Formatter parity` fails on this PR's own repro, and it is a hard gate

```
[fmt-verify] 1 NEW failures (not in baseline)
  - pattern/issues/3130-nth-of-in-any-pseudo-class.svelte [diff] line 4
      oracle:   .a:is(2n) {
      actual:
```

The rule is not dropped — line 4 is a **blank line** the formatter inserts, and the rule sits at line 5 unformatted. Cause traced to `crates/rsvelte_formatter/src/style_css.rs:81-82`: the CSS parse-error fallback returns the body verbatim, but the body begins at the newline after `<style>` while the caller splices its own. Filed as #3628; #3629 fixes that half.

This is **not** caused by this PR's compiler changes — `rsvelte_formatter` sets `defer_script_parse: true`, so `parse_css_strict` never runs under the formatter and `1_parse/read/style.rs` is not on that path. The repro is simply the first input to reach the defect. Note that `fmt-known-failures.json` / `fmt-oracle-excluded.json` contain **zero** pattern-corpus entries, so listing it has no precedent.

The other two check failures are not code: `Language server` died on a binaryen 117 download, and `Tests` is the aggregator for that same run.

## What passed

`cargo test -p rsvelte_core --test css` — **181 compiled, 181 matched**, read from the run's own counter. No regression on the official CSS suite.

## Suggested next step

The 12 falsifier rows plus the live `:not` control are written and their expectations come from the pinned oracle (5.56.9 — asserted from the runtime `VERSION` export on every invocation, because the npm `svelte` on disk resolves to 5.56.10). They are not pushed to this branch; say the word and they can be handed over. Adding them would make the multi-compound-parent axis part of the suite, so whichever direction the fix goes it stays covered.

PR #3481, opened 2026-08-22
fix(client): give a custom-element attribute its own memoizer

### LINKED ISSUE #3479: client: a custom-element attribute skips the memoizer, so an `await` in its value is inlined into a non-async arrow
`<async-custom-element bar={await 'bar'}>` compiles to client output that no JavaScript parser accepts:

```js
$.template_effect(() => $.set_custom_element_data(async_custom_element, 'bar', await 'bar'));
```

`await` inside a non-async arrow. OXC rejects it with ``` `await` is only allowed within async functions and at the top levels of modules ```.

## Which side is wrong, measured rather than assumed

Running the **official** compiler on the same input answers it, and the answer changed between two upstream patch releases:

| compiler | output for the `bar` attribute |
|---|---|
| official 5.56.9 | `$.template_effect(() => $.set_custom_element_data(el, 'bar', await 'bar'))` — same unparseable form |
| official 5.56.10 | `$.template_effect(($0) => $.set_custom_element_data(el, 'bar', $0), void 0, [() => 'bar'])` |
| rsvelte (before this fix) | matches 5.56.9 |

So this is **not** a divergence rsvelte introduced — it mirrored official byte-for-byte. Upstream fixed it in 5.56.10, and `packages/svelte/tests/runtime-runes/samples/async-custom-element-attribute` is upstream's own regression test for that fix. It does not exist at 5.56.9, which is why the defect only surfaces once the submodule is bumped (#3195).

## The one line that separates the paths

`RegularElement.js` `build_custom_element_attribute_update_assignment` gives a custom-element attribute its **own** `Memoizer` and passes `sync_values()` / `async_values()` / `blockers()` to `template_effect`. rsvelte took that one branch through an identity closure and a bare thunk instead:

```rust
let result = build_attribute_value(&attr.value, context, |expr, _metadata| expr);
…
vec![b::thunk(&context.arena, call)]
```

Every other attribute path already routes through a memoizer — the existing `async-attribute` fixture emits `[async () => $.clsx(await …)]` correctly. **One branch bypassed it, and the existing fixtures stayed green**, which is why nothing caught it until upstream added a sample that combines a custom element with an awaited attribute.

This is the `CLAUDE.md` "output no JS parser accepts" class: the output is not merely a text mismatch, it is not JavaScript.

## Blast radius

The bypass also dropped `sync_values` and `blockers`, so a custom-element attribute with state diverged from official even without an `await` — the second `template_effect` argument was missing entirely.

baseballyama:
Closes #3479.

`<async-custom-element bar={await 'bar'}>` compiled to

```js
$.template_effect(() => $.set_custom_element_data(el, 'bar', await 'bar'));
```

`await` inside a non-async arrow — output that is not JavaScript, not merely output that differs.

## The side that was wrong is not the one it looks like

My first reading was "rsvelte skips the memoizer where upstream uses it, so rsvelte is broken." Running the **official** compiler on the same input refuted the second half:

| compiler | `bar` attribute |
|---|---|
| official **5.56.9** | `$.template_effect(() => $.set_custom_element_data(el, 'bar', await 'bar'))` — the same unparseable form |
| official **5.56.10** | `$.template_effect(($0) => $.set_custom_element_data(el, 'bar', $0), void 0, [() => 'bar'])` |
| rsvelte, before | matches 5.56.9 |

rsvelte was mirroring official byte-for-byte. **Upstream fixed this in 5.56.10**, and `runtime-runes/samples/async-custom-element-attribute` is upstream's own regression test for that fix — it does not exist at 5.56.9. That is why the defect appears only when the submodule is bumped (#3195), and why this is a port of an upstream change rather than a repair of a local one.

## One branch, and why the fixtures stayed green

`RegularElement.js` `build_custom_element_attribute_update_assignment` gives the attribute its own `Memoizer` and passes `sync_values()` / `async_values()` / `blockers()` to `template_effect`. rsvelte took that branch — and only that branch — through an identity closure and a bare thunk:

```rust
let result = build_attribute_value(&attr.value, context, |expr, _metadata| expr);
…
vec![b::thunk(&context.arena, call)]
```

Every other attribute path already memoizes; the existing `async-attribute` fixture emits `[async () => $.clsx(await …)]` correctly and passes. **The bypass was invisible to the fixture population** until a sample combined a custom element with an awaited attribute.

The bypass also dropped `sync_values` and `blockers`, so a custom-element attribute carrying state diverged from official with no `await` involved at all — the second `template_effect` argument was simply absent.

## Verification

- Output for the upstream sample is **byte-identical to official 5.56.10** (`diff` against the compiler run from the 5.56.10 submodule commit, exit 0).
- Two regression tests, both **confirmed to fail without the fix**. The first version of the second test did *not* — `output.contains("async () =>")` passes on the broken output too, because the top-level `$derived(await …)` contributes one from `$.run([async () => …])`. Rewritten to assert that no `await` survives inside the `set_custom_element_data` call, and that `[() => 'bar']` reaches the async slot.
- The expected shape was read off the upstream compiler, not off this port. Written from my own output it would have said `[async () => await 'bar']`, which official does not emit — `async_arrow` collapses `await X` to `() => X` when `X` holds no further await. **A test written that way would have been green while the output was wrong.**


baseballyama:
Superseded by #3195, which carries this branch's `regular_element.rs` byte-for-byte (`8427d8429` == `2fad46841`).

This PR cannot go green on its own. The change ports 5.56.10's `build_custom_element_attribute_update_assignment`, but this PR's merge ref pins the submodule at 5.56.9, so `Compiler parity` compares 5.56.10 output against a 5.56.9 oracle:

```
snapshot/samples/dynamic-attributes-casing/main.svelte [js-mismatch] (client, client-dev)
  expected: $.template_effect(() => $.set_custom_element_data(custom_element_1, "fooBar", y()));
  actual:   $.template_effect(($0) => $.set_custom_element_data(custom_element_1, "fooBar", $0), [() => y()]);
```

The "expected" side is 5.56.9, which the PR description already identifies as the version that emits the unparseable form. Only the branch that bumps the submodule can hold this change, and that is #3195.

Leaving this open until #3195 lands, then closing it — no work is lost.

baseballyama:
Closing — #3195 landed and carries this change.

Verified against `main` at the time of closing, file by file:

- `.changeset/custom-element-attribute-memoizer.md` — byte-identical
- `crates/rsvelte_core/tests/custom_element_await_attribute.rs` — byte-identical
- `crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/regular_element.rs` — the memoizer change is present (`Memoizer::with_parent_conflicts(&context.state.memoizer)` and its ungrouped `template_effect`, with the comments this PR wrote). The only remaining difference from this branch is in the opposite direction: `main` has since refactored the element-name lowercasing to `super::shared::utils::html_lowercase(name_str)`, where this branch still carries the older inline `to_lowercase()` form.

So merging this now would revert someone else's refactor while adding nothing, and its `Compiler parity` could never have gone green on its own: the expected outputs it is diffed against were generated by Svelte 5.56.9, while the fix requires 5.56.10 — the version bump that only #3195 carried.

No work is lost. The implementation, the test, and the changeset are all in `main`.

PR #3485, opened 2026-08-22
fix(preprocess): hand hooks their attributes in source order

### LINKED ISSUE #3397: preprocess(): the `attributes` object passed to script/style hooks is in hash order, not source order
## Summary

The `attributes` object handed to a preprocessor's `script` / `style` hook has its keys in an
arbitrary (hash) order instead of source order.

```svelte
<script context="module" foo="bar">let a = 1;</script>
```

```js
// the `attributes` argument the script hook receives
official: { "context": "module", "foo": "bar" }
rsvelte:  { "foo": "bar", "context": "module" }
```

Key order is observable in JavaScript — `Object.keys`, `Object.entries`, `for…in` and
`JSON.stringify` all expose it — so a preprocessor that rebuilds a tag from `attributes`, or hashes
them for a cache key, produces different output under the two compilers.

## It is not a sort, and it is stable

| source attributes | official | rsvelte |
|---|---|---|
| `context="module" foo="bar"` | `context, foo` | `foo, context` |
| `foo="bar" context="module"` | `foo, context` | `foo, context` (agrees) |
| `c="3" a="1" b="2"` | `c, a, b` | **`a, c, b`** |
| `c="3" a="1" b="2" d="4" e="5"` | `c, a, b, d, e` | **`e, d, a, c, b`** |
| `lang="ts" zz="1" aa="2"` | `lang, zz, aa` | `lang, aa, zz` |
| `async defer lang="ts"` | `async, defer, lang` | same (agrees) |
| `lang="scss" zz="1" global` (style) | `lang, zz, global` | same (agrees) |

`a, c, b` is neither alphabetical nor source order, and `e, d, a, c, b` from `c a b d e` rules out
both as well — the order tracks the **key set**, which is the signature of a hash map. It is
deterministic: three separate processes produced the identical order, so the hasher is fixed.

The two agreeing rows are the controls — the same hook, the same plumbing, a key set whose hash
order happens to coincide with source order.

## Where it was found

A `preprocess()` grid: 7 preprocessor shapes × 15 source shapes = 105 cells, comparing both the
resulting `code` and a log of every hook invocation with its arguments. **3 cells diverge**, all of
them this one source shape, and all on the hook-argument half — the `code` matches everywhere.
Everything else about the entry point agreed: which regions go to which hook, the order the hooks
run in, `filename`, the `markup` argument, multiple groups, hooks that return nothing, and every
edge source (empty `<script>`, self-closing `<script src>`, two `<style>` blocks, CRLF, no trailing
newline, a `<style>` spelled inside markup text).

baseballyama:
Closes #3397

## What

The map behind a preprocess attribute set was `FxHashMap`, so the `attributes` object handed to a `script` / `style` hook came out in hash order. Upstream's is a plain JS object filled by the attribute regex in source order (`parse_tag_attributes`) and stringified through `Object.entries` (`stringify_tag_attributes`), so insertion order is the contract on both sides of the hook.

`PreprocessAttributeMap` is now `IndexMap<K, V, FxBuildHasher>` — insertion-ordered, same hasher — with an `AttributeMap` alias for the applied `String`/`AttributeValue` pair. Every consumer already named the exported alias, so `rsvelte_preprocess` and the tests needed no change; only the NAPI bridge, which had spelled `FxHashMap` out again, had to follow.

## Measurement

Oracle: `svelte.preprocess` from `submodules/svelte` (5.56.9), driven with a hook that logs `Object.keys(attributes)`. Nine attribute layouts, plus two no-attribute controls.

| source attributes | official | before | after |
|---|---|---|---|
| `context="module" foo="bar"` | `context, foo` | **`foo, context`** | `context, foo` |
| `foo="bar" context="module"` | `foo, context` | `foo, context` | `foo, context` |
| `c="3" a="1" b="2"` | `c, a, b` | **`a, c, b`** | `c, a, b` |
| `c="3" a="1" b="2" d="4" e="5"` | `c, a, b, d, e` | **`e, d, a, c, b`** | `c, a, b, d, e` |
| `lang="ts" zz="1" aa="2"` | `lang, zz, aa` | **`lang, aa, zz`** | `lang, zz, aa` |
| `async defer lang="ts"` | `async, defer, lang` | `async, defer, lang` | `async, defer, lang` |
| `lang="scss" zz="1" global` (style) | `lang, zz, global` | `lang, zz, global` | `lang, zz, global` |
| `z="1" y="2" x="3" w="4"` (style) | `z, y, x, w` | `z, y, x, w` | `z, y, x, w` |
| `m="1" l="2" k="3" j="4" i="5" h="6"` | `m, l, k, j, i, h` | `m, l, k, j, i, h` | `m, l, k, j, i, h` |

The rows that already agreed are kept in the test as controls — the same hook and the same plumbing, on key sets whose hash order happens to coincide with source order. Reverting the type change on this tree makes the first row fail with `["foo,context"]` against `["context,foo"]`, so the test moves for the reason claimed.

## A second symptom the issue does not report

The issue says the resulting `code` matched everywhere, because its preprocessors returned nothing. `stringify_tag_attributes` iterates the same map, so a hook that **returns** `attributes` had them written back in hash order too: building `zz`, `aa`, `mm` and returning them produced `<script aa="2" zz="1" mm>` where upstream emits `<script zz="1" aa="2" mm>`. That is now the second test, and it is the half a preprocessor actually ships to disk.

## Which gate sees this

None, before this PR. The corpus comparator reads `js.code` / `css.code` of a **compiled** component; the preprocess hook arguments are not an artifact any gate collects, and the returned-attribute write-back only reaches output through a preprocessor that sets `attributes`, which no corpus source does. The two new tests are the gate.

## Tests

`crates/rsvelte_core/tests/preprocess_attribute_order.rs` — 2 tests. Existing `preprocess`, `preprocess_attributes` and the whole `rsvelte_preprocess` suite pass unchanged.

Note on the public API: `PreprocessAttributeMap` keeps its name and its two type parameters, so downstream spellings of the alias are unaffected; the underlying type changes from `HashMap` to `IndexMap`, which is a deliberate break under the project's internal-API policy and is the fix itself.


baseballyama:
## This PR is now stale against `main`, and its own test proves it

Merged into a 29-PR integration branch and run with the full `rsvelte_core` suite, this PR's own new test fails:

```
---- returned_attributes_are_written_back_in_insertion_order ----
  left:  "<script q=\"0\">let a = 1;</script>"
  right: "<script zz=\"1\" aa=\"2\" mm>let a = 1;</script>"
```

The returned attributes are not written back at all.

### Cause: #3340 landed on `main` and decided the opposite question

`7e500f386 fix(preprocess): catch a throwing preprocessor, and discard an attribute-only result (#3340)` (2026-08-24T07:36:59+09:00) rewrote the early-return guard in `preprocess/mod.rs` and left this comment on it:

> Upstream discards the whole result here, `attributes` included, so an attribute-only change never takes effect. Re-emitting the tag for it replaces the attribute list wholesale and drops `module` / `lang`, which changes what the component compiles to.

So on current `main` a hook that returns only `attributes` (no `code` change, no map) is discarded by design. This PR's test asserts the opposite.

### What is *not* in dispute

The rest of this PR — `FxHashMap` → an insertion-ordered `AttributeMap` in `parse_tag_attributes` / `stringify_tag_attributes`, and the matching napi change — is orthogonal to #3340 and is what the PR title describes. Upstream stringifies `Object.entries`, i.e. insertion order, so that part is right. **The `test_stringify_tag_attributes` unit test in `mod.rs` (` lang="ts" defer` in exactly that order) passes.**

Only `returned_attributes_are_written_back_in_insertion_order` in `tests/preprocess_attribute_order.rs` conflicts, and it conflicts on a question this PR did not set out to answer.

### Suggested resolution

Rebase onto `main` and either drop that one test or restate it against the post-#3340 contract (e.g. a hook that changes `code` *and* returns `attributes`, which is not short-circuited). The insertion-order change can then land on its own merit.

### How this was measured

`origin/main@5349df880` + 29 open PRs, `cargo test --release -p rsvelte_core -p rsvelte_esrap --no-fail-fast` → 3469 passed / 5 failed. Four of the five other failures have separate causes (#3527's unpinned diagnostic, and two tests whose expectations upstream contradicts); this one is the only one attributable to a semantic conflict between two PRs.

I have **not** verified this PR in isolation against `main`, so "the merge lost something" is excluded by reading the diffs, not by a second run: this PR's own diff does not touch the guard #3340 rewrote.

PR #3493, opened 2026-08-22
fix(compiler): lower a class expression in a rune argument and in extends

### LINKED ISSUE #3071: client: a class expression inside a rune argument is not lowered (and loses $.proxy)
A class expression that appears **inside a rune's argument** is not lowered: its own `$state` fields keep their public names and get no accessor pair. For `$state(class { … })` the `$.proxy(…)` wrapper is dropped as well.

`T.svelte.js` (also reproduces in a component's instance script):

```js
export class A {
	held = $state(class {
		deep = $state(1);
	});
}
```

client, prod:

```js
// rsvelte
	#held = $.state(class {
		deep = $.state(1);
	});

// official
	#held = $.state($.proxy(class {
		#deep = $.state(1);

		get deep() {
			return $.get(this.#deep);
		}

		set deep(value) {
			$.set(this.#deep, value, true);
		}
	}));
```

Two separate differences in one shape:

1. the nested class body is never handed to the class-field transform, so `deep` stays a plain public field — the emitted class has no reactive `deep` at all;
2. `$.proxy` is missing around the class expression.

`new (class { deep = $state(1); })()` inside a rune argument (`held = $state(new (class { … })())`) reproduces (1) the same way.

This is adjacent to, but not the same site as, the splitter fix for a parenthesised nested class in a plain field initializer: there the class body reaches the member scan and is transformed, here the enclosing member is itself consumed as a rune field and its argument text is copied through verbatim.

Reference: `submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/ClassBody.js` — upstream visits the initializer expression, so a nested `ClassBody` is reached by the ordinary walk.

### LINKED ISSUE #3072: client: an inline class expression in `extends` is parenthesised and the subclass's rune fields are not lowered
When a class's `extends` clause is an **inline class expression**, rsvelte wraps it in parentheses and then fails to lower the subclass's own rune fields.

`T.svelte.js` (also reproduces in a component's instance script):

```js
export class Sub extends class {
	inline = $state('i');
} {
	own = $derived(this.inline + '!');
}
```

client, prod:

```js
// rsvelte
export class Sub extends (class {
	#inline = $.state('i');

	get inline() { … }
	set inline(value) { … }
}) {
	own = $.derived(() => this.inline + '!');
}

// official
export class Sub extends class {
	#inline = $.state('i');

	get inline() { … }
	set inline(value) { … }
} {
	#own = $.derived(() => this.inline + '!');

	get own() {
		return $.get(this.#own);
	}

	set own(value) {
		$.set(this.#own, value);
	}
}
```

Two differences:

1. the added `(` `)` around the heritage expression — a byte divergence that survives oxfmt normalization;
2. `own = $derived(…)` in the **subclass** body is left as a plain public field, so the class has no reactive `own`. The heritage class's own field is lowered correctly, so the member scan is treating the inline heritage body as the class body and never reaches the real one.

Reference: `submodules/svelte/packages/svelte/src/compiler/phases/3-transform/client/visitors/ClassBody.js`.

baseballyama:
Closes #3071
Closes #3072

Both issues are the same shape one level apart: upstream reaches every `ClassBody` through the ordinary zimmerframe walk, and rsvelte's class-field lowering is a text scan that never re-enters a region it has already consumed. They share the fix surface, so they land together.

## #3071 — a class expression inside a rune argument

The member scan takes a rune field's argument as **text** (`ClassStateField::value`) and copies it through verbatim, so a `ClassBody` inside one is never scanned. `transform_nested_class_expression` re-enters it.

Two independent halves, and the second is not implied by the first:

```js
export class A {
	held = $state(class { deep = $state(1); });
}
```

| | before | official |
|---|---|---|
| the nested body | `deep = $state(1)` — a plain public field, and in a **component's instance script** a live reference to the `$state` global (output that parses and throws on first render) | `#deep = $.state(1)` + accessor pair |
| the wrapper | `$.state(class { … })` | `$.state($.proxy(class { … }))` |

The missing `$.proxy` is its own cause: `expression_needs_proxy` is an **allow-list** where upstream's `should_proxy` is a **deny-list** (`Literal` / `TemplateLiteral` / arrow / function / unary / binary / `undefined`), and `ClassExpression` is on neither, so upstream proxies it and rsvelte did not. Only the `$state` arm is touched; `$derived` has its own pattern file to pin that.

## #3072 — an inline class expression in `extends`

```js
export class Sub extends class { inline = $state('i'); } { own = $derived(this.inline + '!'); }
```

Two independent halves again.

**1. Which brace opens the body.** The client transform located it with `memmem::find(b"class ")` + `.find('{')` — the first `{` after the keyword, which is the *inline superclass's*. So the member scan lowered the heritage body and left the subclass's own rune fields alone. It now goes through the lexical `find_class_header` (the `#2986`-hardened scan the server already used), which learns to count the brace-delimited bodies a header can contain — `extends class { … }` and `extends function () { … }`, stacked — and reports where the `extends` keyword ended so the superclass expression can be lowered on its own.

**The server has the same defect and is fixed in the same commit** (`transform_script.rs`), per the #3027 rule; it was measured, not assumed — see below.

**2. The parentheses.** `esrap` prints `class [id ][extends <visit superClass> ]{…}`, visiting the superclass with **no parenthesisation at all**; rsvelte ran it through `child_with_parens(…, 19)`, which parenthesises anything below a `MemberExpression`. `extends` takes a `LeftHandSideExpression`, so a `ClassExpression` / `FunctionExpression` / `ObjectExpression` needs none — they are `PrimaryExpression`s — and those now print bare, matching official. This half is independent of runes: `class Sub extends class { inline = 1; } {}` reproduced it.

The other sub-`MemberExpression` shapes keep their parentheses, because dropping them is how official emits `class E extends Base ?? class {} {}` — text no parser accepts. That is recorded in `compatibility/deliberate-divergences.md` with the measured grid and pinned by a test.

## Measurement

A generated grid of 25 class shapes × {`.svelte.js` module, component instance script} × {client, client-dev, server} = **150 cells**, compared byte-for-byte against `submodules/svelte`'s compiler:

- before: **62** divergent cells
- after: **6**, and all six are `static n = $state(0)` — a case **both** compilers reject with the same code (`state_invalid_placement`), where only my harness's error formatting differs. No output cell diverges.

The 56 closed cells cover: the rune-argument class (client + client-dev, module and component), `new (class …)()` inside the argument, a `$derived` argument, an inline heritage class named and anonymous, a heritage class with a plain subclass, a heritage class with a plain base, two stacked heritage classes, and an already-parenthesised heritage class — on client, client-dev **and server**.

### Server (the #3027 rule)

Measured, not asserted. `#3072` reproduces on the **server** (`mod_extends_inline [server]` and five sibling rows were divergent before, identical after) because the client and server transforms share `find_class_header`. `#3071` does **not**: the server's own grid rows for every rune-argument case were identical before and after, since SSR lowers `$state(x)` to `x` and needs no proxy or accessor pair.

## Regression tests

- `crates/rsvelte_core/tests/nested_class_expression_3071_3072.rs` — 8 tests, including the server half of #3072, the `$derived`-must-not-proxy control, and the parenthesisation grid with both its matching and its deliberately-diverging rows.
- `compatibility/pattern-corpus/issues/307{1,2}-*` — 9 files, `0 divergent` on client / client-dev / server. The control (`3072-extends-shapes-legal.svelte.js`) carries a name, a call, a member, a sequence and a function expression as superclasses, so the printer change is priced rather than merely described.

## Also in this PR

`compile_one` routes a `.svelte.js` / `.svelte.ts` path through `compile_module`. Without it neither issue can be reproduced from a single file, since `compileModule` is a different entry point from `compile`.

## Checks

`cargo fmt --all` and `cargo clippy -p rsvelte_core -p rsvelte_esrap -p rsvelte_devtools --all-targets --all-features -- -D warnings` both run clean. `rsvelte_esrap` (96 tests), `rsvelte_core` `ssr` / `css` / `validator` / `print` / `compiler_fixtures` all pass. The pre-commit hook is skipped (`--no-verify`) because a full-workspace clippy does not finish in the available window on this box.


baseballyama:

Issue #3473, opened 2026-08-22
Phase 2 matches upstream's `globals` table by PREFIX, so `Math.hypot(...)` drops a `?? ''` — 18 cells


baseballyama:
Upstream's `globals` table (`phases/scope.js`) is an **exact key set** — 35 `Math.*`
functions, 6 `Number.*`, `Number`, `String`, `String.fromCharCode`,
`String.fromCodePoint`, `BigInt` — and a keypath outside it evaluates to
`UNKNOWN`, so `scope.evaluate(...).is_defined` is false and an interpolation of
it keeps its `?? ''`.

rsvelte has **two ports of that table**. The client's is exact
(`3_transform/client/visitors/shared/utils.rs:4405`, name for name). Phase 2's
is a **prefix match**:

```rust
// 2_analyze/visitors/variable_declarator.rs:15
fn is_known_defined_global_call(keypath: &str) -> bool {
    keypath.starts_with("Math.")
        || keypath == "Number"
        || keypath.starts_with("Number.")
        || keypath == "String"
        || keypath == "String.fromCharCode"
        || keypath == "String.fromCodePoint"
        || keypath == "BigInt"
}
```

Its doc comment says it "mirrors `is_known_defined_global_call` in the client
transform". It does not: `Math.` and `Number.` are open prefixes there and closed
key sets upstream. So every `Math.<anything>` and `Number.<anything>` call sets
`binding.initial_is_defined`, `identifier_is_defined` then reports the binding as
provably non-nullish, and the `?? ''` coercion guard is dropped.

```svelte
<script>
	const g = Math.hypot(3, 4);
</script>
<b class="a{g}"></b>
```

```js
// official
	$.template_effect(() => $.set_class(b, 1, `a${g ?? ''}`));
// rsvelte
	$.template_effect(() => $.set_class(b, 1, `a${g}`));
```

## Measured scope: 18 client cells

Generated grid, 17 global keypaths × 4 hosts × 2 targets (136 cells), official vs
rsvelte on generated JS. Of the six keypaths that are prefix-matched but absent
from upstream's table, **all three interpolating hosts diverge and none of the
others do**:

| keypath | `class="a{g}"` | `x{g}` | `title="a{g}"` |
|---|---|---|---|
| `Math.hypot(3,4)` | ✗ | ✗ | ✗ |
| `Math.sumPrecise([1])` | ✗ | ✗ | ✗ |
[...]

Issue #3439, opened 2026-08-22
`void <unknown>` is not folded to `undefined` in the client — 16 cells


baseballyama:
`void <anything>` is `undefined` — a **known** value whatever the operand is —
so upstream folds it and renders nothing. rsvelte folds it only when the operand
itself is known, and otherwise emits a reactive `$.template_effect`.

```svelte
<script>
	let v = $effect.tracking();
</script>
<b>{void v}</b>
```

```js
// official
	text.nodeValue = '';
// rsvelte
	$.template_effect(() => $.set_text(text, void v));
```

## Upstream

`phases/scope.js`, `Evaluation`'s `UnaryExpression` branch. The known-operand
fast path is first, and then the **unknown**-operand switch still gives `void`
a single concrete value:

```js
const argument = scope.evaluate(expression.argument);
if (argument.is_known) { this.values.add(unary[expression.operator](argument.value)); break; }

switch (expression.operator) {
    case '!': case 'delete':   this.values.add(false); this.values.add(true); break;
    case '+': case '-': case '~': this.values.add(NUMBER); break;
    case 'typeof':             this.values.add(STRING); break;
    case 'void':               this.values.add(undefined); break;   // ← one value ⇒ is_known
    default:                   this.values.add(UNKNOWN);
}
```

rsvelte has only the first half: an unknown operand makes the whole unary
expression unknown, so nothing folds and the read stays reactive.

## Measured scope: 16 cells

Generated grid: 10 declaration shapes × 5 `void` reference bodies × 2 targets
(100 cells), official vs rsvelte on generated JS. **16 diverging, all client.**

| declaration of `v` | diverging bodies |
|---|---|
| `let v = $derived(1);` | `{void v}`, `x{void v}`, `title={void v}` |
| `let { v } = $props();` | `{void v}`, `x{void v}` |
| `let v = $effect.tracking();` | `{void v}`, `x{void v}` |
| `let v = $effect.root(() => {});` | `{void v}`, `x{void v}` |
| `let v = $state.snapshot(1);` | `{void v}`, `x{void v}` |
| `function v() {}` | `{void v}`, `x{void v}` |
| *(no declaration — `v` is a free identifier)* | `{void v}`, `x{void v}`, `{void v()}` |

The `x{void …}` row is the one that shows the value is wrong and not only the
reactivity: official emits `b.textContent = 'x'` (the `void` part contributes
nothing), rsvelte emits `` `x${void v ?? ''}` `` inside an effect.

[...]

Issue #3437, opened 2026-08-22
A `$derived` with a known-constant argument is reactive in rsvelte and not upstream — 12 cells, and the rule is `is_known`


baseballyama:
A `$derived` whose argument evaluates to a **known constant** is not reactive
upstream: the read is a one-shot `b.textContent = $.get(v)` with a `<b></b>`
template. rsvelte wraps every `$derived` read in `$.template_effect` and
templates a text node for it, so the generated template string differs too.

```svelte
<script>
	let v = $derived(1);
</script>
<b>{v}</b>
```

```js
// official
var root = $.from_html(`<b></b>`);
	b.textContent = $.get(v);
// rsvelte
var root = $.from_html(`<b> </b>`);
	var text = $.child(b, true);
	$.template_effect(() => $.set_text(text, $.get(v)));
```

## The rule is `is_known`, not `typeof`

#3213's row 2 reads this as a disagreement about the reactivity of `typeof` over
a derived. It is not — `typeof` is incidental. Upstream
(`2-analyze/visitors/Identifier.js`):

```js
context.state.expression.has_state ||=
    binding.kind !== 'static' &&
    (binding.kind === 'prop' || binding.kind === 'bindable_prop' ||
     binding.kind === 'rest_prop' || !binding.is_function()) &&
    !context.state.scope.evaluate(node).is_known;
```

and `scope.evaluate` on an identifier whose binding is not updated recurses into
`binding.initial`, where the rune cases are (`scope.js`):

```js
case '$state': case '$state.raw': case '$derived':
    if (arg) scope.evaluate(arg, this.values); else this.values.add(undefined);
    break;
case '$derived.by':
    if (arg?.type === 'ArrowFunctionExpression' && arg.body.type !== 'BlockStatement') {
        scope.evaluate(arg.body, this.values);
        break;
    }
    this.values.add(UNKNOWN);
    break;
```

So `$derived(1)` and `$derived.by(() => 1)` are *known* and therefore not
reactive, while `$derived.by(() => { return 1; })` is unknown and stays
reactive. The `is_known` gate is the whole rule.

## Measured scope: 12 client cells, not 3

Generated grid, 14 declaration shapes × 7 reference bodies × 2 targets (196
cells), official vs rsvelte on generated JS. Every divergence in this family:

| declaration | reference bodies that diverge |
|---|---|
[...]

PR #3501, opened 2026-08-22
refactor(client): answer `is_known` with the shared scope.evaluate port (1/4)

baseballyama:
First instalment of the evaluator unification tracked by #3437 / #3439. It removes
**one** implementation of "is this expression's value known at compile time" and
leaves the shared walk in place for the next three.

## The inventory this is built on

"Is this value known" is not one question with six implementations. It is **three**
questions:

| | question | implementations |
|---|---|---|
| **Q1** | what JS value does this evaluate to / is it known | `server/evaluate.rs` (faithful port), the client fold `get_literal_value_json`, the client's `is_expression_known_json`, `is_initial_value_literal_or_known`, and the SSR text fold `extract_constant_vars` |
| **Q2** | is it provably non-nullish | `is_js_expr_defined` / `identifier_is_defined` / `is_expression_defined_json`, plus `is_expression_defined_typed` in phase 2 — a **derived property of Q1** (`Evaluation::is_defined`) |
| **Q3** | does reading it read reactive state | `has_reactive_state_json` and phase 2's `has_state` — a **different** question that *contains* Q1 as a term |

`2_analyze/visitors/shared/utils.rs::is_pure_node` is not in the family: it is
upstream's `is_pure` ("does evaluating this have side effects").

So Q3 must keep its own predicate and *call* Q1; only Q1 and Q2 collapse. This PR
collapses one Q1 implementation.

## #3027 shared the value type but not the value MODEL

#3027 replaced the client fold's `Option<Option<String>>` with the server's
`EvalValue`, which fixed eleven folds that printed the wrong text. What it did not
share is `Evaluation` — upstream's **set** of possible values, which is what makes
`is_known` mean "exactly one value, and not a type marker". The client kept
`Option<EvalValue>`, i.e. "did the folder succeed", and that representation cannot
hold the two states upstream distinguishes from failure:

* a **marker** — `NUMBER` / `STRING`: not known, but defined and of a known type;
* a **one-value result from an unknown operand** — `void x`, `typeof x`, `!x`.

Both are ordinary arms of upstream's `Evaluation` constructor. Measured, on
`origin/main`:

```svelte
<b class="a{Math.random()}"></b>
```

```js
// official — Math.random() is NUMBER: not known, so no fold, but defined, so no `?? ''`,
// and it has no reactive dependency, so no effect
	$.set_class(b, 1, `a${Math.random()}`);
// rsvelte — no representation for "unknown but defined", so it memoises and guards
	$.template_effect(($0) => $.set_class(b, 1, `a${$0 ?? ''}`), [() => Math.random()]);
```

## What this PR changes

`EvalCtx::evaluate_estree` asked its environment for exactly two things —
resolve an identifier, and tell whether a name is bound. Those are now a trait
(`EvalScope`), and the 300-line walk is a free generic function. `EvalCtx`
implements the trait with its existing methods and keeps inherent wrappers, so
**every server call site is unchanged**. `evaluate_binding_initial` is generic the
same way.

The client then gets a `ClientEvalScope` that supplies its own identifier
resolution — Phase 2's reference-position lookup (`binding_at_reference`, which is
scope-correct where the server walks a scope-index chain) plus the each-block
shadow — and `is_expression_known_json`'s 230-line recursion is deleted in favour
of `evaluate_estree(&ClientEvalScope { context }, node, 0).is_known()`.

## Measurement

Discriminating case — `void <unknown operand>`, where the deleted recursion asked
whether the ARGUMENT was known and upstream gives `void` one value regardless:

```svelte
{#if true}{@const c = void p}<b>{c}</b>{/if}
```

```js
// official and this PR
	b.textContent = $.get(c);          // template `<b></b>`
// origin/main
	var text = $.child(b, true);
	$.template_effect(() => $.set_text(text, $.get(c)));   // template `<b> </b>`
```

Regression grids (official v5.56.10 as the oracle, generated cases, `client` and
`server`):

| grid | cells | `origin/main` | this PR |
|---|---|---|---|
| #3437 — 14 declarations × 7 reference bodies | 196 | 23 | 23 |
| #3439 — 10 declarations × 5 `void` bodies | 100 | 19 | 19 |
| globals — 17 keypaths × 4 hosts | 136 | 56 | 56 |

**The three grids are deliberately unmoved.** They are driven by the *fold* and by
phase 2's `has_state`, not by `is_expression_known_json`; routing them is PR-2 and
PR-4. The positive control that this PR's new path is live and correct is the
`void` case above, which diverges on `origin/main` and matches here — verified by
running both binaries against the same file.

Negative controls, in the test and in the repro: `!p` and `-s` over an unknown
operand are genuinely unknown and must stay reactive, so a fix that made every
unary known would fail. And upstream's `scope.evaluate` has **no**
`SequenceExpression` case (#3323 / #3454) — the shared walk has none either, and a
test pins that it stays that way.

## What is NOT removed

**Implementation 5 survives every PR in this series.** `3_transform/server/helpers.rs`
— `extract_constant_vars` + `try_evaluate_with_constants` — is the SSR constant
fold, and it is a *different domain*: it harvests `name → source text` by scanning
the instance script line by line and folds expressions as **strings**, with name
equality where `EvalScope` has scope resolution. It is the #3175 site. Putting it on
this model needs the harvest side rebuilt first (AST rather than a line scan), which
is its own brief.

So the endpoint of PR-1..4 is **"three of Q1's four AST-domain implementations become
one"**, not "six became one".

## Next in the series

* **PR-2** — route the client *fold* through the same walk, deleting
  `get_literal_value_json`'s recursion, `get_literal_value_complex` and
  `is_initial_value_literal_or_known` (the `memmem::find(json, b"Literal")` proxy).
  This is what #3439's grid needs.
* **PR-3** — one globals table (#3473: phase 2 matches it by prefix).
* **PR-4** — give phase 2's `has_state` its missing `!scope.evaluate(node).is_known`
  term (#3437, and #3228 which is the same cell).

Closes nothing on its own by design; #3437 / #3439 / #3228 close with PR-2 and PR-4.


baseballyama:

Issue #3432, opened 2026-08-22
Seventeen JS parse errors both compilers raise diverge in message, position or code


baseballyama:
Seventeen inputs that **both** compilers reject carry a different `js_parse_error` message, a different byte offset, or a different error *code* than the official compiler. They were found while closing #3243, by a sweep built from a source the issue tracker does not otherwise use: the **90 distinct `raise` / `raiseRecoverable` message literals in acorn 8.17.0**, one input written per context-dependent rule, run against `svelte.compileModule` and rsvelte.

That sweep's *acceptance* half is closed — after #3243's fix, **0 of 50 inputs are accepted by rsvelte and rejected by official**. Everything below is the residue: the verdict agrees and a field of the diagnostic does not.

This matters more than the wording suggests. The corpus gate reports `code` as **saturated** — 0 divergences over the 2,843 `(id, target)` pairs both compilers reject — and four of the rows below are `code` divergences. A population of deliberately illegal JavaScript is not in that corpus, which is why a saturated field still has these in it.

## `code` diverges (4)

| input (`.svelte.js`) | official | rsvelte |
|---|---|---|
| `export class K { a = arguments; }` | `js_parse_error` "Cannot use 'arguments' in class field initializer" @1:21 | `invalid_arguments_usage` @1:21 |
| `export class K { static { this.a = arguments; } }` | `js_parse_error` "Cannot use arguments in class static initialization block" @1:35 | `invalid_arguments_usage` @1:35 |
| `export { nope };` | `js_parse_error` "Export 'nope' is not defined" @1:9 | `export_undefined`, **no position at all** |
| `let a; let a;` | `js_parse_error` "Identifier 'a' has already been declared" @1:11 | `declaration_duplicate` @1:11 |

The `export_undefined` row is worth separating: `start`, `end` and `position` are all absent, so there is nothing for the CLI frame, an LSP diagnostic or an editor squiggle to anchor to — the same shape as the `render_tag_invalid_expression` half of #3247.

The pattern in all four is that rsvelte answers from a **later phase** (analysis) with its own Svelte error code, where upstream answers from acorn during the parse. Whichever direction the fix runs, the two have to agree on which phase owns the input.

## message and/or position diverge, `code` agrees (13)

| input (`.svelte.js`) | official | rsvelte |
|---|---|---|
[...]

Issue #3350, opened 2026-08-21
expected_token vs js_parse_error is decided per template slot: two over-acceptances and three code/message divergences


baseballyama:
A 28-input probe crossing "a JS expression with a trailing token" / "a broken JS expression" with the eleven template slots that read one turned up five divergences that share a cause: **which of `expected_token` and `js_parse_error` a slot reports is decided per slot in rsvelte, and upstream decides it once** — acorn parses one maximal expression and whatever is left over is `expected_token` at the leftover token, everything else is `js_parse_error` at `err.pos`.

All rows are client and server alike, and each was read off the official compiler on the same source.

## Over-acceptances — official rejects, rsvelte compiles

```svelte
{#each a as b c}x{/each}
```
official `expected_token` / `Expected token }` at `1:14` · rsvelte **compiles**

```svelte
{#await a b}x{/await}
```
official `expected_token` / `Expected token }` at `1:10` · rsvelte **compiles**

Both headers stop reading at the first token they recognise and drop the rest, so a typo in a header silently changes what the block iterates over or awaits. The `{#if}`, `{#key}`, `{#each … as}` (expression half) and `{@html}` spellings of the same shape all report correctly, which is what makes this two missing tail checks rather than one missing rule.

## Code divergences — both reject, different code

```svelte
<div title={a b}></div>
```
official `expected_token` / `Expected token }` at `1:14` · rsvelte `js_parse_error` / ``Expected `,` or `)` but found `Identifier` `` at `1:14`

The *directive* spelling of the same value — `<div on:click={a b}>`, `<div use:x={a b}>` — reports `expected_token` correctly, because directive values go through `parse_head_expression` and a plain attribute value goes through `parse_js_expression_strict`, which has no trailing-token classification at all. Same slot kind, two implementations, and only one of them can produce the code upstream produces.

```svelte
{@render s(a b)}
```
official `js_parse_error` / `Unexpected token` at `1:13` · rsvelte `render_tag_invalid_expression` **with no position**

The render tag never reports the JS failure: the expression falls back to a placeholder and the *placement* rule then rejects the placeholder, so a syntax error is reported as a semantic one, in a later phase, without a span.

```svelte
{@const a = 1 2}
```
official `expected_token` / `Expected token }` at `1:14` · rsvelte `const_tag_invalid_placement` at `1:0`–`1:16`

[...]

Issue #3280, opened 2026-08-21
`directive_missing_name` is missing for `style:` / `animate:` / `let:` / `on:` and wrong for `bind:`; unknown `{@…}` tags compile


baseballyama:
`directive_missing_name` is raised per directive kind, and four kinds never raise
it while a fifth raises the wrong code. Crossing the directive kind with the
empty-name case gives the full table:

| source | rsvelte | official |
|---|---|---|
| `<div use:={x}>` | `directive_missing_name` @1:5..1:9 | same |
| `<div transition:={x}>` | `directive_missing_name` @1:5..1:16 | same |
| `<div class:={x}>` | `directive_missing_name` @1:5..1:11 | same |
| `<div bind:={x}>` | `bind_invalid_name` @1:5..**1:14** | `directive_missing_name` @1:5..**1:10** |
| `<div style:={x}>` | **compiles** | `directive_missing_name` @1:5..1:11 |
| `<div animate:={x}>` | **compiles** | `directive_missing_name` @1:25..1:33 |
| `<C let:={x}>` | **compiles** | `directive_missing_name` @1:3..1:7 |
| `<button on:={h}>` | **compiles** | `directive_missing_name` @1:5..1:10 |

Three correct, one wrong code *and* wrong `end`, four missing. This is the
"one arm per visitor, so the rule drifts wherever the product is unenumerated"
shape the `directive-element` matrix family was added for — the check exists and
is right where someone wrote it.

`on:={h}` was filed as one row of #3274 before the other directives were crossed;
this supersedes that row.

## Second over-acceptance found in the same sweep: unknown `{@…}` tags

| source | rsvelte | official |
|---|---|---|
| `{@bogus x}` | **compiles** | `expected_tag` @1:10 — Expected 'html', 'render', 'attach', 'const', or 'debug' |
| `{@nope}` | **compiles** | `expected_tag` @1:2 |
| `{@}` | **compiles** | `expected_tag` @1:2 |
| `{@ }` | **compiles** | `expected_tag` @1:2 |

54 cells in the grid, uniform across every character and both targets — the
character axis is not involved, it is simply that no `expected_tag` check exists.

## One error-field row, for completeness

```svelte
<p>{a</p>
```

| | code | message | start | end |
|---|---|---|---|---|
| rsvelte | `element_unclosed` | `` `<p>` was left open `` | 1:0 | 1:1 |
| official | `js_parse_error` | `Unterminated regular expression` | 1:7 | 1:7 |

[...]

PR #3504, opened 2026-08-22
fix(parse): report the JS parse error a const/await/render tag swallowed

### LINKED ISSUE #3202: `{@const}`, `{#await}` and `{@render}` swallow every JS parse error

### LINKED ISSUE #3354: Over-acceptance: {#await}, {@const} and a {@render} argument swallow a template-expression parse error
**Over-acceptance.** Three template slots swallow a JS parse error outright, so
a component the official compiler rejects with `js_parse_error` /
`expected_token` compiles. It is not a TypeScript question — plain JavaScript
reproduces every row.

```svelte
{#await y +}<p>x</p>{/await}
```

official `js_parse_error` · rsvelte **compiles**

## Grid

Three deliberately-invalid plain-JS expressions × the slot, with
`<script>let y = 1; let z = 2;</script>` in front. Measured against
`svelte@5.56.9`, client target (server agrees).

| expression | slot | official | rsvelte |
|---|---|---|---|
| `y z` | `{#await …}` | `expected_token` | **OK** |
| `y z` | `{@const c = …}` | `expected_token` | **OK** |
| `y z` | `{@render s(…)}` | `js_parse_error` | `render_tag_invalid_expression` |
| `y +` | `{#await …}` | `js_parse_error` | **OK** |
| `y +` | `{@const c = …}` | `js_parse_error` | **OK** |
| `y +` | `{@render s(…)}` | `js_parse_error` | `render_tag_invalid_expression` |
| `y ===` | `{#await …}` | `js_parse_error` | **OK** |
| `y ===` | `{@const c = …}` | `js_parse_error` | **OK** |
| `y ===` | `{@render s(…)}` | `js_parse_error` | `render_tag_invalid_expression` |

The other twelve expression slots (`{…}`, an attribute value, a spread, an
`{#each}` collection, an `{#each}` body, `{#if}`, `{#key}`, `{@html}`, an event
handler, `style:`, `class:`, a transition parameter object) agree on all three
after #3351.

**The two verdicts are not the same defect wearing two hats.** `{#await}` and
`{@const}` produce *no* diagnostic at all — the component compiles and the
expression becomes an empty identifier. `{@render}` produces the wrong code: the
parse failure is swallowed, the placeholder is not a call expression, and
phase 2 then complains about the *shape* rather than the parse. A ratchet keyed
on `code` cannot tell the third row from a correct rejection.

## Cause

- `{#await}` head and `{@render}` argument both go through
  `parse_js_expression_lenient`
  (`crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs`), whose doc
  comment says it deliberately swallows because `{@render}`'s invalid-call check
  is an analysis-phase error. Upstream does not: `read_tag`'s `{@render}` arm is
  `read_expression(parser)` + `eat('}', true)`, and the `{#await}` head is
  `read_expression(parser, undefined, true)` — `disallow_loose`, i.e. errors
  propagate — with the call-shape check separate from the parse.
- `{@const}` goes through `parse_js_expression_internal`, whose `unwrap_or_else`
  returns an empty identifier on failure. Upstream reads it with
  `read_declaration`, which rethrows in strict mode.

## Why it was not visible

Every grid that would have caught it was built out of *valid* expressions;
`compiler-errors` has one fixture per code, and a code with a passing fixture
reads as covered. It surfaced while building the #3261 grid, where these three
slots were the only cells still accepting TypeScript-only syntax after the
language-mode fix — and the plain-JS control above showed the cause has nothing
to do with TypeScript.

baseballyama:
Refs #3202
Closes #3354

Not `Closes #3202`: four of that issue's five rows are fixed here, and the fifth (`{#await 1 + then v}`) has a different cause that belongs to #3350 — see the last section.

`Closes #3354` because that issue's grid is measured and fully closed on this branch — all nine cells, see [the measurement](https://github.com/baseballyama/rsvelte/issues/3354#issuecomment-5379080540). Its three `{@render}` rows keep a +1 byte offset, and the control in that comment shows the identical +1 in the mustache, attribute and `{#if}` slots this PR does not touch, so it is #3432 rather than an unfixed part of #3354.

## What

`{@const …}`'s initializer, the `{#await …}` head and `{@render …}` all routed their expression through a parse whose contract is `.unwrap_or_else(|_| create_empty_identifier(…))`. Ordinary broken JavaScript compiled:

```svelte
{#if true}{@const c = 1 +}<b>{c}</b>{/if}
{#await 42 = nope then v}<b>{v}</b>{/await}
```

Upstream's `read_expression` **throws** unless `parser.loose`, so all three are hard errors there.

`{@render}` is the row that shows the shape rather than restating it: the swallowed expression became an empty identifier, that identifier is not a call, and so the *downstream* `render_tag_invalid_expression` fired — a second error standing in for the one that was dropped, with a different code.

## Both halves of `read_expression`'s contract, not just the error half

Upstream parses **one** expression and the caller then does `allow_whitespace(); eat('}', true)`. So a failure has two possible verdicts, and a fix that maps every failure to `js_parse_error` is wrong in half the cases. Measured on `svelte.compile`, both verdicts occur at all three entry points:

| input | official / rsvelte after this PR |
|---|---|
| `{@const c = 1 +}` | `js_parse_error` "Unexpected token" @25 |
| `{@const c = 42 = nope}` | `js_parse_error` "Assigning to rvalue" @22 |
| `{@const c = 1 2}` | `expected_token` "Expected token }" @24 |
| `{@const c = 1;}` | `expected_token` "Expected token }" @23 |
| `{#await 42 = nope}` | `js_parse_error` "Assigning to rvalue" @8 |
| `{#await 42 = nope then v}` | same, @8 |
| `{#await 42 = nope catch e}` | same, @8 |
| `{#await 1 +}` | `js_parse_error` "Unexpected token" @11 |
| `{#await a b}` | `expected_token` "Expected token }" @10 |
| `{#await a b then v}` | `expected_token` "Expected token }" @10 |
| `{@render s(42 = nope)}` | `js_parse_error` "Assigning to rvalue" @11 |
| `{@render 1 +}` | `js_parse_error` "Unexpected token" @12 |
| `{@render s() x}` | `expected_token` "Expected token }" @13 |
| `{@render s();}` | `expected_token` "Expected token }" @12 |

## One classifier, because there were three copies of the decision

The leftover-input probe (`trailing_token_offset`) is not sufficient on its own: OXC labels an invalid assignment target **at the target's start**, so `42 = nope` looks exactly like "the expression ended at offset 0 and everything after is leftover". The mustache path already guards against that with a prefix re-parse; `parse_head_expression` and `resolve_lazy.rs`'s `HeadBrace` arm did **not**, so the same input would have been classified two ways depending on whether the expression was deferred.

That matters here rather than in the abstract, because `compile` sets `defer_script_parse: true` — so the path this PR must get right is the **deferred** one, and the eager one is what the tests reach in other configurations. The decision is now one function, `close_token_or_parse_error`, called from all three.

## `{@const}` cannot defer, and the negative control says why

The const initializer keeps parsing **eagerly**. Two things below it read the parsed node: the sequence-expression rejection (`init_expr.node_type() == Some("SequenceExpression")`) and `build_const_variable_declaration`, which takes ownership of it.

**Negative control, measured** — swap `parse_const_initializer` for the deferring `parse_head_expression` and rebuild:

```
test a_sequence_initializer_is_still_rejected ... FAILED
test a_const_tag_initializer_reports_its_parse_error ... FAILED
test a_tag_classifies_exactly_as_a_mustache_carrying_the_same_expression ... FAILED
test leftover_input_after_a_complete_expression_is_a_missing_brace ... FAILED
test valid_tags_still_compile ... FAILED
test result: FAILED. 3 passed; 5 failed
```

Every failure is `panicked at .../state/tag.rs:2904` — `expression_into_node`'s `Expression::Lazy must be resolved before building a declaration`. So the prediction I wrote before measuring ("the rejection goes silently missing") was only half right: `node_type()` does return `None` and the rejection *is* skipped, but the builder panics one statement later, so **every** const tag aborts, not just the sequence one. The three tests that stay green are the ones with no `{@const}` in them, which is the control moving in the right direction.

## Over-reach

Eight valid tags are asserted to still compile, chosen against the specific ways this change could over-reject: `{@render s?.()}` (a chained call), `{#await obj.then}` (`then` inside the expression rather than as the clause keyword), `{#await p}…{:then v}…{:catch e}…{/await}`, `{@const c = (1, 2)}` (a parenthesised sequence, which is legal where a bare one is not) and `{@const c = (a) => a}`.

`{@render 1}` — an expression that *parses* but is not a call — is asserted to still reach the analysis-phase `render_tag_invalid_expression` at @9. The parse error and the invalid-expression error are two verdicts and the fix must not collapse them.

Loose (editor) mode keeps recovering with a placeholder at all three sites, as before.

## The residual, and why it is not this defect

rsvelte's `js_parse_error` message and offset come from OXC, which labels some failures a token away from acorn and spells a truncated input `` Expected `)` but found `EOF` `` where acorn says `Unexpected token`. Measured:

|---|---|---|
| `{f(}` | `Unexpected token` @3 | `` Expected `)` but found `EOF` `` @3 |
| `{s(1 +)}` | `Unexpected token` @6 | `Unexpected token` @7 |

Both rows are **mustaches** — the path this PR does not touch — so this is one pre-existing property of `check_js_parse_error_with_pos`, not something the three tags introduce. `a_tag_classifies_exactly_as_a_mustache_carrying_the_same_expression` states exactly that: for the same expression at four entry points (mustache, `{@render}`, `{#await}` head, `{@const}` initializer) the code, the message and the offset *relative to where the expression starts* must be identical. Whatever the residual is, it is now the same everywhere, and closing it is a change to that one function.

(The `{#await f(}` row is deliberately absent from the unbalanced-`(` group: nothing terminates that head, so it is `block_unclosed` — #3247, not this.)

## The fifth row is not this defect

`{#await 1 + then v}` — official raises `expected_token` "Expected token }" @17. rsvelte finds the ` then ` keyword by **text scan** and splits the head there, so it parses `1 +` and fails inside the expression; upstream parses the expression first and only then looks at what follows, so its `then` is part of the expression and the error lands on `v`. That is the await header's end detection, which #3350 covers, and it is untouched here.

## Tests

`crates/rsvelte_core/tests/tag_expression_parse_errors_3202.rs` — code, message and byte offset asserted for every row, all measured against `svelte.compile`.

## Verification

`cargo fmt --all` and `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` were run by hand (the pre-commit hook was bypassed with `--no-verify`; it builds a second multi-GB `target/` and does not finish inside the shared-machine budget right now).

Green on this branch: `compiler_errors`, `parser_fixtures`, `compiler_fixtures`, `ssr`, `runtime`, `validator`, `css`, `block_head_parse_errors`, `const_tag_pattern_pin`, `const_tag_multibyte`, `await_block_multibyte`, `await_ideographic_space_2409`, `render_tag_invalid_expression`, `render_tag_multiline_args`, `special_tag_bracket_scan`, `parser_hardening`, `expected_whitespace_2581`, `js_whitespace_parser_2561`, `block_close_mismatch`, and the new file.

### DIFF
--- a/.changeset/tag-expression-parse-errors.md
+++ b/.changeset/tag-expression-parse-errors.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Report the JS parse error in a `{@const …}` initializer, a `{#await …}` head and a `{@render …}` tag instead of swallowing it. All three routed their expression through a parse that recovers with an empty identifier, so ordinary broken JavaScript compiled — and in the `{@render}` case the empty identifier then failed the downstream call check, so a second error stood in for the one that was dropped. Upstream's `read_expression` throws unless the parser is loose, and its caller then expects the `}`, so leftover input after a complete expression is an `expected_token` while a malformed expression is a `js_parse_error`; both classifications now reach all three tags
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -1858,6 +1858,41 @@ pub fn trailing_token_offset(content: &str) -> Option<usize> {
     probe(SourceType::ts()).or_else(|| probe(SourceType::mjs()))
 }
 
+/// Classify a failed `read_expression` for a construct terminated by
+/// `close_char`, the way upstream's caller does: acorn parses ONE maximal
+/// expression and the caller then `eat(close_char, true)`, so leftover input
+/// after a *complete* expression is a missing close token while a malformed
+/// expression is a `js_parse_error` at the byte where the parse stopped.
+///
+/// The prefix re-parse is what separates the two: OXC labels an invalid
+/// assignment target at the target's start, which the leftover-input probe
+/// would otherwise read as "the expression ended here".
+pub fn close_token_or_parse_error(
+    msg: String,
+    trimmed: &str,
+    trimmed_offset: usize,
+    close_char: char,
+) -> crate::error::ParseError {
+    let trailing = trailing_token_offset(trimmed).filter(|&off| {
+        off > 0
+            && trimmed
+                .get(..off)
+                .is_some_and(|prefix| check_js_parse_error_with_pos(prefix).is_none())
+    });
+    if let Some(off) = trailing {
+        let mut buf = [0u8; 4];
+        return crate::error::ParseError::expected_token(
+            close_char.encode_utf8(&mut buf),
+            trimmed_offset + off,
+        );
+    }
+    let at = check_js_parse_error_with_pos(trimmed)
+        .map_or(trimmed_offset + trimmed.len(), |(_, pos)| {
+            trimmed_offset + pos
+        });
+    crate::error::ParseError::svelte("js_parse_error", msg, (at, at))
+}
+
 /// Create an identifier for invalid expressions
 fn create_invalid_identifier<'a>(
     start: usize,
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/resolve_lazy.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/resolve_lazy.rs
@@ -506,16 +506,11 @@ fn lazy_parse_error(
         // a missing close token; anything else is a point `js_parse_error`.
         LazyKind::HeadBrace | LazyKind::HeadParen => {
             let close = if kind == LazyKind::HeadParen {
-                ")"
+                ')'
             } else {
-                "}"
+                '}'
             };
-            if let Some(pos) = super::read::expression::trailing_token_offset(content) {
-                return Some(crate::error::ParseError::expected_token(close, start + pos));
-            }
-            let abs_pos = super::read::expression::check_js_parse_error_with_pos(content)
-                .map_or(start + content.len(), |(_, pos)| start + pos);
-            crate::error::ParseError::svelte("js_parse_error", msg, (abs_pos, abs_pos))
+            super::read::expression::close_token_or_parse_error(msg, content, start, close)
         }
     })
 }
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
@@ -1534,33 +1534,42 @@ impl<'a> Parser<'a> {
         let adjusted_end = expr_end - (expr_content.len() - trimmed_content.trim_end_ws().len());
         // For await blocks, we parse the expression with a known end position
         // to avoid find_matching_bracket finding the block's closing }
-        let expression = if let Some(lazy) =
-            self.defer_expression(trimmed_content.trim_ws(), adjusted_start, LazyKind::Lenient)
-        {
-            lazy
-        } else {
-            super::super::expression::parse_expression_with_end(
-                &self.arena,
-                trimmed_content.trim_ws(),
-                adjusted_start,
-                adjusted_end,
-                self.expression_line_offsets(),
-                self.source,
-                self.options.loose,
-                false,
-                '{',
-                self.ts,
-            )
-            .unwrap_or_else(|(_, pos)| {
-                // Return an invalid identifier on parse error (empty name)
-                super::super::expression::create_identifier_with_character(
-                    "",
-                    pos,
+        let head = trimmed_content.trim_ws();
+        let expression =
+            if let Some(lazy) = self.defer_expression(head, adjusted_start, LazyKind::HeadBrace) {
+                lazy
+            } else {
+                match super::super::expression::parse_expression_with_end(
+                    &self.arena,
+                    head,
+                    adjusted_start,
                     adjusted_end,
                     self.expression_line_offsets(),
-                )
-            })
-        };
+                    self.source,
+                    self.options.loose,
+                    false,
+                    '{',
+                    self.ts,
+                ) {
+                    Ok(expr) => expr,
+                    Err((_, pos)) if self.options.loose => {
+                        super::super::expression::create_identifier_with_character(
+                            "",
+                            pos,
+                            adjusted_end,
+                            self.expression_line_offsets(),
+                        )
+                    }
+                    Err((msg, _)) => {
+                        return Err(super::super::read::expression::close_token_or_parse_error(
+                            msg,
+                            head,
+                            adjusted_start,
+                            '}',
+                        ));
+                    }
+                }
+            };
 
         // Parse 'then' value if present
         if has_then {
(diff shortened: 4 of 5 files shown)


baseballyama:
salvJ の交差実行で、このブランチに **wrong-verdict のセルが 1 つ**見つかったので共有します。修正の要否はこの PR の担当判断です。

## セル

```
入力:  {#await 1 + then v}
公式 5.56.9 :  expected_token   @ 1:17
#3504       :  js_parse_error   @ 1:11
```

**accept / reject の枠では採点されません** — 両者とも reject するので、コードと位置を比べるゲートでないと見えません。`compatibility/error-{message,position}-known-failures.*` が別々に ratchet されているのと同じ理由です。

## なぜ位置が違うか

acorn は `1 + then` を 1 つの式として読むので、**余った `v` が `}` を期待した位置**になります。`then` をキーワードとして先に切ると、`1 +` が不完全式として `js_parse_error` になり、位置が `+` の側に寄ります。

## 根拠

期待値は推測ではなく、**68 個のアサーション全部を pinned submodule (5.56.9) から再導出**したものです: `agree 68 / DISAGREE 0`。したがって「テストが間違っている」ではなく、このセルについては rsvelte 側の verdict が公式と違う、と読めます。

## 併せて 1 件

同じ交差実行で、**`{@debug a +}` が全ツリー（main / #3499 / #3504 / #3478 / #3460）でコンパイルされてしまう**ことも分かりました。`{@debug}` は #3202 の本文にも、この PR にも名指しされていない **第 5 の入口**です。この PR の範囲外だと思いますが、`{@const}` / `{@render}` / await head と同じ穴なので記録しておきます（draft PR #3573 が #3247 と #3280 の残りと合わせて拾う予定です）。

PR #3521, opened 2026-08-22
fix(parse): raise the last three early errors, and only where acorn does (#3432)

### LINKED ISSUE #3432: Seventeen JS parse errors both compilers raise diverge in message, position or code

baseballyama:
Closes #3432 (the code section).

**Stacked on #3499** (tsE). `origin/main...HEAD` here contains #3499's three commits plus one of mine; review only `0faafc2`. Merge #3499 first.

## What #3432 measured

The issue's population is "both compilers reject, but the message / position / code differ". I re-measured all 18 divergent rows plus 14 legal controls (32 rows) through `compileModule`, against `submodules/svelte` as oracle, at three tree states:

| tree | agree | **code** | position | message | over-accept | over-reject |
|---|---|---|---|---|---|---|
| `origin/main` | 15 | **4** | 13 | 0 | 0 | 0 |
| #3499 alone | 16 | **3** | 13 | 0 | 0 | 0 |
| #3499 + this PR | **19** | **0** | 13 | 0 | 0 | 0 |

So this PR's isolated contribution is **3 code rows fixed, 0 regressed, 0 changed**. #3499 closes the fourth (`let a; let a;`) on its own.

The 13 position rows are deliberately **not** in this PR — every one of them is a position divergence with no message divergence, and they need their own table (agreed with tsE, who offered the method). There are **0 pure-message rows**.

## The three rows

| input | official | rsvelte before |
|---|---|---|
| `export class K { a = arguments; }` | `js_parse_error` @21 *Cannot use 'arguments' in class field initializer* | compiled |
| `export class K { static { this.a = arguments; } }` | `js_parse_error` @35 *Cannot use arguments in class static initialization block* | compiled |
| `export { nope };` | `js_parse_error` @9 *Export 'nope' is not defined* | `export_undefined`, **`start: None, end: None`** |

The two `arguments` contexts cannot share a table row: OXC gives them one message each, and acorn words them differently from OXC **and from each other** — "class static initialization block" against OXC's "static initialization block".

## ★ The export row is host-dependent, and my first version over-rejected

My grid was `compileModule`-only. That is one direction of the population, and #2583's lesson is that a one-directional population is blind in the other direction. When I checked the component host before committing, the new row rejected **three shapes official compiles**:

```
<script>export { nope };</script>              official OK  →  rsvelte js_parse_error
<script>export { nope as other };</script>     official OK  →  rsvelte js_parse_error
<script>export { nope, alsoNope };</script>    official OK  →  rsvelte js_parse_error
```

That is not an exception to be special-cased — it is upstream behaviour I had not ported. `1-parse/acorn.js` takes an `is_script` flag and, when set, **clears acorn's `undefinedExports` after every statement**, with the comment that a `<script>` "might contain an export for something that doesn't exist directly inside but is inside the component instead". `read/script.js:34` passes `true`; `2-analyze/index.js:263` (which is what `compileModule` reaches) passes `false`.

So the flag is threaded through `ProgramParseParams` and set explicitly at all six construction sites, rather than inferred from `script_tag_start == 0` or any other proxy. All 10 component-host cells now match official exactly, positions included.

## Controls

Four ablations, each rebuilt and run:

| ablated | result |
|---|---|
| `'arguments' is not allowed in class field initializer` needle | only that row fails, named; falls through to analyze's `invalid_arguments_usage` |
| `'arguments' is not allowed in static initialization block` needle | only that row fails, named |
| `Export '` needle | only that row fails, named; falls back to `export_undefined` with no position — the pre-fix behaviour |
| `Hosts::ModuleOnly` → `Hosts::Both` | the row's **instance-script** half fails: `"export { nope };" must still compile` |

The fourth is the one worth keeping: the repro asserts the error on the module entry point **and** that the component still compiles, so both directions of the check live in one row.

Three legal neighbours were added to `LEGAL` and measured against official first: `class K { a = 1; m() { return arguments; } }` (a method body has its own `arguments`), `class K { static { this.a = 1; } }`, `let n = 1; export { n };`.

The needle is `"Export '"` rather than `"is not defined"`: keying on the tail would adopt any future OXC diagnostic ending that way and report it as an export error.

## Corrections to issue #3432, from this measurement

- **`export const a = import.meta;` is not a divergence.** Official compiles it and so does rsvelte. The row is wrong and I have struck it.
- **All 13 remaining rows are position divergences.** The issue implies a message/position/code mix; there are **0 rows where only the message differs**.

## Suites

`early_errors_3243` (7), `compiler_errors`, `parser_fixtures`, `validator`, `ssr` — all green. `cargo fmt --all` and `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` run by hand (exit 0); the commit uses `--no-verify` because the pre-commit hook rebuilds into the worktree's `./target`.


baseballyama:
## Blocking: this PR breaks a pre-existing test — but the test is the thing that is wrong

Measured on a tree of this PR merged with nine other open PRs, `cargo test --release`
over `rsvelte_core` / `rsvelte_formatter` / `rsvelte_esrap` / `rsvelte_napi`:
**380 targets ok, 2 FAILED**, one of which is:

```
duplicate_let_function::function_overloads_still_allowed
  assertion failed: tc("<script>function x(){} function x(){}</script>").is_ok()
```

`crates/rsvelte_core/tests/duplicate_let_function.rs` exists on `main` and no PR in
that batch modifies it. The new `1_parse/read/early_errors.rs` maps OXC's
`"has already been declared"` with `hosts: Hosts::Both`, so the redeclaration is now
raised for a component `<script>` as well.

### The oracle says this PR is right

```
[oracle] svelte VERSION = 5.56.9

js  two impls      =>  ERR js_parse_error   Identifier 'x' has already been declared
ts  two impls      =>  ERR js_parse_error   Identifier 'x' has already been declared
ts  real overload  =>  OK
ts  decl merge     =>  OK
js  var + var      =>  OK
```

where `real overload` is

```svelte
<script lang="ts">function x(a: string): void;
function x(a: number): void;
function x(a: any) { console.log(a); }</script>
```

The official compiler rejects two function *implementations* with the same name in
both JS and TS, and accepts a signature set plus one implementation. So the test's
**intent** — "a TS overload set must still compile" — is correct, and its **input** is
not: `function x(){} function x(){}` is two implementations, which upstream rejects.

### Suggested fix (belongs in this PR, since this PR changes the behaviour)

In `crates/rsvelte_core/tests/duplicate_let_function.rs`, replace the body of
`function_overloads_still_allowed` with a real overload set under `lang="ts"`, and add
the two-implementation case as an expected error so the new rule is pinned rather than
merely un-pinned.

Worth checking while you are there: `AGENTS.md` currently records that
"`function f(){}` twice in a `lang=\"ts\"` script stays accepted here" as a known gap
against acorn-typescript. The measurement above shows upstream rejects that TS case
too, so that sentence describes an over-acceptance that is still live — the `ts two
impls` row is the one to keep as a recorded difference, not the JS one.

This PR was left out of the merge batch for now; the other eight landed. Nothing here
questions the early-error work itself — the eleven-class enumeration and the
allow-list design both look right, and this is the one place a pre-existing test
pinned the old behaviour.

PR #3532, opened 2026-08-22
fix: whitespace-containing literal needles in phase-3 entry-point scans (#3470, #3456, #3490, #3510, #3498)

### LINKED ISSUE #3456: A class field's rune initializer is only lowered when exactly one ASCII space follows the `=`
## Summary

A class field's rune initializer is only lowered when the text between `=` and the rune is
**exactly one ASCII space, or nothing**. A tab, two spaces, a line break or a comment there and the
rune call survives verbatim into the client output — the field never becomes reactive, and
constructing the class throws `ReferenceError: $state is not defined`.

```svelte
<script>
	class K {
		v =	$state(1);
	}
	const k = new K();
</script>
<b>{k.v}</b>
```

(that is a tab after the `=`)

```js
class K {
	#v = $.state(1);
	get v() { return $.get(this.#v); }
	set v(value) { $.set(this.#v, value, true); }
}

class K {
	v = $state(1);
}
```

The accessor pair is gone and `$state` is a free identifier. The output parses, so the parse gate
cannot see it. `server` is correct in every cell below.

## What the separator may be

| between `=` and the rune | client / client-dev |
|---|---|
| one ASCII space | match |
| nothing (`=$state(1)`) | match |
| **two spaces** | **rune leaks** |
| **three spaces** | **rune leaks** |
| **a tab** | **rune leaks** |
| **space + tab**, **tab + space** | **rune leaks** |
| **a line break** (initializer on the next line) | **rune leaks** |
| **space + line break** | **rune leaks** |
| **a block comment** (`= /* c */ $state(1)`) | **rune leaks** |
| **U+00A0, U+FEFF, U+000B, U+000C, U+2000, U+2009, U+202F, U+3000, U+2028, U+2029** | **rune leaks** |

A tab and a wrapped initializer are ordinary formatting, so this is not an exotic-input case.

## Cause

`crates/rsvelte_core/src/compiler/phases/3_transform/client/class_transforms.rs:1079`

```rust
let pattern_eq = format!("= {}(", rune_type);
let pattern_nospace = format!("={}(", rune_type);
let pattern_eq_generic = format!("= {}<", rune_type);
let pattern_nospace_generic = format!("={}<", rune_type);
let has_pattern = trimmed.contains(&pattern_eq)
    || trimmed.contains(&pattern_nospace)
    || …;
```

Two literal spellings of one separator, matched with `contains`. Everything above is a third
spelling.

## Which class members

All four field runes (`$state`, `$state.raw`, `$derived`, `$derived.by`) leak, in five member
forms:

| form | verdict |
|---|---|
| `v = ⟨sep⟩$state(1)` | **leaks** |
| `#v = ⟨sep⟩$state(1)` | **leaks** |
| `'v' = ⟨sep⟩$state(1)` | **leaks** |
| `this.v = ⟨sep⟩$state(1)` in the constructor | **leaks** |
| the same inside a class **expression** | **leaks** |
| `static v = ⟨sep⟩$state(1)` | match |
| `['v'] = ⟨sep⟩$state(1)` | match |

## Controls

The same separators in every neighbouring position are clean on all three targets, which isolates
this to the class-field scan rather than to whitespace handling in general:

- `let v = ⟨sep⟩$state(1)` — a plain variable declarator
- `const d = ⟨sep⟩$derived(n * 2)`
- `$state⟨sep⟩(1)` — before the call parenthesis
- `$state(⟨sep⟩1⟨sep⟩)` — inside the argument list
- `$effect⟨sep⟩(() => { … })`
- `v⟨sep⟩= $state(1)` — the separator **before** the `=` (only U+FEFF diverges there, and not as a leak)

## Related

`let { p = 1 } =⟨U+FEFF⟩$props();` leaks the same way on `client` and `client-dev`. U+FEFF is the
only separator of the ten that reaches the destructured-props path, so it is likely a second scan
with the same assumption rather than the same site.

### LINKED ISSUE #3470: Phase-3 keyword needles assume exactly one ASCII space: `export⟨tab⟩let` emits an unparseable nested `export`, `class⟨tab⟩K` leaks its runes
## Summary

Phase 3 locates two JavaScript keywords by searching for the literal bytes `"export let"`,
`"export "` and `"class "`. Each spelling bakes in *exactly one ASCII space*, so a second space, a
tab, a line break or a comment between the keywords makes the construct invisible to the transform.

### `export ⟨sep⟩ let` — the declaration survives into the component function, which cannot parse

```svelte
<script>
	export	let a = 1;
</script>
<b>{a}</b>
```

(that is a tab between `export` and `let`)

```js
export default function A($$anchor, $$props) {
	let a = $.prop($$props, 'a', 8, 1);
	…
}

/* rsvelte */
export default function A($$anchor, $$props) {
	export let a = 1;          // <- `export` inside a function; no parser accepts this
	…
}
```

The prop is also never wired to `$$props`, so even if it parsed the component would not receive it.

| separator | client / client-dev |
|---|---|
| one space | match |
| **two spaces** | **unparseable** |
| **a tab** | **unparseable** |
| **a line break** | **unparseable** |
| **a comment** — `export /** @type {number} */ let a = 1;` | **unparseable** |

`export ⟨sep⟩ function f() {}` behaves identically. The comment row is the one that occurs in real
Svelte 4 code, where JSDoc types are written inline; official emits
`/** @type {number} */ let a = $.prop(…)` for it.

### `class ⟨sep⟩ K` — the class-field runes are not lowered

```svelte
<script>
	class	K {
		v = $state(1);
	}
	const k = new K();
</script>
<b>{k.v}</b>
```

```js
/* official */  class K { #v = $.state(1); get v() {…} set v(value) {…} }
/* rsvelte  */  class K { v = $state(1); }
```

A tab or a line break after `class` reproduce it; a comment there produces a smaller text
divergence on all three targets. `server` is correct. Output parses, so the parse gate is blind to
it, and `$state` is a free identifier at run time.

## Where the needles are

```
3_transform/…                        needle          count
  server/transform_script.rs         "class "        several
  client/class_transforms.rs         "class "        several
  client/props_transforms.rs         "export let"    1
  …                                  "export "       2
  …                                  "function "     3
```

`import C ⟨comment⟩ from './C.svelte'` and `function ⟨comment⟩ f()` also produce text divergences
on all three targets, from the `" from "` and `"function "` needles.

## Relationship to #3456

#3456 is the same defect one needle over — `"= $state("` and friends — measured on the separator
between `=` and a rune. This issue is the keyword half. They should probably be fixed together: the
underlying rule is that **a literal needle containing a space encodes a formatting assumption the
source is not obliged to honour**.

## Controls

Clean on all three targets with every separator: `export ⟨sep⟩ const` in a `.svelte.js` module,
`await ⟨sep⟩ expr` in a template, `{#each xs as x,⟨sep⟩i}`, a comma inside a destructured
`$props()`, and a comma inside a `$state({…})` object literal.

### LINKED ISSUE #3490: compileModule (server): a class `$derived` field is missed unless `=` and the rune are separated by exactly one space
`transform_class_fields_server` decides whether a class member is a rune field with
`memmem::find(trimmed, b"= $derived(")` — a needle carrying **exactly one ASCII space**.
Any other separator makes the server scan bail, the client transform picks the field up
instead, and the emitted accessor pair is the client's shape mapped onto server names.

The **component** SSR path is unaffected: it is the pure-AST server transform. Only
`compileModule` still reaches the text scan, so this is the `compileModule`-only twin of
#3456.

## Reproduction

```js
// A.svelte.js
class K {
	v = $state(1);
	d  = $derived(this.v * 2);   // two spaces between `=` and the rune
}

export const k = new K();
```

`compileModule(src, { generate: 'server' })`:

```js
// official
	set d($$value) {
		return this.#d($$value);
	}

// rsvelte
	set d(value) {
		this.#d(value);
	}
```

The getter and the `#d = $.derived(() => this.v * 2)` backing field are identical on both
sides — **only the setter diverges**, which is why this is not visible as "the rune was not
lowered". The output parses and runs; assigning to `k.d` returns `undefined` instead of the
derived's own return value.

## Separator grid (`compileModule`, `generate: 'server'`)

| separator | verdict |
|---|---|
| `= $derived(` (one space) | match |
| `=$derived(` | match |
| `=  $derived(` (two spaces) | **DIFF** |
| `=\t$derived(` | **DIFF** |
| `=\n\t\t\t$derived(` | **DIFF** |
| `= /* c */ $derived(` | **DIFF** |
| `= $derived(` (NBSP) | **DIFF** |
| `= ﻿$derived(` (BOM) | **DIFF** |

Control — the same eight separators through `compile(..., { generate: 'server' })` on a
component whose instance script holds the identical class: **8/8 match**.

## Sites

`crates/rsvelte_core/src/compiler/phases/3_transform/server/transform_script.rs`

- `:5285-5290` `is_derived_field`
- `:5344-5346` `is_state_field`
- `:5428-5430` and `:5486-5488` the same two tests repeated in the member loop

All four are `memmem::find(…, b"= $rune(")` over the raw member text.

## Why this is the same defect class as #3456 / #3470

A literal needle containing a space encodes "the source is formatted this way", which no
source is obliged to be. `shared/js_scan.rs`'s `find_code` does **not** close it — the bytes
here already *are* code; what is wrong is the assumption about how the whitespace between
them is spelled. The fix wants a separator-tolerant test (skip JS whitespace **and** block
comments after the `=`, and locate the `=` with a test that rejects `==` / `===` / `=>` /
`!=` / `<=` / `>=`), which is what #3456 is adding on the client side.

### LINKED ISSUE #3498: client: a class rune is not lowered when its value starts on a later line than the `=`
Follow-up to #3456. That issue is about how the separator between `=` and the rune is
*spelled*; this one is about the separator containing a **line break**, which the
class-field/constructor scanners cannot represent at all because they work one source line
at a time.

Two shapes reach it. In both the rune is left untouched, so the field is a plain property
holding a `Source` object — the output parses and runs, and reactivity is silently gone.

## 1. A constructor assignment whose value is on the next line

```svelte
<script>
	class K {
		constructor() {
			this.v =
				$state(1);
		}
	}

	const k = new K();
</script>

<b>{k.v}</b>
```

```js
// official (client)
class K {
	#v;

	get v() {
		return $.get(this.#v);
	}

	set v(value) {
		$.set(this.#v, value, true);
	}

	constructor() {
		this.#v = $.state(1);
	}
}

// rsvelte
class K {
	constructor() {
		this.v = $.state(1);
	}
}
```

A **field** declaration in the same position is already handled (`v =\n\t$state(1);` matches),
so this is the constructor-assignment scanner alone.

## 2. A `//` comment as the separator

```svelte
<script>
	class K {
		v = // c
			$state(1);
	}

	const k = new K();
</script>

<b>{k.v}</b>
```

rsvelte emits `v = // c` / `$state(1);` verbatim — the rune is never lowered. A `//` comment
always puts the value on a later line, so this is the same root cause as (1) reached from the
field path.

## Grid

`compile`, all three targets, over the ten class shapes of the #3456 grid
(`field` / `private-field` / `static` / `raw` / `field-derived` / `derived-by` / `ctor` /
`ctor-private` / `quoted` / `class-expr`):

| separator | diverging targets |
|---|---|
| `=\n\t\t\t` | `ctor`, `ctor-private` only — client + client-dev |
| `= // c\n\t\t\t` | **all ten shapes** — client + client-dev |

`server` is clean on both (the SSR path is pure-AST). `compileModule` reproduces (1) as well.

## Sites

`crates/rsvelte_core/src/compiler/phases/3_transform/client/class_transforms.rs`

- `parse_constructor_state_assignment` takes `line` and never looks past it.
- `skip_ws_and_comments` (added for #3456) skips JS whitespace and `/* … */`, but not `//`,
  and by construction cannot cross the end of the line it was handed.

The field path already solves (1) with a "the initializer starts on a later line" head test;
the constructor path has no equivalent, and neither path admits a `//` comment.

### LINKED ISSUE #3510: client: `$props()` is left unlowered when U+FEFF separates it from the `=`
`canonicalize_props_call` gates every `$props()` lowering, and its regex is

```
=\s*(?:(?://[^\n]*\n|/\*(?s:.*?)\*/\s*))*\$props\s*\(\s*\)
```

Rust's `regex` resolves `\s` to Unicode `White_Space`, which **does not contain U+FEFF**.
JavaScript's `\s` does. So a zero-width no-break space between the `=` and the rune makes the
canonicalisation miss, the `= $props()` byte matchers below it miss in turn, and the
declaration is emitted verbatim.

## Reproduction

```svelte
<script>
	let { a = 1 } =﻿$props();
</script>

<b>{a}</b>
```

(the character after `=` is U+FEFF)

```js
// official (client)

// rsvelte (client)
let { a = 1 } = $props();
```

`$props` is not a runtime export, so the component throws `ReferenceError: $props is not
defined` on first render. The output parses, so the parse oracle cannot see it.

## Scope

Whitespace grid × `$props()` shape, `compile`, three targets:

| separator | verdict |
|---|---|
| one space / none / two spaces / tab / newline | match |
| U+00A0 NBSP / U+3000 / U+2009 / U+2028 / U+000B / U+000C | match |
| **U+FEFF** | **client + client-dev unlowered** |

All five shapes reach it: `let { a = 1 } = $props()`, `let { a } = $props()`,
`let { a, ...rest } = $props()`, `let p = $props()`, `let { a = $bindable(1) } = $props()`.
`server` is unaffected.

**No corpus repro is possible.** The fmt oracle rewrites U+FEFF to a single ASCII space
(verified with `od -c` on the formatter's output), so the committed form of the repro
reproduces nothing — the gate has to be a Rust unit test on `canonicalize_props_call`.

## Fix

`crates/rsvelte_core/src/compiler/phases/3_transform/utils.rs:1275` — spell the class
`[\s\u{feff}]` at all four sites in the pattern.

Related: #3456 (the same "the separator is not one ASCII space" family, class fields),
#3470, #3490, #3498.

baseballyama:
**Draft — claiming the cluster before measuring, so a second agent does not pick up any of the five.**

Closes #3470.
Closes #3456.
Closes #3490.
Closes #3510.
Closes #3498.

## Why one PR

These are five reports of **one** assumption: a phase-3 scan locates a construct with a **literal needle that contains whitespace**, so the needle encodes a formatting convention rather than a lexical fact.

| issue | the needle's assumption |
|---|---|
| #3470 | `export let` / `class K` — exactly one ASCII space between keyword and name |
| #3456 | a class field's rune initializer — exactly one space after `=` |
| #3490 | `compileModule` (server) class `$derived` — exactly one space between `=` and the rune |
| #3510 | `$props()` — U+FEFF between `=` and the call |
| #3498 | a class rune whose value starts on a **later line** than the `=` |

This is the `AGENTS.md` **entry-point scan** family: #2253 hardened the *body* scans (`find_matching_bracket`, `code_bracket_depth`) to be comment- and string-aware, and left the scans that decide **where to start counting** as plain byte searches. #2986/#2987/#2988 were the `opaque-keyword` instalment of the same thing; this is the whitespace instalment.

## What the deliverable is

Not five patched needles. Per the assignment:

1. **The denominator** — every whitespace-containing literal needle under `phases/3_transform/`, enumerated, with how many are covered by these five reports and how many are **unreported**.
2. **A grid on `whitespace kind × needle`** — n spaces / tab / newline / U+FEFF / comment / line continuation — since the five issues are five cells of that product and the rest of the product is unmeasured.

The precedent in this repo for exactly this method is "189 cells, 63 divergences". The result to report is the **set of sites sharing the assumption**, not the sites I happened to touch.

## Status

- [ ] enumerate the needles (denominator)
- [ ] build the whitespace × needle grid against the official compiler
- [ ] fix
- [ ] negative control
- [ ] server / client-dev, per the #3027 two-ports rule
- [ ] pattern-corpus files + README rows
- [ ] changeset


baseballyama:
Closing this draft. It was opened as a **claim** on the cluster before measuring, and the agent that opened it was terminated by a session limit before it wrote any code — the branch carries a single empty commit, so nothing is lost.

Two of the five are already fixed elsewhere: **#3456 and #3510 ship in #3534**, which measured 10 class shapes x 14 separators x 3 targets = 420 cells, 63 -> 22 divergences. The remaining three (#3470, #3490, #3498) are unclaimed again and are not blocked by this PR.

The measurement this draft did produce is worth keeping, so recording it here rather than losing it with the branch. Under `phases/3_transform/`, excluding test code, there are **75 distinct whitespace-containing literal needles across 212 call sites**. Split by whether the input is under rsvelte's own control:

| population | needles | sites |
| --- | --- | --- |
| reads **user source** (encodes a formatting assumption) | **61** | **170** |
| reads rsvelte's own generated text (`js_ast/codegen.rs`, `css.rs`, `client/formatting.rs`) | 31 | 42 |

The five issues touch 7 of the 61; **54 are unreported**.

And the finding that should shape whoever picks up #3470: of 56 needle shapes, exactly three (`await`, `catch`, `finally`) already carry non-space variants — so the "write out the whitespace variants" repair exists in-tree and was never generalised. It also **fails on the axis of the defects it would be copied to fix**: `shared/async_body.rs` enumerates exactly `" "`, `"\n"`, `"\t"`, so U+FEFF, `\r`, `\v` and `\f` pass straight through, and U+FEFF is #3510. Copying that pattern to 53 sites would not have fixed it. Prefer finding the keyword as a **token** and skipping whitespace by predicate; an enumeration is a claim that its author knew the whole domain.

#3534 has since added the two tools that follow from this — `skip_ws_and_comments` (skips JS whitespace and block comments) and `find_assignment_eq` (finds an assignment `=`, excluding `==` `===` `=>` `!=` `<=` `>=`). Note that `shared/js_scan.rs`'s `find_code` does **not** close this class: it answers "is this byte code rather than a string/comment", which says nothing about the spelling of whitespace *within* code.

Issue #3515, opened 2026-08-22
client: a comment between `=` and a top-level rune lands in the wrong place in the output


baseballyama:
Distinct from #3456, which is about a class field's rune not being *lowered*. Here the rune
is lowered correctly and the **comment** ends up somewhere else than official puts it. Only
the declarator paths (`$derived`, `$props`) are affected; a `$state` declarator matches.

Official's own placements are odd — esrap threads the comment onto whatever node its cursor
is at, which for a lowered `$derived` is the synthesized arrow's (empty) parameter list — but
byte equality is the goal, so rsvelte has to reproduce them.

## 1. `const d = /* c */ $derived(…)`

```svelte
<script>
	let v = $state(1);
	const d = /* c */ $derived(v * 2);
</script>

<b>{d}</b>
```

```js
// official (client)
const d = $.derived((/* c */) => v * 2);
// rsvelte
const d = /* c */ $.derived(() => v * 2);
```

client-dev the same, one level in: official `$.tag($.derived((/* c */) => v * 2), 'd')`,
rsvelte `/* c */ $.tag($.derived(() => v * 2), 'd')`.

## 2. `let { a = 1 } = /* c */ $props()`

```js
// official (client)
let a = $.prop($$props, 'a', 3, 1 /* c */);
// rsvelte
/* c */
let a = $.prop($$props, 'a', 3, 1);
```

`canonicalize_props_call` deletes the comment as part of normalising the call, and the
comment then reappears from the source map / comment cursor on its own line.

## 3. A `//` comment as the separator

Same two shapes, and here **`server` diverges too** for `$props()`:

```js
// official (client)
let a = $.prop($$props, 'a', 3, 1 // c
);
// rsvelte
let { a = 1 } = // c
$props();
```

— i.e. with a line comment the props declaration is not lowered at all, which is #3456's
failure mode reached from the declarator path.

## Grid

Separator × host, `compile`, three targets. 11 hosts × 9 separators × 3 targets = 297 cells,
**18 diverging**, all in three hosts:

| host | diverging separators |
|---|---|
| `const d = <sep> $derived(v * 2)` | `/* c */`, `// c\n` (client, client-dev) |
| `let { a = 1 } = <sep> $props()` | `/* c */`, U+FEFF (#3510), `// c\n` (client, client-dev, **server**) |
| `let { a, ...rest } = <sep> $props()` | same |

[...]

PR #3534, opened 2026-08-22
fix(client): lower a class rune whose `=` is not followed by one space

### LINKED ISSUE #3456: A class field's rune initializer is only lowered when exactly one ASCII space follows the `=`

### LINKED ISSUE #3510: client: `$props()` is left unlowered when U+FEFF separates it from the `=`

baseballyama:
Closes #3456
Closes #3510

Two sites of one defect: **a literal needle containing a space encodes "the source is
formatted this way"**, and no source is obliged to be.

## What was wrong

`class_transforms.rs` decided a class member was a rune field with
`memmem::find(b"= $state(")` (and the `format!("= {}(", rune)` variants), and rebuilt the
constructor form with `format!("{} = {}", target, pattern)`. Both spellings carry exactly one
ASCII space. A tab, two spaces, a newline, a non-ASCII JS space (U+00A0, U+3000, U+2009,
U+2028, U+000B, U+000C) or a block comment therefore left the field **unlowered** — the output
parses and runs, and the field holds a `Source` object with no reactivity.

`canonicalize_props_call` in `utils.rs` is the same defect with a different mechanism: its
regex spells whitespace `\s`, which Rust resolves to Unicode `White_Space` and so **excludes
U+FEFF**, while JS includes it. `let { a } =<U+FEFF>$props()` was emitted verbatim and throws
`ReferenceError: $props is not defined`.

## What changed

- `skip_ws_and_comments` skips JS whitespace **and** block comments; `find_assignment_eq`
  locates the assignment `=` while rejecting `==` / `===` / `=>` / `!=` / `<=` / `>=`.
- A comment separator is carried into the emitted initializer (`init_prefix`) instead of being
  dropped, so `v = /* keep */ $state(0)` keeps its comment where official puts it.
- `tag_edit` reflowed the dev-mode `$.tag(…)` wrap for **any** comment; measured against
  official, the rule is that a comment region **spanning a line** reflows and a one-line one
  stays inline. It now reflows only in the first case — which is still what a `//` needs,
  since inline it would swallow the rest of the call.
- The props regex spells the class `[\s\u{feff}]`.
- `phases/1_parse/mod.rs`: `mod parser` → `pub(crate) mod parser`, so phase 3 can use
  `is_js_whitespace` (the predicate that already knows U+FEFF is JS whitespace).

## Measurement

10 class shapes × 14 separators × 3 targets = 420 cells.

| | diverging |
|---|---|
| before | 63 |
| after | **22** |

All 22 remaining are the two shapes filed as **#3498** (a constructor assignment whose value is
on a later line, and a `//` comment separator) — a different mechanism, because those scanners
are line-oriented and cannot represent a separator containing a newline at all.

`compileModule` on the same axis went 12 → 7; the 7 are #3490 (server) and #3498 (client).

`$props()` shapes × 13 separators × 3 targets: the U+FEFF row is gone. The remaining rows are
#3515 (comment placement) and `$props.id()`, which has its own detection and is noted on #3510.

## Gates

- `cargo test -p rsvelte_core --lib` — 1732 passed, 0 failed, including the new
  `canonicalize_props_call` U+FEFF case.
- Every `.svelte` / `.svelte.(js|ts)` in `compatibility/pattern-corpus` (720 files × 3 targets
  = 2160 comparisons) run against both compilers before and after: **identical divergence sets**
  — no regressions and no accidental fixes.
- `cargo fmt --check`, `cargo clippy -p rsvelte_core --all-features -- -D warnings`.

## Why the repro looks the way it does

`3456-class-rune-comment-separator.svelte` uses a **block comment** for every separator, and
that is forced rather than chosen: the fmt oracle rewrites a tab, two spaces, a newline and
every non-ASCII JS space to a single space, so the committed form of any other repro
reproduces nothing. The U+FEFF `$props()` case has **no** possible corpus repro for the same
reason (verified with `od -c` on the formatter's output) — it is pinned by a unit test instead.

Eight hosts in one file because no single one reaches all four changed sites: a public field, a
`#private` one, `$state.raw`, `$derived`, `$derived.by`, a quoted key, a constructor assignment
and a class expression. The two-comment row (`/* a */ /* b */`) is the case that separates
"keep the comment inline" from "reflow the call" in the `$.tag` wrap.

## Also in this PR

One `AGENTS.md` paragraph, because it cost several sessions today: the file records that a
**cancelled** run is indistinguishable from a green one in the branch header, and the other
direction is just as bad — `Tests` is a rollup that exits 1 unless every shard is `success`, so
a mass cancellation shows up as `FAILURE` with `BULK: cancelled UNIT: cancelled …` in the step
env and no test output at all. Four PRs read as red that way this afternoon.


baseballyama:

baseballyama:
**This PR's red CI is not about this PR.** Every job fails, including ones that cannot be affected
by the change (`Format`, `Clippy`, `Playground lockfile`, `Workflow trigger guard`), and they all
fail at **"Prepare all required actions"** — before the checkout, so no job reaches a step of its
own:

```
##[error]The action actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 is not allowed
in baseballyama/rsvelte because all actions must be from a repository owned by baseballyama.
```

`Allow <owner> actions only` was set on the repository for a window on **2026-08-23 ~04:30Z**, which
is when this run started. It is not set now:

```
$ gh api repos/baseballyama/rsvelte/actions/permissions
{"enabled":true,"allowed_actions":"all","sha_pinning_required":false}
```

Nothing re-evaluates a completed run, so the red mark will stand until a **new** run exists. I have
deliberately not touched this branch — it may be checked out in another session, and a server-side
branch update would make that checkout diverge. The re-trigger is the owner's call.

`gh run rerun` is the wrong instrument here: it replays the frozen event payload rather than
re-evaluating current state, and re-entering the `cancel-in-progress` group can kill a good run.
A push (or `gh pr update-branch`) is what produces a fresh verdict.

Context and the full triage are in #3630 — of 58 open PRs, 48 have no verdict at all and this is
one of only two whose failure had any cause other than "the run never happened".

Issue #3175, opened 2026-08-20
server: legacy-mode `$derived(...)` is constant-folded into the template, freezing the rendered value


baseballyama:
In **explicit legacy mode**, the server constant-folds a `$derived(...)` initializer
into the template even though `$derived` is a store subscription there, not a rune. The
rendered value freezes at the folded constant.

## Repro

```svelte
<svelte:options runes={false} />

<script>
	let x = $derived(1);
</script>

<p>{x}</p>
```

`generate: 'server'`:

```diff
-		$$renderer.push(`<p>${$.escape(x)}</p>`);   // official
+		$$renderer.push(`<p>1</p>`);               // rsvelte
```

The **declaration is lowered correctly** — with a `derived` store in scope rsvelte emits
`let x = $.store_get($$store_subs ??= {}, '$derived', derived)(1);`, byte-identical to
official. Only the template read is folded, so the output parses, runs, and silently
renders the wrong thing.

## Measured scope

| axis | result |
|---|---|
| target | **server only** — client output is byte-identical to official |
| rune name | **`$derived` only** — `$state(1)`, `$derived.by(() => 1)`, `$effect(1)`, `$host()`, `$props()` all byte-identical |
| negative control | a plain store call `$foo(1)` (store `foo`) is byte-identical — the fold keys on the **name**, not on "callee is a store subscription" |
| store present | not required — reproduces with no `derived` declaration at all |
| how legacy is selected | reproduces via **both** `<svelte:options runes={false} />` and the `runes: false` **compile option** |
| runes mode | correct (folding is right there) |

`$state` and `$derived` share the fold arm in
`3_transform/server/evaluate.rs:1354` yet behave differently, so that arm is not what
decides this. The value comes from `binding.initial`, which
`2_analyze/visitors/variable_declarator.rs:433` fills with **the `$derived` argument's
estree JSON** — for `$derived` specifically:

```rust
b.initial = extract_literal_string_typed(arg).or_else(|| {
    if rune_name == "$derived" { Some(arg.to_json_string()) } else { None }
});
```

and `evaluate.rs:1145` reads exactly that back (`// estree-JSON dump (from `$derived(...)`
/ `{@const ...}` initials)`). Upstream gates the whole rune branch of that visitor on
`context.state.analysis.runes` (`VariableDeclarator.js:19`), which is the guard this path
needs.

## Why this is newly reachable

Before #3172 this source was **rejected** with `rune_invalid_usage` — it is one of the
[...]

PR #3549, opened 2026-08-22
fix(server): drop a reassigned `let` before resolving a `const` from it

### LINKED ISSUE #3530: server: the SSR constant fold inlines a `let`'s initializer even after the variable is reassigned
`const r = w;` where `w` is a `let` with a literal initializer is folded to that initializer,
whatever writes happened to `w` in between. The server renders the wrong text; the client is
byte-identical to official throughout, so this is the client/server two-ports shape again.

## Reproduction

```svelte
<script>
	let w = 1;
	w += 2;
	const r = w;
</script>

<b>{r}</b>
```

```js
// official (server)
$$renderer.push(`<b>${$.escape(r)}</b>`);   // renders <b>3</b>

// rsvelte (server)
$$renderer.push(`<b>1</b>`);
```

`client` and `client-dev` match official on every row below — the divergence is server-only,
the output parses, and the wrong value is a plain literal in the HTML, so it also produces a
hydration mismatch rather than a crash.

## What separates the failing rows from the passing ones

| shape | verdict |
|---|---|
| `let w = 1; w += 2; const r = w;` → `{r}` | **DIFF** (folds to `1`) |
| `let w = 1; if (…) w = 2; const r = w;` | **DIFF** (folds to `1`) |
| `let w = 0; for (const n of [1, 2]) w += n; const r = w;` | **DIFF** (folds to `0`) |
| `let w = 1; w = 2; const r = w; w = 3;` | **DIFF** (folds to `1`) |
| `let w = 'a'; w += 'b'; const r = w;` | **DIFF** (folds to `a`) |
| `let w = 1; const r = w; w = 2;` | **DIFF** (folds to `1` — the *value* is right here, only the emitted shape differs) |
| `let w = 1; const r = w;` (no write) | ok — both fold |
| `const w = 1; const r = w;` | ok — both fold |
| `let w = 1; w += 2;` → `{w}` **read directly** | ok |
| `let w = { a: 1 }; w = { a: 2 }; const r = w.a;` | ok |

Two things the table pins down:

- The fold is on the **`const` alias**, not on the read. Reading the reassigned `let` directly
  in the template is correct; only routing it through `const r = w` reaches the bad path.
- The initializer has to be a **literal**. An object initializer is already excluded, which is
  why `object-init-write` passes and is not evidence that writes are being tracked.

## Operator coverage

Every write form reaches it. Measured with `let w = <init>; <write>; const r = w;`:

`w = 2`, `w += 2`, `w++`, `w <<= 2`, `w **= 3`, `w ??= 2`, `w ||= 2`, `w &&= 2`, and a write
from inside a function that is then called — **all DIFF (server)**. `o.a ??= 2` on an object
property and a write to a destructured `let [w] = [null]` are both **ok**.

## Where

The premise "this binding's value is known" is answered from the binding's *initializer*
without asking whether the binding is ever written. AGENTS.md already records three client-side
implementations of that question (`is_expression_known_json`, `identifier_has_reactive_state`,
`is_initial_value_literal_or_known`); this is the server fold's copy, and it is the one that
changes rendered output rather than just missing an optimisation.

Measured identically on `main` and on the #3456 branch. Related: #3175 (the same fold, wrong
for a different reason), #3027 (the two ports of `scope.evaluate`).

baseballyama:
Closes #3530

**The reassignment filter was not missing — it was not reaching its target.** That is the whole
bug, and it is easy to read past: `extract_constant_vars` already drops a `let` the source
writes to. What it does not do is drop it *before* the second pass, the one that resolves a
declaration whose initializer is another binding. So

```svelte
<script>
	let w = 1;
	w += 2;
	const r = w;
</script>

<b>{r}</b>
```

removes `w` from the constant set and leaves `r` holding `1` — and nothing removes `r`, because
`r` itself is never written.

```js
// official (server)
$$renderer.push(`<b>${$.escape(r)}</b>`);   // renders <b>3</b>
// rsvelte before
$$renderer.push(`<b>1</b>`);
```

The client was byte-identical to official throughout, so this rendered wrong text and produced
a hydration mismatch rather than a crash — the client/server two-ports shape, same as #3027 and
#3175.

## Second cause

The filter's operator list carried **6 of the 17** compound assignments; `??=` `||=` `&&=`
`<<=` `>>=` `>>>=` `**=` `%=` `&=` `|=` `^=` were all writes it could not see.

Whether an enumeration is the right shape depends on whether the domain is closed. The set of
JS whitespace characters is **not** closed, which is why #3456 replaced its enumeration with a
predicate; the set of assignment operators **is** closed by the language spec, so this stays an
enumeration — collected into one constant rather than a chain of `starts_with`, so the next
operator is added in one place.

## The axis, enumerated

The fold rests on *a declaration's initializer is the declaration's value*. #3175 broke that
premise one way (a legacy `$derived` declares the call's **result**); this breaks it another.
Rather than enumerate more spellings of "write", I enumerated the **reasons the premise can
fail** and measured all of them — 18 declaration shapes × 3 targets:

| reason | before |
|---|---|
| written at the top level, from a closure, in `try`/`catch`, in a labeled block, in `$effect`, in a loop body | **6 DIFF (server)** |
| `var w = 1; var w = 2;` hoisted redeclaration | ok |
| `const { a = 1 } = o`, key present **and** absent | ok |
| an inner scope shadows the name | ok |
| object mutation, array mutation | ok |
| `export let`, `$props()` default | ok |
| `$state` never written, `$state` written from a handler | ok |
| legacy-mode `$derived` (#3175) | ok |

**6 of 54, all one mechanism. There is no third reason** — the premise does not need a general
"can this initializer be trusted" analysis, it needs one missing question.

Two of those rows are the reason this is not a wider change:

- `$state` written from a handler is **already correct**, so a write set exists and this fold
  simply was not consulting it.
- object and array mutation are **already correct only because a non-literal initializer never
  becomes a constant**. That restriction looks over-conservative and is load-bearing: lifting it
  without tracking mutation would widen this bug. It is now stated in `drop_reassigned`'s doc
  comment so a future optimisation PR does not remove it.

## Gates

- Three grids, before → after: reasons 6/54 → **0/54**, write operators 11/39 → **0/39**,
  reachable shapes 6/30 → **0/30**.
- Every `.svelte` / `.svelte.(js|ts)` in `compatibility/pattern-corpus` (721 files × 3 targets)
  compiled with both compilers before and after: the **divergence sets are identical**, not just
  the counts — no regressions, no accidental fixes.
- `cargo test -p rsvelte_core --lib` 1732 passed / 0 failed; `cargo fmt --check`;
  `cargo clippy --all-targets --all-features -- -D warnings`.
- Oracle: `submodules/svelte` @ **5.56.9**, asserted from the `generated by Svelte v…` header in
  the compared output rather than from the import path. The reasons grid was additionally run
  against npm **5.56.10** (asserted from the module's `VERSION` export, which the published
  build exposes instead of the header) and is **0/54** there too, so the pending submodule bump
  in #3195 does not move this result.

## Repro

`3530-fold-after-write.svelte` varies the **write** (`+=`, `??=`, `<<=`, a loop body, a
`try`/`catch`, a closure that is called) rather than the read, because that is where the
operator gap lives, and carries `const alias = frozen` as the negative control — the fix moves
in the direction of folding *less*, so without a row that must still fold, a grid of zeros would
also be satisfied by folding nothing. The ten shapes the enumeration found already correct are
named in the `pattern-corpus/README.md` row so the next person does not repeat the enumeration.


baseballyama:

baseballyama:
Closing in favour of #3559 (merged as `e05521e78`), which is a **strict superset** of this PR — measured, not assumed.

All three trees were built from one common base (`origin/main` = `55176a596`, both PRs merged locally, neither branch pushed) and every row compared **raw-byte** against the official compiler, which is stricter than the corpus gate's oxfmt-normalised comparison.

```
                                  main / +3559 / +3549
3530-fold-after-write.svelte      DIFF / MATCH / MATCH   <- this PR's own repro
f2 f3 f4 f5                       DIFF / MATCH / MATCH
f1 f7 f8 f9 f10 f11               DIFF / MATCH / DIFF    <- #3559 only

total DIFF from official:  main 64/148  ->  #3559 53 (-11)  ->  #3549 59 (-5)
rows made worse by either PR: 0
```

**Every row this PR repairs, #3559 also repairs — including this PR's own repro file.** The six rows in the gap are out of a text scanner's reach in principle: `[a] = [5]` is a write through a destructuring pattern, which the line scan cannot see.

Nothing here is a criticism of the change: five fixes, zero regressions, and the discriminating rows behaved exactly as pre-registered. The two PRs also touch the same region of `3_transform/server/helpers.rs`, so keeping both open would force a rebase on whichever landed second for no additional coverage.

#3559 additionally passed `ssr` 99/99 (the suite enforces its own fixture floor at `ssr.rs:21` — `MIN_SSR_FIXTURES: usize = 99` — so a vacuous pass is not possible) and its own 8 tests.

Thanks for the work — the analysis in this PR is what made the comparison possible.

Issue #3505, opened 2026-08-22
css.map drops every segment inside a rule that nests (declarations before an `&` block, and the outer closing brace)


baseballyama:
`css.map` loses every segment inside a rule that nests: the declarations before an `&`
block, and the outer rule's closing brace, map to nothing.

`css.code` is **byte-identical to official** in all four targets, so nothing that compares
generated text can see this. Only the map differs.

## Repro

```svelte
<b class="a">y</b>
<style>
	.a {
		color: red;
		&:hover {
			color: blue;
		}
	}
</style>
```

`compile(src, { generate: 'client', dev: false, filename: 'X.svelte' })`, comparing
`result.css`:

| | official | rsvelte |
|---|---|---|
| `css.code` | identical | identical |
| `css.map.mappings` | `;AAEA,CAAC,iBAAE,CAAC;AACJ,EAAE,UAAU;AACZ,EAAE,CAAC,MAAM,CAAC;AACV,GAAG,WAAW;AACd;AACA;` | `;AAEA,CAAC,iBAAE;;EAED,CAAC,MAAM,CAAC;AACV,GAAG,WAAW;AACd;;` |
| total segments | **17** | **11** |
| segments per output line | `[0,4,3,5,3,1,1,0]` | `[0,3,0,4,3,1,0,0]` |

The generated stylesheet is

```
1
2	.a.svelte-1lj1c1s {
3		color: red;
4		&:hover {
5			color: blue;
6		}
7	}
8
```

so the two lines that lose all their segments are **line 3, the declaration that precedes
the nested block**, and **line 7, the outer rule's closing brace**. Line 2 loses one of
its four. A devtools user clicking `color: red` in the generated stylesheet lands nowhere.

## Scope

- **All four targets, dev and prod.** `client`, `server`, `client-dev`, `server-dev` all
  diverge; `css.code` is equal in all four. It is not a dev-only artifact.
- **Any nesting, not only `&`.** An `@media` wrapper does the same thing:

  ```
  sv  ;AAEA,CAAC,wBAAwB;AACzB,EAAE,iBAAE,CAAC;AACL,GAAG,UAAU;AACb;AACA;
  rs  ;;AAGA,EAAE,iBAAE,CAAC;AACL,GAAG,UAAU;AACb;;
  ```

  Here rsvelte drops the `@media` prelude's own mapping entirely (output line 2 has zero
  segments against official's one).
[...]

Issue #3406, opened 2026-08-22
CSS declarations: `.a { : red }` is over-rejected, and two `css_*` error positions differ


baseballyama:
## Summary

Three divergences in CSS **declaration** handling — one over-rejection and two error positions.
Found by a 48-case sweep of declaration values and at-rule contents (45 of 48 agree).

### 1. Over-rejection: a declaration with no property name

```css
.a { : red }
```

| | |
|---|---|
| official | compiles, emits `.a.svelte-1lj1c2h { : red }` |
| rsvelte | `css_empty_declaration@3:5` |

A file official accepts does not compile. Official passes the malformed declaration through
verbatim.

### 2. `css_empty_declaration` position: leading vs trailing `;;`

```css
.a { ;; color: red ;; }
```

| | |
|---|---|
| official | `css_empty_declaration@3:20` — the **trailing** `;;` |
| rsvelte | `css_empty_declaration@3:5` — the **leading** `;;` |

Both reject, same code, different span. Official tolerates the leading empty declarations and
reports the trailing pair; rsvelte reports the first one it meets.

### 3. `css_expected_identifier` position on a `{}`-valued custom property

```css
.a { --x: { color: red }; }
```

| | |
|---|---|
| official | `css_expected_identifier@3:9` |
| rsvelte | `css_expected_identifier@3:24` |

Both reject (official's rejection is itself tracked upstream as #3052); only the position differs —
official points at the `{` that opens the value, rsvelte at the `;` that closes it.

## The 45 that agree

Value grammar: `var()` with fallbacks (plain, nested, comma-bearing), empty and
semicolon-bearing custom properties, an unbalanced `(` in a custom property, nested `calc()`,
`calc(1px+2px)`, `url()` with an escaped paren / spaces / a data URI, string escapes (line
continuation, `\1F600`, single quotes), `!important` spaced / uppercased / preceded by a comment, a
comment inside a value, a newline in a value, an uppercase property name, a vendor property, and a
declaration with no value at all.

At-rules: `@media` with `and` / commas / `not` / `only` / a range query / nested parens / an
unknown feature; `@supports` with `and`+`or` and with `font-tech()`; `@keyframes` with percentage
selectors / `from`-`to` / a quoted name / nested in `@media`; a multi-value `animation` shorthand;
`animation-name: var(--n)`; `@layer` inside `@supports`; `@scope` with no prelude; a named
[...]

PR #3553, opened 2026-08-22
fix(css): combinator runs, `||` parsing and scoping, malformed attribute selectors, and empty selector-list entries (#3404, #3405, #3406)

### LINKED ISSUE #3404: CSS: `>>` and `>>>` are silently rewritten to `>`, and `||` is rejected
## Summary

Two divergences in combinator handling, in opposite directions.

### `>>` and `>>>` are silently rewritten to `>`

```css
/* source */    .card >> .a { color: red }
/* official */  .card.svelte-1lj1c2h >> .a:where(.svelte-1lj1c2h) { color: red }
/* rsvelte  */  .card.svelte-1lj1c2h >  .a:where(.svelte-1lj1c2h) { color: red }
```

Same for `>>>` and for the nested form `.card { >> .a { … } }`. Official passes the token through
untouched; rsvelte emits a **different combinator**, so the rule matches a different set of
elements at runtime. `>>>` in particular is a deep-selector convention some toolchains still emit,
and collapsing it to a child combinator changes what it selects.

### `||` is rejected

```css
.a || .b { color: red }
```

official compiles it (`.a || .b.svelte-1lj1c2h { color: red }`); rsvelte raises
`css_expected_identifier@3:3`. An over-rejection — a file official accepts does not compile.

## Matrix

8 cases × 2 targets. **4 diverge.**

|---|---|---|
| `.a \|\| .b` | compiles | `css_expected_identifier@3:3` |
| `.card >> .a` | `>>` preserved | rewritten to `>` |
| `.card >>> .a` | `>>>` preserved | rewritten to `>` |
| `.card { >> .a { … } }` | `>>` preserved | rewritten to `>` |

**Controls that match (4):** `>`, `+`, `~`, and `.card ::deep .a`.

### LINKED ISSUE #3405: CSS: six malformed selectors are accepted, two of them emitting invalid CSS (`[]` and `,,`)
## Summary

Six malformed selectors that official rejects are accepted by rsvelte, and **two of them are
emitted as invalid CSS**.

|---|---|---|
| `[*\|data-k] { … }` | `css_expected_identifier@3:1` | compiles, emits **`[].svelte-X { color: red }`** |
| `[\|data-k] { … }` | `css_expected_identifier@3:1` | compiles, emits **`[].svelte-X { color: red }`** |
| `[svg\|data-k] { … }` | `expected_token@3:4` | compiles, prunes it, warns `css_unused_selector` |
| `.a, { … }` | `css_expected_identifier@3:4` | compiles as `.a.svelte-X { … }` |
| `, .a { … }` | `css_expected_identifier@3:0` | compiles as `.a.svelte-X { … }` |
| `.a,, .b { … }` | `css_expected_identifier@3:3` | compiles, emits **`.a.svelte-X,, .b.svelte-X`** |
| `, { … }` | `css_expected_identifier@3:0` | compiles, emits `/* (unused) , { … }*/` |

**Control:** `{ color: red }` with no selector at all — **both** reject, so the parser is not
uniformly permissive here.

## The two that emit invalid CSS are the worse half

`[*|data-k]` and `[|data-k]` lose the whole attribute name and emit `[]`, which no CSS parser
accepts — a stylesheet that breaks at the browser rather than at compile time. `.a,, .b` carries
the double comma straight through to the output for the same reason.

A namespaced attribute selector is not a valid Svelte input either way, so the fix is to reject
these where official does; but the `[]` output means the current behaviour is not "accepts extra
input", it is "produces a broken stylesheet".

Related: #3134 covers namespaced **type** selectors losing their prefix (`svg|circle`); this is the
attribute-selector half, and it fails differently — the name is dropped, not the prefix.

baseballyama:
CSS combinator tokens, `||` scoping, malformed attribute selectors, empty selector-list
entries, and an attribute selector's source spacing.

Closes #3404
Closes #3405

Addresses **#3406 item 1** (`.a { : red }` over-rejection). Items 2 and 3 are error
*positions* and are untouched, so #3406 stays open.

#3404 now closes on both halves: the parse half (the over-rejection and the `>>` / `>>>`
rewrite, which is what its matrix reports) and the *scoping* half the issue does not
report — see "Where the scoping went wrong" below.

## Provenance: this is the union of two independent attempts

Two agents attacked #3404/#3405 and both were killed before committing. Their work survived
as staged-but-uncommitted diffs in two different worktrees
(`fix/3404-3405-3406-css-selector-parse`, whose only commit was a devtools probe, and
`fix/3405-3404-css-selector-parser`, which had no commits at all). Measured on one 69-cell
grid against the official compiler, they fixed **disjoint** sets:

| family | `main` | attempt A | attempt B | union | **this PR** |
|---|---|---|---|---|---|
| #3404 combinators | 7/14 | 12/14 | 12/14 | 12/14 | **14/14** |
| #3405 malformed selectors | 11/22 | 22/22 | 22/22 | 22/22 | 22/22 |
| #3406 declarations | 14/18 | **15/18** | 14/18 | **15/18** | 15/18 |
| attribute-spacing shapes | 7/15 | 12/15 | **14/15** | **14/15** | **15/15** |
| total | 39/69 | 61/69 | 62/69 | 63/69 | **66/69** |

Named at cell level, because the totals alone would also be consistent with "B is simply
better":

- **A fixes exactly one cell B does not**: `decl/no-property` (`.a { : red }`).
- **B fixes exactly two cells A does not**: `x/attr-inner-space` (`[ x ]`) and
  `x/attr-inner-space-eq` (`[ x = "y" ]`).
- **Neither direction is empty**, so neither attempt contains the other and the answer is
  a union rather than a pick. 24 cells fixed, **0 regressed**, for the union over `main`.

The `||`-scoping commit then fixes three more — `comb/column`, `comb/column-spaced`,
`x/column-both-present` — again with 0 regressed. The 3 residual cells are all #3406
items 2/3: `decl/empty-both-sides`, `decl/leading-empty`, `decl/custom-brace-value`.

## What was wrong: the scanner

**A combinator was a `char`, not a token.** Upstream's `REGEX_COMBINATOR` is
`/(\+|~|>|\|\|)/y`: `||` is a combinator, and a lone `|` is a namespace separator. The
scanner matched `+ > ~` byte by byte, so `.a || .b` — which official compiles — raised
`css_expected_identifier`, and `>>` / `>>>` were read as one `>` followed by a compound
that starts with `>`, i.e. **collapsed to a single child combinator**. That is not a
formatting difference: the emitted rule matches a different set of elements. Upstream's
regex also reads a run one token at a time and keeps the last, but its in-place
MagicString rewrite leaves the whole run in the output, so the run has to survive.

**`parse_attribute_selector` never checked the name and never required the `]`.**
`read_identifier` returned `""` for `[*|data-k]` / `[|data-k]` and the code carried on, so
the whole attribute name was dropped and the output was **`[].svelte-X { color: red }`** —
a stylesheet no CSS parser accepts, i.e. a failure that lands at the browser rather than at
compile time. With no matcher it also skipped forward to the next `]`, so `[x` compiled.
Both are now `read_identifier` + `parser.eat(']', true)`.

Fixing that surfaced a third one the issues do not list: the flags (`i` / `s`) were read
*inside* the matcher branch, so **`[x i]` silently dropped the `i`** and emitted `[x]`.
**Only output equality can see this one.** The result is a valid selector, so the parse
gate and the output-validity (acorn/CSS-parseability) gate are both blind to it, and it
raises no warning and no error — the same shape as #2987 / #2988 in AGENTS.md, where a
rewritten module also parsed cleanly and only byte comparison against official reported it.

**The attribute selector was rebuilt rather than copied.** `name` / `matcher` / `value` /
`flags` cannot carry the author's spacing, so `[ data-k ]` came out `[data-k]`. Upstream
never touches the brackets, so the source text is what belongs in the output.

**An empty comma-separated segment was filtered out silently.** Upstream runs
`read_selector` on every segment, so an empty one reaches `read_identifier` and raises
there. `.a,`, `, .a` and `,` compiled; `.a,, .b` carried the double comma straight into the
output.

**An empty property is not an empty declaration.** Upstream's rule is
`!value && !property.startsWith('--')`, so `.a { : red }` is a declaration whose property
is `''` and whose value is `red`, and it is passed through verbatim. rsvelte required a
non-empty property before it would even build the node.

## Where the scoping went wrong

Making `||` parse only moved the divergence one phase along. `apply_combinator` in
`css-prune.js` handles `' '`, `'>'`, `'+'` and `'~'` and then ends on

```js
default:
    // TODO other combinators
    return true;
```

— it returns `true` **without recursing**. So an unhandled combinator *halts the backward
walk*: every relative selector to its left is never visited, and therefore is never marked
`metadata.scoped` and never consulted when `apply_selector` decides whether the rule is
used. Both consequences are observable in official's output.

rsvelte writes `metadata.scoped` **nowhere** — all three readers in `3_transform/css.rs`
default to `true` — so it scoped every compound and pruned against every compound:

| source | official | rsvelte before |
|---|---|---|
| `.a \|\| .b` | `.a \|\| .b.svelte-X` | `.a.svelte-X \|\| .b:where(.svelte-X)` |
| `.a \|\| .b > .c` | `.a \|\| .b.svelte-X > .c:where(.svelte-X)` | both sides scoped |
| `.zz \|\| .b` (no `.zz` in markup) | `.zz \|\| .b.svelte-X` | rule deleted as unused |
| `.b + .d \|\| .c` (no such sibling pair) | `.b + .d \|\| .c.svelte-X` | rule deleted as unused |
| `.a \|\| :global(.x)` | `.a.svelte-X \|\| .x` | `.a.svelte-X \|\| .x` (already right) |

One `reachable_complex_selector` now answers where the walk starts, and the scoping loop
and **both** unused deciders read it — the second one, `check_complex_selector_unused`, is
why the last two rows above were still pruned after the first attempt at this. It runs over
the `truncate`-ed relative-selector list, so a trailing `:global(...)` cannot become the
subject; without that, `.a || :global(.x)` regresses, which is the fourth row of the
directed set diff below.

Measured on a purpose-built 21-cell grid (the `||` position crossed with what sits on each
side, plus the six handled combinators as controls), counting the left and the right
compound as **separate cells** because both sides differ:

| tree | cells matching official |
|---|---|
| `main` | 4/21 |
| after the parse commits | 12/21 |
| first scoping attempt | 17/21 |
| **this PR** | **21/21** |

with 0 regressed at every step except the intermediate one, which regressed
`col-global-right` — caught by the grid and fixed by the `truncate` ordering above.

## Regression sweeps

Against the official compiler (submodule, **5.56.9**), all **0 regressed**:

- **679 real styled components** (the pinned Svelte test tree, `svelte.dev` content,
  `attractions`, `layercake`, `sveltestrap`) on `css.code`: identical before and after.
- The 283-cell pseudo-class-argument grid from #3130 / #3204 / #3133: 131/283 before and
  after — this PR is orthogonal to #3424's area.
- The 45-cell `@media`/nesting map grid from #3226 / #3505: identical before and after on
  the same base.

21 `css_*` integration test binaries in `rsvelte_core`: all green.
`cargo clippy --all-targets --all-features -- -D warnings`: clean.

## Negative controls

**Parse half** — `crates/rsvelte_core/tests/css_selector_parse_3404_3405_3406.rs` on
`main`: all 7 of the tests that existed at that point fail, each for its own predicted
reason:

```
a_combinator_run_survives_into_the_output
  left: ".card.svelte-70s02x >  .a:where(.svelte-70s02x) { … }"   (`>>` collapsed)
 right: ".card.svelte-70s02x >> .a:where(.svelte-70s02x) { … }"
the_column_combinator_is_read_and_a_lone_bar_is_not
  `||` is a combinator official accepts                            (rsvelte rejected it)
a_namespaced_attribute_name_is_rejected_…
  `[*|data-k] { … }` compiled but official raises css_expected_identifier:
  Some("\n\t[].svelte-70s02x { color: red }\n")                    (the invalid CSS)
an_unterminated_attribute_selector_is_rejected
  `[x { … }` compiled but official raises expected_token
an_empty_selector_list_entry_is_rejected
  `.a, { … }` compiled but official raises css_expected_identifier
well_formed_attribute_selectors_are_unchanged
  left: "[x].svelte-70s02x"    right: "[x i].svelte-70s02x"        (the dropped flag)
a_declaration_with_no_property_name_compiles
  `.a { : red }` was rejected but official compiles it
```

The eighth (`an_attribute_selector_keeps_the_source_spacing`) is the second commit's, and
its own control is the first commit's tree: `[ x ]` came out `[x]` there.

**Scoping half** — `crates/rsvelte_core/tests/css_unhandled_combinator_scope_3404.rs` with
only the `3_transform/css.rs` hunk reverted (the two parse commits left in place): **5 of
7 fail**, each for the predicted reason —

```
only_the_subject_side_of_an_unhandled_combinator_is_scoped
  left: ".a.svelte-70s02x || .b:where(.svelte-70s02x) { color: red }"
 right: ".a || .b.svelte-70s02x { color: red }"
the_walk_resumes_on_handled_combinators_to_the_right      (same, with `> .c` appended)
a_handled_combinator_to_the_left_is_still_out_of_reach    (same, `.a > .b || .c`)
the_unreachable_side_does_not_decide_whether_the_rule_is_used
  left: "/* (unused) .zz || .b { color: red }*/"          (pruned on a compound
 right: ".zz || .b.svelte-70s02x { color: red }"           official never tests)
the_reachable_side_alone_can_still_prune_the_rule
  left: ".a.svelte-70s02x .b:where(…) || .c:where(…) > .d:where(…) { color: red }"
 right: "/* (unused) .a .b || .c > .d { color: red }*/"
```

The two that pass are the deliberate controls: `handled_combinators_scope_both_sides`
(`' ' > + ~` are unaffected) and `a_trailing_global_does_not_become_the_subject`, which is
non-discriminating against `main` and discriminating against the intermediate attempt — it
is in the file to pin the `truncate` ordering, not to fail on `main`.

Every expectation in both files is the official compiler's verbatim answer at the pinned
Svelte revision. All three `compatibility/pattern-corpus` repros were checked byte-for-byte
against official on this tree (MATCH) and on `main` (DIVERGE).

## Not fixed

**#3406 items 2 and 3**, both error *positions*: `.a { ;; color: red ;; }` reports the
leading `;;` where official reports the trailing one (and rsvelte rejects a leading-only
`;;` that official accepts), and `.a { --x: { color: red }; }` points at the `;` where
official points at the `{`. 3 of the 18 declaration cells, listed above by id.

## Conflicts

Test-merged, both clean, no ordering constraint:

- **#3424** (`fix/3130-3133-3204-css-pseudo-class-args`) also rewrites `style.rs`, but in
  `SelectorParser::read_selector` / the An+B path — cherry-picking onto it auto-merges.
- **#3550 / #3310** also touch `3_transform/css.rs` and `style.rs` (the `</style>` block
  scan) — likewise auto-merges.

The third commit also backfills the two `compatibility/pattern-corpus/README.md` rows the
first two commits omitted, which makes that (heavily contended) file part of this PR's
diff.


baseballyama:

baseballyama:
## fmt parity: this PR's two combinator repros diverge from the oxfmt oracle

Measured locally with the formatter-parity gate's own comparator
(`scripts/compat-corpus/fmt-one.mjs`, whose exit code is the verdict), on a tree
that is this PR merged with nine other open PRs:

| file | verdict |
|---|---|
| `pattern-corpus/issues/3404-repeated-combinators.svelte` | **MISMATCH** (20 lines) |
| `pattern-corpus/issues/3404-unhandled-combinator-scope.svelte` | **MISMATCH** (6 lines) |
| `pattern-corpus/issues/3405-attribute-selector-spacing.svelte` | MATCH |

The run is discriminating in both directions: three `pattern-corpus` files taken
from `origin/main` (in neither `fmt-known-failures.json` nor
`fmt-oracle-excluded.json`) returned MATCH, and #3475's known-failing repro
returned MISMATCH.

What is observed — no cause claimed:

```
source :  \t.card >> .a {          (tab indent, spaces around >>)
oracle :  "  .card>>.a {"          reindented to 2 spaces, spaces removed
actual :  "  \t.card >> .a {"      source verbatim, plus a blank line after <style>
```

The `||` file diverges the same way (`.a||.b` oracle vs `.a || .b` actual). The
whole `<style>` block comes out essentially as written, tabs included.

### Why this blocks the merge

Formatter parity is a hard gate, and `compatibility/pattern-corpus` reaches it
through the shared manifest — that is how #3475 was caught. No `pattern-corpus`
entry has ever been listed in either fmt ratchet, so listing these is not an
available route.

### Attribution is NOT settled

This PR changes `1_parse/read/style.rs`, which is also where #3475's change
lives, so the measurement cannot separate "the formatter has always passed these
blocks through" from "this PR's parse change made it start". Deciding that needs
an `rsvelte-fmt` built from `main`, which was not built here.

This PR is therefore **deferred out of the current merge batch**; the other nine
proceed. Nothing here questions the compiler fix itself — only the two repro
files' interaction with the formatter gate.

baseballyama:
### Attribution is now settled — and it is not this PR

The caveat in the comment above ("this PR changes `1_parse/read/style.rs`, so the
measurement cannot separate…") no longer applies. `rsvelte_formatter` never calls
rsvelte's own CSS parser:

- `crates/rsvelte_formatter/src/lib.rs:143-149` sets `defer_script_parse: true`
- `1_parse/parser.rs:181-183`: `should_defer_template_parse = !script_ts && options.defer_script_parse`
- `1_parse/read/style.rs:427-429` then leaves `css_children` empty, so
  `parse_css_strict` never runs

Neither failing fixture contains a `<script>` tag at all (`script_tags=0`), so
`script_ts` is false and the deferral holds. **No change to
`1_parse/read/style.rs` can reach `rsvelte-fmt`'s output for these two files.**
The divergence is a pre-existing `rsvelte_formatter` defect that these repros are
simply the first input to reach. (Mechanism traced independently by another
session; the two signatures it predicts — verbatim body, one inserted blank line —
are both visible in the measured diff for `3404-repeated-combinators.svelte`.)

### The two failures are not the same defect

- `3404-repeated-combinators.svelte` — `oxc-css-parser` cannot parse `>>` / `>>>`,
  bails on the whole block (no per-rule recovery), and `style_css.rs:81-82` maps
  the error to the untouched body while `style.rs:221` prepends a second `\n`.
  Signatures: source tabs preserved, everything shifted one line down.
- `3404-unhandled-combinator-scope.svelte` — **no bail.** Both sides format the
  block (source tabs become two spaces, zero line shift); they disagree only on
  printing: `oxc_formatter_css` emits `.a || .b`, the oxfmt oracle emits `.a||.b`.
  The oracle does not use `oxc_formatter_css` for a `.svelte` `<style>` at all —
  `prettier-plugin-svelte` embeds it as `parser: 'css'` and oxfmt only substitutes
  JS parsers, so the CSS goes through Prettier's postcss.

So a fix to the fallback closes the first file and not the second.

**Do not delete the repros.** This PR stays out of the current merge batch only
because the hard formatter-parity gate would go red on `main` and no
`pattern-corpus` entry has ever been listed in either fmt ratchet. Once the
formatter side is fixed, both files should pass as written.

baseballyama:
I was assigned #3404 and #3405 independently and found this PR before writing anything, so rather than duplicate it I built this branch (`04514f53f`) and ran the two matrices I had already written against the pinned `submodules/svelte` oracle (5.56.9), client + server. Posting the residuals rather than opening a competing PR.

**#3404 — 24 cases: `origin/main` 16 diverge → this branch 2 diverge.** Both remainders are the same rule reached one axis over:

| case | official | this branch |
|---|---|---|
| `.card > > .a` | `.card.svelte-X > > .a:where(.svelte-X)` | `.card.svelte-X > .a:where(.svelte-X)` |
| `:is(.card >> .a)` | `:is(.card.svelte-X >> .a:where(.svelte-X))` | `:is(.card.svelte-X > .a:where(.svelte-X))` |

The first is a combinator run whose members are **separated by whitespace**. Upstream's loop
re-enters `read_combinator`, which calls `allow_whitespace()` before reading, so `> >` is the same
shape as `>>` — the earlier `>` is dropped from the AST and survives only because the sheet is
rewritten in place. A run-detector that requires the characters to be adjacent misses it.

The second is the **entry point** axis: the same `>>` inside a functional pseudo-class argument.
There are two near-identical relative-selector scanners in `1_parse/read/style.rs`
(`parse_relative_selectors_with_combinators` at ~999 and `parse_relative_selectors_from_text` at
~2619) and two printer paths in `3_transform/css.rs`, and a pseudo-class argument goes through the
second of each. Every other `>>` row on the list passes, which is what makes this one worth
naming — reaching the rule at the top level says nothing about the argument-list port.

Rows that this branch **fixed** and that are not in issue #3404's table, for the record:
`.card>>.a`, `.card >>.a`, `.card>> .a` (spacing variants), `.card >>>> .a`, `.a ~~ .b`,
`.a ++ .b`, `.a || .b || .c`, `.card { || .a }`, `:is(.a || .b)`, `.a||.b`. Controls that did not
move: `>`, `+`, `~`, `::deep`, `.card /* c */ > .a`, `svg|circle`, `[a|="b"]`.

**#3405 — 25 cases: `origin/main` 18 diverge → this branch 3 diverge.** All three remainders are
the empty-selector-list-entry check not reaching the pseudo-class argument list:

| case | official | this branch |
|---|---|---|
| `:is(.a,)` | `css_expected_identifier@4:7` | compiles as `:is(.a.svelte-X)` |
| `:is(,.a)` | `css_expected_identifier@4:4` | compiles as `:is(, .a.svelte-X)` |
| `:is(.a,,.b)` | `css_expected_identifier@4:7` | compiles as `:is(.a.svelte-X, , .b.svelte-X)` |

Upstream has one `read_selector_list(parser, inside_pseudo_class)` for both, so the trailing/leading/
doubled-comma rejection is the same code; here the top-level list and `parse_args_selector_list`
are separate. The last row emits `, ,` into the output, so it is the `[]`/`,,` half of #3405 one
level in — invalid CSS rather than merely over-acceptance.

Everything else on the list passes, including the seven rows the issue names, the four controls,
and eight axes it does not list (`[ *|data-k ]` with spaces, `[*|data-k="v"]`, `.a ,, .b`,
`.a,,, .b`, `.a { .b, }`, `.a { , .b }`, `:is([*|data-k])`, `@media (…) { .a, { … } }`).

Reproduction: the matrices are two throwaway scripts that compile each source with
`submodules/svelte/packages/svelte/src/compiler/index.js` and the staged NAPI binding and compare
`{css.code, warnings, error-code@line:col}`. Happy to hand them over if useful.

Issue #3197, opened 2026-08-21
Block-scoped name inside a `$:` statement creates a false `reactive_declaration_cycle`


baseballyama:
A name declared **inside** a `$:` reactive statement — a `catch` parameter, or a
`let`/`const` in a nested block — is attributed to the outer binding of the same
name, so a second reactive statement that assigns that outer name is reported as a
dependency cycle. Official compiles both.

```svelte
<script>
	export let a = 1;
	let d = 0;
	let e = 0;
	$: try { d = a; } catch (e) { d = 0; }
	$: e = d + 1;
</script>
<b>{d}{e}</b>
```

```
official : OK
rsvelte  : reactive_declaration_cycle @ 5:1
```

Same with a block-scoped `let`:

```svelte
	$: { let e = a; d = e; }
	$: e = d + 1;
```

A **function parameter** is scoped correctly, which is what makes this a scoping
bug rather than a missing feature:

```svelte
	$: d = ((e) => e)(a);   // both compile
```

Dropping the second reactive statement also compiles, so the false edge is
`catch (e)` / block `let e` counted as an assignment to the module-level `e`.

Found by a generated grid over `$:` statement shape × source layout × trailing
statement (2160 cells): all four layouts of the `try` row diverge, on `client`,
`server` and `client-dev` alike.

Issue #3115, opened 2026-08-19
A local declaration shadowing a rune name does not turn the call into a store subscription


baseballyama:
## Summary

When a local declaration shadows a rune's name, `$rune(...)` stops being a rune and becomes a store subscription to that declaration. rsvelte only implements this for `$state`; the other five diverge, each in its own way.

## Repro

```svelte
<script>
	let { a = 1 } = $props();
	const props = { x: 1 };
</script>
<p>{a} {props.x}</p>
```

```js
// official (server)
	var $$store_subs;
	let { a = 1 } = $.store_get($$store_subs ??= {}, '$props', props)();

// rsvelte
export default function T($$renderer, $$props) {
	let { a = 1 } = $$props;
```

## The family

Each row is `const <name> = …` in the instance script next to a `$<name>(…)` call.

| rune | server | client | rsvelte's behaviour |
|---|---|---|---|
| `$state` | match | match | already a store subscription |
| `$props` | diverge | diverge | stays the rune |
| `$bindable` | diverge | diverge | stays the rune (`v = 2` instead of the store call) |
| `$derived` | diverge | match | store call is emitted, but the **read** still constant-folds as if `d` were a derived — `<p>0 1</p>` where official escapes `d` |
| `$effect` | diverge | match | the whole call is erased |
| `$inspect` | diverge | diverge | the call becomes `;;` |

`$derived` is the interesting one: the declaration side is right and the read side is not, so the two halves of the same decision disagree with each other, not only with upstream.

## Why no gate saw it

Published code does not name a local variable `props` or `effect` next to the matching rune, so the collected corpus cannot contain it; and no generated family crosses "identifier that shadows a rune name" with "rune call site". The `opaque-keyword` family is the nearest neighbour and it varies where a keyword *appears*, not what the surrounding scope declares.

baseballyama:
## Measurement against the current tree — the table above is stale, and the location named below is not where the defect is

Oracle: the **pinned submodule**, Svelte **5.56.9** (`submodules/svelte` @ `20b341f10048`), i.e. `submodules/svelte/packages/svelte/src/compiler/index.js` — not `node_modules/svelte`, which resolves to 5.56.10 and is a different compiler.

[...]

Issue #1975, opened 2026-07-30
compiler: whitespace run around two adjacent removed HTML comments collapses to one space (official keeps two) when the parent element is nested


baseballyama:
## Summary

Whitespace runs around **removed adjacent HTML comments** collapse differently from official: with two adjacent comments official preserves `</header>  <button` (two spaces), rsvelte emits one. Only reproduces when the comments' parent element is itself nested one level deeper.

## Provenance

Found by registering skeletonlabs/skeleton in the compiler corpus (#1924 / PR #1964); 1 seed in `compatibility/known-failures.client.json` with a minimal repro.

PR #3575, opened 2026-08-22
fix(analyze): visit an `{#each}` key expression inside the each scope

### LINKED ISSUE #3285: A write inside an `{#each}` key expression does not mark the collection as reactive state
A write inside an `{#each}` **key expression** is not counted as a mutation of
the collection, so the collection is never promoted to reactive state.

```svelte
<script>
	let arr = [{ id: 1 }];
	void arr;
</script>
{#each arr as v, i (v = 1)}{i}{/each}
```

```js
// rsvelte                          // official
let arr = [{ id: 1 }];              let arr = $.mutable_source([{ id: 1 }]);
void arr;                           void $.get(arr);
$.each(node, 1, () => arr, …)       $.each(node, 1, () => $.get(arr), …)
```

`v` is the each item, so assigning to it writes through to `arr`. Official's
phase-2 walk reaches the key expression and marks `arr` mutated; rsvelte's does
not, so `arr` stays a plain `let` and every later read misses `$.get`.

This is the same shape as #3026 and #3048 — a write in a host the reactivity walk
never visits — with the key expression as the unvisited host. Found by crossing
the `{#each}` **pattern** with the **key-expression shape**, two axes that are
normally confounded because a key expression is nearly always a plain member
read: 612 cells, 382 compiled on both sides, and this is the only output
divergence among them, the other 194 being acceptance or error-field differences.

Worth checking the sibling hosts the same walk may not reach: an `{#await}`
expression, a `{#key}` expression, and a `{@const}` initializer.

baseballyama:
Closes #3285

Upstream visits an `{#each}` key expression with the each bindings **in scope** — `scope.js`'s `EachBlock` does `if (node.key) visit(node.key, { scope })` — so a write to the item there writes through to the collection and marks it mutated. rsvelte visited the key with those bindings out of scope, so the write reached nothing, the collection stayed a plain `let`, and every later read missed `$.get`.

Six lines in `ScopeBuilder::visit_each_block`.

## The half that was still live

The issue's own repro **passes on `origin/main` today**:

```svelte
{#each arr as v, i (v = 1)}{i}{/each}   <!-- already correct on main -->
{#each arr as v, i (v++)}{i}{/each}     <!-- still wrong on main -->
```

Measured on `origin/main` with `compile_one`, client target, against the **pinned submodule oracle, Svelte 5.56.9** (`submodules/svelte` @ `20b341f10048`) — not `node_modules/svelte`, which resolves to 5.56.10:

| key expression | `origin/main` | official | **this PR** |
|---|---|---|---|
| `(v = 1)` — `AssignmentExpression` | `$.mutable_source` ✔ | `$.mutable_source` | ✔ |
| `(v++)` — `UpdateExpression` | `let arr = [1]; void arr; … () => arr` ✘ | `$.mutable_source` | ✔ |
| `(v.id)` — read only | plain `let` ✔ | plain `let` | ✔ (control) |

So re-checking #3285 against the text of the issue would have reported it fixed. The test file carries both forms and says why.

The read-only row is the control that the fix is not "mark the collection mutated whenever there is a key".

## Negative controls

The recovered work touched two files. Each was reverted on its own and the `(v++)` row re-measured:

| tree | `(v++)` output |
|---|---|
| `origin/main` | `let arr = [1];` / `void arr;` / `() => arr` |
| `scope_builder.rs` hunk only | `let arr = $.mutable_source([1]);` / `void $.get(arr);` / `() => $.get(arr)` — **correct** |
| `2_analyze/visitors/each_block.rs` hunk only | identical to `origin/main` — **no effect** |
| both | identical to `scope_builder.rs` alone |

The `each_block.rs` hunk (moving the key visit before the `{:else}` fallback in the analyze visitor) is therefore **not in this PR**. It changed no output on any of the 5 sources measured, and dropping it keeps the PR clear of `visit_each_block`'s fallback region.

## Sibling hosts checked, no divergence found

The issue asks about the other hosts the same walk may not reach. A write in a `{#key}` expression and in a `{@const}` initializer were compiled with both compilers on both targets: **byte-identical before and after**, so neither is a second instance of this defect. (`{#await}` is not reachable in the same way — its expression has no item binding to write through.)

## Local verification

CI is not dispatching, so everything was run locally on `origin/main` plus this diff.

* `cargo fmt --all` — clean; `cargo clippy -p rsvelte_core --all-targets --all-features -- -D warnings` — clean.
* `cargo test -p rsvelte_core`: `--lib` **1732**, `each_key_expression_write_3285` **3**, `ssr` **2**, `validator` **3**, `compiler_errors` **2**, `parser_fixtures` **3**, `sourcemaps` **2**, **`runtime` 5 (228 s)** — all green.
* `cargo test -p rsvelte_projection -p rsvelte_lint` — 67 green suites, 0 failures.

## Provenance and conflicts

Recovered from an uncommitted worktree lost to a session kill (`salvage/3197-3115-3285-legacy`); the `#3197` and `#3115` work that shared that tree is not here (#3197 is PR #3560).

`2_analyze/scope_builder.rs` — the hunk sits in `visit_each_block` immediately before `self.visit_fragment(&block.body)`. **@metA**'s #3494 adds a `push_scope` around the `{:else}` fallback in the same function, below this point; the two are additive and whichever lands second rebases.


baseballyama:
Added the missing `issues/` table row for this repro (`1d44aab1f`, one line, nothing else moved).

`compatibility/pattern-corpus/README.md` states that a pattern file's provenance belongs in that table and nowhere else — an HTML comment inside the file is itself a whitespace-sensitive compiler input (#1975). This PR added `issues/3285-each-key-expression-update.svelte` without a row, so the file's provenance was undocumented. No gate enforces this, so CI was not going to catch it.

PR #3588, opened 2026-08-22
fix(client): refuse a JS reserved word as a generated variable name

### LINKED ISSUE #3582: Client: an element whose tag name is a JS reserved word emits `var var = …` — 42 of 46 names produce output no parser accepts, including `<var>` and SVG `<switch>`

## Repro — a standard HTML element, no runes, no expression

```svelte
<var>x</var>
```

`generate: 'client'`:

```js
// official
	var var_1 = root();
	$.append($$anchor, var_1);

// rsvelte
	var var = root();
	$.append($$anchor, var);
```

`var var = root();` is a syntax error. `compile()` returns successfully, so the
failure surfaces at bundle time (or as a blank page) rather than as a diagnostic.

## Scope

Every element whose tag name is a JS reserved word. Over 46 candidate names,
**42 produce output acorn rejects**; the four that pass — `async`, `of`, `get`,
`set` — are exactly the four that are not reserved words:

```
<switch> <var> <let> <const> <class> <function> <return> <if> <else> <for>
<while> <do> <new> <delete> <typeof> <void> <in> <instanceof> <this> <super>
<null> <true> <false> <try> <catch> <finally> <throw> <case> <default>
<break> <continue> <with> <yield> <await> <static> <enum> <export> <import>
<extends> <debugger> <arguments> <eval>
```

Two of those are **standard elements that appear in real documents**: HTML's
`<var>` and SVG's `<switch>`. The rest are unknown elements, which Svelte
compiles happily.

|---|---|---|---|
| `<var>{v}</var>` | **UNPARSEABLE** | **UNPARSEABLE** | match |
| `<var>x</var>` — no expression at all | **UNPARSEABLE** | **UNPARSEABLE** | match |
| `<var><b>{v}</b></var>` | **UNPARSEABLE** | **UNPARSEABLE** | match |
| `<var bind:this={el}>x</var>` | **UNPARSEABLE** | **UNPARSEABLE** | match |
| two `<var>` siblings | **UNPARSEABLE** | **UNPARSEABLE** | match |
| `<svg><switch><text>{v}</text></switch></svg>` | **UNPARSEABLE** | **UNPARSEABLE** | match |
| `<svelte:element this="var">{v}</svelte:element>` | match | match | match |

Server is clean throughout, and `<svelte:element>` is clean because its variable
is named from a different path.

## Cause

Upstream's `Scope.unique` refuses a reserved word as a generated name —
`phases/scope.js:728-734`:

```js
while (
	this.references.has(name) ||
	this.declarations.has(name) ||
	this.root.conflicts.has(name) ||
	is_reserved(name)
) {
	name = `${preferred_name}_${n++}`;
}
```

rsvelte's `Memoizer::generate_id`
(`3_transform/client/types.rs:2827`) has the three membership tests and not the
fourth. Its fast path is:

```rust
let mut conflicts = self.conflicts.borrow_mut();
if !conflicts.contains(sanitized) {
    conflicts.insert(sanitized.to_string());
    return sanitized.to_string();
}
```

so the very first `<var>` in a component takes the fast path and returns `var`
verbatim. The suffix loop below it has the same omission, which is why a second
`<var>` does not rescue the first.

`is_reserved` already exists in this tree — `1_parse/utils/mod.rs:77` — so the
fix is to consult it in both places, and `sanitize_identifier`/`generate_id_slow`
should be checked for the same gap.

## Why no gate saw it

The parse oracle (#2591) runs on the corpus and on the shape matrix, and the
population of both is code people wrote or axes someone enumerated: `<var>` is
rare enough to be absent from the 12,523 collected components, and no generated
family varies the **tag name** against the reserved-word list. This is the class
`AGENTS.md` records as "a scanner assuming input it did not get", except the
assumption here is in the name allocator: that a tag name is a legal identifier.

(`elements/nested-namespace-switching.svelte`, via SVG `<switch>`), held out of
`compatibility/pattern-corpus/adversarial/` until this is fixed. A
`tag-name × reserved-word` family in `scripts/compat-corpus/matrix/axes.mjs`
would be the durable form of the grid above.

baseballyama:
Closes #3582.

## The defect

```svelte
<var>x</var>
```

A standard HTML element, no runes, no expression. `generate: 'client'`:

```js
// official
	var var_1 = root();

// rsvelte (before)
	var var = root();
```

`compile()` returned successfully, so this reached the bundler as a syntax
error rather than as a diagnostic.

## Scope, measured

Over 46 candidate tag names, **42 produced output acorn rejects**. The four that
did not — `async`, `of`, `get`, `set` — are exactly the four that are not
reserved words. Two of the 42 are standard elements: HTML's `<var>` and SVG's
`<switch>`.

|---|---|---|---|
| `<var>{v}</var>` | UNPARSEABLE → match | UNPARSEABLE → match | match |
| `<var>x</var>` — no expression | UNPARSEABLE → match | UNPARSEABLE → match | match |
| `<var bind:this={el}>x</var>` | UNPARSEABLE → match | UNPARSEABLE → match | match |
| two `<var>` siblings | UNPARSEABLE → match | UNPARSEABLE → match | match |
| `<svg><switch>…</switch></svg>` | UNPARSEABLE → match | UNPARSEABLE → match | match |
| `<svelte:element this="var">` | match | match | match |

Server was clean throughout — it does not name a variable after the tag — and
`<svelte:element>` names its variable from a different path, which is what
localised the site.

## The fix

Upstream's `Scope.unique` rejects four things when picking a name
(`phases/scope.js:728-734`): a reference, a declaration, a root-conflicts entry,
and `is_reserved(name)`. `Memoizer::generate_id` had the first three and not the
fourth, and its **fast path returns the base name verbatim the first time a base
is seen** — so one `<var>` in a component was enough, and a second one did not
rescue the first. `generate_id_slow` had the same gap. Both now consult
`is_reserved`, which already existed in this tree at
`1_parse/utils/mod.rs:77`.

## Controls

- The repro `compatibility/pattern-corpus/issues/3582-reserved-word-tag-name.svelte`
  is **unparseable on the previous binary and byte-equal on this one** — it fails
  for the predicted reason, not merely fails. It carries `<var>` twice so the
  suffix path runs after the fast path, a `bind:this` host, SVG `<switch>`, and
  `<async>`/`<get>` as negative controls that must keep their bare names.
- All **720** pattern-corpus files compiled with both compilers on client,
  client-dev and server before and after the change: **identical divergence
  sets**, not merely identical counts (35 raw, every one absorbed by the corpus
  normalization).
- The 46-name grid goes 42 diverging → **0**.

## Why no gate saw it

The parse oracle runs over the collected corpus and the shape matrix, and both
populations are code someone wrote or axes someone enumerated: `<var>` is absent
from the 12,523 collected components, and no generated family varies the **tag
name** against the reserved-word list. A `tag-name × reserved-word` family in
`scripts/compat-corpus/matrix/axes.mjs` is the durable form of the grid above and
is not in this PR.

Oracle: `submodules/svelte` @ **5.56.9**, asserted from the `compileModule`
output header.


baseballyama:
Local control run on this branch, since CI has not reached a verdict yet:

```
RUST_MIN_STACK=33554432 cargo test -p rsvelte_core --no-fail-fast
→ 330 test binaries, 3250 passed, 0 failed, 8 ignored (exit 0)
```

Together with the three checks already in the PR description — the 46-name
reserved-word grid going 42 → 0, the 7-host grid matching on every cell, and all
pattern-corpus files × 3 targets producing an **identical divergence set**
before and after the fix — that is the full local evidence. The repro
(`compatibility/pattern-corpus/issues/3582-reserved-word-tag-name.svelte`) is
`MISMATCH+UNPARSEABLE` on the pre-fix binary and byte-equal after.

baseballyama:
Independent verification, run against the pinned official compiler from
`submodules/svelte` (asserted `5.56.9` from the oracle's own output, not from a
lockfile) with raw-byte output comparison — stricter than the corpus gate, which
normalizes with oxfmt + blank-line stripping before comparing.

## The domain is closed, so this is an enumeration and not a sample

`is_reserved` (`phases/1_parse/utils/mod.rs:77`) is a finite `matches!` table of
source literals with **exactly 48 entries**, so the domain is closed and can be
enumerated rather than sampled. I generated one component per tag name over a
52-name grid and checked the grid against that table **in both directions**:

```
comm -23 reserved.txt grid.txt   ->  (empty)              the grid covers all 48
comm -13 reserved.txt grid.txt   ->  async get of set     and adds exactly these 4
```

Both directions are load-bearing. The first alone would show only that the grid
covers the table, not that it contains nothing else; it is the second that makes
those four names the **controls** rather than noise — they are precisely the grid
members the table does not contain. So "48 / 48" below is a count over a closed
domain, not an estimate from a sample.

## Before (`fa78572d2`)

Measured on the tree I was verifying from, **not** on `main`: `c75956b20` plus
#3575 merged locally. Naming it precisely matters because the after half runs on
this PR's head, whose base is 18 commits older, so two things move at once unless
the difference is shown not to reach the output. It does not:

- `is_reserved`'s table is **48 entries at both revisions**, identical — the
  denominator is stable.
- `client/types.rs`, the only source file this PR edits, is **byte-identical** at
  this PR's base and at the before-tree.
- This PR's head contains **none** of #3374, #3414, #3575, all merged after its
  base. That is what the base difference amounts to, and none of the three can
  reach these inputs: #3374 strips a leading BOM (**0 of 52** grid files have
  one), #3575 fixes `{#each}` key scoping (**0 of 52** contain `{#each}`), and
  #3414 is server-side (**all 52** rows are compiled for the client).

Note what that argument is *not*. "Zero files under `3_transform/client/` differ"
would be too weak to carry it: the same run shows four **client** rows flipping
because of #3374, which edits `compiler/mod.rs` and `toolchain.rs` — upstream of
client codegen, and able to change client output without touching a file under
it. Identical codegen sources do not imply identical codegen output. The argument
above is per-input instead, which is why it holds.

The direction is also self-checking: the defect is present on the *newer* tree and
absent on the *older-base-plus-fix* tree, so attributing the change to the base
would require those commits to have introduced the defect, which contradicts the
PR existing.

| | reserved (48) | not reserved (4) |
|---|---|---|
| output differs from official | **48 / 48** | 0 / 4 |
| output rejected by acorn | **48 / 48** | 0 / 4 |

Official is accepted by acorn on **148 / 148** outputs in the same run; rsvelte is
rejected on **49** — the 48 above plus this PR's own
`3582-reserved-word-tag-name.svelte` (`Unexpected keyword 'var' (10:5)`).

## After (`c573dfa8401e83c34f0732a154a7fcdb3b583871`, this PR's head)

| | reserved (48) | not reserved (4) |
|---|---|---|
| output differs from official | **0 / 48** | 0 / 4 |
| output rejected by acorn | **0 / 48** | 0 / 4 |

Every reserved-word row is now **byte-identical** to official, not merely
parseable — `<var>` emits `var var_1 = root();` exactly as upstream does. The
PR's own `3582-reserved-word-tag-name.svelte` parses. Official stays at 0 / 148
rejected, and the grid-shape assertion (48 + 4) did not fire.

**The four non-reserved controls read 0 / 4 before and after, so on their own they
cannot detect a comparator that collapsed to all-PASS** — an unchanged number is
unchanged under both hypotheses. What rules that out is that the same run still
reports 27 rows as differing from official, including three server rows
(`f10`, `f11`, `f14`) that must differ and do. The comparator was live when it
reported 0 / 48.

`runtime` (2335 fixtures) and svelte2tsx TSX parity were **not** measured here:
both are covered by the CI checks queued on this PR, and re-running them locally
would have been the single largest disk expenditure in the verification.

## One correction to the PR body's arithmetic

The changeset and the README row say "42 of 46 reserved words", and the README row
identifies the four that behaved differently as `async`, `of`, `get`, `set` — which
are **not** reserved. So the PR's 46 is 42 reserved plus 4 non-reserved, and its
reserved denominator is **42, not 48**. Six reserved words are outside that count.
All six are broken on main, so the defect is wider than the PR claims, not narrower.
I cannot reconstruct which six from the PR body, which does not enumerate its list.

The README row's parse-ability claim ("produced output no JS parser accepts") is
**confirmed**, not added, by the acorn run above — the author had already made it.
What is new here is the denominator and an oracle run independently of the author's.

## What this did not check

- The `compatibility/pattern-corpus/*.svelte` file added by this PR is read by **no
  Rust test suite** (`pattern-corpus` appears in no `crates/**/tests/**`), only by
  the node corpus harness, which was not run. The 52-component grid above is the
  substitute, and it is a wider population than the file's 12 tags.
- The corpus gate, the matrix gate, and the fmt-parity gate were not run.
- `sanitize_identifier` was not audited for whether it can *produce* a reserved word;
  the same guard is present on `generate_id_slow`, so such a case would be caught.

Issue #3428, opened 2026-08-22
server: a comment trailing a block-bodied $: with a surviving successor is replayed at the component tail, not in the template expression


baseballyama:
A comment trailing a **block-bodied** `$:` on the same line, where a surviving statement follows it in the script, is replayed at the end of the component function instead of landing in the first template expression.

```svelte
<script>
	export let a = 1;
	let b;
	$: { b = a * 2; } // mid
	let z = 1;
</script>
<div>{b}</div>
```

`generate: 'server'`:

```js
// official
	$$renderer.push(`<div>${$.escape(
		// mid
		b
	)}</div>`);

	$.bind_props($$props, { a });
}

// rsvelte
	$$renderer.push(`<div>${$.escape(b)}</div>`);
	$.bind_props($$props, { a });
	// mid
}
```

## Not a #3326 regression

`mid-$:-block-sameline / none / dyn-expr / server` and `mid-$:-block-sameline / same-line / dyn-expr / server` are present in the diverging set **on the merge base and on #3326 alike** — checked as a set difference of the two runs of the same 1078-comparison grid, where the totals move 712 → 487 with **0 new divergences**. The `mid-$:-block-sameline` script shape is also the single row that survived the narrower 28-comparison grid both before and after. This one predates #3326 and #3326 does not touch it.

## Why it is its own shape

#3326 deferred a `$:` statement's same-line trailer only when the `$:` is the **last** statement of the script (`defer_reactive_tail = is_reactive && is_last_stmt` in `transform_script_legacy`). A `$:` with a statement after it keeps the older narrow path, `defer_block_reactive_trailing`, which fires on `reactive_body_has_direct_block` plus "no leading comments in the region" and then anchors through `place_on_position` — that path pins the comment to the reactive body and `mark_deferred_tail_comment_landed` suppresses the later claim, so the template expression never gets offered it.

Upstream's own behaviour here is worth stating because it is not obvious: the `$:` body block keeps its source `loc`, so esrap's `reset_comment_index` **re-syncs forward into it**, and the trailing comment survives past the reordered label to be flushed by the next located node — the template expression — even though `let z = 1` is printed before the block. Any fix has to reproduce that ordering, not just move the anchor.

## What no gate sees

[...]

Issue #3426, opened 2026-08-22
server: a script-tail comment is not claimed by an attribute value, an {#if} test, an {#each} collection or a {@html} argument


baseballyama:
A comment left pending at the end of an instance script is flushed by esrap at the next node upstream keeps a `loc` on. #3326 taught the server to hand that run to the first **text** `ExpressionTag`, so `<p>{b}</p>` and a static template are now right. Four other template positions are still not offered the claim, so the run falls through to the component body's end instead:

* an attribute value — `<div title={b}>`
* an `{#if}` test
* an `{#each}` collection expression
* a `{@html}` argument

```svelte
<script>
	export let a = 1;
	let b = a;
	// tail
</script>
<div title={b}>x</div>
```

`generate: 'server'`:

```js
// official
	$$renderer.push(`<div${$.attr(
		'title',
		// tail
		b
	)}>x</div>`);

	$.bind_props($$props, { a });
}

// rsvelte
	$$renderer.push(`<div${$.attr('title', b)}>x</div>`);
	$.bind_props($$props, { a });
	// tail
}
```

`{#if b}`, `{#each [b] as v}` and `{@html b}` are the same shape — official writes the comment in front of the located expression (`if (// tail⏎b)`, `$.ensure_array_like(⏎// tail⏎[b])`, `$.html(⏎// tail⏎b)`), rsvelte replays it at the tail.

## Not a #3326 regression

**Measured on both sides.** A generated grid of 11 script shapes × 7 tail-comment shapes × 7 template shapes × 2 targets (1078 comparisons against the pinned official compiler), run against the merge base and against #3326:

| column | merge base | with #3326 |
|---|---|---|
| `attr / server` | 61 | 56 |
| `if-block / server` | 56 | 56 |
| `each-block / server` | 61 | 61 |
| `html-tag / server` | 61 | 56 |
| `component / server` | 61 | 61 |

Every row here diverges on the merge base too; the set difference of the two diverging-row sets is **225 fixed, 0 new**. #3326 changed *how* these rows are wrong (dropped → replayed at the tail) without changing the verdict on any of them. Anyone starting from "is this a #3326 regression?" will be starting from the wrong place.

## Where it goes

[...]

PR #3591, opened 2026-08-22
fix(server): let every printed template expression claim the script's trailing comment run

baseballyama:
> **Stacked on #3326.** Base is `fix/3080-3098-3251-comment-placement`, not `main`.
> **Do not merge this before #3326.** This repository has merge commits disabled, so #3326 lands squashed; once it does, **rebase this branch onto `origin/main`** and the diff here becomes only its own 13 source files + tests. Merging it as-is, or retargeting without the rebase, makes #3326's changes appear twice and conflict.
> `Closes` links for #3426 / #3428 are deliberately **not** in this body yet — they go in with that retarget.

Resolves **#3426** (a script-tail comment is not claimed by an attribute value, an `{#if}` test, an `{#each}` collection or a `{@html}` argument) and **#3428** (a comment trailing a block-bodied `$:` with a surviving successor is replayed at the component tail).

## What the issues got wrong about their own size

Both were reported off a **7**-shape template axis and name five positions between them. Re-run against a **21**-shape axis and the server population is not five positions — it is **16 template columns, each failing all 36 of its cells**:

| | columns |
|---|---|
| named in the issues | attribute value, `{#if}` test, `{#each}` collection, `{@html}` argument, component prop |
| also failing, unnamed | quoted attribute value, concatenated attribute value, attribute spread, `class:`, `style:`, `{#await}` expression, `{@render}` callee, `{@const}` initializer, component spread, `<svelte:element>` `this`, `<slot>` prop |

Plus the already-handled text `{expr}` column, still failing 12 cells — its two block-bodied `$:` shapes, which is #3428.

**"Template position" is not a closed set to enumerate.** It is the 41 expression-visit sites under `3_transform/server/ast/visitors/`, 27 of which now offer the claim. `{@render}` is why: its callee is built by `reparse_slice`, not by `visit_expr`, so it is the one column a `visit_expr`-shaped fix does not reach — it needs the separate `claim_on_visited` entry point, and it is the single `claim_on_visited` call site in the tree. (A run of the grid against an intermediate tree that had the `visit_expr` half only is recorded in the predecessor's notes as leaving all 36 `render-tag / server` cells red; I have **not** re-run that arm myself, so read the structural half — the two entry points — as the claim and the 36 as the anecdote.) The fix is therefore at the visit sites, not a list of node kinds.

## Measurement

Grid: 10 script shapes × 7 tail-comment shapes × 21 templates = **1,470 cases**, compiled for `client` and `server` = **2,940 comparisons**. Oracle is the pinned submodule (`submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'`), asserted at run time — not `node_modules/svelte`, which resolves to 5.56.10 in some trees.

| tree | diverging | server | client |
|---|---|---|---|
| merge base (`65fc545c`) | 921 | 632 | 289 |
| + #3326 | 709 | 588 | 121 |
| **+ this PR** | **121** | **0** | 121 |

**588 server cells fixed, 0 new** (set difference of the two diverging sets, not a subtraction of totals). All 1,470 client outputs are byte-identical between the #3326 tree and this one — this PR does not touch the client.

Three positive controls, so none of these numbers is a property the arms share:

1. reverting only this commit's 13 source files onto the same tree moves the server count 0 → 588;
2. the 18 new tests go 18/18 failing → 18/18 passing across that same revert;
3. the whole grid was re-run in a second, clean worktree and its 2,940 outputs are byte-identical to the first run's.

## The 121 client cells that remain

Out of scope here (this PR is server-only). **120 of the 121 are comment attribution; the remaining 1 is a different mechanism entirely** — do not read the total as one cause. Measured and posted to #3415, which owns the client comment cluster; the odd one out has its own issue.

All 120 sit in the two script shapes where the comment is the last thing in the script with no successor statement, and the comment's *spelling* (line, block, same-line, own-line, two-line, JSDoc) separates nothing — each affected column fails all 12 of its cells.

| cells | official | rsvelte | columns |
|---|---|---|---|
| 60 | inside the built thunk's parameter list — `$.html(node, (/* tail */⏎) => b)` | **dropped** | `{#each}`, `{#key}`, `{#await}`, `{@const}`, `<svelte:element>` |
| 12 | same | one statement too late (fragment tail) | `{@html}` |
| 24 | **leading** the first template statement | **dropped** | component prop, `<svelte:head>` |
| 24 | same | **trailing** that same statement | component spread, `{@render}` |

So the 84 "dropped" cells and the 36 "misplaced" cells are two different repairs, not one.

The 121st carries no comment at all: `<svelte:head><title>{b}</title>` with `let b = a + 1` prints `$.document.title = b` upstream and `b ?? ''` here, because upstream's `scope.evaluate` knows a `+` cannot yield nullish. Filed separately as #3599 — it is a `fold-value-type` defect, not a comment one.

## Tests

`crates/rsvelte_core/tests/template_position_tail_comment_3426.rs` — 18 expectations, every one of them the **pinned oracle's own output** for the grid cell, machine-generated rather than transcribed.

## gate-coverage

A new row records the axis every comment family was holding fixed: `comment-slot` and `removed-statement-comment` both **reach** this decision point on every run and neither can separate "the run is claimed" from "the run is claimed by a text `{expr}` and by nothing else" — the 5p shape one axis out. Its letter is **provisional** (`5q` is already used twice in that section); the section gets renumbered in one pass once the merge order of the open comment PRs is settled. Cite the row by its title.

## CI

Account-wide Actions dispatch is stalled, so this carries local evidence: the 18 new tests pass, the 2,940-comparison grid is at 0 server divergences on this tree, `cargo fmt --all -- --check` is clean, and `cargo clippy -p rsvelte_core --all-targets -- -D warnings` exits 0.


baseballyama:
## The matrix ratchet is stale on this tree: 142 entries already pass, 0 new failures

Recorded here so the next person does not have to re-derive it — the gate's own report truncates the list at 20 with `… and 122 more`.

Full run on this branch, all 17 families, all 4 targets, no `--no-fmt`, no `--families` subset:

```
[matrix] cases: 6365  targets: client, server, server-dev, client-dev  comparisons: 25044
  match            22414
  error-parity      2408
  js-mismatch          60
  comment-mismatch    162
  warning-mismatch     24
  output-unparseable    0
❌ 142 baseline entries already PASS — the ratchet is stale.
```

**No new failures**: every current failure is already listed. The two-sided ratchet therefore fails on the stale side only.

### Cross-check

The 142 was obtained twice, independently. The gate prints it; separately, re-implementing its final comparison (`stripBlankLines(expected) === stripBlankLines(actual)`) over the kept `compatibility/matrix-artifacts` tree and applying it to all 388 baseline entries gives the same number, and the partition closes:

```
baseline = 388  =  stale 142  +  still-failing 222  +  unmeasured 24
```

The 24 "unmeasured" are not a hole in the measurement — they are `[warning-missing:<code>]` keys, which is a **second key format** living in the same ratchet file alongside `[js-mismatch]` / `[comment-mismatch]`. The gate handles both; anything else that parses this file has to expect both, or it will silently score 24 entries as absent.

### Breakdown

```
comment-slot                                             88
   8  await-block                    L03   (server)      8  (server-dev)
   8  legacy-reactive                L08   (server)      8  (server-dev)
   8  module-script                  L22   (server)      8  (server-dev)
   8  snippet-render                 L03   (server)      8  (server-dev)
   8  class-private-state            L12   (server)
   8  class-static-block             L13   (server)
   8  const-fold-line-continuation   L06   (server)

removed-statement-comment                                54   (server only)
  18  effect          — all __instance-top__succ-none
  18  effect-pre      — all __instance-top__succ-none
  18  effect-root     — all __instance-top__succ-none
```

Two things fall out of the shape rather than out of the count:

* every stale `comment-slot` entry sits at **one line per seed** — the injection slot at the end of that seed's instance script;
* every stale `removed-statement-comment` entry is `__instance-top__succ-none` — the wrapped statement is at the top of the instance script with **no successor**.

Both are the same shape this PR and #3326 are about: a comment left pending at the end of an instance script with nothing after it to flush it.

### Which PR owns which entries — **not yet measured**

Merge order is #3326 → this PR, and the ratchet is two-sided, so an entry that #3326 fixes **on its own** turns `main` red the moment #3326 lands, before this PR is in. Putting all 142 here would be wrong in that case.

Reading each seed's markup and mapping it onto this PR's own grid columns gives bounds, **not values**:

| seed | markup | first printed expression | that column, #3326 alone |
|---|---|---|---|
| `await-block` | `{#await p}…{:then v}<span>{v}</span>` | the `{#await}` expression | 36/36 diverging → 0 with this PR |
| `snippet-render` | `<ul>{#each items as item}{@render row(item)}{/each}</ul>` | the `{#each}` collection | 36/36 diverging → 0 with this PR |
| `legacy-reactive` | `<p>{b}</p>` | text `{expr}` | 12/36 (the block-bodied `$:` shapes) |
| `module-script` | `<button onclick={…}>{local}</button>` | text `{expr}` | 12/36 |
| `const-fold-line-continuation` | `<p>{cont}{n}</p>` | text `{expr}` | 12/36 |
| `class-private-state`, `class-static-block` | `<button onclick={…}>x</button>` | event-handler attribute only | no column in that grid |
| `removed-statement-comment/effect*` | `<p>{a}</p>` | text `{expr}` | 12/36 |

So: **at least 32** (`await-block` 16 + `snippet-render` 16) cannot be fixed without this PR, and **at most 110** could be fixed by #3326 alone. ⚠ Those two numbers rest on the assumption that a matrix seed's first printed expression and the corresponding grid column go through the same mechanism. **They are an inference from structure, not a measurement, and must not be used to assign work to anyone.**

`--update-baseline` is deliberately **not** run here until the split is measured on a #3326-only tree. Measuring it needs a detached checkout rather than `git checkout <sha> -- src/`: `binding.mjs` attests the staged binding by comparing the commit the binary embedded at build time against `git rev-parse HEAD`, and a path-limited checkout moves neither — a binary built from #3326's sources in a tree whose HEAD is this commit would be certified as this commit.

Issue #3602, opened 2026-08-22
fix(parse): a trailing // comment before a template expression’s closing } is rejected in six hosts


baseballyama:
Oracle: `submodules/svelte` @ **5.56.9** (asserted from the `compileModule` output header).

## Repro

```svelte
<script>
	let flag = $state(true);
</script>

{#if flag // why
}
	<b>a</b>
{/if}
```

```
off | compiles (client and server)
rsv | ERROR js_parse_error
```

A `//` comment as the **last thing before the closing `}`** of a template
expression makes rsvelte reject the component. The comment runs to the newline,
the `}` is on the next line, and official parses it as an ordinary trailing
comment.

## Grid: which hosts reject

| expression | official | rsvelte |
|---|---|---|
| `{#if flag // c⏎}` | ok | **`js_parse_error`** |
| `{#key flag // c⏎}` | ok | **`js_parse_error`** |
| `{flag // c⏎}` | ok | **`js_parse_error`** |
| `data-a={flag // c⏎}` | ok | **`js_parse_error`** |
| `{@html "<i>x</i>" // c⏎}` | ok | **`js_parse_error`** |
| `{@render body(1) // c⏎}` | ok | **`render_tag_invalid_expression`** |
| `{#if // c⏎flag}` (leading) | ok | ok |
| `{#await pending // c⏎then value}` | ok | ok |
| `{@const x = 1 // c⏎}` | ok | ok |
| `{#each rows as row // c⏎}` | `expected_token` | same |

Leading `//` comments are fine and a *block* comment in the same trailing
position is fine, so the axis is specifically "the expression's last token is a
line comment": the scan that finds the expression's end takes the `}` on the
following line but hands the parser a slice that still ends inside the comment —
or stops at the newline and hands it an unterminated one. `{@const}` is the one
trailing-position row that works, and `{@render}` fails with a different code,
so this is per-host rather than one shared helper.

Real code reaches this: a multi-line `{#if}` header or a multi-line expression
tag with a `// eslint-disable-line`-style trailing note is an ordinary shape.

Found by a generated adversarial corpus candidate
(`control-flow/block-header-comments-and-spacing.svelte`), held out of
`compatibility/pattern-corpus/adversarial/` until this is fixed. Related to
#3592, which is the same "the scan does not model this opaque region" class one
phase earlier.

Issue #3092, opened 2026-08-19
rsvelte-fmt corrupts `<svelte:element this={a > b}>`


baseballyama:
`rsvelte-fmt` corrupts `<svelte:element>` / `<svelte:component>` when the `this={…}` expression contains a `>`.

```svelte
<script>
	let n = 0;
</script>

<svelte:element this={n > 0 ? 'p' : 'span'}>{n}</svelte:element>
```

```
rsvelte-fmt: <svelte:element this={n > 0 ? "p" : "span"}> 0 ? 'p' : 'span'}>{n}</svelte:element>
oxfmt:       <svelte:element this={n > 0 ? "p" : "span"}>{n}</svelte:element>
```

`find_open_tag_end` starts its scan after the last entry of `attributes`, and `this={…}` is not in `attributes` — it is a separate field on the node. With no other attribute present the scan starts right after `<svelte:element` and stops at the `>` inside `n > 0`, so everything from there to the real `>` is re-emitted as text.

The same input with any other attribute carrying the `>` (`<svelte:element this="div" data-x={n > 1}>`) is fine, and so is a regular element (`<p title={n > 0 ? 'a' : 'b'}>`), which is what isolates the `this=` slot.

Issue #2988, opened 2026-08-15
SSR/CSR .svelte.ts: a regex literal containing rune-call text is rewritten (`/$derived(x)/` becomes `/$.derived(() => x)/`)


baseballyama:
## Summary

In the `.svelte.js` / `.svelte.ts` module path, a **regex literal** whose body happens to contain
the text of a rune call is rewritten as if it were code. `/$derived(x)/` becomes
`/$.derived(() => x)/` and `/$state(x)/` becomes `/$.state($.proxy(x))/` on the client and `/x/`
on the server — three different regular expressions, all silently wrong at runtime.

Found by the `opaque-keyword` shape-matrix family added in #2986's fix PR; it is the same
raw-byte-scan class as #2986, in the opposite direction — that one failed to find code, this one
finds code where there is none.

## Repro (`generate: 'client'` and `'server'`)

```js
const pattern = /$derived(x)/;
export function f() {
	return pattern;
}
```

| target | official | rsvelte |
|---|---|---|
| client | `const pattern = /$derived(x)/;` | `const pattern = /$.derived(() => x)/;` |
| server | `const pattern = /$derived(x)/;` | `const pattern = /$.derived(() => x)/;` |

and with `$state`:

| target | official | rsvelte |
|---|---|---|
| client | `const pattern = /$state(x)/;` | `const pattern = /$.state($.proxy(x))/;` |
| server | `const pattern = /$state(x)/;` | `const pattern = /x/;` |

`/$derived(x)/` is a perfectly ordinary regex — `$` anchors, `derived` is literal, `(x)` is a
capture group — so this is not a contrived spelling. Both outputs parse, so only output equality
can observe it.

Control: `/class /` in the same position is unaffected, and the same regex inside a **component's
instance script** compiles correctly on both targets. This is the `compileModule` pipeline.

## Suggested fix

The module rune-call rewriters must skip regex literals, which is what
`shared::js_scan::skip_opaque` is for — it already distinguishes a regex from a division through
`slash_starts_regex_at`. `code_bracket_depth` and `find_matching_bracket` consume it; the
rewriters that decide *where a rune call starts* do not.

PR #3631, opened 2026-08-23
fix(fmt): keep a {@const} ending in a line comment parseable

### LINKED ISSUE #3613: rsvelte-fmt: a tag whose expression ends in a `//` comment gets its closing `}` printed inside the comment
`rsvelte-fmt` collapses a template expression onto one line without noticing
that its last token is a line comment, so the tag's closing `}` lands inside
the comment and the document no longer parses.

```svelte
<script>
	let flag = $state(true);
	let rows = $state([1]);
</script>

<b>{flag // c
}</b>
```

`rsvelte-fmt --stdin --stdin-filepath x.svelte`:

```svelte
<b>{flag // c}</b>
```

The official compiler rejects that with `expected_token` — the expression tag
is never closed.

## Grid

Input is accepted by both compilers in every row. The verdict is whether the
formatter's OUTPUT still compiles.

| host | rsvelte-fmt output | oxfmt output |
|---|---|---|
| `<b>{flag // c⏎}</b>` | **`expected_token`** | ok (comment deleted) |
| `<div data-a={flag // c⏎}>a</div>` | **`expected_token`** | ok (comment deleted) |
| `{@const x = r // c⏎}` | **`expected_token`** | **`expected_token`** |
| `{@const { a } = { a: r } // c⏎}` | **`expected_token`** | **`expected_token`** |
| `{#if flag // c⏎}` | ok | ok (comment deleted) |
| `{#key flag // c⏎}` | ok | ok (comment deleted) |
| `{@html "x" // c⏎}` | ok | ok |
| `{#await p // c⏎then v}` | ok | ok |
| `{#snippet body(n // c⏎)}` | ok | ok |

So it is the hosts whose closing delimiter is the tag's own `}` — an expression
tag, an attribute value, and `{@const}` — while a block header (`{#if}`,
`{#key}`) already keeps the break. `{@const}` diverges on both sides; the oxfmt
half is recorded in `upstream_issues/oxfmt-const-tag-ending-in-a-line-comment.md`
(it leaks a synthetic `)` into the document).

Note that oxfmt "ok" in the table means the output parses, not that it is
faithful: for five of these rows it silently **deletes** the comment.

## Not the same defect as the parse wrap

The wrap half of this — `format_core.rs`'s `(<slice>);`, whose `);` the comment
swallowed and which aborted the whole file with `script parse failed: Expected
`)` but found `EOF`` — is fixed alongside #3602. This issue is the remaining
half: the markup printer decides where the tag's `}` goes, and it has no notion
that the printed expression ends in a line comment.

The two candidate answers are to keep the source's newline before the `}`, or
to reflow the comment ahead of the expression the way esrap does for
`$.tag(...)` (#2495). Which one matches the official toolchain is not settled
here, because prettier-plugin-svelte deletes the comment instead of placing it.

baseballyama:
Closes #3613 for `{@const}`. The two other hosts the issue names are **not** in scope — see *Scope*, which is set by measurement rather than by the issue's table.

## What the base actually does

The issue's grid was recorded against a tree that has #3602's parse-wrapper fix. This base does not, so I built `rsvelte-fmt` and measured each host before deciding what to change:

| input | this base |
|---|---|
| `<b>{flag // c⏎}</b>` | **errors**, exit 2 — `` Expected `)` but found `EOF` `` |
| `<div data-a={flag // c⏎}>` | **errors**, exit 2 |
| `{@const x = flag // c⏎}` | **mis-prints**, exit 0 — `{@const x = flag; // c}` |

So `{@const}` is the only one of the three that reaches the printer today, and it is the silent one: exit 0, output no Svelte parser accepts.

## Cause — two defects, not one

A block header (`{#if}`, `{#key}`) is safe for a structural reason rather than a deliberate one: `push_bare_expression` splices **only the expression span**, so the source newline before `}` is never touched. `{@const}` prints its own `}` and has no equivalent protection.

**(a) the statement semicolon survives.** `format_const_declaration` recovers the body with

```rust
let s = s.strip_suffix(';').unwrap_or(s);   // declaration.rs
```

With a trailing comment oxc prints `const x = flag; // c`, so the `;` is no longer the last byte and the strip is a no-op — the `;` ends up inside the tag. Fixed by running the printed text through `drop_statement_semicolon` first, the **existing** quote- and comment-aware forward scanner in `format_core.rs` (added for #3092/#3093). No second scanner was introduced for this.

**(b) the closing brace is commented out.** `push_const_tag` emits `format!("{{@const {formatted}}}")`. When `formatted` ends in an unterminated `//`, that `}` is inside the comment. Fixed with a new `ends_in_line_comment`, built on the same scanning idiom (a naive `lines().last().contains("//")` would false-positive on `{"a//b"}` and on `/* // */`).

```
before: {@const x = flag; // c}
after:  {@const x = flag // c
        }
```

## Verification

`crates/rsvelte_formatter/tests/const_tag_line_comment.rs`, 5 tests. **Each hunk was ablated separately** against an otherwise unchanged tree:

| tree | result |
|---|---|
| hunk (a) reverted | **1 FAILED** — `statement semicolon survived`; the other 4 pass |
| hunk (b) reverted | **2 FAILED** — `closing brace inside the comment`, plus idempotency (the second pass cannot parse the first pass's output); the other 3 pass |
| both applied | 5 passed |

So both hunks are load-bearing, each fails for the reason predicted, and the three control tests never move under either ablation.

The controls are the point of the file, not decoration:

- `a_const_without_a_trailing_comment_is_unchanged` — the direction a blanket "always break before `}`" would break
- `a_block_comment_is_not_treated_as_a_line_comment` — `/* c */` terminates, so the `}` must stay on the line
- `a_slash_slash_inside_a_string_is_not_a_comment` — `{"a//b"}` must not be read as a comment
- `the_fix_is_idempotent` — verified end to end through the CLI as well as in-process

Whole crate: `cargo test -p rsvelte_formatter` **exit 0**; `cargo fmt --all -- --check` clean; `cargo clippy -p rsvelte_formatter --all-targets --all-features -- -D warnings` clean; workspace pre-commit hook passed.

## Scope — what this deliberately leaves

**The expression tag and the attribute value are not fixed here.** Both fail before reaching the printer, in `format_expr_core`'s wrapper `format!("({expr_source});")` whose `);` the comment swallows. The issue records that half as landing with #3602, which has its own PR; adding it here would duplicate that work, and the emission-site change cannot be tested end to end until it lands. When it does, the same `ends_in_line_comment` guard applies at `splice.rs:171` and `markup/value.rs:147,149`.

**Parity is not the goal here, and cannot be.** The oxfmt oracle **deletes** the comment for five of the issue's nine rows, so no fix that preserves it can be byte-identical to the oracle. This change is justified by the output being parseable at all — the `CLAUDE.md` #2590 family, where feeding compiler output to a parser found 35 real defects that no equality gate could see.

## Adjacent, not fixed

`format_declaration_tag_body` (`declaration.rs`) strips its terminator with `trim_end_matches(';')` — the same shape as (a) and, being a `trim_end_matches`, also strips a run rather than one. It serves `{let}` / `{const}` declaration tags. I did not touch it because I have not measured a failing input for it, and a change without one would be speculative.


baseballyama:
## The `Formatter parity` failure was real, and it was mine

The gate reported one NEW failure on `flowbite-svelte/src/routes/layouts/ComponentsLayout.svelte`:

```
oracle: {@const pathWithoutSlash = builder.path.replace(/^\//, "")}
actual: {@const pathWithoutSlash = builder.path.replace(/^\//, "")
```

`ends_in_line_comment` was a byte scan, and the last two bytes of the regex literal
`/^\//` are `//`. It read the rest of the line as a comment, concluded the `}` would be
commented out, and broke a line that must not break. This is the `opaque-keyword` family
`AGENTS.md` already records for the compiler (#2988, where a `/$derived(x)/` regex literal
was itself rewritten as code) — reproduced in the formatter.

## The fix, and a second defect it exposed

I did not add regex handling to the scanner. Whether a `/` opens a regex or is division
cannot be decided without the previous significant token, and this repository has paid for
hand-rolled answers to that question more than once. **Both predicates now ask the parser.**

Writing the regression test surfaced a second, pre-existing instance in the same file.
`drop_statement_semicolon` is regex-blind in exactly the same way, so on

```
{@const p = s.replace(/^\//, "") // c
}
```

it never reaches the statement's real `;` and leaves it in the tag body. That one was not
speculative — I had the failing input, which is the bar `AGENTS.md` sets for touching an
adjacent function. It now takes the last statement's span end from the parse and **falls
back to the existing byte scan when the text does not parse**, so input the parser rejects
keeps its current behaviour exactly.

| test | before this commit |
|---|---|
| `a_regex_literal_is_not_a_line_comment` | FAILED — the `}` was pushed to the next line |
| `a_regex_literal_followed_by_a_real_comment_still_breaks` | FAILED — `…"");` — the `;` survived |
| the other five | passed |

All 7 pass now.

## Stated limit

**The whole-crate suite was not run.** I started `cargo test --release -p rsvelte_formatter`
and killed it when free disk on this machine reached 5.7 GiB with the volume at 99% — that
build produces a dozen new test binaries, and running out mid-link here has twice taken the
shell down with it. So the local evidence for this commit is the 7 targeted tests, not the
crate. `drop_statement_semicolon` is shared with `format_expr_core`, so a regression there
would show up in CI's `Formatter parity` and `Test fmt corpus` jobs rather than locally.
Flagging it rather than implying coverage I do not have.

Issue #3647, opened 2026-08-23
client: a store name inside a regex literal is rewritten to `$name()`, changing what the regex matches


baseballyama:
Phase 3's client store-read rewrite turns a `$name` spelled inside a **regex literal** into
`$name()`, silently changing what the user's regex matches. Output parses and runs, so no
parse gate can see it.

Oracle: `submodules/svelte` @ `20b341f10048` — **svelte 5.56.9**, asserted from the compiler's
own `VERSION` export.

## Repro

```svelte
<script>
	import { readable } from 'svelte/store';
	const re = /\$s/;
	const s = readable(1);
</script>
<b>{$s}</b>
```

```js
// official
const re = /\$s/;
// rsvelte
const re = /\$s()/;
```

Everything else in the two outputs is byte-identical, including the `$s` subscription itself —
which is **correct here**. `$s` is genuinely read in the template, so phase 2 is right to
create the store binding. The defect is only that the rewrite of `$s` → `$s()` is applied to
an occurrence inside a regex literal.

## Why this is filed rather than folded into the #3620 fix

#3620 is the phase-2 half: the `$`-reference **collector** had no regex state, so
`/\$mystore/` was counted as a reference even when no store existed. That is fixed by having
phase 2 call `js_scan::skip_opaque`, and it takes a 14-case matrix from 11 divergences to 1.

This is the remaining 1, and it is a different scan in a different phase. #3620's own text
asks for exactly this separation:

> Keep it separate so its effect is attributable.

The two halves are also **independent**: the phase-2 fix does not reach this, and a phase-3
fix would not have fixed the ten cases the phase-2 fix did, because in those the store does
not exist at all.

## What is already known about the boundary

Measured while scoping #3620, on the base tree:

| shape | verdict |
|---|---|
| a **prop** name (no `$`) spelled inside a regex, prop read rewritten to `name()` elsewhere | **byte-identical to official** |
| a **store** name spelled inside a regex, store read rewritten to `$name()` elsewhere | **diverges** |

So the prop-read rewrite is *not* regex-blind and the store-read rewrite *is*. Whatever the
fix is, the prop case is the control that must not move.

## Not measured

- Whether the server target diverges the same way. The #3620 matrix compiled both targets and
  this case diverged on the client; the server column for this specific shape was not isolated
  from the phase-2 symptom, so treat it as unknown rather than as clean.
[...]

Issue #3645, opened 2026-08-23
AGENTS.md's Svelte version (v5.56.8) is stale and, unlike README's, is maintained by nothing


baseballyama:
## Summary

`AGENTS.md:772` opens the Test Status section with

> Source: `pnpm run compatibility-report` (Svelte **v5.56.8**). Re-run `pnpm run test-and-update` to refresh.

The pinned submodule is **5.56.9** (`submodules/svelte` @ `20b341f10048`, whose
`packages/svelte/package.json` and `src/version.js` both say `5.56.9`).

## The stale number is the symptom; the asymmetry is the bug

`README.md:75` carries the same fact and is **correct**:

> **Targeting Svelte `v5.56.9`** ([`20b341f10048`]…). This line is updated by `pnpm run update-docs`.

The two files disagree, and the one that is right is the one a script owns. `scripts/dev/update-docs.mjs`
writes the README marker (`:187`), and it additionally has a `--check` mode that **fails** when the
marker drifts from the submodule (`:265-282`).

`AGENTS.md` has none of that: no marker block, no writer, no check. Grepping the script for
`AGENTS.md` returns nothing. So the line reads as machine-maintained — it sits directly above the
sentence telling you to re-run `pnpm run test-and-update` — while nothing has ever written it and
nothing will ever notice it drifting.

That matters more than one digit, because this is the version number an agent quotes when it
reports what it measured against. I hit it on #3471: I asserted my oracle's identity from the
compiler's own `VERSION` export (5.56.9) and only then noticed the doc said 5.56.8. Had I trusted
the doc, every number in that PR would have carried the wrong provenance while being perfectly
correct — the failure mode is a true measurement with a false label, which no test can catch.

## Suggested fix

Two parts, and the second is the one worth having:

1. Correct `5.56.8` → `5.56.9`.
2. Give the line the same treatment README's has — a marker block written by `update-docs.mjs` and
   covered by its existing `--check`, so the two files cannot disagree again. The check already
   exists; it just needs a second target.

If a marker in `AGENTS.md` is unwanted (it is a prose document and a marker is intrusive), the
cheaper alternative is to **delete the version from the prose** and point at README's line instead.
A fact stated in two places with one owner will drift again; stating it once is a fix, restating it
by hand is not.

## Deliberately not fixed in a PR of my own

`AGENTS.md` is on essentially every in-flight PR's merge path, and the queue is backed up — a
[...]

Issue #3127, opened 2026-08-19
compiler: a `$`-prefixed class member name is read as a store reference and rejected


baseballyama:
A `$`-prefixed **class member name** is read as a store reference, so the compiler rejects code the official compiler accepts.

```svelte
<script>
	class P {
		$abc() {
			return 1;
		}
	}
</script>
```

rsvelte: `global_reference_invalid` — "`$abc` is an illegal variable name". Official `svelte/compiler`: compiles. Fields and getters behave the same way.

## Mechanism

The `$`-reference collector in `crates/rsvelte_core/src/compiler/phases/2_analyze/store_subscriptions.rs` is a lexical scan. It already excludes object-literal keys, member-expression properties, string literals and comments — a **class body** is the shape it does not exclude. Upstream reads `module.scope.references`, which only ever holds real references, so a declaration slot can never appear there.

## How it was found

By the new `scripts/compat-corpus/lint-severity.mjs` gate (#3111), which runs both linters with no explicit configuration and compares the process **exit code**. `rsvelte-lint` merges the compiler's own diagnostics into its report, so a compiler over-rejection shows up there as an exit-code divergence that no finding-level comparison can see. Repro: `compatibility/lint-adversarial/no-inspect/15-class-members.svelte`.

It is recorded as an accepted entry in `compatibility/lint-severity-known-failures.json`, justified under "Defect A" in the paired `.md`. That ratchet is two-sided, so the entry fails CI the day this is fixed rather than rotting.

## Note for whoever picks this up

A first attempt at this also changed `phases/3_transform/client/rune_transforms.rs`, reverting the `$inspect` hole change from #3063, which had landed on `main` in the meantime. Merge `main` before touching that file.

baseballyama:
A regression test for this was written during the lint campaign but is **not** on `lint/adversarial-parity` — that branch reverted the compiler-side fix, so the test asserts behaviour the tree does not implement and was removed in `5965ea09c` to keep CI honest. Preserving it here so it travels with the fix.

```rust
// crates/rsvelte_core/tests/class_member_dollar_name.rs
//! A `$`-prefixed class member NAME is a declaration slot, never a reference,
//! so upstream's `module.scope.references` — the set both the store-subscription
//! loop and runes-mode auto-detection read — never holds it. rsvelte reached the
//! same names through a lexical scan and rejected `class P { $abc() {} }` with
//! `global_reference_invalid`, while `$inspect` as a member name flipped the
[...]

PR #3656, opened 2026-08-23
fix(analyze): a $-name inside a regex literal is not a store reference

### LINKED ISSUE #3620: phase 2's $-reference collector re-implements js_scan's lexical substrate, and has no regex-literal handling
`2_analyze/store_subscriptions.rs` re-implements the lexical substrate that
`3_transform/shared/js_scan.rs` already provides, and the two have drifted. **The predicates
built on top are answering different questions and should NOT be unified** — the duplication is
one level down.

## The two questions are genuinely different

| | question | why it exists |
|---|---|---|
| phase 2 | is this `$name` occurrence a **reference**, or a **declaration slot**? | upstream reads `module.scope.references`, which by construction holds no declaration slot, so upstream never asks |
| phase 3 | is this `$rune(` a **real rune call position**? | upstream's `get_rune` is handed a resolved callee |

Phase 2's predicates are all in the exclusion direction — every one answers "this is a declaration
slot, not a reference":

```
store_subscriptions.rs:700   is_dollar_ident_object_property_key
                      :752   is_dollar_ident_variable_declaration
                      :813   is_dollar_ident_destructuring_declaration
                      :869   is_dollar_ident_import_specifier
                      :897   is_dollar_ident_type_declaration
                      :1219  class_keyword_at      :1236 class_body_open
                      :1330  word_ending_at        :1345 starts_a_class_member
```

Merging these with phase 3's rune-site predicates would carry #3127's hardest constraint — a class
member ends by **ASI**, not `;`, so `a = 1⏎$abc()` must not read as a reference — into a pipeline
that has no such requirement. **That is not the fix being proposed here.**

## What IS duplicated: the substrate underneath

| substrate concern | phase 2 | phase 3 |
|---|---|---|
| string / template-literal / comment state | own machine, `store_subscriptions.rs:992-1115` (`in_string`, `in_line_comment`, `in_block_comment`, `template_stack`, `brace_depth`) | `js_scan.rs`: `skip_opaque` (146), `skip_template` (216) |
| previous significant token | own tracking (`:838`, `:1345`) | `code_bytes` (15), `prev` threaded through `skip_opaque` |
| identifier boundary | own (`word_ending_at`, `:1330`) | `is_ident_byte` (107), `is_ident_char` (342) |
| representation | `Vec<char>` | `&[u8]` |
| **regex literals** | **absent — 0 occurrences of any `slash_starts_regex`** | **present — 7 occurrences** (`slash_starts_regex` 61, `slash_starts_regex_at` 113) |

## The measured asymmetry

Phase 2's scanner is not a naive one — it tracks template literals with a nesting stack and
decides class-member slots from the previous significant token, which is #3127's fix. **But it has
no notion of a regex literal**, while `js_scan` does.

That is the shape #2988 was: a `/$derived(x)/` regex literal read as code. Whether it actually
reproduces through this collector is **not tested here** — this issue records a structural
difference, not a confirmed defect. A probe would be a script containing a regex literal whose body
spells a `$`-name, e.g.

```svelte
<script>
	const re = /\$state\(/;
	let count = 0;
</script>
```

and asking whether the collector counts a `$state` reference (which would feed both the store
subscription decision and runes-mode auto-detection).

## Why this is filed rather than folded into an open PR

Both #3438 and the `salvage/3233-3235-3236-rune-name-resolution` branch touch rune-name resolution
and neither changes this substrate. Folding it into either one would mix a latent-defect change
into a comparison between two implementations, and #3128 is the precedent for why that is bad: an
over-rejection is loud and its repair is quiet, so a change to the `$` collector breaks in a way
that still compiles. Keep it separate so its effect is attributable.

## Not measured

- whether the regex gap reproduces (above)
- whether phase 2 could adopt `js_scan`'s `&[u8]` substrate without a `Vec<char>` → byte-offset
  conversion at every predicate boundary — the representations differ, and that conversion is the
  real cost of the unification, not the scanning itself

baseballyama:
Closes #3620.

`2_analyze/store_subscriptions.rs` decides which `$name` occurrences are *references*
by scanning the script's characters. That scan tracks strings, template literals with a
nesting stack, line and block comments, and — since #3127 — class bodies and the previous
significant token. It has **no notion of a regex literal**, so `/\$mystore/` was scanned as
code and its body counted as a store reference.

The fix does not add a second answer. `3_transform/shared/js_scan.rs` already decides
whether a `/` opens a regex from the preceding token (`slash_starts_regex_at`) and already
skips the literal including character classes and escaped `/` (`skip_opaque`); phase 2 now
calls it. A second implementation here would be free to disagree with the first, which is the
class `compatibility/two-ports-inventory.md` exists to track.

## Oracle

`submodules/svelte` @ `20b341f10048`, **svelte 5.56.9**, asserted from the compiler's own
`VERSION` export at the top of every probe run — not from a filename and not from `AGENTS.md`,
whose Test Status section still says 5.56.8 (#3645). Every expectation in the test file was
generated by running the input through official and read off; none was written from reasoning
about what upstream should do.

## The issue's own probe does not reproduce

#3620 proposed

```svelte
<script>
	const re = /\$state\(/;
	let count = 0;
</script>
```

and that case is **byte-identical** on both compilers. A rune name is excluded from store
treatment downstream, so the collector's mistake is absorbed before it reaches the output.
The shape that discriminates is a `$`-name that is **not** a rune name — a store name. Four
of the five regex cases in the issue's own suggested direction (`$state`, `$derived`,
`$inspect`, `$props`) are silent for the same reason.

## Before / after

Fourteen cases crossing the lexical arm a `/` can end on, the host, and the mode. All
fourteen compile on official.

| | diverging |
|---|---|
| base | **11 of 14** |
| with this change | **1 of 14** |

The eleven that moved are every case carrying a regex literal, including `<script module>`,
runes mode, inside a `${…}` substitution, after `return`, a `/` inside a character class, an
escaped `\/`, and a literal that follows a division earlier in the same script.

## Two symptoms, and the second is the one to notice

`/\$mystore/` did not merely add a spurious `$.setup_stores()` and `$mystore` thunk. Because
the name then resolved to a store, phase 3 rewrote **the literal itself**:

```js
// official
const re = /\$mystore/;
// rsvelte, before
const re = /\$mystore()/;
```

The regex now matches something else. That output parses and runs, so the parse gate is blind
to it and only output equality reports it.

## The isolating control

`plainname-regex-prop` — the same file with a prop name that carries no `$` — was
**byte-identical on the base tree**. So the literal rewrite is a *consequence* of the phase-2
misdetection, not an independent regex blindness in phase 3's prop-read rewrite. That
distinction is what scopes this PR to one file.

## Controls

The over-broad direction is what the new tests are mostly for.

- `a_real_store_beside_a_regex_still_subscribes` — `$other` is spelled **only** inside the
  regex and `$s` **only** outside it, so skipping too much loses a real subscription while a
  regex-only file still passes.
- `division_is_not_read_as_a_regex_literal` — `a / b / 1` must survive. Reading the preceding
  token wrong here does not produce a wrong name, it swallows the rest of the script.
- The same test's second row is `let 日本 = 6; let c = 日本 / 2;`. `js_scan` is byte-based and
  phase 2 is char-based, so the previous token has to be mapped to a byte; a non-ASCII
  identifier char maps to `b'x'` on the reasoning that **JS spells no operator outside ASCII**.
  That row is the control for that assumption, and official's output for it (`let c = 日本 / 2;`,
  verbatim) is what it is pinned against.

**Revert control, run on this tree.** With the fix reverted and the tests kept:

| test | reverted |
|---|---|
| `a_dollar_name_inside_a_regex_literal_is_not_a_store` | **FAIL** — `prop/client: regex body became a store subscription` |
| `the_regex_literal_survives_byte_for_byte` | **FAIL** — ``prop/client: `/\$mystore/` was rewritten`` |
| `a_real_store_beside_a_regex_still_subscribes` | **FAIL** — `COMPILE_ERROR: … "global_reference_invalid" … "`$other` is an illegal variable name"` |
| `division_is_not_read_as_a_regex_literal` | **pass** |

**1 passed, 3 failed**, each for its predicted reason, and the one control that a fix cannot
move stays green. The third row came back louder than predicted: on the base tree that file is
not merely mis-rewritten, it is **rejected outright** — a regex spelling a `$`-name with no
binding raises `global_reference_invalid`. So the defect has an over-rejection arm as well as a
wrong-output arm. The restored file is byte-identical to the pre-ablation copy.

## Scope — the one remaining divergence is a different site, and it is filed

`store-regex-imported` still diverges. There the store is **real** — `$s` is read in the
template, so phase 2 is right to create the subscription — and phase 3's **store-read**
rewrite then rewrites the `$s` spelled inside `/\$s/` to `$s()`. That is a different scan in a
different phase, and #3620 is explicit that this half should not be folded in ("Keep it
separate so its effect is attributable"). Filed as #3647 with the measured repro.

Note that `3620-regex-literal-beside-real-store.svelte` deliberately spells a **different**
name inside the regex from the real store, so it is a control for this PR and not a repro for
that one.

## Verification, and what was not run

Measured on **this branch**, whose base is the current `origin/main` (`d4b6e4e4f`) — not on the
tree the matrix figures were first taken on. The branch was cut from `origin/main` with a clean
tree and `git diff origin/main...HEAD --stat` is exactly these six files.

- `cargo test --release -p rsvelte_core --test regex_literal_store_ref_3620` — **4 passed**
- `cargo fmt --all -- --check` — clean, exit 0
- `node scripts/release/check-changeset-package-names.mjs`, `check-core-consumer-changesets.mjs`,
  `check-esrap-version-bump.mjs` — all exit 0

**Not run: the rsvelte_core crate suite.** This machine has three other agents and a peer session
compiling; a `rsvelte_core` rebuild is currently taking over ten minutes and the full suite is a
few hundred more test binaries on a volume that has run out twice tonight. The local evidence for
this change is the four targeted tests plus the 14-case oracle matrix, not the crate.
`cargo clippy -p rsvelte_core --lib --tests --all-features -- -D warnings` is in flight; I will
comment its verdict rather than state one I do not have.

The corpus, matrix and mutation gates are CI's. Worth flagging for whoever re-baselines: the two
new `pattern-corpus` files are new corpus entries, so they add ids to the manifest — both are
byte-identical to official on client and server on this tree, measured directly.


baseballyama:
## Clippy verdict

`cargo clippy -p rsvelte_core --lib --tests --all-features -- -D warnings` — **exit 0**, zero
warnings, on this branch.

**With its own positive control**, because a clean clippy run and a clippy run that never
reached the file are the same empty output. Injecting a `return` before the final expression of
`regex_literal_end` — the function this PR adds — takes the same command to **exit 101** with

```
error: unneeded `return` statement
```

so the run does reach the new code. The file was then restored and `git diff` is empty for it.

## What is still not run

The **rsvelte_core crate suite**. Three other agents and a peer session are compiling on this
machine; a `rsvelte_core` rebuild is taking over ten minutes and the full suite is a few hundred
more test binaries on a volume that has run out of space twice tonight. The local evidence for
this change is the four targeted tests plus the 14-case oracle matrix — not the crate, and not
the corpus, matrix or mutation gates, which are CI's.

Issue #3652, opened 2026-08-23
A member access on `import.meta` in a template is folded as static, and official treats it as dynamic


baseballyama:
## Summary

A member access on `import.meta` in a template expression is treated as a **static** value by
rsvelte and as a **dynamic** one by the official compiler. The whole component shape changes as a
result: on the client official emits `$.template_effect(…)` (plus the legacy `$.push($$props,
false)` / `$.pop()` pair), rsvelte assigns once; on the server official wraps the body in
`$$renderer.component((…) => …)` with a `$$props` parameter, rsvelte emits the bare form.

`import.meta` **on its own** matches. It is the member access on it that flips the decision, and
it does so in every template slot tried.

```svelte
{import.meta.url}
```

```js
/* official, generate: 'client' */
import 'svelte/internal/disclose-version';
import 'svelte/internal/flags/legacy';
import * as $ from 'svelte/internal/client';

export default function A($$anchor, $$props) {
	$.push($$props, false);
	$.init();
	$.next();

	var text = $.text();

	$.template_effect(() => $.set_text(text, import.meta.url));
	$.append($$anchor, text);
	$.pop();
}

/* rsvelte */
import 'svelte/internal/disclose-version';
import 'svelte/internal/flags/legacy';
import * as $ from 'svelte/internal/client';

export default function A($$anchor) {
	$.next();

	var text = $.text();

	text.nodeValue = import.meta.url;
	$.append($$anchor, text);
}
```

```js
/* official, generate: 'server' */
export default function A($$renderer, $$props) {
	$$renderer.component(($$renderer) => {
		$$renderer.push(`<!---->${$.escape(import.meta.url)}`);
	});
}

/* rsvelte */
export default function A($$renderer) {
	$$renderer.push(`<!---->${$.escape(import.meta.url)}`);
}
```

## Axis

Oracle is `submodules/svelte` at the pin, `VERSION` read off the compiler's own export = **5.56.9**.

| source | client | server |
|---|---|---|
| `{import.meta}` | match | match |
| `{import.meta.url}` | **DIFF** | **DIFF** |
| `{import.meta.env.MODE}` | **DIFF** | **DIFF** |
| `<div title={import.meta.url}></div>` | **DIFF** | **DIFF** |
| `{#if import.meta.url}y{/if}` | **DIFF** | **DIFF** |
| `{@html import.meta.url}` | **DIFF** | **DIFF** |
[...]

PR #3657, opened 2026-08-23
fix(analyze): decide the snippet hoist from references, not an expression-kind list

### LINKED ISSUE #3269: An optional `{@render x?.()}` inside a snippet blocks the module-scope hoist official still performs
A snippet with no component-state dependency is hoisted to module scope by both
compilers — unless its body contains an **optional** `{@render inner?.()}`, in
which case rsvelte keeps it inside the component function. All three targets.

```svelte
<script>
	let flag = $state(true);
</script>
{#snippet outer()}{#snippet sn()}<i>x</i>{/snippet}{@render sn?.()}{/snippet}{@render outer()}
```

`generate: 'server'`:

```js
// official — hoisted
function outer($$renderer) {
	function sn($$renderer) { $$renderer.push(`<i>x</i>`); }
	sn?.($$renderer);
}
export default function A($$renderer) {
	let flag = true;
	outer($$renderer);
}

// rsvelte — nested
export default function A($$renderer) {
	let flag = true;
	function outer($$renderer) {
		function sn($$renderer) { $$renderer.push(`<i>x</i>`); }
		sn?.($$renderer);
	}
	outer($$renderer);
}
```

Client and client-dev are the same divergence with `const outer = ($$anchor) => …`.
Not just a byte difference: the hoisted form allocates the closure once per
module, the nested form once per component instance.

## Scope

| shape | verdict | official hoists | rsvelte hoists |
|---|---|---|---|
| `{@render sn?.()}` inside `outer` | **DIFF** | yes | no |
| `{@render sn()}` inside `outer` | match | yes | yes |
| `{@render outer?.()}` at the call site | match | yes | yes |
| both calls optional | **DIFF** | yes | no |
| three levels, innermost optional | **DIFF** | yes | no |
| `{@render sn?.(1)}` (with an argument) | **DIFF** | yes | no |
| `{@render sn?.()}` with no outer snippet (flat) | match | no | no |
| `outer` reads `$state` (`{n}` in the inner body) | match | no | no |
| the optional render wrapped in `{#if}` | match | no | no |

So it is the optional call **inside** the hoistable snippet that flips the
decision, not the optional call as such — `{@render outer?.()}` from outside
leaves the hoist intact on both sides. The two rows where neither hoists (state
dependency, `{#if}` wrapper) are the built-in controls: the hoist analysis is
reached and agrees there.

## Found by

Adversarial snippet grid: 12 declaration sites × 13 parameter shapes × 6 render
shapes × 3 targets = 2,808 cells, 2,340 accepted by official, 36 divergences —
**all of them** in the (`in-snippet`, `optional`) cell, uniformly across 12 of
the 13 parameter shapes. The 13th (`a = n + 1`) is the one whose default reads
`$state`, which suppresses the hoist on both sides.

### LINKED ISSUE #3444: A snippet containing `{@render s?.()}` is not hoisted to module scope
## Summary

A snippet whose body renders another snippet with an **optional call** — `{@render s?.()}` — is
hoisted to module scope by the official compiler and left inside the component function by
rsvelte. Writing the same render without `?.` matches byte for byte.

```svelte
<script>
	let v = $state(1);
</script>
{#snippet outer()}
	{#snippet s()}x{/snippet}
	{@render s?.()}
{/snippet}
{@render outer()}
```

```js
/* official, generate: 'client' — `outer` is a module-scope binding */
const outer = ($$anchor) => {
	const s = ($$anchor) => { … };
	…
};

export default function A($$anchor) { … }

/* rsvelte — `outer` stays inside the component */
export default function A($$anchor) {
	const outer = ($$anchor) => {
		const s = ($$anchor) => { … };
		…
	};
	…
}
```

`generate: 'server'` diverges the same way (`function outer($$renderer)` at module scope vs
nested), and so does `client-dev`.

## Controls

| source | verdict |
|---|---|
| `{@render s?.()}` inside a snippet | **DIFF on all three targets** |
| `{@render s()}` inside a snippet — the same shape without `?.` | match (both hoist) |
| `{@render s?.()}` at the fragment root | match |
| `{@render s?.()}` inside `{#if}` | match |
| `{@render v?.()}` inside a snippet, where `v` is `$state` | match (both decline to hoist) |
| a snippet declared outside that renders `s?.()` | **DIFF** |

The `$state` row is what makes this a hoisting decision rather than a rendering one: when the
callee genuinely is component state, both compilers agree not to hoist. Only the optional-call
form of a *snippet* reference is misread.

## Scope of the sweep

Found by a template-block grid — 27 block constructs × 7 nesting hosts × 3 targets, 567 cells.
This is the **only** divergence in it: snippets with rest/destructured/defaulted/array parameters,
recursive snippets, `{@render}` through a ternary and through a member expression, every
`{#await}` arm shape including a destructured `then`, `{#key}`, `{#each}` with index and key,
`{:else if}` chains, and all four `<svelte:boundary>` forms are byte-identical everywhere.

baseballyama:
Closes #3444. Closes #3269.

## They are the same defect, and neither issue's stated mechanism is the cause

#3444 reports `{@render s?.()}` inside a top-level snippet not being hoisted to module scope;
#3269 reports the same thing with an argument (`{@render sn?.(1)}`) and through three levels of
nesting. Both attribute it to the **optional call**. Both are the same defect and the optional
call is not the cause.

`can_hoist` is decided in `2_analyze/visitors/snippet_block.rs` by
`expression_only_uses_params_node` — a `match` over expression node kinds ending in `_ => false`.
So a top-level snippet was pinned inside the component function by any expression kind the arm
list did not name, whatever that expression actually referenced. `ChainExpression`, which is what
every `?.` parses to, is one such kind; it is not the only one.

Upstream never enumerates expression kinds. `2-analyze/visitors/SnippetBlock.js`'s
`can_hoist_snippet` iterates `scope.references`, so it only ever asks about **identifiers** and an
expression kind is transparent to it. Two measured rows show the operator is not the variable:

| body of a top-level `{#snippet outer()}` | official | rsvelte before |
|---|---|---|
| ``{tag`x`}`` — a tagged template, no `?.` anywhere | hoists | pinned |
| `{new (class { m() { return M; } })().m()}` | hoists | pinned |
| `{(mo as any)?.a}` | hoists | pinned |
| `<div onclick={() => mo?.a}></div>` — the same `?.` | hoists | **hoists** |

The last row is the discriminator. The predicate has a third copy, `refs_hoistable`, used for
arrow-function bodies; it is a generic reference walk, so `?.` was already right there. The
identical `{mo?.a}` was pinned standing alone and hoisted inside `onclick={() => …}` — which
nothing about optional chaining can explain, and which is the "two ports of one upstream function,
and no gate compares the ports" shape from `compatibility/two-ports-inventory.md`, one level down.

The correction is posted on both issues, because the issue text outlives the PR.

## The fix

Make the unnamed kinds transparent instead of fatal, in both copies of the predicate:

* the typed walk (`expression_only_uses_params_node`) gets an explicit `ChainExpression` arm —
  the common shape, so it costs no allocation — and its `_` arm now walks the node for references
  via `refs_hoistable` instead of rejecting it;
* the JSON twin (`expression_only_uses_params`, reached from `{@const}` / `{let}` initializers)
  gets the same two changes.

This can only ever be **more** conservative than upstream, never less: upstream blocks on a
reference to a non-hoistable binding, and a walk of the whole subtree is a superset of the
reference set (it additionally sees identifiers in non-reference positions, which
`is_identifier_hoistable` rejects). It cannot turn a pinned snippet into a wrongly-hoisted one.

## Before / after over the whole matrix

Blocker × host × target grid: 60 subtree contents × 5 hosts (top level, `{#if}`, `{#each}`, an
element, `<svelte:boundary>`) × 4 targets (client, client-dev, server, server-dev) = **1,200
comparisons**. Each row is scored three ways: whether `outer` is declared at module scope in the
output, byte equality of the whole output, and an acorn parse of **both** outputs. The oracle is
`submodules/svelte` at the pin, with `VERSION` read off the compiler's own export — **5.56.9**
(AGENTS.md says 5.56.8; it is stale).

| | before | after |
|---|---|---|
| `match` | 1,041 | **1,125** |
| `hoist-diff` | 84 | **0** |
| `other-diff` (unrelated, see below) | 15 | 15 |
| `both-error` (not comparable) | 60 | 60 |
| distinct diverging shapes | 21 | **0** |
| outputs no JS parser accepts | 0 | 0 |

The 84 rows are 21 shapes × 4 targets, all at the top-level host and all in the
"official hoists, rsvelte does not" direction. Besides the two shapes the issues report, they
include a bare `{mo?.a}`, `{mo?.[0]}`, `{mf?.()}`, `title={mo?.a}`, `{@const v = mo?.a}`,
`{#if mo?.a}`, `{#each mo?.a ?? [] as z}`, `{@html mo?.a}`, `{(0, mo?.a)}`, `{mo?.a ?? M}`, a
snippet parameter default `(a = mo?.a)`, a snippet parameter called as `{f?.()}`, a tagged
template and a class expression.

## Controls

Both directions are in the grid, because a hoist that happens and should not is a **correctness**
bug — the hoisted function loses the bindings it closed over — while a hoist that does not happen
is a missed optimisation.

| control | rows | before | after |
|---|---|---|---|
| top-level snippet that **must not** hoist (`$state` read, store, instance callee, an instance object behind every optional form, instance `$state` in a parameter default, class expression closing over instance state) | 100 | rsvelte hoisted 0 | rsvelte hoisted 0 |
| top-level snippet that already hoisted correctly (`{@render s()}`, `{mo.a}`, `{mf()}`, `{M}`, static text, `onclick={() => mo?.a}`) | 44 | matched official | matched official |
| snippet **not** at top level, so never hoistable (4 hosts × 57 shapes) | 912 | neither side hoists | neither side hoists |
| official's own answer, per row | 240 | 128 hoist / 100 pin | unchanged |

`crates/rsvelte_core/tests/snippet_hoist_expression_kinds_3444.rs` pins 56 of those cells (31
"must hoist", 25 "must not") on client, client-dev and server. Every `hoist:` value in it was
**generated from the official compiler's output**, not transcribed: the file is emitted from the
grid's JSON, and the generator refuses to emit a row whose measured oracle answer disagrees with
the row's declared direction.

### Revert control

With the fix reverted (`git checkout --` on `snippet_block.rs`) and the tests kept, all three
tests go red, and they go red *for the predicted reason and only in the predicted direction*:

```
test hoist_decision_matches_official_on_server ... FAILED
test hoist_decision_matches_official_on_client_dev ... FAILED
test hoist_decision_matches_official_on_client ... FAILED

21 of 56 cells disagree with the official hoist decision:

inner-optional [server]: official hoists = true, rsvelte hoists = false
--- source ---
{#snippet outer()}{#snippet s()}x{/snippet}{@render s?.()}{/snippet}
{@render outer()}
--- output ---
export default function Component($$renderer) {
	function outer($$renderer) {
…
test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out
```

63 failing rows (21 shapes × 3 targets), **all** of them `official hoists = true, rsvelte hoists =
false`; **0** of them the opposite direction, so no control moved. The other 35 cells per target —
the 25 that must stay pinned and the 10 that already hoisted — stayed green with the fix reverted,
which is what makes the 21 attributable to this change rather than to the file being new. The fix
was then restored and `cmp`'d byte-identical against the copy taken before the revert.

## Corpus repros

The collected corpus does not contain this shape, and the reason is worth recording: both halves
were already there, uncrossed. `pattern-corpus/adversarial/snippets/render-optional-chain.svelte`
has `{@render header?.()}` but the callee is a prop, so neither compiler hoists;
`pattern-corpus/issues/3075-nested-snippet-hoist.svelte` has a nested snippet inside a hoistable
top-level snippet but renders it without `?.`. Binding kind and render form were confounded across
two files, exactly the `write-host` lesson in AGENTS.md. Three files land here:

* `3444-optional-render-snippet-hoist.svelte` — `{@render inner?.()}` inside a top-level snippet,
  with the `$state`-reading twin as the control that must stay pinned (the #3444 shape);
* `3269-optional-render-snippet-hoist-nested.svelte` — three levels with an argument on the
  optional render (the #3269 shape; the argument list is a different arm of the recursion);
* `3444-unenumerated-expression-hoist.svelte` — the same defect with no render tag at all, through
  an `{@const}` initializer and a computed optional member, which land in the predicate's **JSON**
  copy rather than its typed one. A fix applied to one port leaves this file diverging.

All three compile byte-identically to official on client, client-dev, server and server-dev.

## Separate defect found, filed not folded

#3652 — a member access on `import.meta` in a template expression (`{import.meta.url}`,
`{import.meta.env.MODE}`) is folded as static by rsvelte and treated as dynamic by official
(`$.template_effect` on the client, a `$$renderer.component(…)` wrapper on the server). It is the
grid's only `other-diff` verdict, present identically before and after this change. It was
separable from the hoisting defect because it fires at every host, including the four where
neither compiler hoists at all. `{import.meta}` on its own matches; `{zzz.url}` — an ordinary
member access on an unresolvable name — matches too, which is the control that names the cause.

## Deliberately left out

* **The predicate is still three implementations of one upstream function.** This PR makes them
  agree on the axis that was measured; it does not collapse them into one. Doing that is a larger
  change and, per `compatibility/two-ports-inventory.md`, wants a test whose oracle is upstream
  rather than the other port.
* **`{await …}` in a snippet** is `both-error` (`experimental_async`) on 60 rows, so this grid
  cannot say anything about the hoist decision for an async binding — upstream's
  `binding.blocker` arm of `can_hoist_snippet` is unreached here. Recorded, not fixed.
* **No new matrix family.** The grid lives in the PR's measurement, not in
  `scripts/compat-corpus/matrix/`; what lands in-tree is the Rust test plus the three repros.
* **`{/ab/.test('a')}`** is rejected by both parsers in a template expression, so the regex-literal
  cell is a both-error row rather than a comparison.

## Verification

* `cargo fmt --all -- --check` — clean.
* `cargo test --release -p rsvelte_core --test snippet_hoist_expression_kinds_3444` — 3 passed.
* `cargo clippy --release -p rsvelte_core --lib -- -D warnings` — clean, **with a positive
  control**: the first run printed no `Checking rsvelte_core` line, which is the shape of a target
  that was never checked, so a deliberate `len_zero` violation was appended to the touched file and
  the run went to `error: length comparison to zero` / exit 101. Probe removed, file `cmp`'d
  identical, clippy re-run clean. Scope is the one crate this PR touches, not the hook's
  `--all-targets --all-features` (which builds ~248 test binaries and has filled this machine's
  disk before) — the workspace-wide lint verdict is CI's. The commit was made with `--no-verify`
  for the same reason.
* The three new corpus files were formatted with `rsvelte-fmt` (`--config
  scripts/fixtures/fmt-corpus.oxfmtrc.json`) and with the `oxfmt` oracle the fmt gate uses; the two
  outputs are **byte-identical**, so the hard fmt-parity gate has nothing new to fail on. The
  committed files are tab-indented like every other file in the directory, and oxfmt makes no
  structural change to them.
* After-state sweep over every checked-in `pattern-corpus` component (717 files, 1,258 compiled
  comparisons, client + server): 17 raw byte differences, all pre-existing and all absorbed by the
  gate's normalisation — 14 layout-only (identical acorn token streams) and 3 redundant-parenthesis
  cases such as `($$value) => ($.get(todo).text = $$value)`. None involves a snippet or a hoist,
  and the client/server output ratchets are empty (0 entries) on this tree.
* `node scripts/release/check-changeset-package-names.mjs`, `check-core-consumer-changesets.mjs`
  (`✓ @rsvelte/compiler (touched: crates/rsvelte_core/src/)`) and `check-esrap-version-bump.mjs` —
  all exit 0 against the committed diff.
* `node scripts/compat-corpus/known-failures-md-check.mjs` — 45 declared ratchets across 34 docs
  match their JSON.
* Not run locally: the full corpus, matrix, mutation, svelte2tsx and LSP gates. Those are CI's.


baseballyama:
Merged `origin/main` (`5e5fe6b8b`) into the branch — merge commit, no rebase, no force push.

**The conflict was one block and only one block.** #3656 (#3620) appended two rows to the
`pattern-corpus` README's `issues/` table; this branch appends three to the same block. Resolved by
keeping all five, main's two first. `git status` reported exactly one `UU` path; every other file
main touched (`store_subscriptions.rs`, `remove_typescript_nodes.rs`, `ast_state_transform.rs`,
`Cargo.lock`, five changesets, four test files) auto-merged, and none of them is a file this PR
touches.

Checked rather than assumed:

* `git diff origin/main...HEAD --stat` (three dots) shows **exactly the 7 files this PR owns**,
  +449/−2, and nothing of main's — the merge added no drift to the PR's delta.
* All four fix hunks survived: the two `ChainExpression` arms and the two reference-walking `_`
  arms are present, and `_ => false` no longer occurs anywhere in that predicate.
* All five README rows survived and each names a file that exists — cross-checked by script
  against `issues/`, since nothing in the repo checks that a row and its file agree.
* The three repro files are **byte-unchanged** by the merge (`git diff 05894aab1 HEAD --
  issues/` lists only #3620's two new files).
* **Fmt parity re-confirmed after the merge without a rebuild**: the `oxfmt` oracle
  (`scripts/fixtures/fmt-corpus.oxfmtrc.json`) was re-run on the post-merge files and its output is
  byte-identical to the `rsvelte-fmt` output captured before the merge. Given #3658 went red on
  exactly this, that is the check worth naming.

**Incidental, pre-existing, not fixed here:** seven files in `pattern-corpus/issues/` have no row
in the README table — `2163-legacy-props-only-destructure`, `2535-root-has-with-class`,
`2626-comment-delimiter-in-initializer`, `2671-comment-between-assign-and-value`,
`2719-is-argument-specificity`, `2720-is-leading-unused-branch`, `2721-functional-pseudo-compound`.
Verified against `origin/main`'s README, so this predates both this branch and #3656. It is the
same gap the README's own convention is meant to close, and nothing gates it; I have left it alone
rather than invent justifications for files I did not write.

Gates: 35 checks re-triggered on the merge commit — 32 pending, 3 deliberately skipping, **0
cancelled** — and `Corpus Compat` (which carries `Compiler parity` / `Formatter parity` /
`Lint parity` / `svelte-check parity`) is queued. `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

No rebuild was run: free disk was 8.1 GiB on resume and is 7.7 GiB now, all of the change from
other agents. The local verification quoted in the PR description still stands on unchanged
sources.

PR #3660, opened 2026-08-23
ci: say whether the Tests legs failed or were cancelled

### LINKED ISSUE #3630: ci: the LSP full-population ratchet reports FAILURE on 34 PRs with zero input artifacts, and its 16-shard matrix sets the queue ceiling
Two separate problems, measured this morning while triaging why 146 open PRs are red.

## 1. `LSP full-population ratchet` reports FAILURE on 34 open PRs, and none of them broke the ratchet

The failing step is not a ratchet violation:

```
Found 0 artifact(s)
Filtering artifacts by pattern 'lsp-current-*'
Total of 0 artifact(s) downloaded
    at visit (scripts/compat-lsp/artifacts.mjs:196:28)
    at readArtifacts (scripts/compat-lsp/artifacts.mjs:202:3)
```

In the run I traced (`32582399567`, `Corpus Compat`, `event=pull_request`, `run_attempt=5`, created 2026-08-22T15:40Z, cancelled 2026-08-23T05:05Z):

- all 16 `LSP real-world parity (stable shard N/16)` jobs are `cancelled` with **`steps=0`** — they never started
- the run produced **`total_count=0` artifacts**
- `LSP full-population ratchet` nevertheless **ran** (`steps=6`) and exited failure

The job's `if:` already carries `&& needs.lsp-corpus.result == 'success'`, so it should have skipped. Why it did not is **undiagnosed** — I am not guessing at it here. What is established is that a job whose inputs do not exist reports `FAILURE`, and that verdict is indistinguishable in `gh pr checks` from a PR that actually shrank the ratchet.

**Of 146 open PRs, 49 have failures whose only entries are this job plus the `Tests` aggregator that mirrors it.** That is a third of the open backlog reading as broken for a reason unrelated to its content.

Worth separating from the fix: the polarity is right (a shrink-only ratchet should fail closed on absent input), but the *message* is a stack trace from `artifacts.mjs`, so the check name and the failure text both say "ratchet" when the fact is "no data". `CLAUDE.md` records the opposite-polarity version of this hazard — comparisons that score `match` when there is nothing to compare — and the same one-line question applies: can this gate tell "no input" from a verdict?

## 2. The LSP matrix is the capacity constraint on every other gate

Measured, not estimated:

| quantity | value |
|---|---|
| jobs in one `Corpus Compat` run | **26** |
| of those, `lsp-corpus` shards | **16 (62%)** |
| `timeout-minutes` on each shard | **180** |
| `Corpus Compat` share of a 100-run sample of the queued backlog | **17 runs** |
| queued runs, total | **608** |
| non-skipped completions | **48 / hour** (64 completions between 05:03:46Z and 06:24:06Z) |

So a single PR that trips the `lsp-ratchet` path filter can enqueue 16 jobs each permitted three hours, and the run I traced sat from 15:40Z to 05:05Z before being cancelled — 13.4 hours, on its fifth attempt, having produced nothing.

At 48 real completions/hour against 608 queued, a verdict for any PR opened now is **5–13 hours** out. `CLAUDE.md`'s own note that this gate takes 86% of Actions capacity for a 13% verdict-reach rate is consistent with what is measured above.

## Not proposed here

I am deliberately not proposing that the gate be weakened, made non-blocking, or moved off PRs — that trades correctness coverage for throughput and is a product decision. What this issue asks for is that the two facts be on the record with numbers attached, because the second one currently sets the ceiling on how fast anything else can be merged.

baseballyama:
Refs #3630. Part of it — see **Scope** below for what this deliberately does not do.

## The measurement that prompted it, and the correction it forces on #3630

#3630 counted **34** open PRs whose only failures were `LSP full-population ratchet` plus the
`Tests` aggregator that mirrors it. Re-measured now over all open PRs:

```
gh pr list --state open --limit 200 --json number,statusCheckRollup \
  -q '… select every PR whose FAILURE set is a subset of {LSP full-population ratchet, Tests} …'
→ 47
```

**But the population is not what the issue assumed, and I traced one to find out.** PR #3626's
failing run (`32622286812`, `event=pull_request`, **`attempt=1`**, created 2026-08-23T06:11Z)
contains **no LSP job at all**. Every one of its 26 jobs is `cancelled` with **`steps=0`**, and
the only non-cancelled job is `Tests`, `failure`, `steps=3`.

Two things follow.

**#3630's part 1 is not explained by a stale workflow file, and I checked rather than assumed.**
The `needs.lsp-corpus.result == 'success'` guard landed in `613fa1eb9` at 2026-08-22T14:29Z, and
`git merge-base --is-ancestor 613fa1eb9 0e078724b` confirms it **is** present in that PR's head.
So a run carrying the guard still produced this shape. That half stays open.

**And the 47 are a mixed population**, not 47 instances of the LSP defect. At least some — #3626
among them — are runs cancelled wholesale by `concurrency: cancel-in-progress`, where the red
check is the `Tests` aggregator doing exactly what it was designed to do.

## What this changes

The aggregator's polarity is right and is not touched: a cancelled leg ran no tests, so it is not
a pass, and a required check must fail closed. `AGENTS.md` records the opposite-polarity version
of this hazard (comparisons that score `match` when there is nothing to compare) and the correct
answer there was the same — keep failing, but be able to say why.

What was wrong is that the **message** was `One or more test jobs did not succeed.` for both
cases. In `gh pr checks` a cancelled-run red and a broken-tests red are the same single word, so
the log line is the only place the difference can be recovered — and it did not carry it.

The inline bash is now `scripts/ci/tests-aggregate-verdict.mjs`, with the **exit codes
unchanged**: 0 only when every leg succeeded, 1 for anything else. It classifies instead:

| situation | message |
|---|---|
| a leg reported `failure` | `Test job(s) FAILED: <names>.` (plus a count of any cancellations) |
| no failure, ≥1 `cancelled` | `NO VERDICT — no test job failed, but N were CANCELLED (…). A cancelled job ran no tests, so this is not a pass; it is also not a test failure.` |
| no failure, only `skipped`/absent | `Test job(s) did not run: <names>. A skipped test job is not a pass.` |
| an unrecognised result string | exit **2**, `Unrecognised job result(s): …` |

The last row is new behaviour rather than a rename: the old condition was `!= "success"`, so a
result GitHub might add later fell into the generic failure branch silently.

## The test is the point of the change

A verdict message is read by nothing else in CI, so it is precisely the kind of thing that
regresses without anyone noticing. `scripts/ci/test-tests-aggregate-verdict.mjs` pins each case's
exit code **and** its discriminating phrase, and is wired into the `Workflow trigger guard` job
next to the other guard self-tests.

The discriminating case is `a failure among cancellations reports the failure` — a script that
answered "no verdict" whenever anything was cancelled would pass every other row and fail that
one. There is also a wiring check (`ci.yml` must invoke the script and declare every env var it
reads), because a verdict script nothing calls is worth nothing; it is the same shape as the
existing trigger-guard allowlist check.

**Positive control, run on this tree:** eight checks pass. Disabling the no-verdict branch alone
(`if (false && …)`) takes it to **1 failure**, and specifically the `every leg cancelled` row,
which then reports the fall-through message — so the suite is measuring that branch and not
merely running. Restored and re-run green.

`node scripts/ci/workflow-trigger-guard.mjs` and its own self-test both still pass against the
edited `ci.yml` (20 workflows checked, no violations).

## Scope — what this does not do

- **It does not fix #3630 part 1.** Why `LSP full-population ratchet` runs with zero input
  artifacts despite its `needs.lsp-corpus.result == 'success'` guard is still undiagnosed, and I
  am not guessing at it. The measurement above narrows it: the guard was present, so the
  explanation is not "old workflow file".
- **It does not touch #3630 part 2** (the 16-shard matrix setting the queue ceiling). That is the
  product decision the issue explicitly declined to make, and this PR does not make it either.
- **It changes no verdict.** Every input that failed before still fails, with the same exit code.
  If that is not what you want — if a wholly-cancelled run should be `neutral` rather than red —
  that is a different and larger change, and it belongs with part 2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -828,8 +828,17 @@ jobs:
     needs: [test, test-unit, test-runtime, test-thread-safety, test-fmt-corpus, oxc-formatter-probe, language-server]
     if: always()
     runs-on: ubuntu-latest
-    timeout-minutes: 1
+    timeout-minutes: 5
     steps:
+      # This job had no checkout while its verdict was inline bash; the script
+      # needs one. Sparse, because `scripts/ci` is all it reads.
+      # This job had no checkout while its verdict was inline bash; the script
+      # needs one. Sparse, because `scripts/ci` is all it reads.
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          submodules: false
+          sparse-checkout: scripts/ci
+
       - name: Verify all test jobs passed
         shell: bash
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -840,18 +840,11 @@ jobs:
           FMT_CORPUS: ${{ needs.test-fmt-corpus.result }}
           OXC_FORMATTER_PROBE: ${{ needs.oxc-formatter-probe.result }}
           LANGUAGE_SERVER: ${{ needs.language-server.result }}
-        run: |
-          echo "test (shards): $BULK"
-          echo "test-unit: $UNIT"
-          echo "test-runtime: $RUNTIME"
-          echo "test-thread-safety: $THREAD_SAFETY"
-          echo "test-fmt-corpus: $FMT_CORPUS"
-          echo "oxc-formatter-probe: $OXC_FORMATTER_PROBE"
-          echo "language-server: $LANGUAGE_SERVER"
-          if [[ "$BULK" != "success" || "$UNIT" != "success" || "$RUNTIME" != "success" || "$THREAD_SAFETY" != "success" || "$FMT_CORPUS" != "success" || "$OXC_FORMATTER_PROBE" != "success" || "$LANGUAGE_SERVER" != "success" ]]; then
-            echo "::error::One or more test jobs did not succeed."
-            exit 1
-          fi
+        # Same exit codes as before; the script exists so the MESSAGE can say
+        # whether a leg failed or was merely cancelled, which reads identically
+        # in `gh pr checks` and is the difference between "tests are broken" and
+        # "this run produced no verdict".
+        run: node scripts/ci/tests-aggregate-verdict.mjs
 
   compatibility-report:
     name: Compatibility Report
@@ -1224,6 +1217,11 @@ jobs:
       # these are the controls that show it can still fail.
       - name: Run workflow trigger guard tests
         run: node scripts/ci/test-workflow-trigger-guard.mjs
+
+      # The `Tests` aggregator's verdict is a message nothing else reads, so it
+      # can only regress silently.
+      - name: Run Tests aggregator verdict tests
+        run: node scripts/ci/test-tests-aggregate-verdict.mjs
 
       - name: Run differential corpus workflow tests
         run: node scripts/ci/test-differential-corpus-workflow.mjs
--- a/scripts/ci/test-tests-aggregate-verdict.mjs
+++ b/scripts/ci/test-tests-aggregate-verdict.mjs
@@ -1,0 +1,110 @@
+#!/usr/bin/env node
+// Self-test for scripts/ci/tests-aggregate-verdict.mjs.
+//
+// The behaviour this file defends is a *message*, and a message is exactly the
+// kind of thing that regresses silently — nothing else in CI reads it. So each
+// case pins both the exit code and the discriminating phrase, and the pairs are
+// chosen so that a script which returned one constant verdict would fail here.
+//
+// The exit codes must stay identical to the inline bash this replaced: 0 only
+// when every leg succeeded, 1 for anything else. Cancellation is not a pass.
+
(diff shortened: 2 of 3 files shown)


baseballyama:
Correction to this PR's body, from a census posted on #3630.

The body says #3630's part 1 "is not explained by a stale workflow file, and I checked rather than assumed" — that was generalised from **one** sample (#3626) and it does not hold. Over all open PRs, 33 show `LSP full-population ratchet` as FAILURE, and **32 of those runs were created before the guard landed** (2026-08-22T14:29Z). The one created after it, #3591's, used a merge ref still carrying the pre-guard `if: always()`. #3671 and #3657, whose runs use today's file, both show the job SKIPPED.

So the guard on `main` is correct and part 1 needs no workflow change. Nothing in this PR changes — it was always about the aggregator's *message*, and the mixed-population point it makes still stands (a cancelled-run red and a broken-tests red are the same word in `gh pr checks`). Only the sentence about part 1's cause was wrong.

Issue #3396, opened 2026-08-22
Function-valued compile options: `runes` and `warningFilter` are silently ignored, `css`/`customElement`/`cssHash` are rejected


baseballyama:
## Summary

Upstream lets five compile options take a **function**: `warningFilter` and `cssHash` must be one
(`validate-options.js` `fun()`), and `css`, `customElement` and `runes` accept either a function or
a plain value (`parametric()`). Through rsvelte's NAPI `compile()`, none of the five works — two
are silently ignored and three are rejected.

Measured against a binding built from `origin/main` (`583756d9`, i.e. after #3362).

| option | official | rsvelte | |
|---|---|---|---|
| `runes: () => true` | evaluated → runes mode → rejects `export let` (`legacy_export_invalid`) | **compiles as legacy** | silently wrong output |
| `warningFilter: (w) => !w.code.startsWith('a11y')` | a11y warnings dropped | **all warnings kept** | silently ignored |
| `cssHash: ({hash, css}) => 'x-' + hash(css)` | `x-1w4ihw5` | throws: *"A function-valued `cssHash` cannot be called from this entry point; use `compileWithCssHash`…"* | deliberate, but a divergence |
| `css: () => 'injected'` | evaluated | throws: *"css should be either "external" … or "injected""* | function form unsupported |
| `customElement: () => true` | evaluated | throws: *"customElement should be true or false, if specified"* | function form unsupported |
| `cssHash: 'a-string'` | throws `options_invalid_value` | throws, but the error carries **no `code` field** | error shape |

## The two silent ones are the problem

`cssHash`, `css` and `customElement` at least fail loudly, and the `cssHash` message names the
supported entry point. `runes` and `warningFilter` produce a successful compile with the wrong
result:

- `runes: () => true` is how a project opts a subset of files into runes mode from
  `svelte.config.js`. Under rsvelte those files compile in legacy mode and the divergence surfaces
  later as wrong reactivity, not as an error.
- `warningFilter` is how a project suppresses warnings. Under rsvelte the suppression does nothing,
  so a build configured to be warning-clean is not.

## Documentation says otherwise

`apps/npm/compiler/README.md` currently advertises both of these as working through `compile()`:

```js
compile(source, {
  // Filter compiler warnings (applied by the compiler).
  warningFilter: (w) => !w.code.startsWith('a11y'),
  // A constant scope hash, or a dynamic `cssHash({ hash, css, name, filename })`.
[...]

Issue #3384, opened 2026-08-22
No gate varies `compilerOptions`: a 456-cell grid over the option axis returned 127 divergences on its first run


baseballyama:
## The axis

`scripts/compat-corpus/targets.mjs` defines four targets that vary **`generate` and `dev`**. Every
other field of `CompileOptions` is a constant across all of them — and therefore across the
output gate, the warning gates, the error gates, the parse gate, the matrix families, the mutation
fuzz, the svelte2tsx gates and the fmt gates, all of which draw their compile options from that
one place or from an equivalent hardcoded pair.

So "which compiler options is rsvelte compatible under" has never been measured. It is not an
exotic population: `compilerOptions` in `svelte.config.js` is forwarded to `compile()` by the Vite
plugin, and `accessors`, `immutable`, `css`, `customElement`, `preserveComments`,
`preserveWhitespace`, `namespace`, `hmr` and `discloseVersion` are all documented user-facing
options.

## First run

A throwaway grid — 19 option variants × 12 components × 2 generate targets = **456 cells** —
returned **127 diverging cells across 85 keys**, in a tree whose collected corpus is at 0 output
failures for `client`/`server`.

Filed from it:

| | option | what diverges |
|---|---|---|
| #3380 | `accessors` / `immutable` | warns on every compile; upstream warns once per process |
| #3382 | `customElement: true` | no registration, CSS not moved to `$$css`, state folded away |
| #3383 | no `filename` | named `Component` not `_unknown_`; dev drops `[$.FILENAME] = …` |
| #3226 | `css: 'injected'` | already filed — the injected string has ` ;}` where official has `}` |
| #3396 | function-valued `runes` / `warningFilter` / `css` / `customElement` / `cssHash` — two silently ignored, three rejected |

Options that came back clean against the same components: `runes: true`, `css: 'external'`,
`namespace: 'svg'`, `namespace: 'mathml'`, `preserveComments`, `preserveWhitespace`, both
together, `discloseVersion: false`, `fragments: 'tree'`, `sourcemap`/`outputFilename`/
`cssOutputFilename`, and `name`.

## Design note for whoever builds the gate

The grid needs a **`baseline: {}` row**, and it is not decoration. Two of the twelve components
diverge with no options set at all, and without the baseline row those two would have been read as
a divergence under all nineteen options — 38 of the 85 keys are that one artefact. Subtracting the
baseline row is what leaves the six real findings. This is the same shape as the
[...]

Issue #3382, opened 2026-08-22
`customElement: true` as a compiler option is ignored: no registration, CSS not moved to `$$css`, state folded away


baseballyama:
## Summary

`customElement: true` passed as a **compiler option** (no `<svelte:options customElement>` tag) is
ignored by rsvelte. Official compiles a custom-element component; rsvelte compiles a plain one.
Three things diverge at once.

```svelte
<script>
	let n = $state(1);
</script>
<b class="a">{n}</b>
<style>.a { color: red }</style>
```

```js
/* official, { customElement: true, generate: 'client' } — 21 lines */
const $$css = { hash: 'svelte-1lj1c2h', code: '.a.svelte-1lj1c2h {color:red}' };

export default function A($$anchor) {
	$.append_styles($$anchor, $$css);
	let n = $.state(1);
	var b = root();
	b.textContent = $.get(n);
	$.append($$anchor, b);
}

$.create_custom_element(A, {}, [], [], { mode: 'open' });

/* rsvelte — 12 lines */
export default function A($$anchor) {
	let n = 1;
	var b = root();
	b.textContent = '1';
	$.append($$anchor, b);
}
```

1. **No `$.create_custom_element(…)` registration** — the component is never defined as a custom
   element, so the option has no effect at all.
2. **CSS is not moved into `$$css`** — official returns `css.code === ''` and puts the CSS in the
   JS behind `$.append_styles`; rsvelte returns `css.code` as usual and emits no `append_styles`.
3. **State is not preserved** — official keeps `let n = $.state(1)` / `$.get(n)`, rsvelte folds it
   to `let n = 1` / `'1'`. A custom element's properties can be written from outside the
   component, so the "never written, do not wrap" fold is not valid here.

## Matrix

| case | official | rsvelte | |
|---|---|---|---|
| no option (control) | `let n = 1` / `'1'`, no registration | same | ok |
| `customElement: true`, no `<style>` | `$.state(1)` / `$.get(n)` + registration | `let n = 1` / `'1'` | **DIFF** |
| `customElement: true`, with `<style>` | + `$$css` / `append_styles`, `css.code === ''` | plain, `css.code` populated | **DIFF** |

The no-option row is the control that separates the fold from the option: both compilers fold
identically without `customElement`, so item 3 is the option changing official's decision and not
changing rsvelte's.

`customElement: 'my-el'` is rejected by both (`options_invalid_value`) — the string form is only
[...]

PR #3661, opened 2026-08-23
fix(napi): reject an unrecognised compile option and the six removed ones

### LINKED ISSUE #3535: NAPI compile options: an unknown key and the six `removed` options are silently ignored (no `options_unrecognised`, no `options_removed`)
## Summary

Upstream's `validate-options.js` rejects two whole classes of compile option that the NAPI
boundary accepts in silence:

| input | official | rsvelte NAPI |
|---|---|---|
| `compile(src, { nonsense: 1 })` | throws `options_unrecognised` — *"Unrecognised compiler option nonsense"* | **compiles, no warning** |
| `{ legacy: {} }` | throws `options_removed` | **compiles, no warning** |
| `{ format: 'esm' }` | throws `options_removed` | **compiles, no warning** |
| `{ tag: 'x-a' }` | throws `options_removed` | **compiles, no warning** |
| `{ sveltePath: 'svelte' }` | throws `options_removed` | **compiles, no warning** |
| `{ errorMode: 'warn' }` | throws `options_removed` | **compiles, no warning** |
| `{ varsReport: 'full' }` | throws `options_removed` | **compiles, no warning** |

Control, which moves: `{ hydratable: true }` is `warn_removed` upstream, and rsvelte does emit
`options_removed_hydratable` for it — so the removed-option machinery exists and reaches the
boundary; these seven are the ones with no field at all.

Measured against `origin/main` + #3467's branch (the fix there changes the *shape* of an option
error, not which options are checked), and official 5.56.10.

## Why a user cares

`options_unrecognised` is what turns a typo in `svelte.config.js`'s `compilerOptions` into an
error. Under rsvelte a misspelled option is accepted, ignored, and the build succeeds with the
default — the failure surfaces later as wrong output, or never.

## Two different difficulties, and they should not be conflated

**`options_removed` is the easy half.** The six names are fixed and known. They can be declared
as fields on `NapiCompileOptions` exactly the way `hydratable` / `enableSourcemap` /
`loopGuardTimeout` already are (those three exist *only* to raise their diagnostic), and raised
through the same `invalid_option`-shaped path. Upstream's are `removed()` rather than
`warn_removed()`, so they throw rather than warn — otherwise identical.

**`options_unrecognised` is structurally different.** `NapiCompileOptions` is
`#[napi(object)]`, whose generated `FromNapiValue` reads the fields it declares by name and
never enumerates the object. There is no key it does not declare for it to see, so this is not
a missing arm — it needs a different mechanism (enumerate the JS object's own keys at the
boundary and diff against the declared set, or take the options as a `Value` first).

Note the interaction with the option-alias surface: `cssHashOverride` is an rsvelte-only test
hatch and would have to be excluded from any such check, as would anything the vite plugin
forwards.

## Why no gate sees it

`scripts/dev/test-napi-compile-options.mjs` reconciles *declared* keys against covered ones —
its denominator is what `lib.rs` declares, so a key upstream has and rsvelte does not is
outside the denominator by construction. The compiler-error gates compare errors from
*compiling a source* and every one of them supplies a valid option set. And no gate varies
`compilerOptions` at all (#3384).

The `compiler-option` matrix family added for #3384 generates *valid* option values, so an
unrecognised key is not one of its variants either.

baseballyama:
Closes #3535.

## What was wrong

`NapiCompileOptions` is `#[napi(object)]`, whose derived `FromNapiValue` reads the fields it
declares **by name** and never enumerates the JavaScript object. There is therefore no key it
does not declare for it to see, so every key upstream's `validate-options.js` rejects reached the
compiler as silence: an unrecognised key, and the six Svelte-4 options upstream declares *only*
in order to reject.

The user-visible half is `options_unrecognised`. It is what turns a typo in `svelte.config.js`'s
`compilerOptions` into an error. Under rsvelte a misspelled option was accepted, ignored, and the
build succeeded with the default — the failure surfaced later as wrong output, or never.

## What the oracle actually does

Everything below is **read off a run**, not from the issue text. Official `VERSION`, asserted
from the compiler's own export rather than from a filename: **5.56.9** (`AGENTS.md` says 5.56.8
and the repo-root `node_modules` devDependency is 5.56.10; the submodule is the one every gate
here compiles against).

The issue treats one behaviour; there are three, and they differ on the entry point:

| input | `compile` | `compileModule` | `parse` |
|---|---|---|---|
| `{ nonsense: 1 }` | `options_unrecognised` | `options_unrecognised` | accepted |
| `{ nonsense: undefined }` | **`options_unrecognised`** | **`options_unrecognised`** | accepted |
| `{ legacy: {} }` and the other five | `options_removed` | **accepted, silently** | accepted |
| `{ legacy: undefined }` | **accepted** | accepted | accepted |
| `{ legacy: null }` | `options_removed` | accepted | accepted |
| `{ namespace: 'nope' }` | `options_invalid_value` | **accepted, silently** | accepted |
| `{ warningFilter: 'nope' }` | `options_invalid_value` | `options_invalid_value` | accepted |
| `{ experimental: { nope: 1 } }` | `options_unrecognised experimental.nope` | same | accepted |

Four things there are not guessable from the issue and each changed the fix:

1. **Presence, not definedness, for an unknown key.** `object()` walks `for (const key in input)`,
   so `{ nonsense: undefined }` throws. `removed()` is the opposite — it tests `input !== undefined`,
   so `{ legacy: undefined }` does not. The scan uses `napi_get_property_names`, which enumerates
   the same set `for…in` does; the `removed` arms test `is_some()`, which `#[napi(object)]` already
   maps `undefined` to `None`.
2. **`compileModule` is a different validator.** `validate_module_options` is `common_options` plus
   every *component* key mapped to a no-op, so a removed option is an error on `compile` and
   silence on `compileModule`, while an unrecognised key is an error on both and `warningFilter`
   is type-checked on both. One recognised-key list serves both entry points; only the six
   `common_options` are validated on the module one.
3. **`parse()` validates nothing**, upstream and here. It is the control that must not move.
4. **`css` and `customElement` are `parametric()`, and their normalizer runs on the first CALL,
   not during validation.** So upstream reports them only after every plain validator has passed.
   Measured: `{css:'none', varsReport:'x'}` → `options_removed`; `{css:'none', name:3}` → `name`;
   `{customElement:'x-a', css:'none'}` → `customElement`. My first draft put both at their
   declaration position, which would have **regressed** those cells — the grid caught it before
   the commit, not review.

## The fix

- A key scan at the boundary. `NapiCompileOptionsArg` / `NapiModuleCompileOptionsArg` wrap the
  derived decoders and read the object's own key list alongside the fields, so the conversion can
  raise `options_unrecognised` **first** — the position upstream's `object()` reports it from.
  `RECOGNISED_COMPILE_OPTIONS` is the list, and it is reconciled against the two structs' declared
  fields **in both directions** by `test:napi-options` (a key dropped from it would make the
  boundary reject an option the compiler still supports — the failure mode that hurts).
- Seven new declared fields: the six `removed()` options, raising `options_removed` with upstream's
  exact prose, plus `warningFilter`, type-checked like upstream's `fun()`. `warningFilter` **had** to
  be declared — without it the new scan would have rejected every call the vite plugin makes.
- `into_compile_options` now runs in upstream's key-declaration order, with `customElement` and then
  `css` last for the `parametric()` reason above.
- `css: 'none'` gets upstream's own message instead of the generic one.
- Nested `object()` children (`experimental.*`, `compatibility.*`) report `options_unrecognised`
  under their own keypath.

## The grid

**option × value × entry point**, official vs the raw NAPI addon, every cell in its own process
because upstream's `warn_once` latches are per-process and a shared one would score the second
cell as silent. 69 option variants × 3 shared entry points = 207 cells, plus 3 rsvelte-only
entries. The key is **the thrown error and the warning set as well as the output**, because an
ignored option produces no error, no warning and often identical output — a key of `(output)`
cannot see this defect at all.

| verdict | before | after |
|---|---|---|
| `match` (both compiled; same warnings, js, css) | 133 | 133 |
| `both-reject` (same `code` **and** same `message`) | 31 | **63** |
| `both-reject-msg` (same code, different prose) | 2 | **0** |
| `both-reject-code` | 6 | **2** |
| **`silent`** (official rejects, rsvelte compiles) | **29** | **3** |
| `rs-rejects` (official compiles, rsvelte rejects) | 3 | 3 |
| `output-differs` | 2 | 2 |
| `warn-differs` | 1 | 1 |
| **divergent (everything but match/both-reject)** | **43** | **11** |

The verdict is not one bit, for #3384's reason: collapsing `silent` and `rs-rejects` into
"divergent" would score this PR as *no change at all* on the cells it converts.

**All 11 remaining cells are accounted for and none is attributable to #3535:**

| cells | what | status |
|---|---|---|
| 5 | function-valued `css` / `customElement` / `cssHash` / `warningFilter` / `runes` | **#3396**, open, not closed here |
| 1 | `css: 'injected'` output | **#3226**, open |
| 1 | `customElement: true` output | **#3382**, open |
| 2 | `cssHashOverride` | rsvelte-only by design — official rejects it as unrecognised. Asserted as an rsvelte-only row rather than a shared one, with the oracle's refusal pinned so the row cannot silently become a parity claim |
| 2 | `filename: 3` | official throws a bare `TypeError` with `code: undefined` from `state.reset` **before** its own `string()` validator runs; rsvelte throws the coded `options_invalid_value` that validator specifies. rsvelte is the more faithful of the two |

Per entry point after the fix: `parse` **69/69 match** (the control — neither side validates);
`compile` and `compileBoth` identical to each other; `compileModule` 49 match + 18 both-reject.

## Controls

Every row below was measured against the oracle before being asserted, and every one stayed green
through the revert control.

| control | why it is here | before | after |
|---|---|---|---|
| `parse` with each of the 69 variants | upstream validates nothing here; the scan must not leak into it | 69 match | 69 match |
| `{ legacy: undefined }`, `{ format: undefined }` | `removed()` tests `!== undefined`; an eager `is_some()` would reject these | accepted | accepted |
| `compileModule` with all six removed options, `namespace:'nope'`, `css:'none'`, `name:1`, `compatibility:{nope:1}` | the module validator declares component keys as no-ops | accepted | accepted |
| `warningFilter: (w) => …`, `sourcemap: <v3 map>` | the vite plugin forwards both; the scan must not reject them | accepted | accepted |
| `hydratable` / `enableSourcemap` / `loopGuardTimeout` | `warn_removed()` warns, it must not throw | warn | warn |
| the 27 pre-existing `INVALID_OPTIONS` rows and every discrimination case | no regression in what already worked | green | green |
| `test:vps-shim` (62), `test:napi-csshash` (29), `test:vps-fixture` (6), `test:vps-unit` (32), `test:envelope`, `test:parse-envelope` | the shim pre-resolves function options and deletes `warningFilter`; these are the consumers the new refusal could break | green | green |

Three precedence rows — `prec-name-beats-css`, `prec-name-beats-customElement`,
`prec-customElement-beats-css` — **passed before the fix too**. They are controls, not evidence:
the old order happened to agree with upstream on them. The rows that discriminate the reorder are
`prec-removed-beats-namespace` and `prec-varsReport-beats-css`, and both were red.

## Revert control

Pre-fix binary, post-fix tests: **29 failures, 0 control failures**, each for the predicted reason.
A sample of the actual text:

```
FAIL invalid.unknown-key: rsvelte rejects it too — rsvelte compiled it
FAIL invalid.removed-varsReport: rsvelte rejects it too — rsvelte compiled it
FAIL invalid.css-none: message — official "…css: \"none\" is no longer a valid option…"
                                 vs rsvelte "…css should be either \"external\" … or \"injected\"…"
FAIL invalid.prec-unknown-beats-dev: code — official "options_unrecognised" vs rsvelte "options_invalid_value"
FAIL invalid.prec-varsReport-beats-css: code — official "options_removed" vs rsvelte "options_invalid_value"
FAIL module.invalid.warningFilter-string: rsvelte rejects it too — rsvelte compiled it
```

Restoring the fixed artifact returns `400 passed, 0 failed`.

## Where the coverage lives, and why there is no `pattern-corpus` repro

**A `pattern-corpus` file cannot carry a compile option.** That corpus compiles every entry with
`scripts/compat-corpus/targets.mjs`'s fixed targets, which vary `generate` and `dev` and nothing
else (#3384) — there is no place in a `.svelte` file to say `{ nonsense: 1 }`. The coverage is
therefore in `scripts/dev/test-napi-compile-options.mjs`, the gate that already owns this
boundary, which gains: the recognised-key reconciliation, 26 new `INVALID_OPTIONS` rows compared
on `code` + `name` + `filename` + `message`, 10 new `VALID_OPTIONS` rows, an rsvelte-only
`cssHashOverride` row that also pins the oracle's refusal, and a **new `compileModule` half** —
the entry-point axis, which no gate had.

## What this does *not* close

**#3396 stays open, deliberately.** Its five function-valued options are five of the eleven
remaining cells. Two of them (`runes`, `warningFilter`) are still accepted and ignored by the raw
addon. Making them refuse is a behaviour decision about a published entry point, and it interacts
with this PR only in that `warningFilter` now has to be *declared* — so the decision is now one
`match` arm away, in a place a gate watches, instead of absent.

**#3384 stays open.** This PR gates the option axis at the NAPI boundary; it does not add the
`compiler-option` matrix family that would cross option × component *shape* over the corpus
targets. Its own lesson applies to what I did here, though, and I applied it: the grid's source
carries an `` precisely so `warningFilter` has a warning to filter — without it
that row is green whether or not the callback is honoured.

## Separate defects found, filed rather than folded in

Four, none of them touched by this PR: three rsvelte-only NAPI entry points that answer
differently from `compile()`, and a release-plumbing gap. Filed as their own issues; see the
comment below.

### DIFF
--- a/.changeset/napi-unrecognised-and-removed-options.md
+++ b/.changeset/napi-unrecognised-and-removed-options.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/vite-plugin-svelte-native": patch
+---
+
+Reject the compile options upstream rejects. `compile()` and `compileModule()` now raise `options_unrecognised` for a key `validate-options.js` does not declare — so a typo in `svelte.config.js`'s `compilerOptions` is an error instead of a silently-ignored default — and `compile()` raises `options_removed` for the six Svelte-4 options upstream declares only in order to reject (`legacy`, `format`, `tag`, `sveltePath`, `errorMode`, `varsReport`). `warningFilter` is now declared and type-checked like upstream's `fun()`, `css: 'none'` carries upstream's own message rather than the generic one, and when two options are bad the one reported is the one upstream reports: the unrecognised-key scan runs before every validator, the rest run in upstream's key-declaration order, and `customElement`/`css` come last because their `parametric()` normalizer runs on first call rather than during validation. `compileModule` follows upstream in accepting every component-only key as a no-op, so a removed option is an error there only on `compile`.
--- a/crates/rsvelte_napi/src/lib.rs
+++ b/crates/rsvelte_napi/src/lib.rs
@@ -360,7 +360,7 @@ fn position_object<'env>(
 pub fn napi_compile(
     env: Env,
     source: String,
-    options: Option<NapiCompileOptions>,
+    options: Option<NapiCompileOptionsArg>,
 ) -> napi::Result<Value> {
     let opts = options_to_compile(Some(&env), options)?;
     let filename = opts.filename.clone();
@@ -386,7 +386,7 @@ pub fn napi_compile(
 pub fn napi_compile_both(
     env: Env,
     source: String,
-    options: Option<NapiCompileOptions>,
+    options: Option<NapiCompileOptionsArg>,
 ) -> napi::Result<Value> {
     let opts = options_to_compile(Some(&env), options)?;
     let filename = opts.filename.clone();
@@ -452,7 +452,7 @@ fn compile_result_to_json(result: rsvelte_core::compiler::CompileResult) -> Valu
 #[napi(js_name = "compileWithCssHash", catch_unwind, ts_return_type = "any")]
 pub async fn napi_compile_with_css_hash(
     source: String,
-    options: Option<NapiCompileOptions>,
+    options: Option<NapiCompileOptionsArg>,
     #[napi(
         ts_arg_type = "(input: { hash: (str: string) => string, css: string, name: string, filename: string }) => string"
     )]
@@ -463,7 +463,7 @@ pub async fn napi_compile_with_css_hash(
     // way the synchronous entries do.
     let mut options = options;
     if let Some(o) = options.as_mut() {
-        o.css_hash = None;
+        o.inner.css_hash = None;
     }
     // An `async` export's arguments must be `Send`; `Env` is not, so this is
     // the one entry whose option failure cannot carry the coded shape.
@@ -893,6 +893,26 @@ fn invalid_option(detail: impl std::fmt::Display) -> OptionError {
     }
 }
 
+/// `validate-options.js`'s `removed()`: an option that still has a name but no
+/// behaviour. Unlike `warn_removed()` (which only warns) this throws.
+fn removed_option(detail: &str) -> OptionError {
+    OptionError {
+        code: Some("options_removed"),
+        message: format!("Invalid compiler option: {detail}\nhttps://svelte.dev/e/options_removed"),
+    }
+}
+
+/// `validate-options.js`'s `object()` reports a key it does not declare, before
+/// running any per-option validator.
+fn unrecognised_option(keypath: &str) -> OptionError {
+    OptionError {
+        code: Some("options_unrecognised"),
+        message: format!(
+            "Unrecognised compiler option {keypath}\nhttps://svelte.dev/e/options_unrecognised"
+        ),
+    }
+}
+
 /// A value upstream accepts that this entry point cannot honour. There is no
 /// upstream code for it because upstream never raises it.
 fn unsupported_option(detail: impl std::fmt::Display) -> OptionError {
@@ -996,6 +1016,9 @@ fn coerce_css(v: &LenientScalar) -> OptionResult<CssMode> {
         LenientScalar::Str(s) => match s.as_str() {
             "external" => Ok(CssMode::External),
             "injected" => Ok(CssMode::Injected),
+            "none" => Err(invalid_option(
+                "css: \"none\" is no longer a valid option. If this was crucial for you, please open an issue on GitHub with your use case.",
+            )),
             _ => Err(invalid_option(
                 "css should be either \"external\" (default, recommended) or \"injected\"",
             )),
@@ -1040,6 +1063,7 @@ fn coerce_experimental(v: &LenientScalar) -> OptionResult<ExperimentalOptions> {
     if !v.is_object() {
         return Err(invalid_option("experimental should be an object"));
     }
+    reject_unrecognised_child(v, "experimental", &["async"])?;
     let mut exp = ExperimentalOptions::default();
     if let Some(a) = v.field("async") {
         exp.r#async = coerce_bool("experimental.async", a)?;
@@ -1051,10 +1075,25 @@ fn coerce_compatibility(v: &LenientScalar) -> OptionResult<rsvelte_core::compile
     if !v.is_object() {
         return Err(invalid_option("compatibility should be an object"));
     }
+    reject_unrecognised_child(v, "compatibility", &["componentApi"])?;
     v.field("componentApi").map_or_else(
         || Ok(rsvelte_core::compiler::ComponentApi::default()),
         coerce_component_api,
     )
+}
+
+/// A nested `object()` validator reports an unknown key under its own keypath,
+/// at the point the parent walks it rather than before every other option.
+fn reject_unrecognised_child(v: &LenientScalar, keypath: &str, known: &[&str]) -> OptionResult<()> {
+    let LenientScalar::Object(fields) = v else {
+        return Ok(());
+    };
+    fields
+        .iter()
+        .find(|(k, _)| !known.contains(&k.as_str()))
+        .map_or(Ok(()), |(k, _)| {
+            Err(unrecognised_option(&format!("{keypath}.{k}")))
+        })
 }
 
 /// Typed mirror of `CompileOptions` for the NAPI boundary.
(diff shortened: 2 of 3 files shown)


baseballyama:
Four separate defects fell out of the grid and are filed rather than folded in:

- **#3662** — `compileBuffers` throws a `{:?}`-formatted Rust struct with `code: 'GenericFailure'`
  for every compile failure. That is exactly the shape #3536 fixed for `compile()`; the option
  *parsing* is shared (one `options_to_compile`, so this PR reaches every entry point), but the
  *error rendering* is per-entry-point and only one entry point has a gate. Same entry point also
  refuses `modernAst`, which official accepts everywhere.
- **#3663** — `compileWithCssHash` clears `cssHash` unconditionally before validating, so a
  *non-function* `cssHash` is swallowed where official raises `options_invalid_value`. Same
  accepted-in-silence class as this PR, on the one entry point this PR's fix cannot reach because
  the field is gone before the validator runs. One-line fix, deliberately not taken here.
- **#3664** — `validate-options.js` is ported **three** times (`rsvelte_napi`, `rsvelte_capi`,
  `rsvelte_lint_bindings/compiler_wasm`) and only the NAPI port is compared to upstream by
  anything. The three have drifted in *different* directions — the wasm port validates nothing
  (`namespace: 'nope'` silently becomes `html`) but is the only one that honours function-valued
  `runes`/`css`/`customElement`/`warningFilter`; the C ABI rejects `generate: 'dom'`, which
  official accepts.
- **#3665** — release plumbing: `crates/rsvelte_napi` ships only in
  `@rsvelte/vite-plugin-svelte-native`, which is its own `fixed` group with no cascade edge, and
  `check-core-consumer-changesets.mjs` has no rule for the crate. #3536's still-pending changeset
  names `@rsvelte/compiler`, which does not contain the code — so on the next release the coded
  `CompileError` fix would be built, staged, and not published. This PR's changeset names the
  right package.

One measurement worth recording separately, because it is not a defect on this side:
`{ filename: 3 }` makes **official** throw a bare `TypeError` (`code: undefined`) out of
`state.reset` *before* its own `string()` validator runs, while rsvelte throws the coded
`options_invalid_value` that validator specifies. Those two cells are counted as divergent in the
table above and rsvelte is the more faithful of the two; I have not filed it upstream.

Issue #3669, opened 2026-08-23
`export {};` in a `<script module lang="ts">` survives into the generated output


baseballyama:
`export {};` in a `<script module lang="ts">` survives into the generated output

Oracle: `submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'` (read from the compiler's own export). rsvelte at `f22ea71e7` plus the #3566 fix.

## Repro

```svelte
<script module lang="ts">
export {};
</script>
<script lang="ts">
let k = 1;
</script>
{k}
```

official (`generate: 'client'`, prod):
```js
import 'svelte/internal/disclose-version';
import 'svelte/internal/flags/legacy';
import * as $ from 'svelte/internal/client';

export default function C($$anchor) {
```

rsvelte:
```js
import 'svelte/internal/disclose-version';
import 'svelte/internal/flags/legacy';
import * as $ from 'svelte/internal/client';

export {};

export default function C($$anchor) {
```

Diverges on both targets, dev and prod — 4 cells. The instance-script spelling of the same source
(`<script lang="ts">export {};</script>`) is byte-identical after the #3566 fix, which is what
isolates this to the module-script emission path.

## Cause

Upstream's `ExportNamedDeclaration` visitor in `remove_typescript_nodes.js` empties an export whose
specifier list filters to nothing — `if (specifiers.length === 0) return b.empty` — and that
includes an export written with none. rsvelte's typed strip now does the same, so the statement is
an `EmptyStatement` in the module program; but the `<script module>` body reaches the output as
**source text**, not from the stripped program, so the original `export {};` is spliced back in.

That makes it an instance of the class `AGENTS.md` records for the client instance-script pipeline
— a text path deciding what the AST path already decided — one script kind over.

## Note on gates

The corpus comparison normalizes blank lines away but not statements, so this WOULD be visible to
the output-equality gate if any corpus source carried the shape. `export {}` in a `<script module>`
is a TypeScript idiom (forcing module scope) that no current corpus source happens to use.

## Found by

The declaration-form matrix in #3566's fix, `empty-export|plain|module-script` row.

Issue #3668, opened 2026-08-23
`export declare global { … }` compiles in rsvelte where official raises `js_parse_error`


baseballyama:
`export declare global { … }` compiles in rsvelte where official raises `js_parse_error`

Oracle: `submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'` (read from the compiler's own export). rsvelte at `f22ea71e7`.

## Repro

```svelte
<script lang="ts">
export declare global { interface W {} }
let k = 1;
</script>
{k}
```

official: `js_parse_error`.
rsvelte: compiles, emitting `export default function C($$anchor) { … }`.

Diverges in an instance script and in `<script module>`, on both targets, dev and prod — 8 cells.
The un-exported `declare global { interface W {} }` is accepted by both and is byte-identical, so
only the `export` spelling is affected.

## Cause

`export` on a global augmentation is not legal TypeScript (TS1038 / "a 'declare' modifier cannot
be used in an already ambient context" family). acorn-typescript rejects it and upstream surfaces
that as `js_parse_error`; oxc accepts it, so rsvelte proceeds and strips the declaration.

This is an **under**-rejection: rsvelte accepts a program official refuses. It is the mirror of
the over-rejection class the `bind:` / `param-default` matrix families exist for, and the same
gate blind spot applies — a collected corpus cannot contain it, because published code compiles.

## Found by

The declaration-form matrix in #3566's fix, `declare-global|export` row (29 forms ×
`export`/no-`export` × 3 entry points × 2 targets × dev/prod).

Issue #3667, opened 2026-08-23
An `import` inside a TypeScript `namespace` is a `js_parse_error` in rsvelte where official compiles it


baseballyama:
An `import` inside a TypeScript `namespace` is a `js_parse_error` in rsvelte, where official compiles it (or rejects it with a different code)

Oracle: `submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'` (read from the compiler's own export). rsvelte at `f22ea71e7` plus the #3566 fix — this is a **parser**-level divergence, unaffected by that change.

## Repro

```svelte
<script lang="ts">
namespace N { import type { A } from 'm'; }
let k = 1;
</script>
{k}
```

| source | official | rsvelte |
|---|---|---|
| `namespace N { import type { A } from 'm'; }` | compiles (the import is type-only, so `remove_typescript_nodes` empties it and the namespace strips) | `js_parse_error: Import declarations in a namespace cannot reference a module.` |
| `namespace N { import 'm'; }` | `typescript_invalid_feature` (the import is a value, so the namespace is not type-only) | `js_parse_error`, same message |

Both spellings diverge in an instance script and in `<script module>`, on both targets, dev and
prod — 32 cells.

## Cause

`Import declarations in a namespace cannot reference a module` is a **TypeScript semantic**
diagnostic (TS1147) that oxc raises during parsing. acorn-typescript, which upstream parses with,
does not implement it, so upstream reaches `remove_typescript_nodes` and answers on the namespace
body instead.

The first row is an over-rejection: valid, published-shaped TypeScript that upstream compiles.
The second is an error-**code** divergence — both reject, and a corpus error ratchet keyed on
`code` would see it.

## Scope

Every other TS1xxx-class diagnostic oxc raises that acorn-typescript does not is the same shape.
This one was found by enumerating what can appear inside a namespace body; the family has not been
swept.

## Found by

The declaration-form matrix in #3566's fix, `nsbody-import-type` and `nsbody-bare-import` rows.

Issue #3666, opened 2026-08-23
`export import x = require('m')` in an instance script adds a `$$props` parameter official does not


baseballyama:
`export import x = require('m')` in an instance script adds a `$$props` parameter official does not — and the un-exported form is hoisted out of the component function.

Oracle: `submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'` (read from the compiler's own export). rsvelte at `f22ea71e7`.

## Repro

```svelte
<script lang="ts">
export import ie = require('m');
let k = 1;
</script>
{k}
```

official (`generate: 'client'`, prod):
```js
export default function C($$anchor) {
	import ie = require('m');
	…
```

rsvelte:
```js
export default function C($$anchor, $$props) {
	export import ie = require('m');
	…
```

The `$$props` parameter is a change to the component's **calling signature**, not a byte
difference, and it is present on `client` and `server` in prod. In dev both compilers emit
`$$props` anyway, so the divergence is prod-only — a matrix that only varies `dev` cannot see it.

## Cause

`convert_declaration_for_program` (`1_parse/read/expression.rs`) has no
`Declaration::TSImportEqualsDeclaration` arm, so the declaration falls to `_ => Value::Null` and
becomes a `JsNode::Null`. `strip_export_named_declaration_typed` empties an export whose
declaration strips to an `EmptyStatement`; a `Null` is neither, so the `ExportNamedDeclaration`
survives the TypeScript strip, `process_legacy_exports` counts it, and `analysis.needs_props`
is set.

oxc's `ExportDeclaration::export_kind()` returns `Value` for `TSImportEqualsDeclaration`
(`declare()` is `false` and `is_type()` is `false`), so the `exportKind === 'type'` short-circuit
that covers `export type` / `export interface` / `export declare …` does not fire here. This is
the last declaration form in that `_` arm that is reachable with a non-type `exportKind`.

## The awkward part

**Both compilers emit output no JS parser accepts for this form.** Over the 29-form matrix, all
24 unparseable outputs are this shape: official produces 16 (it leaves `import ie = require('m')`
inside the component function body on every target) and rsvelte 8. So "match official" is not
available as a specification here — official's own answer is not JavaScript. Fixing the `$$props`
half is still well-defined; deciding what to *print* needs an upstream decision first.

The un-exported spelling is a byte divergence of the same family: rsvelte hoists
`import ie = require('m')` to module scope while official keeps it in the function body. Both are
unparseable.

## Found by

[...]

PR #3679, opened 2026-08-23
fix(parse): judge a named export empty only after its declaration is stripped

### LINKED ISSUE #3566: `export namespace N { …type-only… }` in an instance script adds a `$$props` parameter official does not
インスタンススクリプトの `export namespace N { <型だけの本体> }` が、公式には無い `$$props` 仮引数を rsvelte のコンポーネントに付けます。**TypeScript だけの宣言に付いた `export` が、コンポーネントの export（＝プロップ）として数えられています。**

## 再現

```svelte
<script lang="ts">
export namespace N { type T = 1; }
let k = 1;
</script>
{k}
```

オラクル = `submodules/svelte` @ `20b341f10048`、`VERSION === '5.56.9'`
（`git ls-tree origin/main submodules/svelte` と一致 = CI が読むツリー）。

公式:
```js
export default function C($$anchor) {
```

rsvelte（`origin/main` = c8a67c5aa、および PR #3417 のブランチ — 出力はバイト一致）:
```js
export default function C($$anchor, $$props) {
```

本体はどちらも同一で、**シグネチャだけが違います**。

## 軸: `export` の対象が型だけの宣言かどうか

`export` の付く TypeScript 宣言 8 種を instance script と `compileModule` の 2 入口 ×
36 本体で回した結果、乖離するのは **`export namespace` の 2 行だけ**です:

| body（instance script） | official | rsvelte |
|---|---|---|
| `export type T = 1;` | `$$props` 無し | 一致 |
| `export interface I {}` | `$$props` 無し | 一致 |
| `export declare const c: number;` | `$$props` 無し | 一致 |
| `export declare namespace N { type T = 1; }` | `$$props` 無し | 一致 |
| **`export namespace N { type T = 1; }`** | `$$props` 無し | **✗ `$$props` あり** |
| **`export namespace N { interface I {} }`** | `$$props` 無し | **✗ `$$props` あり** |
| `export namespace N { let x = 1; }` | 両者 reject（`typescript_invalid_feature`） | 一致 |

つまり `export type` / `export interface` / `export declare …` は正しく「型だけ」と
判定されていて、**`export namespace` だけが判定から漏れています**。
`declare` が付くと（`export declare namespace`）正しくなるのも同じ穴の裏返しです。

## 影響

`$$props` の有無はコンポーネントの**呼び出しシグネチャ**なので、バイト差ではなく
生成物の API 差です。出力はパースも実行もできるため、パースゲートからは見えません。

## 見つけ方（次の人向け）

PR **#3417**（`fix/3244-typescript-invalid-feature`）と、それに包含される
`fix/3217-3243-3244-js-early-errors` ブランチを**両方ビルドして 72 セル（36 本体 × 2 入口）
を公式と突き合わせた**ときの副産物です。両ツリーの出力は 72 行とも**バイト一致**なので、
これは **#3417 の退行ではなく、どちらも直していない既存欠陥**です。

同じ 72 セルで見つかったもう 1 つは #3568（公式側が生の `TypeError` で落ちる形）。

baseballyama:
Closes #3566. Also carries the two open halves of #3568 (an `upstream_issues/` write-up and a test
that pins rsvelte's chosen behaviour) — the fix touches it, and the choice is stated below rather
than derived.

## The regression window matters, because #3566 did not reproduce when it was filed

Measured against the pinned oracle (`submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'`
asserted from the compiler's own export, not from a path or a lockfile):

- on `f22ea71e7` — `main` when I started — `export namespace N { type T = 1; }` produced
  **byte-identical** output on client/server × dev/prod × instance/`<script module>`;
- on `5e5fe6b8b` (**#3417**, merged mid-task) it reproduces exactly as the issue describes.

That is not a contradiction, it is the cause. Before #3417 the parse conversion collapsed an
exported `TSNamespaceDeclaration` into an `EmptyStatement` *before* `remove_typescript_nodes` ever
saw it, so the export was already empty when the strip judged it. #3417 made the body survive parse
— which it must, so a value inside the namespace can be rejected — and that removed the accident
this code was relying on. #3566 is the defect #3417 uncovered, not one #3417 introduced.

The harness was not blind before the window either: driven over the same matrix on `f22ea71e7` it
reported `export import ie = require('m')` producing exactly the spurious-`$$props` divergence
#3566 describes, so it could see the class when it was there.

## Cause

Upstream's `ExportNamedDeclaration` visitor decides **after** it visits:

```js
const result = context.next();
if (result?.declaration?.type === 'EmptyStatement') return b.empty;
```

`strip_export_named_declaration_typed` did it the other way round: it checked whether the
declaration was *already* an `EmptyStatement`, filtered specifiers, and only then recursed. So an
export whose declaration becomes empty **during** the visit — every type-only declaration form once
its body survives parse — reached `process_legacy_exports`, which sets `analysis.needs_props` for
any surviving `ExportNamedDeclaration`, and the component gained a `$$props` parameter.

`$$props` is the component's calling signature, so this is an API divergence, not a byte
divergence: the output parses, runs, and in **dev mode both compilers emit `$$props` regardless**,
so only a production target discriminates. A matrix that varies `dev` but not the declaration form
cannot see it.

## Fix

- **`remove_typescript_nodes.rs`** — recurse into the declaration first, then judge the export
  empty, mirroring upstream's order. The specifier half of the same visitor comes with it: an
  export whose specifier list filters to nothing — including one written with none, `export {}` —
  is empty, as upstream's `if (specifiers.length === 0) return b.empty` says.
- **`1_parse/read/expression.rs`** — a dotted `namespace N.M { … }` is nested as its desugaring
  `namespace N { namespace M { … } }` instead of having its body dropped at parse (see #3568
  below).

## Measured, whole matrix

The axis is the declaration **form**, not the `namespace` keyword: 29 forms
(`namespace` / `module` / `declare namespace` / `declare module` / `declare global` / `interface` /
`type` / `enum` / `const enum` / `abstract class` / `declare const|function|class` /
`import type` / `import x = require()` / `export {}` / dotted spellings / 12 namespace-**body**
forms) × `export` / no-`export` × {instance script, `<script module>`, `compileModule` on a
`.svelte.ts`} × {client, server} × {dev, prod} = **1,960 compared pairs**. Every expectation is
read off the oracle; none is written from reasoning about what upstream "should" do. Every accepted
output on both sides is run through acorn.

| | before | after |
|---|---|---|
| compared pairs | 1,960 | 1,960 |
| verdict divergences | **128** | **82** |
| byte-only divergences (same verdict) | 22 | 22 |
| unparseable outputs | 24 | 24 |

**46 cells fixed, 0 regressed** (set-differenced per cell, not compared as totals). The 24
unparseable outputs are all `import x = require('m')` — 16 official, 8 rsvelte — unchanged in both
directions and reported as #3666.

Fixed, by form (all in a production target; the dev cells were already equal because both
compilers emit `$$props` there):

| form | before → after |
|---|---|
| `export namespace N { type T = 1 }` | `fn($$anchor, $$props)` → `fn($$anchor)` |
| `export namespace N { interface I {} }` | same |
| `export namespace N { export type T = 1 }` | same |
| `export namespace N { namespace M { type T = 1 } }` | same |
| `export namespace N { }` | same |
| `export namespace N { declare const/function/class … }` | same |
| `export module N { type T = 1 }` | same |
| `export {}` (instance) | same |
| `namespace N { export {} }`, all 4 spellings × both entry points | `ERR typescript_invalid_feature` → `fn(…)` |

## #3568 — the decision, stated

Official crashes on **every** dotted spelling with a raw
`TypeError: node.body.body.map is not a function`: `e.code` is `undefined`, there is no `start`, so
it is not a Svelte diagnostic. `remove_typescript_nodes.js` assumes `node.body` is a
`TSModuleBlock`; for a dotted name it is another `TSModuleDeclaration`.

rsvelte does **not** reproduce the crash, and that is a choice. `namespace N.M { … }` is the source
spelling of `namespace N { namespace M { … } }`, and upstream compiles the nested spelling
correctly — a type-only body strips, and a value in it raises a coded `typescript_invalid_feature`
positioned on the inner namespace. rsvelte therefore treats the dotted form **as its desugaring**,
so both halves of upstream's own behaviour carry over. Reproducing the crash was the alternative and
was rejected: an uncoded exception has no `code` and no span for the error ratchets to compare, and
every embedder (`rsvelte-check`, the language server, the Vite plugin) would surface a panic instead
of a diagnostic.

This is a *tightening*, not a new permission. Before it, rsvelte dropped the dotted body at parse
without looking at it, so `namespace N.M { let x = 1; }` compiled — silently **more** permissive
than the desugaring. It now raises `typescript_invalid_feature`, which is what the un-dotted
spelling gets from upstream.

Recorded in `compatibility/deliberate-divergences.md` with the row-by-row measurement, reported in
`upstream_issues/3568-svelte-dotted-namespace-crash.md`, and pinned by
`crates/rsvelte_core/tests/ts_export_type_only_declaration.rs` so a later reader does not "fix"
rsvelte towards the crash.

**A correction to #3568's own text.** It offers "the same `namespace N.M` passes through
`compileModule` on the official side" as evidence that the crash is instance-script-specific. That
control does not hold on the pinned oracle: `compileModule` parses with `typescript: false`
(`2-analyze/index.js` → `parse(source, comments, false, false)`), so **every** TypeScript
declaration form — `type`, `interface`, `namespace`, dotted or not — is a `js_parse_error` there,
measured over all 29 forms × 2 export spellings. The visitor is never reached from that entry
point, so it says nothing about whether the crash is path-specific.

## Controls

Written to fail if the fix over-reaches, i.e. if "empty every export" would pass:

| control | asserted |
|---|---|
| `export let v = 1` | keeps `fn($$anchor, $$props)` / `fn($$renderer, $$props)` |
| `export const kk = 1` | keeps it |
| `export function fn() {}` | keeps it |
| `export class Cls {}` | keeps it |
| `const w = 1; export { w }` | keeps it |
| `const w = 1; export { w as renamed }` | keeps it |
| `export namespace N { type T = 1 }` **beside** `export let v = 1` | keeps the parameter **and** still emits `$.prop(` |
| `namespace N.M { let x = 1 }` | `typescript_invalid_feature` (the dotted form did **not** become a blanket accept) |

**Revert control.** With both source files reverted to the base and the new tests kept, the tests go
red for the predicted reason:

```
running 5 tests
test a_type_only_export_beside_a_real_one_keeps_the_real_one ... ok
test an_exported_type_only_declaration_adds_no_props_parameter ... FAILED
test a_dotted_namespace_with_a_value_is_rejected ... FAILED
test a_dotted_namespace_strips_when_its_body_is_type_only ... ok
test a_real_export_keeps_its_props_parameter ... ok

---- an_exported_type_only_declaration_adds_no_props_parameter stdout ----
panicked at crates/rsvelte_core/tests/ts_export_type_only_declaration.rs:125:9:
`export namespace N { type T = 1; }` produced a props parameter:
...
export default function C($$anchor, $$props) {
	let k = 1;
...

---- a_dotted_namespace_with_a_value_is_rejected stdout ----
panicked at crates/rsvelte_core/tests/ts_export_type_only_declaration.rs:85:29:
`namespace N.M { let x = 1; }` compiled (instance, Client, dev=false); expected `typescript_invalid_feature`:

test result: FAILED. 3 passed; 2 failed; 0 ignored
```

and the three controls stay green.

**One of the four new tests does not measure the fix, and that is worth saying out loud.**
`a_dotted_namespace_strips_when_its_body_is_type_only` passes on the **reverted** tree too, because
the old code dropped the dotted body wholesale — so its stripping was accidental rather than
derived. Only `a_dotted_namespace_with_a_value_is_rejected` discriminates that half of the change.
A reader counting four passing tests would otherwise assume all four are load-bearing.

Restored afterwards; `cargo fmt --all -- --check` is clean on the restored tree.

## Deliberately not done

- **The dotted form is not made to crash.** Stated above; it is the one place here where rsvelte's
  behaviour is a decision rather than a measurement.
- **`declare global` is left alone** (#3668), as are the namespace-`import` rows (#3667), the
  `import x = require()` rows (#3666) and the `<script module>` `export {}` residue (#3669). Each is
  a different check from the one this PR moves — I confirmed that by cell, not by inference: all
  four have byte-for-byte the same verdicts before and after this fix.
- **No ratchet is re-baselined.** Nothing in the corpus, matrix, warning or error baselines moves;
  the repro is a new pattern-corpus file, which the gates compile for the first time.

## Separate defects, filed as their own issues

Found by the same matrix, all pre-existing and untouched here:

- **#3666** — `export import x = require('m')` adds a `$$props` parameter official does not, and
  both compilers emit output acorn rejects for the bare form (16 official cells, 8 rsvelte).
- **#3667** — an `import` (bare or `import type`) inside a `namespace` body is a `js_parse_error` in
  rsvelte where official compiles it or raises `typescript_invalid_feature`. 32 cells.
- **#3668** — `export declare global { … }` compiles in rsvelte where official raises
  `js_parse_error`. 8 cells. Opposite direction to #3667.
- **#3669** — `export {};` in a `<script module lang="ts">` survives into the generated output.

## Repro

`compatibility/pattern-corpus/issues/3566-export-type-only-declaration.svelte`, with its README row.
Verified byte-identical to official on client/server × dev/prod (4/4), and oxfmt-idempotent so the
formatter-parity gate is a real comparison rather than a rewrite.

## Verification

- **Revert control**: run, quoted above. Two tests red for the predicted reason, three green.
- **Differential matrix**: 1,960 pairs, before and after, against the pinned oracle. Numbers above.
- **Repro**: byte-identical to official on client/server × dev/prod (4/4), oxfmt-idempotent.
- **`cargo fmt --all -- --check`**: clean.
- **Changeset gates**: `check-changeset-package-names.mjs`, `check-core-consumer-changesets.mjs`,
  `check-esrap-version-bump.mjs` — all pass locally.
- **Not run locally, stated rather than implied**: the wider `rsvelte_core` TypeScript suites
  (`ts_namespace_3244`, `ts_erasure_visitor_coverage`, `module_compile_ts_strip`,
  `ssr_typescript_erasure`, `parse_ts_types`, `compiler_errors`, `compiler_fixtures`, …) and
  `cargo clippy`. The dev machine is thrashing — a single `rsvelte_core` release rebuild is taking
  ~29 minutes against 13+ concurrent `rustc` processes and 4-7 GiB of free disk — so I pushed with
  the load-bearing evidence (revert control + 1,960-pair before/after) rather than hold a verified
  fix in a worktree a disk-full could destroy. **The verdict on those suites and on clippy is
  CI's.**


baseballyama:
The `Compiler parity` failure here is **not a regression from this PR** — it is a pre-existing svelte2tsx divergence that this PR's new fixture is the first thing to expose.

The failing check is `Verify svelte2tsx (oxfmt-normalized TSX + source maps)`, one NEW entry:

```
pattern/issues/3566-export-type-only-declaration.svelte [ts-mismatch]
  ts line 3
    expected:   export namespace Types {
    actual:     namespace Types {
```

Measured on `main` (a tree with none of this PR's commits in it), rsvelte's instance-script projection already drops the keyword:

```
0: ///<reference types="svelte" />
1: ;function $$render() {
2:
3: 	 namespace Types {
4: 		type Id = string;
5: 	}
```

Note the leading space on line 3 where `export ` was blanked. Official svelte2tsx keeps `export namespace`. Ten-line repro, no corpus needed:

```rust
use rsvelte_projection::svelte2tsx::{Svelte2TsxOptions, svelte2tsx};
let src = "<script lang=\"ts\">\n\texport namespace Types {\n\t\ttype Id = string;\n\t}\n\n\texport {};\n\n\tlet k = 1;\n</script>\n\n{k}\n";
println!("{}", svelte2tsx(src, Svelte2TsxOptions::default()).unwrap().code);
```

So the compiler-side fix in this PR is fine; what the fixture additionally reaches is the projection's blanket `export`-stripping in the instance script (correct for `export let x`, which is a prop — wrong for a namespace/enum declaration, which official leaves exported).

Two ways forward, and which is right is the author's call:

1. Teach the instance-script projection to keep `export` for a `TSNamespaceDeclaration` / `TSEnumDeclaration`. That is the real fix and clears the entry.
2. Add the fixture to `compatibility/svelte2tsx-known-failures.json` with the divergence documented, and leave the projection for its own PR. The ratchet is two-sided, so the entry has to actually still fail — it does.

I have not touched this branch. Flagging it because I hit the same pattern on #3195 today: a submodule/fixture change makes a pre-existing divergence newly visible, and the failure reads like the PR caused it.

PR #3693, opened 2026-08-23
fix: remove duplicate chain expression arms

baseballyama:
Removes two duplicate match arms introduced by overlapping merges. The duplicates make the latest main fail every Rust job under `-D warnings`. Semantics are unchanged because the retained arms are identical.\n\nValidated with:\n- `cargo fmt --all -- --check`\n- `cargo clippy --all-targets --all-features -- -D warnings`


baseballyama:
Verified locally, and the two remaining red checks are both independent of this change.

**The fix compiles.** Applied this PR's diff to a clean `origin/main` worktree and built the crate the way CI does:

```
$ RUSTFLAGS="-D warnings" cargo check -p rsvelte_core --lib
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 57.08s
$ echo $?
0
```

Without the patch the same command fails with the two `unreachable pattern` errors, so the control moves.

**The retained arms are semantically identical to the removed ones**, which is the claim the PR body makes and the part worth checking rather than taking:

- Node walk — `snippet_block.rs:389` retained vs `:509` removed: byte-identical bodies (`expression_only_uses_params_node(arena.get_js_node(*expression), param_names, context)`).
- JSON walk — `:1479` retained is `obj.get("expression").is_none_or(|e| expression_only_uses_params(e, …))`; `:1583` removed is `if let Some(inner) = obj.get("expression") { return expression_only_uses_params(inner, …) } true`. They agree on both branches: `None` → `true`, `Some` → recurse.

**The two failures left are pre-existing on `main`, not caused here.**

`Changeset` — the gate's own message names the escape hatch: *"Add the 'skip-changeset' label to bypass."* This PR removes unreachable duplicate arms, so nothing observable ships and there is no version bump to describe. I have not added the label; that is the author's call.

`Shape matrix parity (generated inputs)` — the ratchet is stale in the **shrink** direction:

```
[matrix] ❌ 110 baseline entries already PASS — the ratchet is stale.
  - comment-slot/class-private-state__L12__block.svelte       [comment-mismatch] (server)
  - comment-slot/class-static-block__L13__line.svelte         [comment-mismatch] (server)
  - comment-slot/const-fold-line-continuation__L06__block.svelte [comment-mismatch] (server)
  … and 90 more
```

All 110 are `comment-mismatch` on `server`. The only recent commit touching server comment placement is `ad470065f` — *"follow esrap's comment cursor at the end of an instance script (#3326)"* — which changed `3_transform/server/ast/{comments,mod,script,visitors/shared}.rs` and touched no `known-failures` file. That is a shape match, **not a proof**: confirming it means running the matrix at `ad470065f^` and checking the 110 return to failing, which I have not done. Either way it needs its own PR, because a two-sided ratchet has to be re-baselined by the change that moved it.

This one blocks every branch in the repository right now, not just this PR.

PR #3757, opened 2026-08-25
fix: build server component slots in source order

### LINKED ISSUE #3681: server: `each_array` suffixes follow the slot build order, not the source order
On the server, `each_array` suffixes are allocated in the order the `$$slots`
entries are *built*, which puts the default slot first. Upstream allocates in the
order the children appear in the source, so a component whose first child carries
`slot="…"` gets the suffixes the other way round.

### Repro

```svelte
<script>
	import C from './C.svelte';
	let n = 7;
	const pr = Promise.resolve(1);
	let items = [1, 2];
</script>

<C><p slot="beta">{#each items as n}<b>{n}</b>{/each}</p><p slot="alpha">{#await pr then n}<i>{n}</i>{/await}</p><span>{#each items as _, n}<u>{n}</u>{/each}</span></C>
```

The children are on one line on purpose: a newline between them is a whitespace
text node, which is a default-slot child, which puts `default` first and makes the
two orders agree.

### `generate: 'server'`, svelte 5.56.9 as the oracle

```diff
 	C($$renderer, {
 		children: ($$renderer) => {
 			$$renderer.push(`<span><!--[-->`);
-			const each_array = $.ensure_array_like(items);            // rsvelte
+			const each_array_1 = $.ensure_array_like(items);          // official
 		},
 		$$slots: {
 			beta: ($$renderer) => {
-				const each_array_1 = $.ensure_array_like(items);      // rsvelte
+				const each_array = $.ensure_array_like(items);        // official
```

Everything else on this input matches byte for byte, including the `$$slots` key
order (`beta, alpha, default`) and the `let:`-shadow resolution — so this is the
only remaining divergence on it.

### Where

`3_transform/server/ast/visitors/component.rs`. #3518 changed only the *assembly*
of the `$$slots` object (entries are collected into `slot_entries` and emitted in
`slot_order`); the two loops that *build* them still run default-slot first, and
the counter is allocated during the build. So the numbering predates #3518 and is
unaffected by it — verified by reading its diff, which touches no part of the loop
order.

Fixing this means allocating in child order, which is the order `slot_order`
already records.

baseballyama:
Fixes #3681.

- drive server slot-body construction from first source occurrence instead of default-first grouping
- preserve existing children props, let shadows, and serialized slot order
- cover server prod/dev plus a four-target pattern-corpus repro

Validation: cargo fmt --all -- --check; git diff --check. Compilation and full tests are left to CI as requested.


baseballyama:

PR #3758, opened 2026-08-25
fix(fmt): hug inline elements glued inside flow blocks

### LINKED ISSUE #3675: rsvelte-fmt: an inline element written on one line inside a block body never hugs, where the oracle does
Oracle: `oxfmt -c scripts/fixtures/fmt-corpus.oxfmtrc.json` with `svelte: true`
(prettier-plugin-svelte for the Svelte structure), the same oracle the
formatter-parity corpus uses, against `target/release/rsvelte-fmt --stdin
--stdin-filepath`.

## Repro

An inline element written on one line inside a block body it is glued to on both
sides:

```svelte
{#if flag}<span>aaaaaaaa bbbbbbbb cccccccc dddddddd eeeeeeee ffffffff gggggggg</span>{/if}
```

```
ora|{#if flag}<span
ora|    >aaaaaaaa bbbbbbbb cccccccc dddddddd eeeeeeee ffffffff gggggggg</span
ora|  >{/if}
rsv|{#if flag}<span>aaaaaaaa bbbbbbbb cccccccc dddddddd eeeeeeee ffffffff gggggggg</span>{/if}
```

rsvelte-fmt leaves the whole thing on one 89-column line. The oracle hugs the
`<span>`: the open `>` drops to its own indented line, the close `</span` glues to
the content, and the block's `{/if}` glues after the final `>`.

## It is element-level, not tag-level

Three payloads, same block, same result — so nothing about the *content* is the
variable:

| content of the `<span>` | oracle | rsvelte |
|---|---|---|
| plain prose text | hugs | **one line** |
| a run of expression tags with nothing breakable (`{o.aaaaaaaa}{o.bbbbbbbb}…`) | hugs | **one line** |
| a run containing a breakable group (`{o["aaaa"]}{o.bbbbbbbb}…`) | hugs, then breaks the group | **one line** |

The unbreakable-run row is the discriminating one: no tag in it *can* break, so the
divergence cannot be about how a tag is measured. It is that the element never hugs.

## Not #3423 / #3565

Those are about a content tag inside an element that **has** hugged, and are fixed
in #3674. This shape is one level out: the element does not hug in the first place.
Written with the `<span>` on its own line inside the block, both sides agree at every
width:

```svelte
{#if flag}
	<span>{o["aaaa"]}{o.bbbbbbbb}{o.cccccccc}{o.dddddddd}{o.eeeeeeee}{o.ffffffff}</span>
{/if}
```

```
ora|{#if flag}
ora|  <span
ora|    >{o[
ora|      "aaaa"
ora|    ]}{o.bbbbbbbb}{o.cccccccc}{o.dddddddd}{o.eeeeeeee}{o.ffffffff}</span
ora|  >
ora|{/if}
```

(that one matches on `main` + #3674). `{#each}` behaves identically to `{#if}` in
both spellings.

## Where it probably lives

`collapse/hug.rs`'s `try_hug_block_inline_body` is the pass for "a block whose body
is exactly one huggable inline element glued to both block tags", and it produces the
`{#each …}<span>{x}</span\n  >{/each}` form. It requires `body.nodes.len() == 1` and
the block to sit at the start of its line; the `{#if}` spelling above satisfies both,
so either the pass is not reached for `IfBlock` or its fit test declines. Worth
checking against `{#key}` and `{#await}` too — I only measured `{#if}` and `{#each}`.

Found while measuring the host axis for #3423; filed rather than folded in because
the fix is a different pass and the discriminating input (the unbreakable run) shows
it is a different defect.

baseballyama:
Fixes #3675.\n\nRoutes if blocks and single-branch await blocks through the existing block-body hug pass, and distinguishes an overflow through the element close tag from an overflow caused only by the final bracket and block close. Covers if, each, key, and await blocks with formatter regressions and a pattern-corpus fixture.\n\nLocal verification was intentionally limited to formatting and static diff checks; CI owns builds and tests.


baseballyama:

PR #3759, opened 2026-08-25
fix(css): replace universal scope point before global

### LINKED ISSUE #3690: CSS: a universal selector followed by `:global(...)` is not replaced by the scoping class
## Summary

When a compound is `*` followed by `:global(...)`, rsvelte keeps the `*` where official
replaces it with the scoping class.

```svelte
<div class="card wide"><p class="a">x</p><span class="b" data-k="v">y</span></div>

<style>
	*:global(.g) { color: red }
</style>
```

```css
/* official */  .svelte-1lj1c1s.g { color: red }
/* rsvelte  */  *.svelte-1lj1c1s.g { color: red }
```

`css.code` only — the scoped element set agrees, and neither side warns.

## Why it is separate from #3402

#3402 is the opposite direction of the same decision: there `*` sits **before** another simple
selector in the compound and rsvelte *replaced* it where official keeps it. The fix for #3402
(`*` is consumed by the modifier only when it is the compound's last non-pseudo selector, mirroring
upstream's backwards walk in `phases/3-transform/css/index.js:344-364`) leaves this row **unchanged
byte for byte** — measured before and after on the same 22-case matrix, `*:global(.g)` diverges
identically in both. It is reached through a different branch of rsvelte's printer: the
`is_global` arm, which concatenates the compound's simple selectors with no universal handling at
all, rather than the `should_scope` arm the #3402 fix touches.

Upstream has no such split. `:global(.g)` mid-compound is stripped by `remove_global_pseudo_class`
and the compound then goes through the ordinary backwards walk, which reaches the `*` (a
`PseudoClassSelector` does not stop it) and `code.update(selector.start, selector.end, modifier)`
replaces it.

## Matrix

Measured against `<div class="card wide"><p class="a">x</p><span class="b" data-k="v">y</span></div>`,
client + server (identical):

|---|---|---|
| `*:global(.g)` | `.svelte-X.g` | `*.svelte-X.g` |

**Controls that match:** `*` alone, `*.a` (post-#3402), `*:first-child`, `:global(.g)` alone,
`.a:global(.g)`, `*:global(.g) .a`.

## Note

Found while measuring #3402; it is the one row of that issue's 22-case matrix the #3402 fix does
not close. Filing it so a divergence noticed in passing is not rediscovered later at full cost.

baseballyama:
Fixes #3690.\n\nThe partial-global CSS transform now applies the same bare-universal replacement rule as the regular scoped-selector path. It replaces the universal only when that node is the compound scoping point, preserving earlier and namespaced universal selectors.\n\nAdds crate-level controls, a pattern-corpus reproducer, and a compiler patch changeset.\n\nValidation: cargo fmt --all; git diff --check. Compilation and test execution are intentionally delegated to CI.


baseballyama:

PR #3762, opened 2026-08-25
fix(analyze): ignore import-equals component exports

### LINKED ISSUE #3666: `export import x = require('m')` in an instance script adds a `$$props` parameter official does not

baseballyama:
Fixes #3666

Represents the declaration inside export import-equals as an empty TypeScript declaration so the existing export stripper removes it before legacy component exports are counted. The shared analysis therefore keeps the official production component calling signature without an unnecessary props parameter.

This intentionally does not decide how import-equals text should be emitted: both compilers currently emit output that is not valid JavaScript for this construct. The focused regression observes the shared `needs_props` result through client production output; the server transform now deliberately rejects the same invalid erased source at its classification reparse.

Local verification was limited to `cargo fmt --all -- --check` and `git diff --check`; compilation and tests are delegated to CI per the project request.

### DIFF
--- a/.changeset/ts-import-equals-props.md
+++ b/.changeset/ts-import-equals-props.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+fix: avoid adding a props parameter for an exported TypeScript import-equals declaration
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -9071,6 +9071,22 @@ fn convert_declaration_for_program_as_node(
             offset,
             line_offsets,
         ),
+        // `export import x = require('m')` has value export-kind in oxc, but
+        // upstream's TypeScript visitor still removes the declaration before
+        // legacy component exports are counted. Represent only the declaration
+        // inside the export as empty so the wrapper follows the existing
+        // strip_export_named_declaration_typed path. The unexported spelling is
+        // deliberately untouched: what either compiler should print for it is
+        // an upstream question (both currently emit invalid JavaScript).
+        Declaration::TSImportEqualsDeclaration(decl) => {
+            let start = offset + decl.span.start as usize;
+            let end = offset + decl.span.end as usize;
+            JsNode::EmptyStatement {
+                start: start as u32,
+                end: end as u32,
+                loc: create_typed_loc(start, end, line_offsets),
+            }
+        }
         _ => JsNode::from_value(convert_declaration_for_program(
             arena,
             decl,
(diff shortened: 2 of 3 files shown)


baseballyama:

PR #3764, opened 2026-08-25
fix(server): hoist awaited element attributes

### LINKED ISSUE #3648: server: an awaited attribute, directive or spread value is not hoisted into `$$renderer.child(async …)`
Found by the new `async-attribute-slot` matrix family (200 cases / 792 comparisons, `experimental.async: true`). Oracle: `submodules/svelte` at `VERSION 5.56.9`, asserted from the compiler's own export.

## What happens

When an attribute value, a `class:` / `style:` directive value, or a spread on an element contains `await`, upstream's server transform hoists it into a nested async renderer:

```js
// official, <div style:color={await p}>, generate: 'server'
$$renderer.child(async ($$renderer) => {
	const $$0 = (await $.save(p))();

	$$renderer.push(`<div${$.attr_style('', { color: $$0 })}></div>`);
});
```

rsvelte pushes the `await` straight into the template string of a **non-async** renderer callback:

```js
// rsvelte
$$renderer.push(`<div${$.attr_style('', { color: await p })}></div>`);
```

## Two failure modes, and only one of them is loud

**230** rows of the family diverge on this. **80** of them are output no JS parser accepts (`Cannot use keyword 'await' outside an async function`) — the enclosing function is not async, so the module is a `SyntaxError` and the build fails at bundle time. The other **150** land inside a `child(async …)` that happens to exist for another reason, so the text parses and the only thing missing is the `(await $.save(x))()` wrapper — output that runs and is silently wrong.

## Population

Every host in the family (`<div>`, a custom element, `<svelte:element>`, a component) and every slot (plain attribute, `class`, `style`, `class:`, `style:`, spread), on `server` and `server-dev`. `<svelte:element>` is the largest single share. Controls (`static`, `state`) are green.

Entries are listed in `compatibility/matrix-known-failures.json` under `async-attribute-slot/`, justified in the paired `.md` as cause 1.

## Why no existing gate saw it

`scripts/compat-corpus/compile.mjs` passes exactly `{ generate, dev, filename }`, so the collected corpus never sets `experimental.async` and this whole population is unreachable there at any corpus size. See gate-coverage 5j.

baseballyama:
## Summary

- route awaited class, style, and spread values through the server `PromiseOptimiser`
- include directives and object-valued spreads in async detection for regular, custom, dynamic, and component hosts
- give dynamic `<svelte:element>` lowering an isolated per-element optimiser matching upstream
- add oracle-parity coverage for inline awaits and nested async-IIFE awaits
- remove all 230 stale generated-matrix entries and six stale Appwrite server/server-dev corpus entries cleared by the implementation

The implementation and every observed ratchet shrink are kept in this PR so the two-sided baselines remain mergeable atomically.

## Validation

- `git diff --check`
- `node scripts/compat-corpus/known-failures-md-check.mjs`
- Full compile/test, corpus, and generated-matrix verification are delegated to CI per the current no-local-test constraint.

Closes #3648


baseballyama:

PR #3766, opened 2026-08-25
test(corpus): cover function declaration aliases

### LINKED ISSUE #3636: Client: a const aliasing a function declaration is folded to a one-shot `textContent`, dropping the reactive text placeholder
## Summary

A `const` whose initializer is a bare identifier naming a **function declaration** is treated as a
known, non-reactive value by the client text pass, so `{alias}` is emitted as a one-shot
`b.textContent = alias` and the element's text placeholder disappears from the `from_html`
template. Official keeps the `$.template_effect(() => $.set_text(text, alias))` and the ` `
placeholder.

Output parses and the rendered text happens to be the same, so **only output equality can see
this** — the parse gate is blind to it, and `server` is byte-identical, which makes it another
instance of the client/server two-ports shape.

## Repro

```svelte
<script>
	function f() {
		return 1;
	}

	const g = f;
</script>

<b>{g}</b>
```

```js
…
	var text = $.child(b, true);

	$.reset(b);
	$.template_effect(() => $.set_text(text, g));

/* rsvelte */
…
	b.textContent = g;
```

## What separates the diverging case from its neighbours

Measured on one file, `generate: 'client'`, official **5.56.9**
(`submodules/svelte` @ `20b341f10048cf1016a2028ac7eee5595cfef6a5`):

| declaration | `{x}` in markup | verdict |
|---|---|---|
| `function f() {…}` then `const g = f;` | `{g}` | **DIFF** — rsvelte folds, official does not |
| `const arrow = () => 2;` | `{arrow}` | match (both fold) |
| `const plain = 1;` then `const alias = plain;` | `{alias}` | match (both fold) |
| `const literal = 'x';` | `{literal}` | match (both fold) |
| `function f() {…}` | `{typeof f}` | match (both fold) |

So it is not "a const alias" in general and not "a function value" in general: it is specifically a
`const` bound to an **identifier that resolves to a function declaration**. It reproduces in legacy
mode and in runes mode, on `client` and `client-dev`; `server` matches on every row above.

## Why this is filed separately

Found while measuring #3470 (keyword needles assuming one ASCII space). It reproduces with a single
ASCII space everywhere and with no `export` and no `class` in the source, so none of the code paths
that issue touches are reachable from it.

Adjacent, and probably the same question one port over: #3530 / PR #3559 fixed the **server** SSR
fold inlining a const alias of a *written* binding. This is the client side of "which initializers
make a `const` a known value", and `CLAUDE.md` already records that "is this value known" has
several separate implementations in the client (`is_expression_known_json`,
`identifier_has_reactive_state`, `is_initial_value_literal_or_known`).

baseballyama:
## Summary

- pin the function-declaration alias case from #3636 in the output-equality corpus
- cover both legacy and explicit runes mode
- keep direct-arrow, plain-alias, and `typeof` controls beside the discriminating case

The implementation was corrected after the issue was filed by #3501, which moved the client `is_known` decision onto the shared `scope.evaluate` port. A function marker conveys a known type, not a concrete compile-time value, so the alias now remains a dynamic client text chunk.

## Verification

- `git diff --check`
- local builds/tests intentionally skipped; compatibility CI is the oracle for all corpus targets

Closes #3636


baseballyama:
Superseded by #3799, which includes the legacy/runes corpus cases from this PR and fixes the shared evaluator used by the affected path.

PR #3768, opened 2026-08-25
fix(client): recognize defined option value expressions

### LINKED ISSUE #3259: Client: `<option value={…}>` keeps `?? ''` for arrow, function, binary and unary expressions official knows are non-nullish
Upstream drops the `?? ''` guard on `<option value={…}>` when the expression
cannot be nullish. rsvelte's defined-ness check does not recognise several
expression kinds, so the guard is emitted where official omits it.

```svelte
<script>
	let s = $state('v');
</script>
<select><option value={() => 1}>x</option></select>
```

`generate: 'client'`:

```js
// official
option.value = option.__value = () => 1;
// rsvelte
option.value = (option.__value = () => 1) ?? '';
```

## Which kinds

A 22-value probe over `<option value={V}>`:

| diverges | matches |
|---|---|
| `() => 1` | `1`, `'a'`, `true`, `null`, `undefined`, `void 0` |
| `function () {}` | `s`, `fn`, `obj.k`, `sink(1)` |
| `1 + 1` | `{}`, `[]`, `new Date()`, `` `t${s}` `` |
| `-1` | `s ?? "d"`, `s || "d"`, `Symbol.iterator` |

Literals and identifiers are already handled; the gap is
`ArrowFunctionExpression`, `FunctionExpression`, `BinaryExpression` and
`UnaryExpression`. Note `fn` (an identifier bound to an arrow) matches while the
inline arrow does not — the alias path resolves, the direct node kind does not.

## Same shape as #3249

This is a second **enumeration of node kinds** that is missing arms — there,
`init_needs_expr_json` omits `LogicalExpression` and the fold never runs; here,
the defined-ness predicate omits four kinds and the guard is emitted anyway.
Worth auditing the two together: a per-kind enumeration written against the
cases someone had in front of them is exactly the shape that goes stale, and
neither predicate has a gate that enumerates its domain.

## Found by

Adversarial attribute grid (39 names × 23 value shapes × 8 hosts × 2 targets);
`value / expr-fn / option` was the single diverging cell in the `value` column,
which the 22-value follow-up expanded to four kinds.

baseballyama:
## Summary

- teach the option-value definedness check about arrow functions, function expressions, binary expressions, and non-`void` unary expressions
- preserve the nullish fallback for `void` and unknown identifiers
- add a pattern-corpus regression covering all four missing expression kinds plus both negative controls

Fixes #3259.

## Validation

- `node scripts/ci/check-pattern-corpus-docs.mjs`
- `node scripts/ci/test-check-pattern-corpus-docs.mjs`
- `git diff --cached --check`

Per the requested workflow, I did not run a local build or compiler test suite. The compatibility and build results are delegated to CI, including all configured corpus targets.


baseballyama:
Superseded by the broader shared-definedness fix in #3801, which covers option values together with the title and assignment/sequence consumers.

PR #3769, opened 2026-08-25
Fix shadowed binding diagnostic parity

### LINKED ISSUE #3297: Shadowed bindings: `export_let_unused` is not emitted, and `snippet_parameter_assignment` has no position
Two diagnostics diverge on shadowed template bindings.

## 1. `export_let_unused` is not emitted when every read is shadowed

```svelte
<script>
	export let v = "OUTER";
</script>
{#each ["A"] as v (v)}{String(v)}{/each}
```

```
official : export_let_unused @ 2:11
rsvelte  : (no warning)
```

Every occurrence of `v` in the template resolves to the each item, so the prop
really is unused; rsvelte's usage tracking counts the shadowed reads as uses of
the prop. 14 cells — the `export let` outer × `each_keyed` × seven read sites ×
both targets.

## 2. `snippet_parameter_assignment` is raised without a position

```svelte
<script>
	let v = "OUTER";
</script>
{#snippet s(v)}<button onclick={() => { v = "W"; }}>b</button>{/snippet}{@render s("A")}
```

| | code | message | start | end |
|---|---|---|---|---|
| rsvelte | `snippet_parameter_assignment` | `Cannot reassign or bind to snippet parameter` | **null** | **null** |
| official | `snippet_parameter_assignment` | same | 6:40 | 6:47 |

The code and message agree; only the span is missing. 20 cells — every outer
declaration form × `snippet_param` × both targets — so it is one rule, not a
per-outer gap.

Third instance of the same shape after #3225 (`options_deprecated_immutable`
without a position) and the `<svelte:options runes />` + `$:` error in #3276.
Worth checking whether they share a helper before fixing them one at a time.

## Population

shadow-read 2,288 cells (2,288 compiled on both sides), shadow-write 440 cells
(224 compiled on both sides, 216 errored on both). The no-collision control at
1,694 cells produced neither of these signatures, so both are shadowing-specific.

baseballyama:
## Summary

- attach the assignment/update/bind node span to `snippet_parameter_assignment`
- pin that an `{#each}` item shadowing an exported prop does not count as a use of the prop
- cover both snippet mutation and `bind:` validation paths

The shadowed-reference half is now handled by the lexical each-scope resolution landed in #3594. This PR adds the missing regression coverage and fixes the remaining diagnostic-position divergence.

Fixes #3297.

## Validation

- `git diff --check`

Per repository guidance, I did not run local builds or tests; CI is the build/test oracle.

### DIFF
--- a/.changeset/shadowed-diagnostic-spans.md
+++ b/.changeset/shadowed-diagnostic-spans.md
@@ -1,0 +1,5 @@
+---
+"rsvelte": patch
+---
+
+Match upstream diagnostics for shadowed each bindings and snippet parameter assignments.
--- a/.changeset/shadowed-diagnostic-spans.md
+++ b/.changeset/shadowed-diagnostic-spans.md
@@ -1,5 +1,5 @@
 ---
-"rsvelte": patch
+"@rsvelte/compiler": patch
 ---
 
 Match upstream diagnostics for shadowed each bindings and snippet parameter assignments.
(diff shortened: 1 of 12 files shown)


baseballyama:

PR #3772, opened 2026-08-25
fix(client): preserve each-item mutation sequence parentheses

### LINKED ISSUE #3616: Client: an each-item mutation is not wrapped in the one-element SequenceExpression upstream builds, so the parentheses are missing

## Repro

```svelte
<script>
	let rows = $state([{ picked: [], n: 0 }]);
</script>

{#each rows as row, i}
	<button onclick={() => row.picked = 1}>x</button>
{/each}
```

`generate: 'client'`:

```js
// official
$.delegated('click', button, () => ($.get(row).picked = 1));

// rsvelte
$.delegated('click', button, () => $.get(row).picked = 1);
```

## Cause

The parentheses are not a printer preference — they are an artefact of the node
upstream builds. `EachBlock.js:240-243` (and `:276-278` for a destructured
context) registers

```js
mutate: (_, mutation) => {
    uses_index = true;
    return b.sequence([mutation, ...sequence]);
}
```

so **every** mutation of an each-block item comes back wrapped in a
`SequenceExpression`. `sequence` is empty for a plain block, which leaves a
ONE-element sequence, and esrap prints `SequenceExpression` as
`'(' + … + ')'` (`esrap/src/languages/ts/index.js:1646-1650`). rsvelte returns
the bare assignment, so the parens are missing everywhere the mutation is the
whole value of an expression position.

## Scope: the mutation, not the host — 16 of 36

9 hosts × {plain, keyed} × {client, server}:

| host | verdict |
|---|---|
| `bind:group={row.picked}` | **diverges** |
| `bind:value={row.name}` | **diverges** |
| `onclick={() => row.picked = 1}` | **diverges** |
| `onclick={() => { row.picked = 1; }}` | **diverges** (statement position too) |
| `onclick={() => row.n++}` | **diverges** (an update, not only an assignment) |
| `{@const q = (row.picked = 1)}` | **diverges** |
| `onclick={() => (() => row.picked = 1)()}` | **diverges** |
| `onclick={() => 1 ? row.picked = 1 : 0}` | **diverges** |
| `onclick={() => rows[i].picked = 1}` | match |
| every one of the above, `server` | match |

Both the keyed (`{#each rows as row, i (row)}`) and the plain form diverge, and
the keyed form is what shows the wrapping is not about `$.get(...)`: there the
text is `(row.picked = 1)`, parenthesised with no `$.get` in sight. The
`rows[i].picked` row is the negative control — its object is not an each-item
binding, so no `mutate` transform is registered and both sides print it bare.

`server` matching across the board is the second control: the SSR transform has
no `transform.mutate`, so this is a client-only shape rather than a difference
in how the two compilers read the source.

## Why no gate reports it

The output parses and evaluates identically — `(x = 1)` and `x = 1` are the same
expression — so the parse oracle and every runtime test are blind, and the
corpus comparison's oxfmt normalization removes redundant parentheses. It is
visible only in a raw byte comparison.

Found while fixing #3576, whose repro contains one instance of it; the group
half of that issue is fixed and this half is not.

baseballyama:
## Summary

- preserve the one-element `SequenceExpression` that upstream's each-item `mutate` transform always builds
- locate the owning each context for simple and destructured item names, including ancestor each blocks
- keep legacy inner/store invalidations after the mutation while retaining parentheses when that tail is empty
- add raw client-output coverage for assignments, updates, bindings, keyed and destructured items, plus an indexed-collection negative control

## Why

Upstream returns `b.sequence([mutation, ...sequence])` even when `sequence` is empty. rsvelte only built a sequence when legacy invalidations were present, so runes and keyed each-item mutations lost the parentheses emitted by esrap. Comparison-side formatting removes those parentheses, so the regression tests inspect raw compiler output.

Closes #3616.

## Validation

- `git diff --check`
- full builds/tests intentionally left to CI to avoid the project's heavy local workload

### DIFF
--- a/.changeset/each-item-mutation-sequence.md
+++ b/.changeset/each-item-mutation-sequence.md
@@ -1,0 +1,5 @@
+---
+"rsvelte": patch
+---
+
+Preserve upstream's sequence-expression parentheses around each-item mutations in client output.
--- a/.changeset/each-item-mutation-sequence.md
+++ b/.changeset/each-item-mutation-sequence.md
@@ -1,5 +1,5 @@
 ---
-"rsvelte": patch
+"@rsvelte/compiler": patch
 ---
 
 Preserve upstream's sequence-expression parentheses around each-item mutations in client output.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
@@ -1252,51 +1252,56 @@ pub fn apply_transforms_to_expression_with_shadowed(
                 // Also handle legacy mode each item mutation: append $.invalidate_inner_signals()
                 if let JsExpr::Identifier(name) = &base_object
                     && !local_scope.contains(name)
-                    && context.state.each_item_names.contains(name)
+                    && let Some(each_ctx) = context
+                        .state
+                        .each_binding_context
+                        .iter()
+                        .rev()
+                        .find(|ctx| {
+                            ctx.item_name == *name
+                                || ctx.destructured_update_paths.contains_key(name.as_str())
+                        })
+                        .cloned()
                 {
-                    context.state.each_item_assign_or_mutate.set(true);
+                    if each_ctx.context_is_identifier {
+                        each_ctx.binding_used.set(true);
+                    }
 
-                    // In legacy mode, wrap the mutation with $.invalidate_inner_signals()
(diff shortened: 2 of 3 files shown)


baseballyama:
Superseded by #3824. The implementation here reconstructed mutation sequences inside the transform walker and regressed store/index handling in CI (`each-bind-store-no-warning` and six compiler-parity entries). The replacement only wraps a final bare each-item mutation, preserves existing invalidation sequences unchanged, and covers typed/JSON assignment, update, and binding paths.

PR #3778, opened 2026-08-25
fix(compiler): lower class runes across separators

### LINKED ISSUE #3490: compileModule (server): a class `$derived` field is missed unless `=` and the rune are separated by exactly one space

### LINKED ISSUE #3498: client: a class rune is not lowered when its value starts on a later line than the `=`

baseballyama:
Closes #3490.
Closes #3498.

## Summary

- share class assignment/rune separator parsing between client and server transforms
- recognize JavaScript whitespace, block comments, and line comments between `=` and class runes
- reject equality, comparison, and arrow tokens as assignment operators
- lower client class fields and constructor assignments whose rune initializer starts on a following line
- retain line comments without commenting out the generated `$.state` / `$.derived` expression
- add focused unit coverage and pattern-corpus cases for server modules and components

## Checks

- `git diff --check`
- builds and tests deferred to CI per request


baseballyama:

PR #3781, opened 2026-08-25
fix(client): fold dev equality declaration initializers

### LINKED ISSUE #3570: client-dev only: a declaration whose initializer contains `===`/`!==`/`==`/`!=` is never folded, and gets a spurious `template_effect`

## Repro — no `$state` anywhere, two literals

```svelte
<script>
	const v = "a" === "a" ? 1 : 2;
</script>

<b>{v}</b>
```

`client` matches. `server` matches. `client-dev`:

```
off|var root = $.add_locations($.from_html(`<b></b>`), Eq[$.FILENAME], [[5, 0]]);
off|	b.textContent = '1';
rsv|var root = $.add_locations($.from_html(`<b> </b>`), Eq[$.FILENAME], [[5, 0]]);
rsv|	var text = $.child(b, true);
rsv|	$.template_effect(() => $.set_text(text, v));
```

rsvelte does not fold it in dev, and the consequence is not layout: it emits a
**`$.template_effect`** for a value that is a compile-time constant. The same
component in prod client folds to `b.textContent = '1'`, so rsvelte's dev and
prod outputs disagree with each other as well as with official.

## The axis is the operator, and it is exactly the four equality operators

A ternary test `n <op> 2` with `let n = $state(2)`:

| op | client | client-dev | server |
|---|---|---|---|
| `===` | match | **DIFF** | match |
| `!==` | match | **DIFF** | match |
| `==` | match | **DIFF** | match |
| `!=` | match | **DIFF** | match |
| `>` `<` `>=` `<=` | match | match | match |
| `+` `-` `*` | match | match | match |
| `&&` `\|\|` `??` | match | match | match |
| `instanceof` `in` | match | match | match |

4 of 16, all in one column.

## The ternary is incidental — the host is any non-inline binding

|---|---|---|---|
| `const v = n === 2` | match | **DIFF** | match |
| `const v = n === 2 ? 1 : 2` | match | **DIFF** | match |
| `const v = !(n === 2)` | match | **DIFF** | match |
| `const v = "a" === "a" ? 1 : 2` (no state at all) | match | **DIFF** | match |
| `const v = $derived(n === 2 ? 1 : 2)` | match | **DIFF** | match |
| `{@const v = n === 2 ? 1 : 2}` | match | **DIFF** | match |
| `<b>{n === 2 ? 1 : 2}</b>` — inline in the template | match | match | match |

So the fold reaches an equality comparison from an inline template expression on
every target, and fails to reach it from a declaration's initializer in dev
only. `const v = (n === 2) && 1` diverges on all three targets, but that is
#3249 (a logical expression as the initializer never folds) and is a different
row — it is not dev-specific.

## Why it is worth its own issue

Every existing fold issue here (#3249, #3439, #3437, #3471, #3448, #3228) is
target-symmetric or client-only-versus-server. This one splits **client from
client-dev**, which is a pair no gate compares to each other, and it is the one
direction where the divergence is a spurious reactive effect rather than a
layout or a text difference.

Found while screening a generated adversarial corpus candidate
(`expressions/operator-precedence-zoo.svelte`), which is held out of
`compatibility/pattern-corpus/adversarial/` until this and #3249 are fixed.

baseballyama:
## Summary
- keep dev equality folding disabled for converted template expressions
- allow constant folding after evaluation follows an identifier into its original declaration initializer
- cover all four equality operators plus inline-expression controls

## Validation
- cargo fmt --all -- --check
- git diff --check
- heavy tests intentionally deferred to CI

Closes #3570

### DIFF
--- a/.changeset/calm-equalities-fold.md
+++ b/.changeset/calm-equalities-fold.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/core": patch
+---
+
+Fold constant equality expressions reached through declaration initializers in development builds.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/tests.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/tests.rs
@@ -2767,6 +2767,56 @@ fn globals_fold_client_code(source: &str) -> String {
     .code
 }
 
+fn dev_fold_client_code(source: &str) -> String {
+    crate::compiler::compile(
+        source,
+        crate::compiler::CompileOptions {
+            generate: crate::compiler::GenerateMode::Client,
+            filename: Some("dev-fold.svelte".to_string()),
+            dev: true,
+            ..Default::default()
+        },
+    )
+    .expect("compiles")
+    .js
+    .code
+}
+
+#[test]
+fn dev_equality_in_declaration_initializers_still_folds_template_reads() {
+    for (operator, expected, helper) in [
+        ("===", "2", "$.strict_equals"),
+        ("!==", "1", "$.strict_equals"),
+        ("==", "1", "$.equals"),
+        ("!=", "2", "$.equals"),
+    ] {
+        let declaration = dev_fold_client_code(&format!(
+            "<script>\n\tconst v = '1' {operator} 1 ? 1 : 2;\n</script>\n<b>{{v}}</b>\n"
+        ));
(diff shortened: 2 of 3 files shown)


baseballyama:
Superseded by #3799, which fixes dev equality initializers through the shared typed evaluator and carries the regression coverage.

PR #3784, opened 2026-08-25
Cover assignment and sequence nullish guards

### LINKED ISSUE #3601: client: a const initialized by an assignment or a sequence drops the template chunk's `?? ''` guard
## Summary

A `const` whose initializer is an `AssignmentExpression` or a `SequenceExpression` is judged **defined** by rsvelte and **not defined** by the official compiler, so the client template chunk drops the `?? ''` guard upstream emits.

Upstream's `Evaluation` (`phases/scope.js`) has no `case 'AssignmentExpression'` and no `case 'SequenceExpression'` — both fall to `default: values.add(UNKNOWN)`, which sets `is_defined = false`. `build_template_chunk` then keeps `?? ''`. rsvelte's ported lattice evaluates something narrower for those shapes and reports the value as defined.

**This is pre-existing, not a regression.** A `compile_one` binary built before #3585's row-3 fix produces byte-identical output on the repro, so a reviewer of that PR does not need to consider it.

## Repro

```svelte
<script>
	let { cond, un } = $props();
	let obj = $state({ a: 0 });
	const v = (obj.a = 1);
</script>

{v}{cond}
```

`{v}{cond}` (two chunks) is required — a single-chunk `{v}` compiles to `$.set_text(text, v)`, where the guard has nowhere to appear.

Official (dev or prod, client):

```js
$.template_effect(() => $.set_text(text, `${v ?? ''}${$$props.cond ?? ''}`));
```

rsvelte:

```js
$.template_effect(() => $.set_text(text, `${v}${$$props.cond ?? ''}`));
```

## Measured

Official 5.56.9 (`submodules/svelte`, `20b341f10048`, `VERSION` export asserted) vs rsvelte, client, over nine initializer shapes in the host above — **4 of 9 diverge**, and the divergent set is exactly the two node types upstream does not model:

|---|---|---|---|
| `(obj.a = 1)` | `?? ''` | bare | **DIFF** |
| `(obj.a = cond ? 1 : 2)` | `?? ''` | bare | **DIFF** |
| `(obj.a = (un, 1))` | `?? ''` | bare | **DIFF** |
| `(un, 1)` | `?? ''` | bare | **DIFF** |
| `(obj.a = 1 \|\| un)` | `?? ''` | `?? ''` | match |
| `(obj.a = un)` | `?? ''` | `?? ''` | match |
| `(cond ? 1 : 2)` | bare | bare | match |
| `obj.a` | `?? ''` | `?? ''` | match |
| `un` | `?? ''` | `?? ''` | match |

`(obj.a = 1 || un)` and `(obj.a = un)` match only because the *right-hand side* is independently not-defined — the assignment wrapper is still being seen through.

## Where

`EvalCtx::evaluate_estree` (`crates/rsvelte_core/src/compiler/phases/3_transform/server/evaluate.rs`). It deliberately unwraps `ParenthesizedExpression` and the TypeScript wrappers, which **is** upstream-faithful (Svelte erases TS before analysis — `obj.x = (1 as number)` is `no-wrap` on both sides, measured). The `AssignmentExpression` / `SequenceExpression` shapes need the opposite: they must reach the `UNKNOWN` default rather than resolve to an inner value.

## Why it was not caught

Found while writing a repro for #3585's row 3 — the first draft rendered the assignment results in one multi-chunk text node, which is the only shape that exposes the flag. The corpus output ratchets are all empty (`[]`), so this shape does not occur in the collected corpus at all; it took a hand-written file to reach it.

baseballyama:
## Summary
- pin #3601's full nine-row initializer matrix in client prod and dev
- verify assignment and sequence initializers stay UNKNOWN for concatenated template chunks
- retain positive controls for conditional expressions and unknown reads

The shared scope.evaluate path on current main already fixes the reported divergence; this PR prevents it from returning.

Fixes #3601

## Validation
- cargo fmt --all -- --check
- git diff --check
- full build/test deferred to CI as requested

### DIFF
--- a/crates/rsvelte_core/tests/assignment_sequence_nullish_guard_3601.rs
+++ b/crates/rsvelte_core/tests/assignment_sequence_nullish_guard_3601.rs
@@ -1,0 +1,53 @@
+//! #3601: upstream `scope.evaluate` has no `AssignmentExpression` or
+//! `SequenceExpression` arm. Both are UNKNOWN even when their result can be
+//! inferred from a child, so a concatenated template chunk must retain its
+//! `?? ''` guard. The shared evaluator introduced after the report already has
+//! that behavior; this pins the complete measured matrix, including controls.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile, compiler::CssMode};
+
+fn client(initializer: &str, dev: bool) -> String {
+    let source = format!(
+        "<script>\n\tlet {{ cond, un }} = $props();\n\tlet obj = $state({{ a: 0 }});\n\tconst v = {initializer};\n</script>\n\n{{v}}{{cond}}\n"
+    );
+
+    compile(
+        &source,
+        CompileOptions {
+            filename: Some("NullishGuard.svelte".to_string()),
+            generate: GenerateMode::Client,
+            dev,
+            css: CssMode::External,
+            ..Default::default()
+        },
+    )
+    .expect("compile")
+    .js
+    .code
+}
+
+#[test]
+fn assignment_and_sequence_initializers_stay_unknown() {
+    let cases = [
+        ("(obj.a = 1)", true),
+        ("(obj.a = cond ? 1 : 2)", true),
+        ("(obj.a = (un, 1))", true),
+        ("(un, 1)", true),
+        ("(obj.a = 1 || un)", true),
+        ("(obj.a = un)", true),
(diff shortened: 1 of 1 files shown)


baseballyama:
Superseded by #3801, which keeps assignment and sequence expressions unknown and includes the discriminating guard tests.

PR #3785, opened 2026-08-25
Use shared evaluator for title definedness

### LINKED ISSUE #3599: client: <svelte:head><title> emits a `?? ''` upstream drops when the value cannot be nullish
`<svelte:head><title>` is the one place the client emits a `?? ''` coercion, and rsvelte emits it unconditionally. Upstream drops it when `scope.evaluate` can prove the value is never nullish — and a `+` is enough to prove that.

```svelte
<script>
	export let a = 1;
	let b = a + 1;
</script>
```

`generate: 'client'`:

```js
// official
$.document.title = b;

// rsvelte
$.document.title = b ?? '';
```

The output parses, runs, and differs from official only when the value is nullish — which the `+` guarantees it is not. So this is byte divergence with no runtime symptom, the class that only output equality can see.

## The trigger is the `+`, not the store

Probed directly against the pinned oracle (`submodules/svelte` @ `20b341f10048`, `VERSION === '5.56.9'`), one initializer per row, everything else held fixed:

| initializer | official | rsvelte |
|---|---|---|
| `let b = a` (a prop) | `b ?? ''` | `b ?? ''` |
| `let b = a + 1` | **`b`** | `b ?? ''` |
| `let b = $s` (bare store read) | `b ?? ''` | `b ?? ''` |
| `let b = $s + a` | **`b`** | `b ?? ''` |

Rows 1 and 3 are the negative controls: they rule out "any prop" and "any store read" as the trigger. Only the rows with a `+` move, and they move in both the store and the non-store case. Upstream's rule is the ordinary one — `a + b` evaluates to a string or a number, never `null`/`undefined`, so the coalesce is dead code and it is not printed.

## Where it belongs

This is the "is this value known" family AGENTS.md records under #3027 / `fold-value-type`. That fix gave the client fold one model of a folded value by routing it through the server's `EvalValue`, and the same section names **three implementations that were left behind on the client**:

* `is_expression_known_json`
* `identifier_has_reactive_state`
* `is_initial_value_literal_or_known`

`+` being the trigger points at the last of those rather than at the fold: `binding.initial` keeps a non-literal initializer as **source text**, and the JSON branch is what re-parses it. A `BinaryExpression` initializer is exactly the shape that survives as text and is then judged by a scan rather than by an evaluator — the same reason `{1 || 2}` folds while `const c = 1 || 2; {c}` does not.

## How it was found, and what no gate sees

It fell out of the #3426 / #3428 template-position grid (10 script shapes × 7 tail-comment shapes × 21 templates, 2,940 comparisons). Of the 121 client cells still diverging there after the server fix, **120 are comment attribution and this is the 1 that is not** — it is the only diverging cell whose source contains no comment at all, which is how it was separated out rather than absorbed into the total.

That is also the reason it needs its own issue: nothing in that grid was aimed here. The grid varies the *template position*, and `<svelte:head><title>` happens to be the only position whose codegen asks "can this be nullish". No comment family, and no corpus entry, is looking at that question.

Worth stating explicitly for whoever picks this up: **the coalesce is emitted from one site, but the predicate behind it is not the one the server uses.** Fixing the `<title>` arm alone would leave the split intact; the interesting question is which of the three client "is it known" implementations answers here and whether it can be routed through `EvalValue` like the fold was.

baseballyama:
## Summary
- route client title/template definedness checks through the shared `scope.evaluate` model
- remove the duplicate expression-kind table that disagreed with upstream on binary expressions
- cover prop and store reads in bare and binary forms in prod and dev modes

Fixes #3599

## Validation
- `git diff --check`
- full build and test suite deferred to CI as requested

### DIFF
--- a/.changeset/share-title-definedness-evaluator.md
+++ b/.changeset/share-title-definedness-evaluator.md
@@ -1,5 +1,5 @@
 ---
-'@rsvelte/core': patch
+"@rsvelte/compiler": patch
 ---
 
 Use the shared Svelte expression evaluator when deciding whether title values need nullish coercion.
--- a/.changeset/share-title-definedness-evaluator.md
+++ b/.changeset/share-title-definedness-evaluator.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/core': patch
+---
+
+Use the shared Svelte expression evaluator when deciding whether title values need nullish coercion.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/shared/utils.rs
@@ -4621,13 +4621,10 @@ pub(crate) fn is_js_expr_defined(
 }
 
 /// Resolve a bare identifier to its binding and decide whether upstream's
-/// `scope.evaluate(<identifier>).is_defined` would hold. Shared by the
-/// original-expression path (`is_expression_defined_json`) and the
-/// transformed-value path (`is_js_expr_defined`): upstream's `scope.evaluate`
-/// resolves identifiers through the scope in BOTH cases, so a non-reactive
-/// binding that survives transformation as a bare identifier (e.g. a legacy
-/// `let iconAsc = "↑"` inside a `cond ? iconAsc : iconDesc`) must resolve the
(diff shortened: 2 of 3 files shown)


baseballyama:
Superseded by #3801, which routes title definedness through the same shared evaluator and includes the title regression coverage.

PR #3790, opened 2026-08-25
fix(client): resolve dev console arguments in lexical scope

### LINKED ISSUE #3619: Client dev: the console-wrap decision resolves a shadowing parameter to the outer binding, and treats a regex literal as unknown
Oracle: `submodules/svelte` @ **5.56.9**.

The dev `console.log` wrap follows `scope.evaluate(arg).has_unknown` (#2028). A **function
parameter** is unknown to that evaluation, so upstream wraps a read of it; rsvelte resolves the
name to the component-scope binding of the same name — which it can evaluate — and skips the
wrap.

```svelte
<script>
	let a = $state(1);

	function f(a) {
		console.log(a);
	}

	f(2);
</script>
<b>{a}</b>
```

`generate: 'client'`, `dev: true`:

```js
// official                                            // rsvelte
console.log(...$.log_if_contains_state('log', a));     console.log(a);
```

Rename the parameter and both sides wrap, so this is name resolution and not the wrap decision:
the argument is a *shadowing* binding, and the predicate answers for the outer one.

A second row of the same shape, from the same grid, is a **regex literal**:

```svelte
<script>
	let a = $state(1);
	const r = /\$inspect\(a\)/;
	console.log(r);
</script>
<b>{a}</b>
```

```js
// official                 // rsvelte
console.log(r);             console.log(...$.log_if_contains_state('log', r));
```

Here it runs the other way: a regex literal is a KNOWN value to `scope.evaluate`
(`EvalValue::Regex`), so upstream does not wrap, and rsvelte does. Both directions of one
predicate, which is why they belong in one issue.

Measured on the 384-cell `$inspect` host grid behind #3606 and the 60-cell carrier grid behind
#3605.

baseballyama:
Fixes #3619.\n\nThe generated-script console pass now builds lexical scope information and records references resolved below program scope. Function parameters and local declarations therefore no longer fall through to same-named component bindings when applying the upstream scope.evaluate wrap rule.\n\nThis also pins the issue's regex-known and nested-console directions in integration coverage and adds a pattern-corpus repro.\n\nLocal verification was intentionally limited to cargo fmt --all -- --check and git diff --check; build and test execution is delegated to CI.

### DIFF
--- a/.changeset/console-wrap-local-scope.md
+++ b/.changeset/console-wrap-local-scope.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Resolve dev console arguments against function-local bindings before same-named component bindings.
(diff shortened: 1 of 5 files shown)


baseballyama:

PR #3799, opened 2026-08-25
fix(client): share evaluator for template folding

### LINKED ISSUE #3439: `void <unknown>` is not folded to `undefined` in the client — 16 cells

### LINKED ISSUE #3466: Client: the static-text decision for a {@const} disagrees with upstream's globals table in both directions
## Summary

The **client**'s "is this `{@const}`'s value static?" decision disagrees with upstream in
**both** directions on a global keypath. When it says "not static" the text node gets a
spurious `$.template_effect(() => $.set_text(...))` where official assigns
`text.nodeValue` once; when it says "static" for a keypath upstream does not know, the
effect that official emits is missing.

```svelte
{#if true}{@const c = Number.isInteger(1)}{c}{/if}
```

```js
// official (client)
const c = $.derived_safe_equal(() => Number.isInteger(1));
var text = $.text();

text.nodeValue = $.get(c);

// rsvelte (client)
const c = $.derived_safe_equal(() => Number.isInteger(1));
var text = $.text();

$.template_effect(() => $.set_text(text, $.get(c)));
```

## Measurement

16 `{@const}` initializers, `generate: 'client'` (identical on `client-dev`), one compiler
process per case. "assignment" = `text.nodeValue = …`, "effect" = `$.template_effect(…)`.

|---|---|---|---|
| `1 + 1` | assignment | assignment | ok |
| `"a" + "b"` | assignment | assignment | ok |
| `true` | assignment | assignment | ok |
| `Math.max(1, 2)` | assignment | assignment | ok |
| `Math.min(1, 2)` | assignment | assignment | ok |
| `Math.abs(-1)` | assignment | assignment | ok |
| `Math.PI` | assignment | assignment | ok |
| `Boolean(1)` | effect | effect | ok |
| `Math.random()` | effect | effect | ok |
| `JSON.stringify(1)` | effect | effect | ok |
| `Number.isInteger(1)` | assignment | **effect** | under |
| `Number.parseInt("3")` | assignment | **effect** | under |
| `String("A")` | assignment | **effect** | under |
| `String.fromCharCode(65)` | assignment | **effect** | under |
| `Math.f16round(1.337)` | assignment | **effect** | under |
| `Number.MAX_SAFE_INTEGER` | effect | **assignment** | **over** |

The last row is the opposite direction and the one a "rsvelte emits extra effects" reading
would miss: upstream's `global_constants` (`phases/scope.js:77-86`) holds exactly the eight
`Math.*` constants and **not** `Number.MAX_SAFE_INTEGER`, so official keeps it unknown while
rsvelte folds it.

The five under-reads are all keypaths that *are* in upstream's `globals`
(`scope.js:64-73`) — `Number.isInteger`, `Number.parseInt`, `String`,
`String.fromCharCode`, plus `Math.f16round` — so the client's table is a different, smaller
set than upstream's rather than a different *rule*.

## Scope and relation to #3388

Server output for the same 16 cases is unaffected — the server folds by value through
`3_transform/server/evaluate.rs`, which consults the real table. This is the client's own
known-ness machinery, which AGENTS.md already records as having three implementations
(`is_expression_known_json`, `identifier_has_reactive_state`,
`is_initial_value_literal_or_known`).

#3388 is a different defect in the same neighbourhood and does **not** cause this: the rows
above reproduce on an unmodified tree, and `Number.isInteger(1)` / `String("A")` — neither of
which #3388 touches — are among them.

## Why nothing caught it

A spurious `template_effect` produces the same rendered output and the same reactivity for a
value that never changes, so no runtime test separates them; only output equality can. The
`constant-fold` matrix family reaches the fold on every run but its rows are all shapes where
the two answers agree.

### LINKED ISSUE #3570: client-dev only: a declaration whose initializer contains `===`/`!==`/`==`/`!=` is never folded, and gets a spurious `template_effect`

### LINKED ISSUE #3636: Client: a const aliasing a function declaration is folded to a one-shot `textContent`, dropping the reactive text placeholder

baseballyama:
Closes #3439.
Closes #3466.
Closes #3570.
Closes #3636.

## What changed

- route client template constant folding through the same typed `Evaluation` walk used by SSR and the client `is_known` path
- remove the second recursive client evaluator and the source-text/JSON literal heuristic
- preserve client-only semantics for call memoization, transformed root identifiers, and dev equality lowering via a narrow evaluator override
- extend the pattern repro and unit coverage to the direct `x{void p}` path; keep `!p` and `-p` as negative controls
- evaluate binding initializers as unconverted source AST so all four equality operators still fold in client-dev
- use the complete globals table for `{@const}` knownness and preserve the unknown FUNCTION marker through aliases of function declarations

`void <unknown>` now has the single known value `undefined`, matching upstream, while operand-dependent unary expressions remain unknown. Client-dev also no longer emits a spurious effect for a constant whose initializer contains equality. Foldable `Number`, `String`, and `Math` calls take the one-shot `{@const}` path, while marker-only globals and function-declaration aliases remain reactive.

## Validation

- `git diff --check`
- `jq empty .changeset/config.json`

Per repository instructions, no local build or test suite was run; CI is the execution oracle.


baseballyama:

PR #3800, opened 2026-08-25
fix(analyze): evaluate known derived reads

### LINKED ISSUE #3437: A `$derived` with a known-constant argument is reactive in rsvelte and not upstream — 12 cells, and the rule is `is_known`

baseballyama:
Closes #3437.

Stacked on #3799, which makes the existing `scope.evaluate` port reusable by non-server callers.

## What changed

- add a Phase 2 resolver for the shared `EvalScope` value model
- apply upstream's complete Identifier `has_state` condition in both Phase 2 identifier paths
- preserve lexical resolution without initializing Phase 3's immutable reference cache during analysis
- pin the reported 12-cell grid plus transitive-known and block-body-unknown controls
- add a pattern-corpus regression and changeset

## Verification

- `git diff --check`
- `jq empty compatibility/known-failures.client.json`

Per project instructions, compilation and test execution are delegated to CI.


baseballyama:

PR #3801, opened 2026-08-25
fix(client): unify definedness evaluation

### LINKED ISSUE #3259: Client: `<option value={…}>` keeps `?? ''` for arrow, function, binary and unary expressions official knows are non-nullish

### LINKED ISSUE #3599: client: <svelte:head><title> emits a `?? ''` upstream drops when the value cannot be nullish

### LINKED ISSUE #3601: client: a const initialized by an assignment or a sequence drops the template chunk's `?? ''` guard

baseballyama:
## Summary

- route original-expression `is_defined` checks through the shared typed `Evaluation`
- reuse the transformed-AST definedness helper for special element values and cover function expressions
- keep memo placeholders unknown while matching upstream for title bindings and assignment/sequence initializers
- add focused unit coverage and a four-target pattern-corpus repro

Closes #3259.
Closes #3599.
Closes #3601.

## Validation

- `cargo fmt --all`
- `git diff --check`
- no local build or test run (CI is the requested execution environment)

## Stack

This is stacked on #3800 (`fix/3437-known-derived-reactivity`).


baseballyama:

PR #3803, opened 2026-08-25
fix(css): accept custom property block values

### LINKED ISSUE #3052: upstream(svelte): CSS custom property with a {}-block value is rejected (css_expected_identifier)
### Summary

The official Svelte CSS parser rejects a custom property whose value contains a `{}` block:

```svelte
<p>x</p>
<style>
p { --tpl: { color: red; }; }
</style>
```

```
css_expected_identifier — Expected a valid CSS identifier
```

Per css-variables, a custom property's value is `<declaration-value>`, which admits arbitrary
token streams including blocks; browsers and PostCSS/Prettier accept this input. rsvelte
deliberately reproduces the rejection byte-for-byte (error parity), so this is not an rsvelte
defect — it is tracked here so the upstream report has an anchor.

Found by the adversarial CSS pattern sweep (PR #3051 campaign); the candidate file is held out
of `compatibility/pattern-corpus/adversarial/css/` until upstream decides the intended behavior.

### Action

Report to sveltejs/svelte; record in `upstream_issues/`.

baseballyama:
Closes #3052.

## Summary

- parse balanced `{ ... }` and `[ ... ]` blocks inside CSS custom-property values without treating their internal semicolons as declaration terminators
- preserve string, escape, and comment boundaries while scanning those values
- recognize custom-property declarations in declaration-taking at-rules such as `@font-face`
- keep ordinary invalid declarations such as `color: { red }` on the existing error path

Upstream currently rejects this valid custom-property syntax. This PR keeps rsvelte's semantically correct behavior, records the deliberate divergence, and updates the upstream-issue report accordingly.

## Verification

- `cargo fmt --all`
- `git diff --check`

Per repository instructions, Rust builds and tests were not run locally; CI is the execution oracle.


baseballyama:

PR #3804, opened 2026-08-25
fix(esrap): preserve side-effect import attributes

### LINKED ISSUE #3635: A side-effect import loses its import-attributes clause (upstream esrap; rsvelte matches)
Found while fixing #3352. **rsvelte matches official byte-for-byte here, so this is not a parity failure** — it is a defect both compilers have, and the question is whether to keep reproducing it.

A **side-effect** import loses its import-attributes clause in the generated output, on every target and whatever line the clause was written on:

```svelte
<script>
	import "./d.json" with { type: "json" };
	let z = 1;
</script>
{z}
```

| target | official 5.56.9 | rsvelte |
|---|---|---|
| client | `import "./d.json";` | same |
| server | `import "./d.json";` | same |

An import with any specifier keeps its clause (`import d from "./d.json" with { type: "json" };` is emitted intact), so this is specific to the no-specifier form.

## Cause (upstream, in esrap)

`esrap@2.2.12`, `src/languages/ts/index.js:1282`:

```js
ImportDeclaration(node, context) {
    if (node.specifiers.length === 0) {
        write_keyword(context, node, 'import', ' ');
        context.visit(node.source);
        context.write(';');
        return;                       // ← never writes node.attributes
    }
    …
```

The early return for the specifier-less form skips the clause; the general path below it writes one.

## Why it matters

The attribute is not decoration. A JSON module imported without `with { type: "json" }` is a **runtime** failure in a host that enforces the check (Node, and browsers), so the emitted module can throw on import where the source would not have. Nothing in the compile reports it.

Same shape, different keyword: `import "./sheet.css" with { type: "css" };` for a constructable-stylesheet side-effect import is the case most likely to occur in real code, since side-effect CSS imports are common.

## What this repo currently does

`crates/rsvelte_core/tests/import_attributes_clause_3352.rs::a_clause_after_a_side_effect_import_matches_official` **pins the current behaviour**: the clause is absent, and nothing is left behind in the component body. That test was written to hold parity, not to bless the output — it is the place to change if we decide otherwise.

Two options, and they are not equivalent:

1. **Report upstream and keep reproducing it** (`upstream_issues/`), which is what byte-equality with official requires today. The corpus gate would go red on the day upstream fixes esrap, which is the intended signal.
2. **Diverge deliberately** and emit the clause, accepting a known corpus divergence in exchange for output that does what the source says.

Filing so the choice is recorded rather than made by default.

baseballyama:
Closes #3635.

## Summary

- preserve an import-attributes clause when printing a side-effect import
- reuse the existing attribute printer used by imports with specifiers
- reverse the compiler regression test so client, server, and client-dev must all retain the clause
- record the upstream esrap defect and the deliberate semantic divergence

Official esrap returns before printing attributes for an import with no specifiers. Dropping the clause can make the emitted JSON or CSS module fail to load, so this PR keeps the source program's semantics instead of reproducing the byte output.

## Verification

- `cargo fmt --all`
- `git diff --check`

Per repository instructions, Rust builds and tests were not run locally; CI is the execution oracle.

### DIFF
--- a/.changeset/side-effect-import-attributes.md
+++ b/.changeset/side-effect-import-attributes.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/core": patch
+---
+
+Preserve import attributes on side-effect imports.
--- a/.changeset/side-effect-import-attributes.md
+++ b/.changeset/side-effect-import-attributes.md
@@ -1,5 +1,5 @@
 ---
-"@rsvelte/core": patch
+"@rsvelte/compiler": patch
 ---
 
 Preserve import attributes on side-effect imports.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2641,7 +2641,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_esrap"
-version = "0.10.35"
+version = "0.10.36"
 dependencies = [
  "compact_str",
  "memchr",
--- a/compatibility/deliberate-divergences.md
+++ b/compatibility/deliberate-divergences.md
@@ -1,9 +1,9 @@
 # Deliberate divergences from the official compiler
 
 Output must match the official compiler exactly, because upstream is the specification.
-That rule does not extend to reproducing bytes that are **not valid JavaScript**: a module
-that no parser accepts is a defect a byte match cannot pay for. Where the two conflict,
-correctness wins.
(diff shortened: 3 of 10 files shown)


baseballyama:

PR #3805, opened 2026-08-25
Fix awaited autofocus and event attributes

### LINKED ISSUE #3651: upstream: `autofocus={await …}` and an event attribute `on*={await …}` emit `await` in a non-async function (5.56.9 and 5.56.10)
Found while establishing the mechanism for #3621. **This is an upstream defect that rsvelte reproduces faithfully**, so no gate here can see it — every comparison in this repo is rsvelte against the pinned official compiler.

Measured against **both** `svelte@5.56.9` (the pin) and `svelte@5.56.10`, `generate: 'client'`, `experimental.async: true`, prod and dev:

```svelte
<script>
	const p = Promise.resolve(true);
</script>

<input autofocus={await p} />
```

```js
// official, both versions
$.autofocus(input, await p);
```

```svelte
<div onclick={await p}></div>
```

```js
// official, both versions — non-delegated (`onfocusin`) is identical
$.delegated('click', div, function (...$$args) {
	(await p)?.apply(this, $$args);
});
```

acorn on either: `Cannot use keyword 'await' outside an async function`. The enclosing function is not async in both cases, so the emitted module is a `SyntaxError` and the build fails at bundle time.

## Why these two slots

`RegularElement.js` dispatches attributes in the order `autofocus` → `class` → `style` → custom element → everything else. `autofocus` calls `build_attribute_value(attribute.value, context)` with no memoize callback and pushes `$.autofocus(node_id, value)` straight into `state.init`; event attributes go to `visit_event_attribute`, which wraps the value in a plain `function (...$$args)`. Neither path touches `Memoizer`, so neither can hoist the `await` into `template_effect`'s async list the way every other attribute slot does.

The **third** slot with the same shape — a custom element's attribute — is fixed in 5.56.10 by giving `build_custom_element_attribute_update_assignment` a `Memoizer`; these two were not touched by that change.

## What to do here

Report upstream, and record it under `upstream_issues/` so the divergence has a decision attached rather than silently agreeing with a broken oracle. Note that this class is *structurally* invisible to every gate in the repo: `matrix/run.mjs:268-270` routes an unparseable **official** output to a run abort rather than to a verdict, so "upstream emits non-JavaScript" can never become a ratchet row. Recorded as gate-coverage blind spot 5r.

baseballyama:
## Summary

- route awaited autofocus values through a local template-effect memoizer
- resolve awaited event handlers before registering them, including dev-mode call expressions
- record the deliberate divergence from Svelte 5.56.10 and pin parseable output

## Validation

- `git diff --check`
- Rust builds/tests intentionally deferred to CI per repository workflow

Closes #3651

### DIFF
--- a/.changeset/async-autofocus-events.md
+++ b/.changeset/async-autofocus-events.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Emit parseable client code for awaited `autofocus` and event attributes.
--- a/compatibility/deliberate-divergences.md
+++ b/compatibility/deliberate-divergences.md
@@ -284,3 +284,24 @@ uncoded panic instead of a diagnostic.
   compiler at all, so no published source carries the shape.
 - **Output-parseability gate**: rsvelte's output is valid JavaScript either way — the divergence is
   whether the input is accepted, which that gate does not ask.
+
+---
+
+## Awaited `autofocus` and event attributes (client)
+
+**Pinned by** `crates/rsvelte_core/tests/async_autofocus_event_3651.rs`.
+**Reported upstream** in
+`upstream_issues/3651-svelte-async-autofocus-and-event-output-is-unparseable.md`.
+
+With `experimental.async: true`, official Svelte 5.56.10 emits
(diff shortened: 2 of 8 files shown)


baseballyama:

PR #3809, opened 2026-08-25
fix(parser): accept let declarations without initializers

### LINKED ISSUE #3705: `{let x}` without an initializer is rejected, and official accepts it
`{let x}` — a declaration tag with no initializer — is rejected, and official accepts it.

## Repro

```svelte
<script>
	const obj = { a: 1 };
</script>

{let x}
```

| | verdict |
|---|---|
| official | compiles — a `VariableDeclaration` whose single declarator has `init: null` |
| rsvelte | `declaration_tag_invalid_type: Declaration tags can only contain \`let\` or \`const\` variable declarations` |

An over-rejection: rsvelte refuses a document the official compiler accepts, which
is a population no collected corpus can hold.

## Cause

`try_parse_declaration_tag` (`phases/1_parse/state/tag.rs`) decides the shape of a
declaration by splitting the body at the first top-level `=`. Upstream instead
parses the whole statement with acorn and reads the node it gets back, where an
absent initializer is simply `init: null`.

The localisation is what makes it small: the **multi**-declarator path already
handles a missing initializer —

```rust
None => (seg.to_string(), String::new(), seg_off + seg.len()),
…
let init_value: Value = if init_str.is_empty() { Value::Null } else { … };
```

— so `{let x, y}` and `{let x = 1, y}` are both correct today. Only the single
declarator, which returns early before reaching that builder, is wrong.

## Measured

`g-declinit.mjs` — 13 bodies × 3 slots (top level / `{#if}` / `{#each}`) × 2
targets = **78 cells, 6 diverging**, and all 6 are the one `{let x}` row (three
slots × two targets). The controls move nothing: `{let x = 1}` and `{const x = 1}`
compile, `{let x, y}` and `{let x = 1, y}` already agree, and `{const x}` /
`{let }` / `{const }` are `js_parse_error` on both sides — a `const` without an
initializer is not JavaScript.

## Not in this issue

`{const x}` and `{let }` agree on the error **code** and diverge on its message
and position:

|---|---|---|
| `{const x}` | `js_parse_error` @4:8 `Unexpected token` | `js_parse_error` @4:7 `Missing initializer in const declaration` |
| `{let }` | `js_parse_error` @4:1 `The keyword 'let' is reserved` | `js_parse_error` @4:4 `Unexpected token` |

The grid above compares `code`, so it scores those two as matching — the same
key limitation the compiler-error gates split `message` / `start` / `end` apart
for. The cause is different (upstream's `parse_statement_at` runs acorn over the
**rest of the template**, not over the tag's slice, so its errors are reported
against text rsvelte never hands to the parser) and it belongs in its own issue.

baseballyama:
Closes #3705.

## What changed

- recognize a successfully parsed single `let` variable declaration even when it has no initializer
- reuse the existing declarator builder so the ESTree initializer is `null` with source-accurate spans
- keep non-declaration statement shapes and initializer-less `const` declarations rejected
- add AST and client/server coverage for top-level, `{#if}`, and `{#each}` slots

## Verification

- `git diff --check`
- changeset JSON configuration validation

Per project guidance, Rust builds and tests were not run locally; CI is the execution oracle.

### DIFF
--- a/.changeset/calm-otters-declare.md
+++ b/.changeset/calm-otters-declare.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Accept declaration tags containing a `let` declarator without an initializer.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -1981,6 +1981,27 @@ pub fn check_js_statement_parse_error(content: &str, ts: bool) -> Option<(String
     })
 }
 
+/// Whether `content` parses as a JavaScript/TypeScript variable declaration.
+///
+/// Upstream's declaration-tag reader distinguishes a successfully parsed
+/// `VariableDeclaration` from an `ExpressionStatement`; parse success alone is
+/// not enough to commit a `{let ...}`-shaped tag to the declaration path.
+pub fn is_js_variable_declaration(content: &str, ts: bool) -> bool {
+    with_oxc_allocator(|allocator| {
+        let source_type = if ts {
+            SourceType::ts()
+        } else {
+            SourceType::mjs()
+        };
+        let result = OxcParser::new(allocator, content, source_type).parse();
+        result.diagnostics.is_empty()
+            && matches!(
+                result.program.body.first(),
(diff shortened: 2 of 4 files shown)


baseballyama:
Closing as a duplicate of #3748, which implements #3705 and already has a full CI run.

PR #3810, opened 2026-08-25
fix(parser): reject super and await template references

### LINKED ISSUE #3694: `{super.x}` and `{await}` are accepted in a template expression
## Summary

Two identifiers the template-expression parser accepts and acorn does not.
Both are over-acceptances — rsvelte compiles a document the official compiler
rejects — and neither is reached by the reserved-word gate, because the gate's
job is to hand the expression to the real parser and the real parser accepts
them.

|---|---|---|
| `{super.x}` | `js_parse_error` | compiles, emits `super.x` |
| `{await}` | `js_parse_error` | compiles, emits `await` |
| `{await.x}` | `js_parse_error` | compiles, emits `await.x` |

`{super}` on its own already agrees — oxc rejects a bare `Super` and accepts it
as a member object, which is why only the `member` shape survives.

## Where

`check_js_parse_error_with_pos` (`1_parse/read/expression.rs`) already exists
for exactly this class — "the restrictions acorn applies and OXC does not" —
and `acorn_only_violation` is its AST scan, currently covering decorators,
`with`, and TypeScript class modifiers. These two belong there.

Both need a condition rather than a bare node match, which is why they are not
a one-line addition:

- `super` is legal inside a class method, and a `<script>` with
  `class A extends B { m() { return super.m(); } }` is ordinary code. The scan
  has to track method depth, or it becomes an over-**rejection**.
- `await` is only reserved because Svelte parses with `sourceType: 'module'`.
  `obj.await` is a legal property name, so the check has to be on
  `IdentifierReference`, not on the text.

## Measurement

44 reserved words × 3 shapes (`{w}`, `{w.x}`, `{obj.w}`), client target: after
the reserved-word gate is widened, **3 of 147** cells remain over-accepted and
all three are these. The other 9 residual cells are #3692.

The population is programs official rejects, which no collected corpus holds —
0 of the 12,523 collected `.svelte` files reach either shape.

baseballyama:
Closes #3694.

## Summary

- reject top-level `super` references in template expressions with Acorn-compatible wording and position
- reject module-reserved `await` identifier references without rejecting legal property names such as `obj.await`
- preserve legal `super` use in class and object methods
- cover mustache, attribute, and `{@const}` expression slots plus positive controls

## Verification

- `git diff --check`

Per repository instructions, compilation and tests were left to CI.

### DIFF
--- a/.changeset/sharp-cats-wait.md
+++ b/.changeset/sharp-cats-wait.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Reject `super` and module-reserved `await` references in template expressions with Acorn-compatible diagnostics.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -7145,6 +7145,11 @@ fn acorn_only_violation(
         with_at: Option<u32>,
         export_declare_global_at: Option<u32>,
         check_ts_modifier: bool,
+        check_super: bool,
+        super_at: Option<u32>,
+        method_depth: u32,
+        check_await_identifier: bool,
+        await_identifier_at: Option<u32>,
         content: &'c str,
         ts_modifier_at: Option<u32>,
     }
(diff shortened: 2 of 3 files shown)


baseballyama:
Closing as a duplicate subset of #3746, which covers #3694 together with the adjacent template-expression validation gap and already has CI coverage.

PR #3811, opened 2026-08-25
Fix restricted identifiers in templates

### LINKED ISSUE #3707: `arguments` in a template expression is accepted; `invalid_arguments_usage` never fires outside the instance script
`arguments` in a template expression is accepted; `invalid_arguments_usage` never fires outside the instance script

## Repro

```svelte
<script>
	const ok = 1;
</script>

{arguments}
```

| | verdict |
|---|---|
| official | `invalid_arguments_usage: The arguments keyword cannot be used within the template or at the top level of a component` |
| rsvelte | compiles, on both `client` and `server` |

An over-rejection's opposite: rsvelte **accepts a document official rejects**, so
the population is programs that do not compile upstream and no collected corpus
can hold it — published code compiles.

## Cause

The check is not missing. `2_analyze/visitors/identifier.rs:80-92` has it:

```rust
if name == "arguments" {
    let is_in_function = context.js_path.iter().any(|n| {
        matches!(n.get_type_str(), Some("FunctionDeclaration") | Some("FunctionExpression"))
    });
    if !is_in_function {
        return Err(errors::invalid_arguments_usage().at(start, end));
    }
}
```

What is missing is the **visit**: the identifier visitor is never reached for an
identifier that lives in a template expression, only for one in the instance
script. The error message names the template first — "cannot be used within the
template or at the top level of a component" — and the template half is exactly
the half that does not run.

This is the same family as #3652: what a template expression's leading token is
allowed to be, decided by a path that does not see what the script path sees.

## Measured

`g-arguments.mjs` — 7 expression shapes × 8 hosts × 2 targets = **112 cells, 60
diverging**, and the split is total rather than partial:

| host | diverging |
|---|---|
| `{…}` text | 10 / 14 |
| `title={…}` attribute | 10 / 14 |
| `{#if …}` test | 10 / 14 |
| `{#each […] as n}` collection | 10 / 14 |
| `{@const c = …}` | 10 / 14 |
| `{@html …}` | 10 / 14 |
| instance script, top level | **0 / 14** |
| instance script, inside `function f()` | **0 / 14** |

Every template host diverges on every shape that is not inside a `function`
(`arguments`, `arguments.length`, `String(arguments)`, `arguments[0]`, and
`(() => arguments)()` — an arrow has no `arguments` of its own, so upstream
rejects it too). The two script hosts are the positive control naming the path:
the identical expression is rejected there. `(function () { return arguments; })()`
and the ordinary identifier `ok` are the negative controls, and they agree in
all 8 hosts.

### LINKED ISSUE #3728: Template binding hosts accept `arguments` / `eval` as a binding name; the pattern form is unchecked everywhere but `{#snippet}`
### Summary

`arguments` and `eval` are restricted as *binding names* in strict mode, and every Svelte template construct that declares a binding is supposed to reject them. rsvelte rejects them in some hosts and accepts them in others, and the split is not arbitrary: **the plain-identifier check exists in three hosts and the destructuring-pattern check exists in one**.

An over-acceptance, so the population is programs the official compiler rejects — nothing a collected corpus can hold, because published code compiles.

### Repro

```svelte
<script>
	const v = { a: 1, arguments: 1 };
</script>
{#each [1] as { arguments }}<span>ok</span>{/each}
```

| | verdict |
|---|---|
| official | `js_parse_error: Assigning to arguments in strict mode` |
| rsvelte | compiles, on both `client` and `server` |

And the plain form in a const tag:

```svelte
{#if true}{@const arguments = v}<span>ok</span>{/if}
```

| | verdict |
|---|---|
| official | `unexpected_reserved_word: 'arguments' is a reserved word in JavaScript and cannot be used here` |
| rsvelte | compiles |

### Grid

5 template binding hosts × 4 pattern forms × 2 names = **42 cells, 28 diverging**, every one in the same direction (rsvelte accepts what official rejects).

| host | plain | `{ n }` | `[n]` | `{ a: n }` |
|---|---|---|---|---|
| `{@const n = v}` | **diverges** | **diverges** | **diverges** | **diverges** |
| `{#each [1] as n}` | agrees | **diverges** | **diverges** | **diverges** |
| `{#each [1] as q, n}` | **diverges** | — | — | — |
| `{#await v then n}` | agrees | **diverges** | **diverges** | **diverges** |
| `{#await v catch n}` | agrees | **diverges** | **diverges** | **diverges** |
| `{#snippet s(n)}` | agrees | agrees | agrees | agrees |

Two sub-gaps, which the table separates:

* **the plain identifier** is checked in `each as` / `await then` / `await catch` and **not** in `{@const}` or the `{#each … as q, INDEX}` slot — 4 cells;
* **every destructuring form** is unchecked in all four of those hosts — 24 cells.

### Controls

Both directions are present and both move the right way.

* **Ordinary name.** The identical 21 host × pattern cells with the name `q` diverge **0 / 21**, so this is about the name and not about the pattern parser being wrong.
* **The 14 agreeing cells all agree on a REJECTION, not on acceptance** — which is the check worth stating, because a grid that only records disagreement cannot tell "both reject" from "both accept". `each as` / `await then` / `await catch` with a plain name are `unexpected_reserved_word` on both sides.
* **`{#snippet}` rejects all four pattern forms on both sides** (`js_parse_error`). That is the positive control that locates the defect: the pattern path *can* reject these names and does so in one host, so what is missing is the call, not the capability.

### Notes

Measured at Svelte 5.56.10 against `submodules/svelte`, `generate: 'client'`; the server target behaves identically (errors are raised in parse/analyze).

`is_reserved` (`phases/1_parse/utils/mod.rs:77`) already lists both names, and `{@const}` already consults it — but only on the **missing-`=` error path** (`state/tag.rs:2487`), never on the success path.

Official's two codes are not interchangeable and a fix should reproduce both: a plain identifier is `unexpected_reserved_word` positioned at the start of the name, and any destructuring form is acorn's `js_parse_error` with the message `Assigning to <name> in strict mode`.

Found while fixing #3707; the 2 residual cells of that issue's grid are this defect, and they were already diverging before that fix (as an acceptance rather than as a wrong code).

baseballyama:
## Summary

- reject `arguments` references reached through the typed template-expression walker
- distinguish ordinary functions, which provide `arguments`, from arrows, which inherit it
- apply the official restricted-word diagnostic to each indices and `{@const}` bindings
- report Acorn-compatible strict-mode assignment diagnostics for restricted names inside destructuring patterns
- add client/server regression grids for expression hosts, binding hosts, and neighbouring accepted controls

Closes #3707.
Closes #3728.

## Validation

- `git diff --check`

Per repository guidance, Rust builds and tests were not run locally; CI is the execution oracle.

### DIFF
--- a/.changeset/quiet-badgers-restrict.md
+++ b/.changeset/quiet-badgers-restrict.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Match Svelte's diagnostics for `arguments` in template expressions and for restricted template binding names.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -11730,6 +11730,28 @@ fn is_plain_ascii_identifier(s: &str) -> bool {
     bytes.all(|b| b.is_ascii_alphanumeric() || b == b'_' || b == b'$')
 }
 
+struct RestrictedBindingIdentifier {
+    first: Option<(u32, String)>,
+}
+
+impl<'a> oxc_ast_visit::Visit<'a> for RestrictedBindingIdentifier {
+    fn visit_binding_identifier(&mut self, id: &oxc_ast::ast::BindingIdentifier<'a>) {
+        if self.first.is_none() && matches!(id.name.as_str(), "arguments" | "eval") {
+            self.first = Some((id.span.start, id.name.to_string()));
+        }
+    }
+}
+
+fn restricted_binding_identifier(
+    pattern: &oxc_ast::ast::BindingPattern<'_>,
+) -> Option<(u32, String)> {
+    use oxc_ast_visit::Visit;
+
+    let mut visitor = RestrictedBindingIdentifier { first: None };
+    visitor.visit_binding_pattern(pattern);
+    visitor.first
+}
+
 pub fn parse_binding_pattern<'a>(
     arena: &ParseArena,
     content: &str,
@@ -11799,6 +11821,15 @@ pub fn parse_binding_pattern<'a>(
             result.program.body.first()
             && let Some(decl) = var_decl.declarations.first()
         {
+            if let Some((wrapped_start, name)) = restricted_binding_identifier(&decl.id) {
+                let start = offset + wrapped_start as usize - 4;
+                return Err(crate::error::ParseError::svelte(
+                    "js_parse_error",
+                    format!("Assigning to {name} in strict mode"),
+                    (start, start),
+                ));
+            }
+
             if let oxc_ast::ast::BindingPattern::BindingIdentifier(id) = &decl.id {
                 let start = offset + id.span.start as usize - 4;
                 let end = offset + id.span.end as usize - 4;
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/state/tag.rs
@@ -1402,7 +1402,18 @@ impl<'a> Parser<'a> {
                     self.advance();
                 }
             } else {
-                index = Some(CompactString::from(&self.source[idx_start..idx_end]));
+                let name = &self.source[idx_start..idx_end];
+                if super::super::utils::is_reserved(name) {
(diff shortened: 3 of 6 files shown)


baseballyama:
Closing as duplicate coverage of #3746 and #3747, which separately cover #3707 and #3728 and already have CI results.

PR #3813, opened 2026-08-25
Fix legacy export const diagnostic precedence

### LINKED ISSUE #3262: Error parity: a write to an `export const` reports `constant_assignment` at the write, not `state_invalid_export` at the declaration
Writing to an `export const` binding raises a different error code, at a
different position, than upstream.

```svelte
<script>
	export const a = 1;
</script>
<button onclick={() => a = 2}>{a}</button>
```

```
official  state_invalid_export  2:1–2:20  (the declaration)
          "Cannot export state from a module if it is reassigned. Either export
           a function returning the state value or only mutate the state
           value's properties"

rsvelte   constant_assignment   4:23–4:28 (the write)
          "Cannot assign to constant"
```

Upstream checks the *export*: a re-assigned binding may not be exported, so the
diagnostic points at the declaration. rsvelte reaches the generic
const-assignment check first and points at the assignment.

## Scope

|---|---|---|
| `export const a = 1;` + `a = 2` in a handler | `state_invalid_export` @2:1 | `constant_assignment` @4:23 |
| `export const a = 1;` + `a = 2` in a script function | `state_invalid_export` @2:1 | `constant_assignment` @3:16 |
| `export const a = 1;` + `a++` in a handler | `state_invalid_export` @2:1 | `constant_assignment` @4:23 |
| `export const a = 1;` + `<input bind:value={a} />` | `state_invalid_export` @2:1 | **`constant_binding`** @4:7 |
| `const a = 1;` (not exported) + `a = 2` | `constant_assignment` @4:23 | same |
| `export let a = 1;` + `a = 2` | accepted | accepted |
| `export function a() {}` + `a = 2` | accepted | accepted |
| `export class a {}` + `a = 2` | accepted | accepted |

The non-exported control matches exactly, code and both endpoints — so the
generic path is right and only the `export`ed case takes the wrong branch. The
`bind:` row shows a third code, so this is not a single mis-mapped constant.

## Why no gate sees it

The 145 `compiler-errors` fixtures carry one input per code, and both
`constant_assignment` and `state_invalid_export` have a passing fixture, which
reads as covered. Nothing compares *which* of two valid codes an input that
could trigger either should get.

## Found by

Adversarial component-interface grid (24 declaration forms × 17 template uses ×
3 targets = 1,224 cells, 1,017 accepted by official): `export-const` diverged in
exactly the three columns that write to the binding, on all three targets.

baseballyama:
## Summary

- carry scope-resolved template binding references from scope construction into early analysis
- promote directly declared legacy `export const` bindings before the instance visitor when upstream's updated-and-template-referenced criterion applies
- turn all 12 pinned #3262 divergences into exact client/server diagnostic parity assertions

## Why

Upstream constructs its complete scope reference graph before analysis and performs legacy state promotion before visiting exports. rsvelte populated full references during the visitor walk, so a later write produced `constant_assignment` or `constant_binding` before the export could produce `state_invalid_export`.

The pre-analysis set stores binding indices rather than names, preserving template shadowing behavior. The promotion is intentionally restricted to direct legacy `export const` declarations, leaving renamed specifier exports, runes mode, and constants without a qualifying template reference unchanged.

## Validation

- `git diff --check`
- full build and test validation deferred to CI as requested

Fixes #3262

### DIFF
--- a/.changeset/quiet-dragons-agree.md
+++ b/.changeset/quiet-dragons-agree.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Match Svelte's legacy export-const diagnostic precedence when an exported constant is updated and referenced from the template.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/mod.rs
@@ -413,6 +413,7 @@ pub(crate) fn analyze_prepared_component_with_retained(
     let has_export = memchr::memmem::find(source.as_bytes(), b"export").is_some();
     if !analysis.runes && has_export {
         process_legacy_exports(ast, &mut analysis);
+        promote_legacy_export_const_state_bindings(ast, &mut analysis);
     }
 
     // Validate and analyze scripts (JavaScript AST)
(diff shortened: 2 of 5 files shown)


baseballyama:

PR #3814, opened 2026-08-25
test(parse): pin malformed mustache error positions

### LINKED ISSUE #3333: A mustache's extent is decided by a brace scan, so a broken expression reports the wrong js_parse_error position
rsvelte locates a mustache's closing `}` by scanning the source for it; upstream never
searches. `read_expression` hands the rest of the template to acorn, takes whatever one
expression it consumes, and `eat('}', true)` then demands the brace wherever that
stopped. So when the expression itself is what is broken, upstream reports **acorn's**
error at **acorn's** position, and rsvelte reports whatever its scan concluded.

#3210 closed 8 of its 10 rows. These two are the residue, and they are one cause:

| input | official | rsvelte (after #3210) |
|---|---|---|
| `<b>{v</b>` | `js_parse_error` 8:7 | `js_parse_error` 8:9 |
| `{#if flag}{@const c = 1<b>x</b>{/if}` | `js_parse_error` 8:29 | `js_parse_error` (root) / `block_unclosed` (inside a block) |

`find_mustache_close` now reproduces upstream's *decision* when the scan finds no brace
— a complete leading expression makes the first leftover token the point, anything else
is the JS parser's own error — so the **code** agrees on the root shape. Two things it
cannot reproduce:

**1. The position, on this input class.** Both rows fail on an unterminated regular
expression: acorn lexes `v < /b>…` and reports one past the `/`, oxc reports the `/`
itself. Every sample of this shape is off by exactly one, which is exactly why a `+1`
would be wrong — it is a property of the two lexers' error conventions on *this* rule,
not a constant offset, and two samples cannot tell those apart.

**2. The code, once the input is nested in a block.** `find_matching_bracket` balances
the `{` of an enclosing `{/if}` against its `}` and so does find *a* brace, which means
the "no brace anywhere" branch is never reached and the block's own close consumes the
error instead. The scan cannot distinguish a brace that closes this mustache from one
that belongs to the markup after it, because that question is answered by parsing.

## What would actually fix it

Give the expression readers a "parse a prefix expression at this index and tell me where
it ended" entry point, the way `parse_expression_at` works upstream, and stop deciding
the mustache's extent textually. Eight call sites go through `find_mustache_close` today
and would use it; `parse_head_expression` / `parse_js_expression_strict` would take the
slice the parse chose rather than the slice the scan chose.

That is the same "premise of a raw text scan" shape `AGENTS.md` records for
`compileModule` (#2986 / #2987 / #2988) and for the SSR constant fold (#3175), one level
down: the scan is not wrong about braces, it is answering a question that only the JS
parser can answer.

## Why no gate sees it

Both outputs are errors, so the parse oracle has nothing to parse. The corpus output
verdict compares `code` only, which now agrees on the root shape. `message` and `start`
ratchet separately, and `js_parse_error` messages are already the largest listed cluster
(15 of 59 on `error-position-known-failures`), so a new instance enrols into a bucket
that reads as known. The shapes are truncated templates, which no collected corpus holds
— they occur 0 times in the 7,186 collected entries.

Repro: `crates/rsvelte_core/tests/malformed_markup_positions.rs` has the sibling shapes
that do agree; these two are deliberately absent from it.

baseballyama:
Closes #3333

The two residual malformed-mustache shapes are already handled on `main` by the close-tag regular-expression diagnostic recovery retained in #3736. This PR pins both exact upstream points so that behavior cannot regress silently:

- a top-level expression followed by `</b>` reports `js_parse_error` at byte 7;
- the same failure inside `{@const}` in an `{#if}` block reports byte 29 instead of letting the enclosing `{/if}` turn it into a block error.

The assertion deliberately checks `(code, start, end)` through the existing malformed-markup suite. This is rule-specific coverage for Acorn's unterminated-regexp coordinate; it does not introduce a generic position shift.

Static checks run locally:

- `git diff --check`

Per project instruction, compiler tests/builds were not run locally and are left to CI.

### DIFF
--- a/.changeset/fuzzy-rabbits-report.md
+++ b/.changeset/fuzzy-rabbits-report.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Match Svelte's diagnostic position for malformed mustache expressions followed by an enclosing block close.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/parser.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/parser.rs
@@ -1114,7 +1114,20 @@ impl<'a> Parser<'a> {
             expr_start,
             '{',
         ) {
-            return Ok(end);
+            let candidate = &self.source[expr_start..end];
+            let swallowed_block_close = candidate.rfind('{').is_some_and(|block_open| {
+                let block_name = candidate[block_open + 1..].trim();
+                matches!(block_name, "/if" | "/each" | "/await" | "/key" | "/snippet")
+                    && memchr::memmem::find(candidate[..block_open].as_bytes(), b"</").is_some()
+            });
+            // In malformed markup such as `{@const c = 1<b>x</b>{/if}`, the
+            // lexical bracket scan reads `</b>` as the start of a regexp and
+            // the slash in `{/if}` as its end. The final `}` then looks like
+            // this mustache's close. Let the recovery below report the HTML
+            // close-tag position Acorn uses instead.
+            if self.options.loose || !swallowed_block_close {
+                return Ok(end);
+            }
         }
         // Loose mode keeps recovering so a half-typed document still yields a tree.
         if self.options.loose {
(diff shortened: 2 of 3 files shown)


baseballyama:

PR #3815, opened 2026-08-25
fix(parse): match plain-JS TypeScript diagnostics

### LINKED ISSUE #3304: TypeScript syntax without `lang="ts"`: 28 message and 10 position divergences over 46 compared rejections
Every TypeScript-only construct in a script with **no** `lang="ts"` is rejected by both compilers with the same code, and diverges on the two fields the code cannot express. Over 26 constructs × 2 hosts (a component instance `<script>` and `compileModule` on a `.svelte.js`):

| field | divergent | compared |
|---|---|---|
| `code` | **0** | 46 |
| `start` position | **10** | 46 |
| `message` | **28** | 46 |

The `code` column reproduces what the collected corpus already measures — that field is saturated — so this is entirely in the two fields that were invisible until they were captured, and on a population no collected corpus holds, because published code with TypeScript syntax declares `lang="ts"`.

## The position rows

| source (in a plain `<script>`) | official | rsvelte |
|---|---|---|
| `let v = 1 as number;` | 2:10 (at `as`) | 2:8 (at the expression start) |
| `let v = [1] as const;` | 2:12 | 2:8 |
| `let v = 1 satisfies number;` | 2:10 | 2:8 |
| `function f(): number { … }` | 2:12 (at the `:`) | 2:0 (at `function`) |
| `interface I { a: number }` | 2:0 (at `interface`) | 2:10 (past the name) |

They go in both directions, which rules out one constant offset: rsvelte reports the enclosing construct's start for `as` / `satisfies` / a return type, and reports *past* the construct for `interface`. `compileModule` behaves identically, one line up.

## The message rows

28 of 46. rsvelte surfaces oxc's TypeScript-aware text where official surfaces acorn's generic one:

| | |
|---|---|
| official | `Unexpected token` |
| rsvelte | `Type assertion expressions can only be used in TypeScript files` |
| rsvelte (satisfies) | `Type satisfaction expressions can only be used in TypeScript files` |
| rsvelte (generic fn) | `Expected \`(\` but found \`<\`` |

rsvelte's text is arguably more useful; byte equality with the official compiler is the goal, so it still has to match.

## Not in this issue

Three constructs are an **over-acceptance** rather than a field divergence — `readonly a = 1`, `accessor a = 1` and `override m() {}` compile in rsvelte and are `js_parse_error` at 2:19 for official. That is #3203, and the rows are named here only so the denominator adds up: 52 cells, 6 of them accept-mismatches, 46 compared.

## Constructs swept

`as`, `as const`, `<T>x`, `satisfies`, `!`, `let v: T`, a typed parameter, a return type, a generic function, a generic call, `interface`, `type`, `enum`, `declare`, `abstract class`, `implements`, a constructor parameter property, `readonly`, `let v!: T`, an optional parameter, an overload signature, `namespace`, `import type`, `export type`, `accessor`, `override`.

baseballyama:
Closes #3304.

## What changed

- realign OXC's TypeScript-in-JavaScript diagnostics to the token where upstream acorn stops
- reproduce upstream's `Unexpected token` and reserved-keyword wording for the 14 message-divergent constructs
- keep searches lexical: strings, comments, template literals, and regex literals cannot donate `as`, `satisfies`, `interface`, or `enum`
- preserve TypeScript-mode behavior for `lang="ts"` components and `.svelte.ts` modules
- add a patch changeset

## Coverage

The regression test pins all 28 message-divergent cells from the issue matrix: 14 syntax shapes across both a plain component `<script>` and `compileModule` with a `.svelte.js` filename. It asserts code, first-line message, start, and end. It also includes TypeScript-mode controls.

## Verification

- `git diff --check`

Per project instructions, no local build or test suite was run; CI is the execution oracle.


baseballyama:

Issue #3603, opened 2026-08-22
fix(transform): a block comment inside a template expression is dropped in 17 of 24 host/target cells


baseballyama:
Oracle: `submodules/svelte` @ **5.56.9** (asserted from the `compileModule` output header).

## Repro

```svelte
<script>
	let flag = $state(true);
</script>

{#if /* leading */ flag}
	<b>a</b>
{/if}
```

```js
// official (server)
	if (/* leading */ flag) {

// rsvelte (server)
	if (flag) {
```

A block comment inside a **template** expression is dropped. Comments inside the
`<script>` are kept — this is the template side only, where the expression is
re-printed rather than carried through.

## Grid: 17 of 24 cells

| expression | client | server |
|---|---|---|
| `{#if /* c */ flag}` | match | **dropped** |
| `{#if flag /* c */}` | match | match |
| `{#each /* c */ rows as row}` | **dropped** | **dropped** |
| `{#each rows as row (/* c */ row)}` | **dropped** | **dropped** |
| `{#key /* c */ flag}` | **dropped** | match |
| `{#await /* c */ pending then v}` | **dropped** | **dropped** |
| `{#await pending /* c */ then v}` | **dropped** | **dropped** |
| `{#snippet body(/* c */ v)}` | **dropped** | **dropped** |
| `{@render /* c */ body(1)}` | **dropped** | **dropped** |
| `<b>{/* c */ flag}</b>` | **dropped** | match |
| `<b data-a={/* c */ flag}>` | **dropped** | **dropped** |

Official's placements are often odd — `$.each(node, 17, (/* c */) => rows, …)`
puts the comment in an empty parameter list, and `$.append($$anchor, b /* c */)`
moves it past the whole template — but byte equality is the goal, and the corpus
gate's oxfmt normalization preserves comments, so every one of these is a
divergence the gate reports.

The per-target pattern is the interesting part: three rows diverge on exactly one
target (`{#if}` leading on server only, `{#key}` leading and a bare expression
tag on client only). So this is not one dropped-comment site — the client and
server visitors each re-print a different subset of these expressions, and they
lose comments in different places.

Found by a generated adversarial corpus candidate
(`control-flow/block-header-comments-and-spacing.svelte`), held out of
`compatibility/pattern-corpus/adversarial/` until this is fixed. The same file
also carries #3602.

baseballyama:
[...]

PR #3816, opened 2026-08-25
fix(compiler): finish template comment cursor parity

### LINKED ISSUE #3070: Comment placement: five classes still diverge in the JS regions (server interior slot, own-line spill, dangling tail, dev tag, template leaks)

### LINKED ISSUE #3098: Comments in a template expression tag and at the end of the instance script are dropped

baseballyama:
## Summary

Carry comments written inside rebuilt template/block expressions through the same source-position cursor the official compiler uses.

This completes the measured #3098 slots for ordinary expression tags, each collections, `{@const}`, event-handler rewrites, `{@html}`, `{#key}`, and `{@render}`, plus the corresponding server if/each/const/html/render paths. It also fixes #3070's remaining server interior binding-name row while preserving the documented deliberate divergence for upstream's unrelated template-comment leaks.

The implementation extends the existing client `SourceAnchor` region mechanism to non-`{expr}` openers and registers the equivalent expression regions in the server chunk registry. Placement stays on the source identifier/expression rather than on its generated wrapper.

This also addresses the matching if/each/key/html/render rows reported in #3603; await headers, snippet parameters, and each keys remain separate generated positions and are not claimed here.

## Verification

- `git diff --check`
- behavioral verification: GitHub Actions only (no local build or test run)

Tracks #3070.
Closes #3098.
Contributes to #3603.

### DIFF
--- a/.changeset/template-block-expression-comments.md
+++ b/.changeset/template-block-expression-comments.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Comments inside template block and directive expressions now follow the official compiler's source cursor on client and server output, including each collections, if and key tests, html and render arguments, and event handlers.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/source_anchor.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/source_anchor.rs
@@ -65,7 +65,11 @@ impl CommentRegion {
             oxc_span::SourceType::mjs().with_typescript(true),
         )
         .parse();
-        if !ret.diagnostics.is_empty() || ret.program.comments.is_empty() {
+        // A caller may deliberately span template punctuation between two
+        // source nodes (for example an each header and a following `{@const}`).
+        // The slice is not necessarily a valid JS program, but the lexer still
+        // reports its comments accurately.
+        if ret.program.comments.is_empty() {
             return None;
         }
         let comments = ret
@@ -101,6 +105,26 @@ impl CommentRegion {
             comments: self.comments.clone(),
             at,
             at_end,
+            preserve_inner_spans: false,
+        }))
+    }
+
+    /// As [`Self::anchor`], but retain source spans on descendants of a
+    /// generated wrapper. This is needed when upstream's cursor reaches an
+    /// original identifier inside a rebuilt array/call before it reaches the
+    /// wrapper itself.
+    pub fn anchor_inner(&self, arena: &JsArena, expr: JsExpr, at: u32, at_end: u32) -> JsExpr {
+        if at < self.start || at_end > self.end {
+            return expr;
+        }
+        JsExpr::SourceAnchored(Box::new(JsSourceAnchor {
+            inner: arena.alloc_expr(expr),
+            region_start: self.start,
+            region: self.text.clone(),
+            comments: self.comments.clone(),
+            at,
+            at_end,
+            preserve_inner_spans: true,
         }))
     }
 }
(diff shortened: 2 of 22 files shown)


baseballyama:

PR #3818, opened 2026-08-25
fix: preserve keyed-each expression comments

### LINKED ISSUE #3603: fix(transform): a block comment inside a template expression is dropped in 17 of 24 host/target cells

baseballyama:
## Summary

- preserve leading block comments in client keyed-each callbacks by anchoring the lowered key expression to the source inside its parentheses
- carry the server-only skipped key region to the next source-located body expression, matching upstream's global esrap cursor without evaluating or emitting the key on SSR
- add a four-target differential corpus fixture, including a `(` inside the comment so the delimiter lookup must remain lexical-aware

## Stack

- base: #3817
- #3816 covers the template-expression rows shared by ordinary expression visitors
- #3817 covers await headers and snippet parameters
- this PR covers the final client/server keyed-each row and completes #3603

## Validation

- `cargo fmt --all`
- `git diff --check`
- no local build or test suite was run; GitHub Actions is the build, unit-test, and four-target differential oracle

Closes #3603


baseballyama:

PR #3820, opened 2026-08-25
fix(client): place props initializer comments after defaults

baseballyama:
## Summary

- retain comments between a destructured `$props()` assignment and the rune call before call canonicalization removes the separator
- feed those comments through the existing esrap-style trailing-comment cursor: after the last default argument when it is source-located, otherwise after the generated declaration
- cover block, line, multiple, multiline, plain-property, and rest-property forms in client and client-dev output

## Scope

This is the client `$props()` placement slice of #3515. The already-merged #3415 covers client `$state`/`$derived` declarators. Server `$props()` placement and the separate `{@const}` comment-retention channel remain separate follow-ups because they have different causes and output slots.

## Validation

- `git diff --check`
- no local build or test run, per repository workflow; CI is the execution oracle

Refs #3515

### DIFF
--- a/.changeset/props-initializer-comments.md
+++ b/.changeset/props-initializer-comments.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Match Svelte's client output placement for comments between a destructured `$props()` assignment and the rune call.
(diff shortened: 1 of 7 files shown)


baseballyama:

Issue #3306, opened 2026-08-21
Official emits unparseable output for a write to an `{#each}` rest binding and after an `{:catch}` shadow — rsvelte is the faithful side


baseballyama:
Three inputs make the **official** compiler emit client output no JS parser
accepts. rsvelte's output parses, so these are cells where rsvelte is the more
faithful side and byte equality is what forces the question.

## 1. A write to an `{#each}` rest binding

```svelte
{#each [["A", "B"]] as [first, ...v]}
	<button onclick={() => { v = "W"; }}>b</button>
{/each}
```

```js
// official
($$array.slice(1) = "W");     // Assigning to rvalue

// rsvelte
v = "W";
```

The object-rest spelling does the same with a different read expression:

```svelte
{#each [{ a: 1, b: 2 }] as { a, ...v }}<button onclick={() => { v = "W"; }}>b</button>{/each}
```
```js
($.exclude_from_object($$item, ['a']) = "W");   // official
```

## 2. A write to an outer binding after an `{#await … :catch v}` that shadowed it

```svelte
<script>
	let v = "OUTER";
</script>
{#await Promise.reject("A")}w{:catch v}{String(v)}{/await}
<button onclick={() => { v = "W"; }}>b</button>
```

```js
$.get(v) = "W";     // official — Assigning to rvalue
```

The write is outside the block and targets the outer `let`, which is correct
source. Only `{:catch}` does it — the control matters here:

| shadowing construct | official output |
|---|---|
| `{:catch v}` | **unparseable** |
| `{#await … then v}` | parses |
| `{#each … as v}` | parses |
| `{#snippet s(v)}` | parses |
| `{@const v = …}` | parses |
| no shadowing at all | parses |

## Common shape

All three put the binding's **read** expression on the left-hand side of an
assignment. `$$array.slice(1)`, `$.exclude_from_object(…)` and `$.get(v)` are
each what a *read* of that binding compiles to.

## Scope

Both compilers **accept** these inputs — no error, no warning — so the parse gate
is the only thing that sees them, and only on official's side. A plain
`{#each … as v}` item write is byte-identical between the two, which is the
control showing this is the rest/catch path rather than each-writes in general.

Found in the no-collision control of the shadowing sweep (#3296): 14 cells over
1,694, both targets.

## What to do here

Byte equality is the goal, so the precedent is #2990 / `client/dead_comments.rs`:
rsvelte reproduces an upstream defect deliberately rather than shipping the
[...]

PR #3821, opened 2026-08-25
fix(client): resolve destructured each writes lexically

### LINKED ISSUE #3296: Template-scope shadowing: a read resolves to the outer binding, a write to a destructured each item is lost, and the outer is wrongly made reactive
When a template construct introduces a binding whose name already exists in the
instance scope, four things go wrong — a read resolved to the wrong binding, a
write that never lands, an outer binding needlessly made reactive, and an
over-broad invalidation. All four need the two bindings to differ in
**reactivity**, which is what #3215 and #3033 each caught one cell of.

## 1. A read resolves to the OUTER binding (server)

```svelte
<script>
	let base = $state(1);
	let v = $derived(base + 1);
</script>
{#each ["A"] as v}{#if true}{@const c = String(v)}{c}{/if}{/each}
```

```js
// rsvelte  (server)          // official
const c = String(v());        const c = String(v);
```

`v()` is the outer `$derived`'s accessor. The `{@const}` renders the derived
value instead of the each item — a wrong value at runtime, not a spelling
difference. 10 cells: the outer `$derived` × `each_item` / `each_index` /
`each_destructured` / `each_rest` / `each_keyed` / `await_then` / `each_in_each`
/ `snippet_param` / `let_directive`, all reading through a `{@const}`, server
target.

## 2. A write to a destructured each item is lost, and the outer is made reactive

```svelte
<script>
	let v = "OUTER";
</script>
{#each [{ v: "A" }] as { v }}<button onclick={() => { v = "W"; }}>b</button>{/each}
```

```js
// rsvelte                                 // official
let v = $.mutable_source("OUTER");         let v = "OUTER";
  let v = () => $$item.v;                    let v = () => $$item.v;
    v = "W";                                   $$item.v = "W",
                                               $.invalidate_inner_signals(() => ($$array()));
```

Two defects in one cell:

- the handler assigns to the **local getter binding** (`let v = () => $$item.v`),
  so the write overwrites a function instead of reaching the array — the write
  is silently lost and no invalidation happens;
- the **outer** `v` is promoted to `$.mutable_source` even though nothing writes
  to it, because the write to the shadowed inner binding is attributed to the
  outer one.

The outer promotion reproduces for `let`, `const` and a store
(`const v = $.mutable_source(writable("OUTER"))`), and the lost write reproduces
for a prop, an imported binding and a function declaration as the outer.

## 3. A write inside nested same-named `{#each}` invalidates both arrays

```svelte
{#each ["A"] as v}{#each ["B"] as v}<button onclick={() => { v = "W"; }}>b</button>{/each}{/each}
```

```js
// rsvelte                                              // official
$.invalidate_inner_signals(() => ($$array_1(), $$array()));   $.invalidate_inner_signals(() => ($$array_1()));
```

The write targets the inner item; rsvelte invalidates the outer array too. Same
mis-attribution as (2), in the opposite direction — over-invalidation rather
than a lost write. 6 cells.

## 4. `deep_read_state` is added for a shadowed legacy prop

```svelte
<script>
	export let v = "OUTER";
</script>
{#each ["A"] as v}{String(v)}{/each}
```

rsvelte emits `[() => ($.deep_read_state(v), $.untrack(() => String(v)))]` where
official emits `[() => (v, $.untrack(() => String(v)))]`, and in several cells
uses the inline-closure `template_effect` form where official uses the
dependency-array form. 20 cells across `each_item` / `each_index` /
`each_keyed` / `each_in_each` × five read sites.

## Population, and what the control removed

| family | cells | compiled both | divergent |
|---|---|---|---|
| shadow-read (13 outer decls × 11 constructs × 8 read sites) | 2,288 | 2,288 | 185 |
| shadow-write | 440 | 224 | 66 |
| control — identical shapes, **no name collision** | 1,694 | 1,694 | 87 |

The control is why the numbers above are smaller than the raw counts. It
produced exactly two divergence signatures, and both also appear in the shadow
families, so neither is about shadowing:

- **88 cells** of `$: $.set(v, base + 1);` — that is #3272, reached because the
  fixture's `reactive_stmt` declaration puts `let base = 1;` and the `$:` on one
  line. My fixture, not a shadowing defect.
- **12 cells** of a `{@const}` chain emitting `template_effect` where official
  assigns `text.nodeValue` directly — real, but present without any collision,
  so filed separately.

Subtracting those leaves the four defects above as shadowing-specific.

## Not an rsvelte defect, recorded so it is not re-found

Two shadow-write shapes make **official** emit output acorn rejects
(`Assigning to rvalue`), while rsvelte's parses:

```svelte
{#each [["A","B"]] as [first, ...v]}<button onclick={() => { v = "W"; }}>b</button>{/each}
{#await Promise.reject("A")}w{:catch v}{String(v)}{/await}<button onclick={() => { v = "W"; }}>b</button>
```

14 cells. rsvelte is the better side; this belongs in `upstream_issues/` rather
than here.

baseballyama:
## Summary

- resolve each-item reads and writes by lexical template ownership when names are reused
- write destructured assignments, compound assignments, updates, and member mutations back through the generated item path
- resolve `item_reassigned` from the exact each-block scope instead of scanning unrelated scopes by name
- preserve rsvelte's deliberate, parseable behavior for upstream Svelte issue #3306 rest-binding writes

## Coverage

- adds focused client/server regression tests for outer-binding shadowing, destructured writes, nested same-name invalidation, and legacy prop deep reads
- adds two pattern-corpus cases for read and write shapes
- adds a patch changeset

## Local checks

- `cargo fmt --all`
- `git diff --check`

Per the repository workflow, no local build or test suite was run; CI owns compilation and compatibility validation.

Closes #3296


baseballyama:

Issue #3702, opened 2026-08-24
parse(): 1.5% of components still attach a comment to the wrong node — 450 of 678 are TS declaration nodes rsvelte never serializes, 228 are a walk-order difference


baseballyama:
## Summary

After #3387 attaches comments to statements, **208 of 14,089 corpus components (1.5%)** still put
at least one comment on a different node than official does. This issue is about the residue, which
splits into two causes with very different sizes.

Measured by aligning nodes on `(type, start, end)` rather than by array index — index alignment
cannot be used here, because a `Program.body` length difference shifts every later index and
manufactures comment keys that are not comment defects:

```
compared 14089 components
208 still diverge when nodes are aligned by (type, start, end)

  node absent from rsvelte:  440
  TS node present on both:    10
  other:                     228
```

## Cause 1 (450 of 678) — the comment has nowhere to land

rsvelte's serialized program drops TypeScript declaration nodes, so a comment attached to one is
lost. This is the ESTree gap AGENTS.md already records under the lint section
(`convert_statement_for_program` has no `TSTypeAliasDeclaration` arm and falls to `_ => None`), now
with a second consumer. Example —
`flowbite-svelte/src/routes/admin-dashboard/utils/JSONView.svelte`:

| node | official | rsvelte |
|---|---|---|
| `TSInterfaceDeclaration@99-266` | `leading: ["Updated props data interface"]` | node absent |
| `TSPropertySignature@201-219` | `trailing: ["This matches your actual…"]` | node absent |
| `VariableDeclaration@291-402` | `leading: ["Component props"]` | same |

The last row is the control: everything that is *not* a TS node agrees in this file.

`phases/1_parse/read/expression.rs` also calls `attacher.skip_past(...)` for a type-only statement,
which **discards** its comments rather than deferring them — correct while the node does not exist,
and the thing to revisit when it does.

## Cause 2 (228 of 678) — the walk reaches a different node first

Not TypeScript, and both sides serialize the node. rsvelte gives the comment to a node with a
**later** start than the one official chooses:

| file | official | rsvelte |
|---|---|---|
| `layercake/src/routes/components/[slug]/+page.svelte` | `ArrowFunctionExpression@2845-3016` leading | `TemplateElement@3037-3037` leading |
| `layercake/src/routes/guide/+page.svelte` | `Literal@1722-1727` trailing | `ArrowFunctionExpression@1754-1774` leading |
[...]

PR #3822, opened 2026-08-25
fix(parse): match template literal comment walk order

baseballyama:
## Summary

- serialize JavaScript template-literal children in Acorn's field-creation order
- make the comment attacher visit `expressions` before `quasis`, matching zimmerframe over Acorn nodes
- pin a comment before an arrow in a template substitution to the arrow rather than a later `TemplateElement`

## Scope

Progresses #3702. This addresses the template-literal walk-order cause visible in two of the issue's representative rows. It intentionally does not claim the remaining non-TypeScript residue; #3761 handles the separate missing-TypeScript-node cause.

## Validation

- `git diff --check`
- Rust builds and tests intentionally deferred to CI per the requested workflow


baseballyama:
Superseded by and incorporated into #3868, which has now been merged into main.

Issue #3572, opened 2026-08-22
compile(): build the legacy AST's `loc`s during conversion, not at parse — hot path 0


baseballyama:
#3571 wires `compile()`'s legacy `ast` by keeping the expression locations the
`compile()` parse path used to discard (`skip_expression_loc: !modern_ast` →
`false`). That is correct and it is what upstream pays too, but it charges **every**
compile for a field most of them never read.

Deterministic counter over the 400 `runtime-legacy` components (from #3571's body — a
count, not a share of time):

| arm | objects | entries | `loc` objects | bytes |
|---|---|---|---|---|
| before (`ast` was `null`) | 0 | 0 | 0 | 0 |
| after #3571 | 54,506 | 179,301 | 10,948 | 2,193,294 |
| after #3571, `skip_expression_loc` hunk ablated | 19,139 | 94,770 | 0 | 1,353,075 |

The `loc`s are **65% of the objects** (a `loc` is three objects and six map entries;
there are ~27 per component).

## What to do

1. **Post-fill `loc` during the legacy conversion instead of at parse.** A `loc` is a
   pure function of `start`/`end` and the source — `Utf8ToUtf16` in
   `compiler/legacy.rs` already computes exactly this and is already built there. So the
   parse can go back to skipping them and the converter can add them, and the hot path
   pays zero.
2. **Build `name_loc` at parse unconditionally.** It is not derivable after the fact (it
   spans the raw attribute token including the directive keyword and any `|modifier`s,
   which the legacy node does not otherwise record), and it is one small struct per
   directive attribute rather than one per expression node.
3. **Give the envelope path an opt-out.** `compileEnvelope` — the Vite plugin's entry
   point — goes through the same `compile()` and the envelope format has no AST field
   (`reject_modern_ast_for_binary_result` refuses `modernAst` for that reason). Today the
   whole tree above is built and discarded there. With (1) and (2) the remaining cost is
   the conversion + `serde_json::to_string`, which still wants suppressing on that entry
   point.

## The trap: enumerate the `as_json()` call sites POSITIVELY

`convert_to_legacy` builds the legacy tree from the parse-time `serde_json::Value`
under `Script.content.as_json()`. There are **32** `as_json()` call sites in `compiler/legacy.rs` (158 across `rsvelte_core/src`). Converting them
by *exclusion* — "everything except the ones that already have `loc`" — breaks silently,
because `legacy.rs` itself constructs ESTree nodes that never came from a parse: the
[...]

PR #3823, opened 2026-08-25
perf(napi): skip unused AST in binary compile APIs

baseballyama:
## Summary

- add an internal core compile path that preserves generated JS, CSS, warnings, metadata, and sourcemaps without materializing `CompileResult.ast`
- use that path for `compileBuffers` and every component envelope/batch variant, including zero-copy and async entry points
- keep the public `compile` and `compileBoth` AST contract unchanged
- add a core regression test that compares every encoded field and asserts only the AST is omitted

This progresses #3572 by removing legacy AST conversion from native binary APIs whose wire formats cannot expose it. It deliberately does not close the issue: moving expression locations off the parser hot path still needs the issue's requested differential measurement.

## Validation

- `git diff --check origin/main...HEAD`
- `node scripts/release/check-changeset-package-names.mjs`
- `node scripts/release/check-core-consumer-changesets.mjs`

Per the task constraint, no local build or test suite was run; CI is the execution oracle.

### DIFF
--- a/.changeset/quiet-binary-ast.md
+++ b/.changeset/quiet-binary-ast.md
@@ -1,0 +1,6 @@
+---
+"@rsvelte/compiler": patch
+"@rsvelte/vite-plugin-svelte-native": patch
+---
+
+Skip materializing the public component AST for native buffer and envelope APIs whose binary formats do not expose it.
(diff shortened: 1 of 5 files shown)


baseballyama:

PR #3825, opened 2026-08-25
Preserve single legacy reactive dependency sequences

### LINKED ISSUE #3411: Legacy `$:` compound assignment: the dependency thunk loses its parentheses

### LINKED ISSUE #3579: A single-dependency `$:` block statement loses the parentheses around its dependency thunk
## Summary

A legacy `$:` statement whose body is a bare **block** and whose dependency thunk has exactly **one** dependency loses the parentheses around the thunk body. Every other `$:` statement form with one dependency — expression, `if`, `for`, `try` — is already correct, and so is the block form with two or more dependencies.

The axis is therefore **which statement form is broken**, not "rsvelte drops parens".

Reproduces on `origin/main` @ `c8a67c5aa`, client target. Oracle: the pinned submodule, Svelte **5.56.9** (`submodules/svelte` @ `20b341f10048`) — not `node_modules/svelte`, which resolves to 5.56.10.

## Repro

```svelte
<script>
	let d = 0;
	$: {
		let x = 1;
		d += x;
	}
</script>

<b>{d}</b>
```

```js
// rsvelte
	$.legacy_pre_effect(() => $.get(d), () => {

// official
	$.legacy_pre_effect(() => ($.get(d)), () => {
```

The rest of the file is byte-identical. The output is valid JS and semantically identical — an arrow body `$.get(d)` and `($.get(d))` evaluate the same — so this is a byte-equality divergence only.

## Which form is broken

All six cells compiled from **one** source file on `origin/main`, so the forms are compared under identical conditions. Only the first diverges; the whole-file diff is that single line.

| `$:` statement form | deps | rsvelte | official | |
|---|---|---|---|---|
| **block — `$: { let x = 1; d += x; }`** | **1** | `() => $.get(d)` | `() => ($.get(d))` | **diverge** |
| block — `$: { let y = 1; d += y + a; }` | 2 | `() => ($.get(d), $.deep_read_state(a()))` | same | agree |
| expression — `$: e = d + 1;` | 1 | `() => ($.get(d))` | same | agree |
| `try` — `$: try { d = a; } catch (q) { d = 0; }` | 1 | `() => ($.deep_read_state(a()))` | same | agree |
| `if` — `$: if (d) { e = 2; }` | 1 | `() => ($.get(d))` | same | agree |
| `for` — `$: for (const z of [d]) e = z;` | 1 | `() => ($.get(d))` | same | agree |

The two-dependency row agrees in both forms because a sequence expression needs the parentheses to parse at all; the single-dependency parenthesis is redundant, and every form except the bare block emits it anyway. So the bug is not "single dependency" and not "block" — it is their intersection, which is why no existing row shows it.

## Not caused by #3560

Found while adding `$:` scope-resolution tests in #3560, but **#3560 is not the cause**: the repro above needs no fix to compile and reproduces on `origin/main` with #3560 absent — the table above was measured on a pristine `origin/main` build.

What #3560 does is make it *reachable on one more shape*. A `class` declaration inside a `$:` block is falsely rejected on `main` today, so it emits nothing to compare; once that over-rejection is removed, the statement is a single-dependency block and shows this divergence.

That is why #3560's `class g {}` row lives in the Rust test (which asserts the compile verdict) and was **removed from** `compatibility/pattern-corpus/issues/3197-reactive-block-declaration-shadow.svelte` (which asserts byte equality) — carrying it there would have enrolled this divergence as a corpus failure filed under a different defect. Both of #3560's pattern-corpus repros are byte-identical to official on client and server as merged.

This is the "removing an over-rejection makes what was behind it reachable" shape recorded in `AGENTS.md`, one instance further on.

## Population: unmeasured

Two things this report does **not** establish, both cheap to answer and neither to be assumed from it:

* how many single-dependency `$:` block statements exist in the collected corpus;
* whether the corpus gate can see this at all — its normalization is oxfmt plus blank-line stripping, and whether oxfmt removes a redundant parenthesis around an arrow body was not checked. If it does, this divergence is gate-invisible and the count above is the only evidence there is.

baseballyama:
## Summary

- preserve upstream's one-element `SequenceExpression` for legacy reactive thunks with exactly one dependency
- carry the sequence identity through intermediate raw-text AST rewrites with the existing internal marker mechanism
- make the text-printer fallback restore the marker too, without touching strings or comments
- cover both the single-statement (#3411) and block-statement (#3579) forms

## Validation

- `cargo fmt --all`
- `git diff --check`
- full builds and tests intentionally delegated to CI

Fixes #3411
Fixes #3579

### DIFF
--- a/.changeset/fair-reactive-sequences.md
+++ b/.changeset/fair-reactive-sequences.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve upstream's parenthesized single dependency in legacy reactive effect thunks after assignment rewrites.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/ast_state_transform.rs
@@ -39,7 +39,7 @@ use super::rune_transforms::{process_derived_destructuring_pattern, wrap_state_v
 use super::{DERIVED_TMP_COUNTER, SCRIPT_ARRAY_COUNTER, STATE_TMP_COUNTER, VAR_STATE_VARS};
 use crate::compiler::phases::phase2_analyze::ComponentAnalysis;
 use crate::compiler::phases::phase2_analyze::types::ScriptProjection;
-use crate::compiler::phases::phase3_transform::js_ast::to_oxc::SINGLE_TARGET_DESTRUCTURE_SEQUENCE_MARKER;
+use crate::compiler::phases::phase3_transform::js_ast::to_oxc::SINGLE_ELEMENT_SEQUENCE_MARKER;
 use crate::compiler::phases::phase3_transform::shared::template::escape_js_string;
 
 thread_local! {
@@ -4278,10 +4278,10 @@ impl<'a, 's> StateVarCollector<'a, 's> {
                 // esrap always self-parenthesizes a `SequenceExpression`. The marker
                 // call keeps that "must be a sequence" decision alive across the
                 // eventual raw-text reparse. See
-                // `SINGLE_TARGET_DESTRUCTURE_SEQUENCE_MARKER`.
+                // `SINGLE_ELEMENT_SEQUENCE_MARKER`.
                 Some(format!(
                     "{}({})",
-                    SINGLE_TARGET_DESTRUCTURE_SEQUENCE_MARKER,
+                    SINGLE_ELEMENT_SEQUENCE_MARKER,
                     assignments.into_iter().next().unwrap()
                 ))
             } else {
@@ -4374,7 +4374,7 @@ impl<'a, 's> StateVarCollector<'a, 's> {
                 // the same "always a real SequenceExpression" rule applies here.
                 format!(
                     "{}({})",
-                    SINGLE_TARGET_DESTRUCTURE_SEQUENCE_MARKER,
+                    SINGLE_ELEMENT_SEQUENCE_MARKER,
                     assignments.into_iter().next().unwrap()
                 )
             } else {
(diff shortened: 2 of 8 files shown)


baseballyama:
Superseded by #3773. Both pull requests cover #3411 and #3579, but #3773 restores the builder-only sequence at the single state-assignment rewrite boundary, has the focused output regressions, and has completed the full CI suite successfully. Retaining that narrower implementation avoids extending the internal marker protocol across unrelated AST/codegen and text-fallback paths.

PR #3827, opened 2026-08-25
Report state references in computed pattern keys

### LINKED ISSUE #3590: fix(parse): a computed property key reports its reference at the `[`, and a destructured one not at all

## Repro

```svelte
<script>
	let s = $state(1);
	const t = { [s]: 1 };
</script>

<b>1</b>
```

```
off | state_referenced_locally @ 3:14-3:15   <- the identifier `s`
rsv | state_referenced_locally @ 3:13-3:14   <- the `[`
```

The code is right, the position is one column early: the reference node's span
starts at the **`[`** of the computed key instead of at the identifier inside
it. `js.code` is byte-identical on both `client` and `server` for every cell
below, so no output gate can see this — only the warning-position comparison.

## Scope: the property-key position only, and one arm emits nothing

Same body, `let s = $state(1)` above it:

| expression | verdict |
|---|---|
| `const t = { [s]: 1 };` | **off by one** (13 vs 14) |
| `class C { [s]() {} }` | **off by one** (11 vs 12) |
| `class C { [s] = 1; }` | **off by one** (11 vs 12) |
| `const t = { a: { [s]: 1 } };` | **off by one** (18 vs 19) |
| `const t = [{ [s]: 1 }];` | **off by one** (14 vs 15) |
| `const { [s]: v } = {};` | **no warning at all** |
| `const t = { s };` | match |
| `const t = { k: s };` | match |
| `const t = window[s];` | match |
| `let o = {}; o[s] = 1;` | match |
| `const t = [s];` | match |
| `const t = String(s);` | match |
| `` const t = `${s}`; `` | match |
| `const t = window.a + s;` | match |

The two computed **member** rows (`window[s]`, `o[s] = 1`) are correct, so the
axis is not "a `[`" — it is the `key` slot of a `Property` /
`PropertyDefinition` / `MethodDefinition`, which is serialized with the
bracket's offset as the identifier's `start`. The destructuring row is a second
arm: a computed key in an `ObjectPattern` produces no reference at all, so the
warning is missing rather than misplaced.

This is the same serialization already recorded in `AGENTS.md` under the lint
port — *"a computed property key serializes with its start at the `[`, which
matches no oxc reference"* — reached here from the compiler's own warning
positions rather than from a lint rule, which shows the span is wrong in the
shared program, not in a rule's reading of it.

## Where

`crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs` builds the
key node; phase 2 reads its span for the warning.

(`expressions/unicode-escape-identifiers.svelte`); the unicode escapes were a
red herring — plain ASCII reproduces it, which is why the row above is `s`.

baseballyama:
## Summary

- treat identifiers evaluated inside computed object-pattern keys as references, not declarations
- preserve declaration treatment for the corresponding binding value
- cover exact warning positions on both client and server targets

The property-key span issue described by #3590 was already fixed by merged PR #3519. The remaining destructuring arm emitted no warning because phase 2 classified every identifier inside `VariableDeclarator.id` as a declaration, including evaluated computed keys.

Fixes #3590

## Validation

- `git diff --check`

Per project policy, compilation and the Rust test suite are delegated to CI.

### DIFF
--- a/.changeset/quiet-computed-keys.md
+++ b/.changeset/quiet-computed-keys.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Report state references used as computed keys in object destructuring patterns.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/identifier.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/identifier.rs
@@ -265,17 +265,33 @@ fn visit_identifier_inner(
     // The official compiler has `node !== binding.node` check to skip warnings for the
     // declaration identifier itself. We approximate this by checking if the identifier
     // is inside a VariableDeclarator's `id` pattern (which is the declaration site).
-    let is_declaration_node = context.js_path.iter().any(|ancestor| {
-        if ancestor.get_type_str() == Some("VariableDeclarator") {
-            // Check if the current node's position falls within the `id` pattern range
-            let id_start = ancestor.get_child_field_start("id", context.parse_arena);
-            let id_end = ancestor.get_child_field_end("id", context.parse_arena);
-            if let (Some(id_s), Some(id_e)) = (id_start, id_end) {
-                return start >= id_s && start < id_e;
+    // A computed key inside the pattern is evaluated rather than bound. Keep
+    // identifiers in that key out of the broad `VariableDeclarator.id` guard;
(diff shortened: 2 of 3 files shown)


baseballyama:
Closing as superseded by #3791. That PR implements the same remaining #3590 fix using the binding's exact declaration node, includes the pattern-corpus regression and positive/negative unit coverage, and has completed CI successfully.

PR #3828, opened 2026-08-25
Fix reference-order dependent lowercase warnings

### LINKED ISSUE #3361: Binding references are collected during the visitor walk, not while building scopes — so no pre-walk decision can read them
Upstream populates a binding's reference list while **building scopes**, before any
analysis visitor runs. rsvelte populates it **during** the visitor walk. Every
decision upstream makes from `scope.references` *before* the walk therefore has no
correct answer available in rsvelte at the point the corresponding visitor runs.

This is the cause of #3262, but it is not that issue's subject — nobody reading
#3262 would arrive at this constraint, and the set of other decisions it affects has
not been measured.

## (a) Where each side collects

| | site | when |
|---|---|---|
| official | `phases/scope.js` — `Scope.reference()`, called from `create_scopes` / `analyze_module` | **before** `walk(ast, state, visitors)` (`2-analyze/index.js` L713 runes / L783 legacy) |
| rsvelte | `2_analyze/visitors/identifier.rs:195` (`add_reference_with_flags`), plus `each_block.rs:310`, `style_directive.rs:38`, `shared/utils.rs:1304`, `component.rs:70` | **during** the visitor walk (`2_analyze/mod.rs:451` module → `:485` instance → `:572` template) |

`scope_builder.rs:2935` is the one pre-walk writer, and it covers directive names
only (`reference_directive_name`), not ordinary identifier reads.

The same timing property holds for `Binding::has_direct_template_read`, which is
set in `identifier.rs:215` / `bind_directive.rs:187` — i.e. also during the walk.

Note the asymmetry that makes this easy to miss: **updates are known pre-walk,
references are not.** `binding.reassigned` / `binding.mutated` are set by the scope
builder, including from template expressions (`process_template_expression`).

## (b) The measurement

Upstream's legacy `normal → state` promotion (`2-analyze/index.js` L627-645) is
pre-walk and needs both halves: *updated* **and** referenced from a `Fragment`, a
`$:` block or a `StyleDirective`. rsvelte's port is
`2_analyze/mod.rs::promote_legacy_state_bindings`, which reads
`binding.references` and therefore has to run after the walk.

Moving that call into the same pre-walk block as `process_legacy_exports` and
instrumenting it on

```svelte
<script>
	export const a = 1;
</script>
<button onclick={() => a = 2}>{a}</button>
```

prints

```
OPTC promote? name=a kind=Normal decl=Const updated=true refs=0
```

— and every one of the 12 diverging cells in `EXPORT_WRITE`
(`crates/rsvelte_core/tests/error_check_precedence_3262_3268.rs`, added in #3359)
is **unchanged** by the move. The promotion silently stops promoting anything,
because the criterion's reference half is empty at that point.

## (c) What reads `Binding::references` in rsvelte, and whether its upstream counterpart is pre-walk

Non-test readers, enumerated by grep over `crates/`:

| site | upstream counterpart | counterpart runs | status |
|---|---|---|---|
| `2_analyze/mod.rs:1862` `promote_legacy_state_bindings` | `2-analyze/index.js` L627-645 | **pre-walk** | **diverges** — this is #3262 |
| `2_analyze/mod.rs:764-780` runes "reassigned non-state read in the template" warning | `2-analyze/index.js` L735-770 | post-walk | matches |
| `2_analyze/visitors/regular_element.rs:543` `component_name_lowercase` (`references.is_empty()`) | `2-analyze/visitors/RegularElement.js` | during the walk on both sides | **unmeasured** — rsvelte reads a list that is only partly populated at that moment, so the answer depends on where in the template the element sits relative to the reference. Whether that produces a divergence has not been tested. |
| `2_analyze/scope.rs:95` `binding_at_reference` (lazy span → binding index) | n/a — rsvelte-only, consumed by phase 3 | post-walk | matches |

`has_direct_template_read` has the same timing and two readers
(`2_analyze/mod.rs:717` and the `non_reactive_update` path); their upstream
counterparts are post-walk, so they are not affected — but they are listed so the
next person does not have to re-derive it.

**What is not measured:** how many *diagnostics* change if reference collection
moves pre-walk. I did not run the corpus or the matrix against a prototype, because
no prototype exists — the change is not a call-site move (that is what (b)
falsifies) but relocating the identifier-resolution walk itself into the scope
builder. Treat the table above as the enumeration of *readers*, not as a bound on
the blast radius.

## Why the cheap workarounds do not work

Both were checked against the oracle rather than assumed:

- **"updated is enough."** No: `export const a = 1;` written only from a script
  function and never read in the template is `constant_assignment` on **both**
  sides. The reference half decides it. (Row
  `export_const_written_not_in_template` in the #3359 grid.)
- **"scan the template for the name."** No: a template binding of the same name
  shadows the outer one, so a name match is a false positive — which is exactly the
  defect class #3296 / #3297 are about.

## Suggested shape of a fix

Move identifier resolution + reference recording into `scope_builder.rs`, so that
`Binding::references` is complete before `2_analyze/mod.rs:451`. Then
`promote_legacy_state_bindings`, `promote_reassigned_store_variables` and the two
`each` promotions can move into the pre-walk block next to `process_legacy_exports`,
where upstream has them, and `ExportNamedDeclaration` sees final binding kinds.

baseballyama:
## Summary

- make `component_name_lowercase` consult the scope-resolved template-reference set built before Phase 2 visitors run
- remove the warning's dependence on whether a real reference appears before or after the lowercase element
- retain the warning when the imported lowercase name has no template reference

## Why this resolves the remaining reader

#3361 enumerates two readers whose upstream counterparts need a complete reference graph before or during the visitor walk. The export-const precedence reader is handled by the base PR, #3813. This stacked change handles the only other enumerated timing-sensitive reader.

The implementation deliberately reuses the binding-index set produced by the scope builder instead of duplicating every `Binding::references` record before the visitor. Full reference records still carry visitor-only flags and diagnostics; the early decisions only need the scope-correct binding identity.

## Stack

- base: #3813
- this PR: the remaining `component_name_lowercase` reference-order case

Closes #3361

## Validation

- `git diff --check`
- no local build or test suite was run; CI is the execution oracle


baseballyama:

PR #3829, opened 2026-08-25
fix(server): preserve props initializer comment cursor

### LINKED ISSUE #3515: client: a comment between `=` and a top-level rune lands in the wrong place in the output

baseballyama:
## Summary

- keep builder-created rune declarators location-less in the server AST, matching upstream
- preserve source spans only for declarators upstream visits verbatim
- add the five discriminating `$props()` line/multiline-comment fixtures from the corrected issue grid

## Why

Upstream retains the source location on the declaration and binding pattern, but rebuilds the rune declarator and `$$props` initializer without locations. Restoring the original declarator span changes esrap's cursor behavior only when the separator crosses a line, producing the five remaining server divergences in #3515.

## Verification

- `cargo fmt --all`
- `git diff --cached --check`
- no local build or test run; the pattern-corpus fixtures delegate exact official-compiler comparison across every configured target to CI

Stacked on #3820, which handles the client/client-dev `$props()` cells. Together with merged #3415 for `$derived` declarators, this completes the reported 18-cell grid.

Closes #3515


baseballyama:

Issue #3569, opened 2026-08-22
`has_call` is decided at 7 sites in phases 2/3, and the current tests cannot distinguish one of them from its absence


baseballyama:
Upstream asks "does this expression contain a call?" in one place per node kind and stores the answer on the expression's metadata. rsvelte writes `set_has_call` at **7 sites** across phases 2 and 3, and while fixing #3275 I found that at least one of them is not reached by the input class its own comment describes.

## What was measured

While measuring #3275 (PR #3548) I ablated `2_analyze/visitors/call_expression.rs::visit_typed`'s `has_call` block two ways:

| ablation | result |
|---|---|
| move the block back to **before** the argument walk (i.e. undo the upstream `CallExpression.js:264` → `:272` ordering) | **0 of 8** rows in `rest_props_memoized_3275` move |
| kill the block entirely — `if false && (!is_pure_call \|\| has_dependencies)` | **0 of 8** rows move, and `rsvelte_core --lib` (1732), `ssr`, `compiler_fixtures` and `validator` all stay green |

Measurement conditions: private `CARGO_TARGET_DIR`, `cargo --manifest-path <worktree>/Cargo.toml`, and each run's log checked for `Compiling rsvelte_core (<that worktree>)` so a warm or foreign artifact could not be mistaken for a result.

## What this does and does not show

**It does not show the block is dead code**, and please do not delete it on the strength of this issue. What was measured is that **the current test population cannot distinguish the tree with that block from the tree without it**. Whether an input exists that reaches it — a script-level expression, a `$:` body, a `compileModule` entry — is **unmeasured**. The honest statement is: the eight rows written specifically to exercise "a call over `$$restProps` / `$$slots` in a template attribute" do not reach this visitor, and neither do the four broad gates above.

The reason that matters is the #3275 fix itself: the rescued branch moved this block to match upstream's ordering, and the change was correct as a reading of upstream and **invisible to every gate**. PR #3548 therefore reverts it rather than shipping an unmeasured behaviour change. The next person who reads `CallExpression.js` will make the same edit unless the reachability question is answered.

## The 7 sites, and which input each answers for

The count alone reads as "unify them", but whether they *can* be unified depends on whether the seven answers agree — which is a separate measurement nobody has taken. The table is what makes that question askable.

| # | site | answers for |
|---|---|---|
| 1 | `2_analyze/visitors/shared/utils.rs:1416` | *(unattributed — see below)* |
[...]

PR #3830, opened 2026-08-25
Remove unreachable phase-two has-call writers

baseballyama:
## Summary

- remove three `has_call` writes that cannot be reached while expression metadata is installed
- keep the reachable template-expression call and spread writers unchanged
- record the static reachability result and the remaining duplicated decision sites in the two-ports inventory

## Reachability proof

`VisitorContext.expression` is initialized to `None`. The only Phase 2 assignment of `Some` is in the `{#if}` visitor, which passes its condition to `walk_js_expression_node`. It never enters the typed script walker.

The removed writes lived only in that typed script walker and its typed `CallExpression` visitor. Therefore no input can observe them while the current single producer and consumer remain disjoint. This answers three of the unattributed rows in #3569 without treating a green test population as proof of dead code.

The three reachable Phase 2 writers remain in `walk_js_expression_node`: ordinary calls, object spreads, and top-level spreads. Phase 3's independent re-derivations remain explicitly unmeasured, so this PR progresses rather than closes the issue.

## Validation

- `git diff --check`
- writer and `context.expression` assignments audited with `rg`
- no local build or test suite was run; CI is the execution oracle

Progresses #3569

### DIFF
--- a/.changeset/remove-unreachable-has-call.md
+++ b/.changeset/remove-unreachable-has-call.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Remove unreachable phase-two call metadata writes.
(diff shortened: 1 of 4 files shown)


baseballyama:

Issue #3420, opened 2026-08-22
Client: a `$state` declared bare in a `case` clause has its references left untransformed by official (rsvelte's output is the correct one)


baseballyama:
Found while grid-testing declarator containers for #3070 / #3146. Not blocking anything;
filing so the decision is recorded rather than rediscovered.

A `let` declared **directly in a `case` clause** — legal JS, no block braces — is lowered
to `$.state(...)` by the official compiler, but its *references* are left untransformed.
rsvelte transforms them. So this is one of the rows where **rsvelte's output is the
correct one and official's is broken at runtime**, like #2990.

```svelte
<script>
	let a = $state(0);
	function f(k) {
		switch (k) {
			case 1:
				let s = $state(1);
				s++;
				return s;
		}
	}
</script>

<button onclick={() => a++}>{a}{f(1)}</button>
```

client, official:

```js
	function f(k) {
		switch (k) {
			case 1:
				let s = $.state(1);
				s++;
				return s;
		}
	}
```

client, rsvelte:

```js
	function f(k) {
		switch (k) {
			case 1:
				let s = $.state(1);
				$.update(s);
				return $.get(s);
		}
	}
```

`$.state(1)` returns a `Source` object. Official's `s++` therefore evaluates
`{…} + 1` → `NaN` and `return s` hands back `NaN`; the component renders `NaN` where
rsvelte renders `2`. The lowering and the reference handling disagree with each other
*within official's own output*, which is what says it is a defect rather than a design
choice: if `s` were meant to stay a plain value, `$state(1)` would have been folded away,
not turned into `$.state(1)`.

## The discriminating control

Adding braces to the same case — changing nothing else — makes official agree with
rsvelte:

```svelte
			case 1: {
				let s = $state(1);
				s++;
				return s;
			}
```

client, official:

```js
			case 1:
				{
					let s = $.state(1);

					$.update(s);

					return $.get(s);
				}
```

So the divergence is not about `switch` and not about the declaration — it is about
whether the case clause introduces a `BlockStatement`. That is the control that rules out
"upstream treats switch-local state as non-reactive on purpose".

## Measured

Six shapes × dev on/off against the pinned official compiler (`submodules/svelte`,
5.56.10). Only the brace-less `case` row diverges:

| shape | verdict |
|---|---|
| `case 1: { let s = $state(1); s++ }` | match |
| `case 1: let s = $state(1); s++` | **diverges, both dev and non-dev** |
[...]

PR #3831, opened 2026-08-25
Cover rune declarations across statement containers

### LINKED ISSUE #3146: meta: container coverage written as a per-rule `match` keeps producing the same defect
Recording the shape these five defects share, because each was found the same way and the next one will be too.

rsvelte answers "what containers does this rule reach into?" with a hand-written `match` over `TemplateNode` variants, once per rule. Upstream answers it once, structurally — one flat `analysis.elements`, one scope chain, one parser stack — so a container it gains is covered everywhere by construction. Every rsvelte list is an independent chance to omit one, and the omissions do not correlate: each rule is missing a *different* container.

Measured on one generated grid (content x container x target, 4600 cells), the omissions found so far:

| rule | walk | container(s) missing |
|---|---|---|
| CSS scoping (#3136) | 5 walks in `css_scoping.rs` | `<svelte:fragment>`, `<svelte:component>`, `<svelte:self>` |
| class/style synthesis (#3136) | `synthesize_class_style_attributes` | `<svelte:boundary>`, `<svelte:fragment>`, and four more |
| `bind:group` registration (#3142) | `mark_group_bindings_in_fragment` | `<svelte:boundary>`, `<svelte:fragment>` |
| root-only meta placement (#3140) | three depth counters | `{#key}`, `<svelte:element>`, `<svelte:head>`, `<slot>` |
| snippet hoistability (#3139, #3143) | `check_hoistable` | `<svelte:element>`, `<svelte:self>` by node type; `let:` names by scope |

Two of the five were fixed by adding the missing arms, which leaves the class intact. The other three replaced the enumeration: #3140 now reads one flag maintained in the single function every container already calls to reach its children, and #3139/#3143 judge references rather than node types, which is what upstream does.

What to take from it: **adding the missing arm closes the instance, not the class.** When a rule's coverage is a list of container types, the useful question is whether the list can be deleted — is there a funnel every container already passes through, or a property (a reference, a scope depth) that answers the question without naming types? Where there is, the next container added to the AST is covered for free; where there is not, it is a fresh omission waiting for a grid to find it.

The grid is `content x container x target`: ~50 content shapes crossed with ~26 containers, compiled by both compilers on all four targets and compared for byte equality. It started at 88 divergences over 23 classes on a tree where the collected corpus was green.

baseballyama:
Closes #3146.

## What changed

- add a generated `rune-statement-container` family crossing `$state` / `$derived` declarations with bare switch, label, if/else, for/for-of/while bodies, braced controls, and both component/module entry points
- make the module dev tag-declarator walker descend through labeled statements and switch consequents
- pin the two newly reached walker shapes with focused unit tests
- document the oracle boundary in the matrix README and gate-coverage inventory

This closes the meta issue rather than claiming every structural walk has been unified: all known concrete container-coverage defects are fixed, the selector grid is at 0 divergences over 8,976 cells, and this PR adds the last missing generated axis recorded on #3146 while fixing the two omissions it reaches. Any future discriminating container class should be filed from its measured divergence rather than keeping the meta issue open without a known residual.

The exact lexical bare-case row from #3420 is intentionally compared only on server targets. Official's client output lowers the declaration but leaves its references untransformed and computes `NaN`; rsvelte's correct client behavior remains pinned by `case_clause_state_3420.rs` and the matching `upstream_issues/` record.

## Verification

- `node --check scripts/compat-corpus/matrix/axes.mjs`
- `node --check scripts/compat-corpus/matrix/generate.mjs`
- `git diff --check`

No local build or test run, per repository instructions; CI is the execution oracle for the new generated cells.


baseballyama:

PR #3832, opened 2026-08-26
refactor(client): consume attachment expression metadata

baseballyama:
## Summary

- consume `AttachTag` expression metadata populated during Phase 2
- remove the duplicate Phase 3 `has_state` and `has_call` tree walks for attachments
- pin the Phase 2 decisions for local calls, global pure calls, and state-only references

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- compile and test execution intentionally deferred to CI per repository workflow

## Stack

- base: #3830
- progresses #3569

PR #3833, opened 2026-08-26
refactor(client): consume spread expression metadata

baseballyama:
## Summary

- add upstream-shaped expression metadata to `SpreadAttribute`
- populate it in Phase 2 on regular elements, dynamic elements, components, `svelte:component`, `svelte:self`, and slots
- consume the saved call/state flags in both client spread transform paths instead of walking the expression again
- cover host traversal and the local-call, pure-global-call, and state-only decisions directly

## Scope

This is the next independently reviewable slice of #3569 and is stacked on #3832. It progresses the issue but does not close the remaining duplicate metadata walkers.

## Verification

- `git diff --check`
- confirmed the only `SpreadAttribute` constructor initializes the metadata
- confirmed the client transform has no remaining spread-specific `expression_has_call` / `expression_has_reactive_state` walk

Per project instructions, compilation and tests were not run locally; CI is the execution oracle.

### DIFF
--- a/crates/rsvelte_core/src/ast/template.rs
+++ b/crates/rsvelte_core/src/ast/template.rs
@@ -900,11 +900,20 @@ impl serde::Serialize for AttributeValuePart<'_> {
 }
 
 /// A spread attribute: `{...props}`.
+#[derive(Debug, Clone, Default)]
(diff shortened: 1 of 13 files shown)


baseballyama:

PR #3834, opened 2026-08-26
refactor(client): consume style directive metadata

baseballyama:
## Summary

- add upstream-style aggregate expression metadata to `StyleDirective`
- populate it once during Phase 2 for shorthand, single-expression, and multi-chunk values
- replace the client transform's duplicate expression walks and binding-kind inference with the stored metadata
- resolve shorthand bindings from the current lexical scope, including each-block bindings

## Scope

This is the next independently reviewable slice of #3569. It is stacked on #3833; only the top commit belongs to this PR.

The existing shorthand-only blocker compatibility path remains intact. Phase 3 now derives its update-routing blocker check from the binding indices recorded in Phase 2 instead of re-scanning generated expressions.

## Validation

- `git diff --check`
- added metadata coverage for every legal host, multi-chunk values, pure global calls, and lexically scoped shorthand bindings

Per project instruction, no local build or test suite was run; CI is the execution oracle.

### DIFF
--- a/crates/rsvelte_core/src/ast/template.rs
+++ b/crates/rsvelte_core/src/ast/template.rs
@@ -1055,6 +1055,13 @@ impl serde::Serialize for ClassDirective<'_> {
 }
 
 /// A style directive: `style:property={expression}`.
+#[derive(Debug, Clone, Default)]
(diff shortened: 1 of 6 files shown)


baseballyama:

PR #3835, opened 2026-08-26
refactor(client): consume style attribute metadata

baseballyama:
## Summary

- consume the Phase 2 `ExpressionTag` metadata for regular `style=` attribute expressions
- share one cached-metadata builder across single values, one-part sequences, and interpolated sequences
- add focused metadata and pattern-corpus coverage for regular and dynamic elements
- update the two-ports inventory without claiming the remaining generic Phase 3 walkers are resolved

## Context

This is a stacked follow-up to #3834 and another bounded slice of #3569. Upstream computes expression metadata during analysis and only reads it in the client transform. The rsvelte style-attribute helper instead re-walked each expression with `analyze_expression_properties`, independently of the metadata already attached to every `ExpressionTag` by Phase 2.

The constant-folding, definedness, and memoizer behavior in the style-value builder are unchanged; only the metadata source is replaced.

## Verification

- `git diff --check`

### DIFF
--- a/compatibility/pattern-corpus/issues/3569-style-attribute-metadata.svelte
+++ b/compatibility/pattern-corpus/issues/3569-style-attribute-metadata.svelte
@@ -1,0 +1,12 @@
+<script>
+	let color = $state('red');
+
+	function local(value) {
+		return value;
+	}
+</script>
+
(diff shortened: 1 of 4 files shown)


baseballyama:

PR #3836, opened 2026-08-26
refactor(client): consume standalone attribute metadata

baseballyama:
## Summary

- consume each standalone regular attribute expression's Phase 2 metadata in the client transform
- preserve the analyzed narrow `has_call` decision and binding-reference set instead of rebuilding broad flags in Phase 3
- cover regular, component, dynamic-component, dynamic-element, slot, and one-part quoted attribute forms
- add a pattern-corpus repro and narrow the remaining two-ports inventory entry to multi-part attributes

## Context

This is a stacked follow-up to #3835 and another bounded slice of #3569. Upstream passes `chunk.metadata.expression` directly into `build_expression`. Rsvelte's standalone attribute branches instead called `extract_metadata_from_tag`, which re-walked the expression in Phase 3 and did not carry Phase 2's binding-reference set.

The existing transformed-state and async-blocker checks are unchanged. Multi-part attribute values also remain on their existing broad Phase 3 walker so this PR does not claim the full two-ports row is resolved.

## Verification

- `git diff --check`

### DIFF
--- a/compatibility/pattern-corpus/issues/3569-single-attribute-metadata.svelte
+++ b/compatibility/pattern-corpus/issues/3569-single-attribute-metadata.svelte
@@ -1,0 +1,13 @@
+<script>
+	let title = $state('title');
+
+	function local(value) {
+		return value;
+	}
+</script>
+
+<div title={local(title)}></div>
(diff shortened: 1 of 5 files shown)


baseballyama:

PR #3837, opened 2026-08-26
refactor(client): consume multipart attribute metadata

baseballyama:
## Summary

- consume each interpolated attribute chunk's Phase 2 expression metadata
- use the upstream-compatible narrow `has_call` decision for multipart attribute memoization
- distinguish a pure global call from an impure local call within the same quoted attribute
- update the pattern corpus and two-ports inventory while leaving helper removal for a separate slice

## Context

This is a stacked follow-up to #3836 and another bounded slice of #3569. Phase 2 already walks every `ExpressionTag` inside an `AttributeValue::Sequence` and writes its metadata in place. The client transform nevertheless rebuilt broad flags from the expression AST and then treated every nested call as impure.

Upstream passes each chunk's `node.metadata.expression` directly to `build_expression` and the memoizer. This change does the same. The later transformed-state check remains in place, and the old broad helper implementations are not deleted in this PR.

## Verification

- `git diff --check`


baseballyama:

PR #3838, opened 2026-08-26
refactor(client): remove obsolete attribute metadata walkers

baseballyama:
## Summary

- remove the now-unreferenced Phase 3 attribute metadata walkers
- delete both typed and JSON implementations of the obsolete broad flag calculation
- remove unit tests that only compared those unused implementations to each other
- retain the attribute builder's boolean/text behavior tests and update the two-ports inventory

## Context

This is a stacked follow-up to #3837 and a cleanup slice of #3569. After every generic attribute chunk consumes its Phase 2 `ExpressionTag` metadata, `extract_metadata_from_tag`, `expression_tag_has_call`, and their typed/JSON support graph have no production caller. Their remaining tests exercised only that closed, unused implementation island.

Removing the code separately keeps the preceding behavior change reviewable and makes it explicit that this PR changes no active transform path.

## Verification

- `git diff --check`
- static reference scan for every removed helper


baseballyama:

PR #3839, opened 2026-08-26
Consume metadata for component CSS properties

baseballyama:
## Summary

- feed each component CSS custom-property expression chunk's Phase 2 metadata to the client memoizer
- preserve upstream handling for calls and awaits in both standalone and multipart values
- remove the last attribute-path caller of the Phase 3 `expression_has_call` re-walk
- add a pattern-corpus discriminator for local versus pure calls in CSS custom properties

## Verification

- `git diff --check`

Local builds and tests were intentionally not run per repository instructions; CI is the test oracle.

Part of #3569. Stacked on #3838.


baseballyama:

PR #3840, opened 2026-08-26
Use analyzed calls in template chunks

### LINKED ISSUE #3569: `has_call` is decided at 7 sites in phases 2/3, and the current tests cannot distinguish one of them from its absence

baseballyama:
## Summary

- read `has_call` from each expression tag's Phase 2 metadata in the shared client template-chunk builder
- remove the builder's independent Phase 3 call calculation while preserving its existing state/member/await decisions
- cover standalone and multipart text expressions with local-call and pure-call controls
- finish the writer/consumer attribution recorded in the two-ports inventory

## Verification

- `git diff --check`

Local builds and tests were intentionally not run per repository instructions; CI is the test oracle.

Closes #3569. Stacked on #3839.


baseballyama:

PR #3842, opened 2026-08-26
refactor(client): carry hoisted import source spans

baseballyama:
> Status: the retained-import implementation in #3853 supersedes this PR because it also handles TypeScript type-only cleanup and the standalone phase-3 API. Static inspection shows that #3854 cannot decide the remaining esrap punctuation work: the generated-token pass visits identifiers and literals, but not import punctuation such as the named-import `{`. The final esrap-only changes from this branch will therefore be split into a small PR stacked after #3853, with their split-coordinate regression test as the direct oracle. This draft remains unmerged and will be closed once that replacement PR exists.

## Summary

- carry retained Phase-1 import declaration ranges alongside hoisted client imports
- emit source-mapped hoisted imports only when the retained declaration formats to the exact extracted output; otherwise preserve the existing unmapped fallback
- remove the client `verbatim_import` generated-text matching pass
- add discriminating source-map coverage for named-import punctuation in both module and instance scripts
- document the removed pass and the gate coverage it now has

This is a focused part of #3015. It does not close the issue.

## Safety

The retained AST supplies the source range. Text equality is used only to prove that an extracted import is the declaration at that range, not to discover a generated position. TypeScript-erased or otherwise changed declarations remain on the prior `Raw` path, so failure to establish that pairing cannot drop or misattribute an import.

The server-side import matcher is independent and remains unchanged.

## Verification

- `cargo fmt --all`
- `git diff --check`

Per repository instructions, I did not run local builds or tests; CI is the execution oracle.

### DIFF
--- a/.changeset/curly-import-maps.md
+++ b/.changeset/curly-import-maps.md
@@ -1,5 +1,5 @@
 ---
-"rsvelte_core": patch
+"@rsvelte/compiler": patch
 ---
 
 Preserve source-map spans for hoisted client imports.
--- a/.changeset/curly-import-maps.md
+++ b/.changeset/curly-import-maps.md
@@ -1,0 +1,5 @@
+---
+"rsvelte_core": patch
+---
+
+Preserve source-map spans for hoisted client imports.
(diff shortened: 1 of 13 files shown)


baseballyama:
Superseded by #3853 for retained-import construction and #3859 for the remaining esrap punctuation mappings. The latter is stacked after #3853 and retains the split-coordinate regression coverage.

PR #3843, opened 2026-08-26
refactor(client): carry component bind key spans

baseballyama:
## Summary

- carry component `bind:` accessor-key locations on dedicated `JsPropertyKey` variants
- preserve both identifier and quoted property-key shapes through the text and OXC printers
- remove the generated-text search that reconstructed `get`/`set` key mappings
- retain and document the separate component-bind interpolation fallback

## Upstream parity

This mirrors upstream `component.js`, which assigns `attribute.name_loc` to the generated getter and setter keys. Parsed directives use the retained `name_loc` directly; only synthetic directives fall back to reconstructing the directive-name range.

The explicit getter/setter SequenceExpression form remains unstamped because upstream does not assign `name_loc` in that branch.

This is an independent follow-up slice for #3015 and does not close the umbrella issue.

## Verification

- `git diff --check`
- added builder shape tests for identifier and quoted keys
- strengthened the client source-map test so the deleted generated-text pass cannot supply accessor mappings

Per project instructions, no local build or test command was run; CI is the execution oracle.

### DIFF
--- a/.changeset/tidy-potatoes-map.md
+++ b/.changeset/tidy-potatoes-map.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Map generated component binding accessor keys to their source directives.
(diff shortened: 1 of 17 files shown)


baseballyama:

PR #3844, opened 2026-08-26
refactor(client): carry element handle source spans

baseballyama:
## Summary

- retain source spans for generated element-handle identifiers in the JS arena
- teach both the handwritten and OXC printers to map those identifier uses directly
- delete the client source-map pass that recovered element runtime uses by scanning generated JavaScript

## Safety

The span registry is out-of-band, so generated handles remain ordinary `JsExpr::Identifier` values and existing lowering matchers keep seeing the variants they expect. Handle names are generated from the component-wide memoizer, making the registry key unique within a compilation.

## Validation

- `cargo fmt --all`
- `git diff --check`
- `git diff --cached --check`
- confirmed the removed template-element runtime scan symbols no longer occur in phase 3

Per the repository instructions, no local build or test suite was run; CI is the execution oracle.

Part of #3015

### DIFF
--- a/.changeset/carry-element-handle-spans.md
+++ b/.changeset/carry-element-handle-spans.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Carry generated element-handle source spans directly through client code generation.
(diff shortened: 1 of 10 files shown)


baseballyama:

PR #3845, opened 2026-08-26
Carry legacy prop read source spans

baseballyama:
## Summary

- carry legacy dependency getter locations on the identifier passed into the read transform
- preserve the template expression's independent use-site span
- delete the generated-text `legacy_prop_read` source-map enrichment pass
- replace its synthetic scanner test with an end-to-end merged source-map regression test

## Why

Upstream clones the located binding identifier and passes that node through `build_getter`. rsvelte instead wrapped the fully lowered getter and then recovered the two generated positions by scanning output text. Keeping the span on the identifier matches the upstream AST shape and avoids mapping the surrounding `foo()` call as though the whole call came from the identifier.

This moves the measured `legacy_prop_read` pass contribution from 16 segments to 0 and removes the pass.

## Validation

- `git diff --check`
- verified no `generate_legacy_prop_read*`, `legacy_prop_read_mappings`, or `map_pass_disabled("legacy_prop")` references remain

Local builds and test binaries were intentionally not run; CI is the execution oracle for this change.

Part of #3015.

### DIFF
--- a/.changeset/bright-props-map.md
+++ b/.changeset/bright-props-map.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve source-map carriers for lowered legacy prop reads.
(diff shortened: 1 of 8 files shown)


baseballyama:

PR #3846, opened 2026-08-26
Carry inline script source spans

baseballyama:
## Summary

- restore inline instance-script declaration locations through the existing `RawMapped.copied_spans` carrier
- remove the generated-text `inline_script` source-map pass and its declaration search
- replace the scanner-only unit test with a compile-level merged source-map regression test
- record the pass deletion in the phase-3 plan and gate-coverage inventory

This is the fifth focused source-map slice for #3015. Normalized script text already records each unchanged run against its original component offset, and the OXC conversion restores those locations onto the reparsed declaration nodes. The post-print search was therefore duplicating information the AST path now retains.

## Validation

- `git diff --check`
- verified the deleted helper/pass names have no remaining references

Per the current workflow, I did not run local builds or tests; CI is the execution oracle.

Part of #3015.

### DIFF
--- a/.changeset/tidy-inline-script-maps.md
+++ b/.changeset/tidy-inline-script-maps.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve source-map spans for normalized inline client scripts.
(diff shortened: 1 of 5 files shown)


baseballyama:

PR #3847, opened 2026-08-26
Carry bind:value accessor source spans

baseballyama:
## Summary

- carry the source span for automatic `bind:value` accessor roots on the completed generated call
- apply that scoped span in both the text and OXC printers without wrapping member-chain nodes
- delete the five-segment `bind_value` generated-text source-map recovery pass
- retain explicit child spans and exclude custom accessors and `<select>` bindings

## Stack

This draft is stacked on #3844, whose element-handle spans replace the other half of the deleted recovery pass.

## Verification

- `cargo fmt --all`
- `git diff --check`
- local builds and tests intentionally not run; CI is the build/test oracle

Part of #3015


baseballyama:

PR #3848, opened 2026-08-26
Delete component bind source-map matching

baseballyama:
## Summary

- delete the final client `component_bind` generated-text source-map pass
- rely on the source span already carried by the transformed getter callee for `${potato() ?? ...}`
- keep the merged accessor/interpolation map regression as the CI oracle

## Why

The read transform already lowers a source `Spanned(Identifier)` to an unlocated call whose callee is `Spanned(OpaqueIdentifier)`. That is the same ownership as upstream's `b.call(b.id(name, loc))`: the identifier maps, not the generated `()` wrapper. The text scanner therefore duplicated information already present in the AST.

This is stacked on #3843, which moves component `bind:` accessor-key locations onto dedicated property-key variants.

## Validation

- `cargo fmt --all`
- `git diff --check`
- confirmed no `component_bind` pass references remain

Per the repository instructions, no local build or test command was run; CI is the build/test oracle.

Part of #3015.


baseballyama:

PR #3849, opened 2026-08-26
perf: rebuild legacy AST locations at conversion

### LINKED ISSUE #3572: compile(): build the legacy AST's `loc`s during conversion, not at parse — hot path 0

baseballyama:
## Summary

- omit ESTree expression locations while compiling the default legacy AST and rebuild them only at its JSON boundary
- retain parse-time locations when `modernAst: true`, because that tree is exposed directly
- cover every legacy converter expression boundary, including scripts and the synthesized `{@const}` assignment
- rebuild script and in-tag comment locations while preserving their distinct `character` shape
- preserve upstream's distinct binding-pattern, directive-name, and script `Program` location shapes
- pin representative each/await/snippet/const patterns, comment locations, and the empty-line binding-column quirk

## Validation

- `git diff --check`
- build and test commands intentionally left to CI because this workstation is under load
- the existing public parse AST gate compares the recursive legacy tree against the pinned official compiler over the full corpus (about 28,000 pairs), superseding the original 400-component audit
- the existing N-API option test compares the compile-time modern AST exactly with the official compiler

Stacked on #3823.

Closes #3572

### DIFF
--- a/.changeset/quiet-binary-ast.md
+++ b/.changeset/quiet-binary-ast.md
@@ -1,0 +1,6 @@
+---
+"@rsvelte/compiler": patch
+"@rsvelte/vite-plugin-svelte-native": patch
+---
+
+Skip materializing the public component AST for native buffer and envelope APIs whose binary formats do not expose it.
--- a/compatibility/gate-coverage.md
+++ b/compatibility/gate-coverage.md
@@ -2601,21 +2601,22 @@ which build the map in Rust. Neither covers `compileBuffers`, `compileModuleBuff
 `*ZeroCopy` entries, or `compileModule`'s map — and no assertion generalizes: each is a named
 field on a named entry.
 
-### 22g — the legacy `result.ast` is compared as a top-level KEY SET, nothing below it [D]
+### 22g — this gate compares the legacy `result.ast` as a top-level KEY SET only [D]
 
 `modernAst: true` is compared to official as a whole canonicalized tree; the **default** `ast` —
-the Svelte-4 legacy tree, wired in #3295 — is compared only as `Object.keys(ast).sort()`, so
-`{css, html, instance, module, _comments}` agreeing is the entire verdict. Everything inside is
-unobserved here and nowhere else: no gate in this repo compares the legacy tree's contents on
-either shape. Measured over 400 `runtime-legacy` components, **22 are byte-identical to official
-and 378 diverge** in nine classes (`.raw`/`.data` on 199 files from upstream's in-place
+the Svelte-4 legacy tree, wired in #3295 — is compared here only as `Object.keys(ast).sort()`, so
+`{css, html, instance, module, _comments}` agreeing is the entire verdict. The later public
+`parse()` AST gate (§39) now compares the legacy tree recursively over the full corpus, but that
+does not broaden this gate: if §39 is removed or filtered, this check cannot notice any value
+below those root keys. Before §39 existed, a measurement over 400 `runtime-legacy` components
+found **22 byte-identical to official and 378 divergent** in nine classes (`.raw`/`.data` on 199
+files from upstream's in-place
 `clean_nodes` mutation, expression `loc` coverage on 311 files, computed-key positions on 15,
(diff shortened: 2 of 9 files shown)


baseballyama:

PR #3850, opened 2026-08-26
Clear async server spread matrix residue

### LINKED ISSUE #3648: server: an awaited attribute, directive or spread value is not hoisted into `$$renderer.child(async …)`

baseballyama:
## Summary

- remove the 38 stale server/server-dev spread entries from the generated async-attribute matrix
- remove the six stale Appwrite server/server-dev corpus entries cleared by async spread hoisting
- cover object-valued spreads with inline awaits and nested async-IIFE awaits across element, dynamic-element, and component lowering
- update both ratchet inventories and their checked partitions

## Stack

- Base: #3764 (`fix/server-async-attribute-hoist`)
- #3764 supplies the promise optimiser routing and scope-aware object-expression await scan; this PR records and gates the spread part of that behavior

## Validation

- `git diff --check`
- `node scripts/compat-corpus/known-failures-md-check.mjs`
- Full compile/test and generated matrix verification are delegated to CI per the current no-local-test constraint.

Closes #3648


baseballyama:
Folded into #3764 so the implementation, all observed ratchet shrinkage, and its focused regression land atomically. A baseline-only child leaves the parent red under the two-sided ratchet.

PR #3851, opened 2026-08-26
Report aligned parse comment owners

baseballyama:
## Summary

- add an opt-in aligned comment-owner report to the public `parse()` AST parity sweep
- join comments only by their immutable source range and owners by `(type, start, end)`, independently of the field-level ratchet key
- classify moved, missing, and extra comments separately, including whether the official owner node is absent
- enable that report in the corpus CI job so the remaining #3702 serializer-order clusters are measured before they are changed

## Measured result

Corpus Compat run 32962432489 completed green across all jobs. Over 66,511 public AST comparisons, the report found 2,598 aligned comment differences:

- modern: 557 moved, 746 missing, 56 extra, 714 expected-owner nodes absent
- legacy: 437 moved, 746 missing, 56 extra, 1,118 expected-owner nodes absent
- loose: 0 moved, 0 missing, 0 extra

The largest transition is `TSPropertySignature.leadingComments -> ABSENT` at 505 occurrences on each component AST axis. The next clusters move comments from import or type declarations onto export declarations and other statement owners. This supports splitting follow-up fixes by owner transition rather than by aggregate comment count.

## Validation

- `node --check scripts/compat-corpus/parse-ast-verify.mjs`
- `git diff --check`
- full Corpus Compat run 32962432489: green

Stacked on #3822. This is diagnostic infrastructure for #3702; it does not close the issue.

### DIFF
--- a/.github/workflows/corpus-compat.yml
+++ b/.github/workflows/corpus-compat.yml
@@ -239,7 +239,7 @@ jobs:
       # 28k compared pairs, against ~59 minutes for one shard of the job that
       # currently costs the most.
       - name: Verify public parse() AST parity (ratchet)
-        run: node scripts/compat-corpus/parse-ast-verify.mjs
+        run: node scripts/compat-corpus/parse-ast-verify.mjs --comment-owners
 
       # Asserts a PROPERTY of the transform rather than comparing its output, so
       # it sees the #3026 class on inputs whose output is byte-perfect: un-seal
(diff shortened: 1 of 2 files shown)


baseballyama:
Superseded by #3868, which measures the aligned comment-owner residual after both #3761 and #3822 are combined and excludes structural/missing-comment differences from owner movements.

PR #3853, opened 2026-08-26
fix(compiler): map hoisted imports from retained spans

baseballyama:
## Summary

- carry retained phase-1 import spans through client import hoisting
- delete the client-side generated-text/source matching pass for verbatim imports
- preserve mappings through TypeScript type-only import cleanup and the standalone phase-3 API

Progresses #3015.

## Validation

- cargo fmt --all -- --check
- git diff --check

Rust builds and tests are intentionally delegated to CI because the local machine is under sustained load.


baseballyama:

PR #3854, opened 2026-08-26
diagnose: identify token source-map owners

baseballyama:
Progresses #3015.

Temporary CI diagnostic. This branch integrates the current source-carrier slices from #3843–#3848 and #3853–#3857, without changing those implementation PRs or `main`. After those slices remove ten generated-text enrichment passes, the source-map integration test starts its own test binary twice, once normally and once with only the remaining client `token` mapping pass disabled. For every generated identifier or literal position owned only by that pass, it compares the mapping against official Svelte when generated code is byte-identical and classifies it as exact, different, absent, or incomparable. The test intentionally fails after printing the inventory.

## CI result

The current macOS diagnostic completed in [run 33015071082](https://github.com/baseballyama/rsvelte/actions/runs/33015071082):

- mappings owned by the remaining token pass: **32**
- official exact: **0**
- official different: **0**
- official absent: **0**
- incomparable: **32**

The build, header checks, and C ABI cargo tests all passed before the diagnostic's intentional failure. The pass therefore still contributes 32 mappings after the carrier slices, but this corpus cannot judge any of them against official because every candidate sits on generated output without a byte-identical official line/source match. The result rules out using this diagnostic alone as a deletion gate. The residual mappings were reviewed with independently decoded official maps; they are absent or attach to the wrong generated token, so #3867 deletes the client token pass and this temporary diagnostic. The older run 32990988188 reported 40 mappings on an earlier head.

Scope boundary: the generated-token pass does not visit punctuation. In particular, it cannot determine ownership of a named import's opening `{`; the independent esrap punctuation mapping and its split-coordinate oracle are isolated in #3859.

Local validation is limited to `cargo fmt --all`, `cargo fmt --all -- --check`, and `git diff --check`; builds/tests are delegated to CI because this machine is under load.

### DIFF
--- a/.changeset/brave-wrappers-map.md
+++ b/.changeset/brave-wrappers-map.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Map comment-bearing client component function braces directly from AST positions.
--- a/.changeset/bright-props-map.md
+++ b/.changeset/bright-props-map.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve source-map carriers for lowered legacy prop reads.
--- a/.changeset/carry-element-handle-spans.md
+++ b/.changeset/carry-element-handle-spans.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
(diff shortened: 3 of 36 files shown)


baseballyama:

PR #3855, opened 2026-08-26
fix(compiler): delete collapsed declaration map pass

baseballyama:
## Summary

- delete the client/server `collapsed_declaration` text-matching source-map pass
- add a compile-level regression that deliberately puts `let` and its name on separate source lines, verifies both generated targets collapse the declaration, and pins the generated name to the source name
- document the constructed population that makes deletion safe despite the upstream sourcemap fixture gate measuring zero contribution
- add a compiler patch changeset

Progresses #3015.

## Validation

- `git diff --check`

Full builds and tests are delegated to CI because the local machine is under high load.

### DIFF
--- a/.changeset/tidy-collapsed-declaration-maps.md
+++ b/.changeset/tidy-collapsed-declaration-maps.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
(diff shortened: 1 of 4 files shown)


baseballyama:

PR #3857, opened 2026-08-26
fix(compiler): delete comment wrapper map pass

baseballyama:
## Summary

- carry component function brace mappings separately from comment-placement spans
- delete the retained comment-only client wrapper text-matching pass
- add a comment-bearing compile regression covering all four brace boundary segments
- correct the source-map refactor and gate-coverage documentation

Progresses #3015.

## Validation

- `git diff --check`

Full builds and tests are delegated to CI because the local machine is under heavy load.


baseballyama:

PR #3859, opened 2026-08-26
fix(esrap): map hoisted import token boundaries

baseballyama:
## Summary

- emit retained import keywords with the same sequential source anchors as upstream esrap
- preserve default, namespace, named-specifier, and module-string node boundaries while printing hoisted imports
- add split-coordinate coverage for every boundary of a representative named import

This is the esrap-only import-token slice stacked after #3853. The retained import construction in #3853 supplies the original spans; this PR makes the printer retain the mappings those spans carry. It addresses part of #3015 and does not close it.

## Why these boundaries

Upstream does not search the source for import punctuation. Its `import ` keyword boundary owns the generated `{` column, the named identifier's end owns the generated `}` column, and the module string's end owns the generated semicolon column. Reproducing those node-boundary writes restores the official segments without adding mappings that official omits.

## Verification

- `git diff --check`

Per repository instructions, I did not run local builds or tests; CI is the execution oracle.

### DIFF
--- a/.changeset/curly-import-maps.md
+++ b/.changeset/curly-import-maps.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Preserve source-map spans for hoisted client imports.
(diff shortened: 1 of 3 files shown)


baseballyama:

PR #3862, opened 2026-08-26
Reject arguments references in template expressions

### LINKED ISSUE #3707: `arguments` in a template expression is accepted; `invalid_arguments_usage` never fires outside the instance script

baseballyama:
Closes #3707.

Template expressions use the typed expression walker and bypassed the script Identifier visitor, so invalid_arguments_usage never fired there. Track ordinary-function depth in that walker: function declarations/expressions introduce arguments, arrows only inherit it.

Regression coverage checks client and server across text, attribute, if, each, const, and html hosts; it also covers top-level arrows, nested arrows, default parameters, and exact diagnostic spans.

Local verification (no local build per project workflow):
- cargo fmt --all -- --check
- git diff --check

Full validation is delegated to CI.

### DIFF
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/visitors/mod.rs
@@ -370,6 +370,10 @@ pub struct VisitorContext<'a> {
     pub parent_element: Option<String>,
     /// Current function depth.
     pub function_depth: usize,
+    /// Depth inside ordinary functions while walking a template expression.
+    /// Arrow functions do not introduce their own `arguments`, but inherit one
+    /// from an enclosing ordinary function.
+    pub template_regular_function_depth: usize,
     /// Depth inside $derived(...) expressions (but not $derived.by(...)) or @const
     pub derived_function_depth: usize,
     /// Whether we have a $props() rune.
@@ -604,6 +608,7 @@ impl<'a> VisitorContext<'a> {
             bind_has_await: false,
             parent_element: None,
             function_depth: 0,
+            template_regular_function_depth: 0,
             derived_function_depth: 0,
             has_props_rune: false,
             component_slots: rustc_hash::FxHashSet::default(),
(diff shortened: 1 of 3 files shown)


baseballyama:
Closing as a duplicate of #3746, which already covers #3707 together with #3694 and has a full green CI run. No merge performed.

PR #3865, opened 2026-08-26
fix(sourcemap): retain copied raw token ends

baseballyama:
Part of #3015.

Depends on #3846.

RawMapped linear ranges represent copied bytes as half-open ranges, but AST spans also use the exclusive end as a source-map boundary. The split-coordinate printer previously looked that end up as the start of the following generated range, dropping exact copied-token end mappings.

This change gives end-position lookup the same boundary ownership as the existing RawMapped fallback restorer: a copied linear range ending at the boundary wins over the following range. Start-position lookup and generated ranges are unchanged. A focused printer test covers a copied identifier followed immediately by a generated call suffix.

Validation: cargo fmt --all -- --check; git diff --check. Per project instruction, no local build or test command was run.


baseballyama:

PR #3866, opened 2026-08-26
fix(sourcemap): retain member root identifier spans

baseballyama:
Part of #3015.

Depends on #3865.

## What changed

- preserve source spans for member-expression root identifiers without wrapping the object in `JsExpr::Spanned`
- store only these uncommon spans in a lazy arena side table, avoiding growth of every `JsExpr` or `JsMemberExpression`
- restore the carrier in both the OXC/esrap path and the handwritten source-map fallback
- cover the IR shape, both printers, and an end-to-end `Math.random()` template expression

This keeps downstream member-chain matchers seeing the original `Member -> Identifier` variants; an in-band wrapper in this position previously changed transform semantics.

## Validation

- `git diff --check`
- `git show --check`

Per project instructions, no local build or test command was run; CI is the test oracle.


baseballyama:

Issue #719, opened 2026-06-07
Panic in svelte2tsx (char-boundary slice) on multibyte/CJK `<script>` content — `--emit-overlay`/`--tsgo` aborts


baseballyama:
## Summary

`svelte-check --emit-overlay` / `--tsgo` panics on `.svelte` files whose `<script>` contains multibyte (e.g. Japanese / CJK) characters in a particular byte layout:

```
thread 'main' panicked at crates/rsvelte_core/src/svelte2tsx/script/mod.rs:1495:32:
start byte index 173 is not a char boundary; it is inside '須' (bytes 172..175 of string)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The whole run aborts (no diagnostics, non-zero via panic). It happens in the **svelte2tsx overlay** generation, so it affects `--emit-overlay` and `--tsgo` (the TypeScript path). Plain default mode (svelte/a11y/css only) is unaffected.

## Minimal repro

```svelte
<!-- src/T.svelte -->
<script lang="ts">
  interface Props {
    /** アバターのコンテンツタイプ */
    content: 'image' | 'initial' | 'count';
    /** 画像のソース (content='image' の場合に必須) */
    imageSrc?: string;
  }
  const { content, imageSrc }: Props = $props();
</script>
<p>{content}{imageSrc}</p>
```

```sh
svelte-check --workspace "$(pwd)" --emit-overlay   # absolute path; see note*
# => thread 'main' panicked ... start byte index 173 is not a char boundary; it is inside '須'
```

`@rsvelte/svelte-check@0.1.6`, macOS arm64. Reproduces with `--tsgo` too.

\* absolute `--workspace` used as a workaround for the separate `--workspace .` = 0-files bug.

## What pins down the cause

It's an **offset miscalculation**, not "Japanese in a comment" per se — the trigger is byte-position-dependent:

- Removing the **first** JSDoc comment (`/** アバターのコンテンツタイプ */`) → no panic.
- Changing `content: 'image' | 'initial' | 'count'` to `content: string` → no panic.
- The **same structure with ASCII-only comments** → no panic.

[...]

PR #4075, opened 2026-09-01
fix(svelte2tsx): a byte offset is not a char boundary, and `..=` cuts one past it

### LINKED ISSUE #4072: svelte2tsx panics (SIGABRT) on a bind: expression preceded by multibyte/astral text — byte offset used as str slice index
## svelte2tsx panics (SIGABRT) on a `bind:` expression preceded by multibyte/astral text — byte offset used as a `str` slice index

`svelte2tsx` (via `rsvelte-check --tsgo` overlay materialization) aborts with a UTF-8 char-boundary panic when a `bind:` value expression is preceded by enough multibyte/astral characters that the expression's byte start offset lands **inside** a multibyte character.

### Panic

```
thread 'main' panicked at crates/rsvelte_projection/src/svelte2tsx/template/utils/opener_spacing.rs:441:38:
end byte index 185 is not a char boundary; it is inside '値' (bytes 184..187 of string)
... svelte-check was terminated by SIGABRT.
```

### Repro

This is your own corpus fixture, `compatibility/lint-adversarial/no-bind-value-on-checkable-inputs/18-ts-and-astral.svelte`, which crashes `svelte-check` even though the lint/compat gates pass on it:

```svelte
<script lang="ts">
  let 値: string = '';
  const kind = 'checkbox' as const;
</script>

<div data-x="𝔘𝔫𝔦👨‍👩‍👧‍👦">👩‍🚀<input type={kind} bind:value={値} /></div>
<div data-y="😀😀"><input type="radio" bind:value={値 as string} /></div>
```

```
rsvelte-check --tsgo --workspace <dir> --no-tsconfig   # panics
```

The astral/ZWJ characters in the earlier attributes accumulate a byte-vs-codepoint drift, so by the second `bind:value={値 as string}` the computed `expr_start` no longer sits on a char boundary.

### Root cause

`crates/rsvelte_projection/src/svelte2tsx/template/utils/opener_spacing.rs:441`:

```rust
let equals = source_offset(source[..=expr_start as usize].rfind('=').unwrap_or(0));
```

`expr_start` is a **byte** offset; `source[..=expr_start]` slices the `str` and panics when that byte is a UTF-8 continuation byte. Since the search target `'='` is ASCII, the slice is unnecessary — a byte search is boundary-safe:

```rust
let equals = source_offset(
    source.as_bytes()[..=expr_start as usize]
        .iter()
        .rposition(|&b| b == b'=')
        .unwrap_or(0) as u32,
);
```

### Sibling site (same class, still unfixed)

`crates/rsvelte_projection/src/svelte2tsx/template/nodes/if_else_block.rs:178`:

```rust
let else_close = source[..alternate_start as usize].rfind('}');
```

Same pattern — a `{:else}` preceded by multibyte/astral text would panic identically. Worth fixing together.

### Note

This looks like a regression of the byte-vs-char offset class swept in #2378 / #2386 (type-level byte/char distinction). `opener_spacing.rs` is recent (#4043) and reintroduced the raw `expr_start as usize` slice at a site the sweep never covered. The svelte2tsx gate doesn't appear to exercise this fixture's overlay path, which is why the corpus is green while `svelte-check` aborts.

Environment: `@rsvelte/svelte-check` 0.5.22 (darwin-arm64 native).

baseballyama:
Fixes #4072.

## The defect

`push_bind_directive_ranges` (`opener_spacing.rs`) located a `bind:` directive's `=` with

```rust
source[..=expr_start as usize].rfind('=')
```

`expr_start` is a **byte** offset and the range is **inclusive**, so the slice ends at
`expr_start + 1` — inside the expression's first character whenever that character is
multi-byte. Under `panic = "abort"` the panic takes the process with it, which is why the
report reads `svelte-check was terminated by SIGABRT` rather than as a diagnostic:
`materialize_overlay` maps a svelte2tsx *error* to one, and a panic never gets there.

**The reported mechanism is not quite the one that fires.** The issue attributes it to
astral/ZWJ text "accumulating a byte-vs-codepoint drift"; there is no drift. `expr_start`
is a valid boundary — it is the `..=` that cuts one byte past it. `<input bind:value={値} />`
is enough, with no astral characters anywhere. Measured: a 512-cell sweep (33 constructs ×
4 character classes in the expression slot × 4 in a padding slot) panics in 56 cells, every
one of them a `bind:`, and 0 after the fix.

Every search target in this class is ASCII, so a byte search returns the same index the
`str` search did on every input the old code did not abort on.

## Enumerating the class

`crates/rsvelte_projection` was swept for the shape rather than taking the issue's table as
a partition. 23 sites index a `str` with an `as usize` bound and 21 more slice one with a
variable; all but four take their bound from an AST span or from a `find`/`memchr` hit on an
ASCII needle, both of which are boundaries by construction. The four that do byte arithmetic:

| site | verdict |
|---|---|
| `opener_spacing.rs:441` — `[..=expr_start]` | **fixed.** The reported defect. Unsafe *by construction*: `..=` cuts one past a boundary. |
| `hoistable_types.rs:805` — `[extends_end - 7 .. extends_end]` | **fixed.** A second live site, found by this sweep, not reported. |
| `hoistable_types.rs:769` — `[start .. start + 9]` | not changed. Measured on a boundary in 6/7 probes that reach it (`assert!(false)` positive control confirms reachability); the nine bytes are inside the ASCII `interface` keyword region. |
| `if_else_block.rs:178` — `[..alternate_start]` | **hardened, no test.** Same spelling, but *exclusive*, and `alternate_start` is either an AST node start or one past an ASCII `}`. |
| `hoistable_types.rs:142` — `[j - 6 .. j]` | already carries an explicit `is_char_boundary` guard (#719). The precedent this fix follows. |

### `hoistable_types.rs:805` — the second live site

`rewrite_interface_to_type_dts` walks back from the first heritage entry over whitespace and
reads the seven bytes before it as the `extends` keyword. A comment between `extends` and the
entry puts those bytes inside the comment's text:

```svelte
<script lang="ts">
  interface B { a: string }
  interface A extends /*日本*/B { b: string }
  export let x: A;
</script>
```

aborts under `--mode dts` (`svelte-package`), which is why `rsvelte-check` never reached it.
`/*😀*/` does **not** reproduce it — 8 bytes puts `-7` on the `*` — which is what says the
arithmetic is the mechanism.

The comment also defeats the keyword walk itself, so neither that case nor its ASCII control
rewrites `extends`. That is a **separate, output-level divergence** from upstream, which takes
the position from `heritageClauses[0].getStart()` (a TypeScript heritage clause starts at the
keyword) and never scans — worth its own issue. The tests here therefore pin the two cases
against **each other**, plus a no-comment control that does rewrite, rather than pinning the
current output as correct.

### `if_else_block.rs:178` — the reported sibling, measured

The issue calls this "same pattern — worth fixing together". It is the same spelling and not
the same defect: the slice is exclusive, so it only panics if `alternate_start` is *itself*
mid-character. Measured with a 1750-cell `{:else}` sweep (14 shapes × 5 character classes in
three slots):

- `assert!(false, "SITE-REACHED")` at the site → **1750/1750** panics. The sweep reaches it in
  every cell.
- `assert!(source.is_char_boundary(alternate_start))` → **0/1750**. The offset is on a boundary
  in every cell.

Hardened anyway, because the shape should not survive in the file. **No test was added for it**
— a test that cannot be made to go red measures nothing.

## Verification

| check | result |
|---|---|
| `cargo test --no-fail-fast -p rsvelte_projection -p rsvelte_check` | exit 0 — 90 test binaries `ok` (3 of them `unittests src/lib.rs`), 0 `FAILED` |
| `RSVELTE_REQUIRE_PREREQS=1 … --test svelte2tsx_fixtures` | 256/256, 0 failed, 0 skipped |
| `cargo fmt --check` | exit 0 |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0, 0 warning/error lines (root workspace) |
| `node scripts/ci/check-pattern-corpus-docs.mjs` | exit 0 |

### Positive controls

Both fixes were ablated one line at a time and the tests measured, rather than assumed:

| ablation | result |
|---|---|
| revert the `opener_spacing` line | **6/6 `bind:` tests red**, each with `end byte index N is not a char boundary; it is inside '値'` at that site — the reported failure mode |
| revert the `hoistable_types` line | **exactly 1 red** (the multi-byte dts test); its ASCII control and the other 8 stay green |
| revert the `opener_spacing` line, `rsvelte_check` overlay test | red, same message — pinning the reported entry point, not just the library call |

Every test fails independently: there is no "one of these panics" assertion that another
site's fix could satisfy.

### The repro file

`compatibility/pattern-corpus/issues/4072-bind-value-multibyte-boundary.svelte`, minimised
from the reported file (the astral padding is not load-bearing) and formatted with the
corpus's own `oxfmt -c scripts/fixtures/fmt-corpus.oxfmtrc.json`. Verified against the
official compiler at the pinned submodule before landing, so it enrols green rather than
seeding a ratchet entry:

```
client:     BYTE-IDENTICAL
server:     BYTE-IDENTICAL
client-dev: BYTE-IDENTICAL
svelte2tsx: BYTE-IDENTICAL
```

(Byte equality *before* the gates' oxfmt normalization, i.e. strictly stronger than what they
require. The fmt-parity gate itself was not run — `oxfmt` is only in the primary checkout —
so that one is unverified locally.)

Official svelte2tsx handles this input fine, so this was purely an rsvelte defect.


baseballyama:
#4076 が `compatibility/pattern-corpus/README.md` の同じアンカーに行を足したため衝突しました。
Integrator 側で `origin/main` にリベースし、**両方の行を残す**形で解決して force-push しています
（この repo の標準解決）。ソースの変更には触れていません。
`node scripts/ci/check-pattern-corpus-docs.mjs` は exit 0 です。

CI が再実行されるので、緑になり次第マージします。

PR #4080, opened 2026-09-01
fix(svelte2tsx): the dts heritage rewrite reconstructed three positions by scanning text

### LINKED ISSUE #4077: svelte2tsx --mode dts: a comment before an interface's heritage entry leaves `extends` in place, emitting TypeScript that does not parse
## `svelte2tsx --mode dts` leaves `extends` in place when a comment sits before the heritage entry, emitting TypeScript that does not parse

`rewrite_interface_to_type_dts` (`crates/rsvelte_projection/src/svelte2tsx/script/hoistable_types.rs`) rewrites `interface X extends Y { … }` into `type X = Y & { … }` for `--mode dts`. The `interface` → `type` half always lands; the `extends` → `=` half is skipped whenever **any** comment sits between `extends` and the first heritage entry, and the result is not valid TypeScript — `type A extends …` is a syntax error.

Found while fixing #4072 (the same function's `extends_end - 7` byte arithmetic panics when that comment contains a multi-byte character). That panic is fixed by #4075; **this is what the fixed code then emits**, so it is a separate defect and this issue is about the output, not the crash.

### Repro

`A.svelte`:

```svelte
<script lang="ts">
  interface B { a: string }
  interface A extends /*hi*/B { b: string }
  export let x: A;
</script>

{x}
```

```js
svelte2tsx(source, { filename: 'A.svelte', isTsFile: true, mode: 'dts', namespace: 'html', version: '5' })
```

### Actual output, both compilers

Official (`submodules/language-tools/packages/svelte2tsx/index.js`, pinned `092af38`):

```ts
  type B ={ a: string }
  type A = /*hi*/B &  { b: string }
```

rsvelte (NAPI binding, `fix/svelte2tsx-bind-multibyte-boundary-4072`):

```ts
  type B ={ a: string }
  type A extends /*hi*/B  & { b: string }
```

Fed back through `ts.createSourceFile(…, ScriptKind.TSX)`, official's whole file yields **0** `parseDiagnostics` and rsvelte's yields **2**, the first being `'=' expected.`

### Which cells reproduce

One axis — the comment between `extends` and `B` — measured on both compilers, with `parseDiagnostics` counted on each full output:

| comment | official | rsvelte | rsvelte syntax errors |
|---|---|---|---|
| *(none)* | `type A = B &  { b: string }` | `type A = B  & { b: string }` | 0 |
| `/*hi*/` | `type A = /*hi*/B &  { b: string }` | `type A extends /*hi*/B  & { b: string }` | 2 |
| `/*abcde*/` (9 bytes) | `type A = /*abcde*/B &  { b: string }` | `type A extends /*abcde*/B  & { b: string }` | 2 |
| `/*😀*/` (8 bytes) | `type A = /*😀*/B &  { b: string }` | `type A extends /*😀*/B  & { b: string }` | 2 |
| `/*日本*/` (10 bytes) | `type A = /*日本*/B &  { b: string }` | `type A extends /*日本*/B  & { b: string }` | 2 |
| `//hi` + newline | `type A = //hi` | `type A extends //hi` | 1 |

So it is **not** about the comment's content, its width, or its kind — every comment reproduces it, and the no-comment row is the control that stays correct on both sides. (In #4072's *panic* the byte width is the mechanism: `/*日本*/` aborts and `/*😀*/` does not, because `extends_end - 7` lands mid-character in one and on the `*` in the other. Here the width is irrelevant, which is what separates the two defects.)

### Cause

Upstream takes the position from the **heritage clause**, which in the TypeScript AST starts at the `extends` keyword, and never scans (`processInstanceScriptContent.ts:430`):

```ts
const extendsStart = node.heritageClauses[0].getStart() + astOffset;
str.overwrite(extendsStart, extendsStart + 'extends'.length, '=');
```

OXC's `TSInterfaceHeritage` span starts at the **type**, not at the keyword, so rsvelte reconstructs the position by walking backwards from `extends[0].span.start`:

```rust
let mut p = first_start;
while p > 0 {
    let prev = bytes[p - 1];
    if prev == b' ' || prev == b'\t' || prev == b'\n' || prev == b'\r' { p -= 1; } else { break; }
}
let extends_end = p;
if extends_end >= 7 {
    let prev_kw = &raw_content.as_bytes()[extends_end - 7..extends_end];
    if prev_kw == b"extends" { … }
```

The walk skips whitespace only. A comment is not whitespace, so `p` stops at the comment's last byte, `prev_kw` is the comment's tail rather than `extends`, and the overwrite is silently declined. Making the backward walk comment-aware (or recovering the keyword position some other way) is the fix; note that upstream keeps the comment *after* the `=` (`type A = /*hi*/B`), so the rewrite has to land on the keyword, not swallow the trivia.

### Why no gate sees it

- The corpus svelte2tsx-parity gate compiles with `mode: 'ts'` (`scripts/compat-corpus/svelte2tsx-compile.mjs:113`). **`mode: 'dts'` is not exercised by any corpus gate.**
- The fixture suite does reach this code: `transforms-interfaces-dts` covers `extends`, `extends A, B` and a generic `extends`, and passes 256/256. None of its interfaces carries a comment before the heritage entry, so the axis that reproduces this is unenumerated rather than uncovered.

### Also noticed, not the subject of this issue

Even the no-comment control differs in whitespace — official writes `B &  { b: string }` and rsvelte writes `B  & { b: string }`. Both parse and the fixture suite is green, so this is recorded rather than claimed as a defect; it may be the same fixture-coverage gap one step over.

baseballyama:
Fixes #4077.

## The defect

Upstream's `transformInterfacesToTypes` is entirely span-based
(`processInstanceScriptContent.ts:413-447`): the `extends` position comes from
`heritageClauses[0].getStart()` — a TypeScript heritage clause **starts at the keyword** — the gap
between two entries is overwritten whole, and ` & ` is appended at the clause's end.

OXC's `TSInterfaceHeritage` span starts at the **type**, not the keyword, so rsvelte rebuilt all
three positions by reading raw bytes: a backward walk that skipped whitespace but not comments, a
`find(',')` for the separator, and a `find('{')` for the body. A comment defeats every one of
them, and two of the three splice the operator into the comment's own body:

| source | rsvelte before |
|---|---|
| `interface I extends /*c*/B {}` | `type I extends /*c*/B  & {}` |
| `interface I extends B /*,*/, C {}` | `type I = B/* &*/ , C  & {}` |
| `interface I extends B /*{*/ {}` | `type I = B/* & {*/ {}` |

None of that is TypeScript. All three positions now come from spans. The keyword is the first
code token after the interface head, found by consuming the trivia **the parser already located**
(`program.comments`) rather than by adding a scan — the shape that caused this in the first place.

## The whitespace difference recorded on #4075 is the same site

`B &  {` (official) vs `B  & {` (rsvelte) is scan 3: rsvelte appended ` & ` before the body's `{`,
upstream appends it at the last entry's **end**. It is not comment-dependent — it fired on **every**
extends-bearing interface. Moving the anchor settles it, which is why the comment-free controls in
the new test file are load-bearing rather than decorative.

## Measurement

202 cells — 6 interface shapes (1/2/3 heritage entries, generic entry, generic interface, no
heritage) × 4 comment slots (after the name, after `extends`, between entries, before the body) ×
5 comment kinds (`/*c*/`, `/*,*/`, `/*{*/`, `/*extends*/`, `//c`) × `dts`/`ts`, plus a comment-free
control per shape — against the pinned official svelte2tsx, with `ts.createSourceFile`
`parseDiagnostics` counted on **both** sides:

| | before | after |
|---|---|---|
| byte-equal | 110 | **200** |
| differ | 92 | **2** |
| rsvelte unparseable | 32 | **0** |
| official unparseable | 2 | 2 |
| panics | 0 | 0 |

**`mode: 'ts'` is the control**: 0 of its 101 cells ever diverged, before or after. The whole class
is `dts`-only.

The 32 unparseable cells were not one bug — they were the three scans, in the proportion
25 / 2 / 5 (`extends` keyword / comma-in-a-comment / brace-in-a-comment).

### The 2 remaining differences run the other way, and stay

Upstream reaches a bodyless interface's `{` with `str.original.indexOf('{', node.getStart())`,
which a comment holding a brace sends into the comment's text:

```
source    interface I /*{*/ { b: string }
official  type I /*={*/ { b: string }     <- 1 parse diagnostic
rsvelte   type I /*{*/ ={ b: string }     <- 0
```

rsvelte takes the position from the body span and is correct. Matching would mean reproducing an
upstream bug, so `controls::a_brace_in_a_comment_does_not_move_the_equals_of_a_bodyless_rewrite`
pins rsvelte's answer instead of official's.

## What the gates could not see — measured, not argued

- The corpus svelte2tsx-parity gate compiles with `mode: 'ts'`
  (`scripts/compat-corpus/svelte2tsx-compile.mjs:113`). **No corpus gate reaches `mode: 'dts'`.**
- The fixture suite *does* reach this code — `transforms-interfaces-dts` covers `extends`,
  `extends A, B` and a generic `extends`. It cannot see either half, and both were checked by
  reverting the fix and re-running rather than by reading the harness:

  | reverted | fixture suite |
  |---|---|
  | the `extends` keyword fix | `PASS: transforms-interfaces-dts`, **256/256** — no fixture writes a comment there |
  | the ` & ` anchor fix | `PASS: transforms-interfaces-dts`, **256/256** — the comparison falls back to whitespace-collapsed equality (`tests/common/svelte2tsx.rs:156,173`) |

  The checked-in expected output carries official's spacing (`type Bar2 = Bar1 &  {`), so that
  fixture had been passing on the *collapsed* comparison and now passes byte-exactly.

**No pattern-corpus repro is added on purpose.** That corpus reaches svelte2tsx only through the
`mode: 'ts'` gate, so a file there would measure nothing about this defect. The new Rust test file
is the gate.

## Verification

| check | result |
|---|---|
| `cargo test --no-fail-fast -p rsvelte_projection -p rsvelte_check` | exit 0 — 90 test binaries `ok`, 0 `FAILED` |
| `RSVELTE_REQUIRE_PREREQS=1 … --test svelte2tsx_fixtures` | 256/256, 0 failed, 0 skipped |
| `cargo fmt --check` | exit 0 |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0, 0 warning/error lines |

### Positive controls

One scan reverted at a time, each producing a distinct signature — so a regression in any one of
the three is detected on its own:

| reverted scan | tests red |
|---|---|
| `extends` keyword | **2** — both `extends` rows. The separator and brace rows stay green. |
| entry separator | **1** — the comma-in-a-comment row. |
| trailing ` & ` anchor | **5** — its own row plus **both comment-free controls**, because this scan fires on every extends-bearing shape. |

## Interaction with #4075 (please merge #4075 first)

This branch is cut from `origin/main`, so it does not contain #4075. The two touch the same
function and #4077's fix **subsumes** #4075's `hoistable_types` half: the `extends_end - 7` slice
that #4075 made boundary-safe is deleted here, so `interface A extends /*日本*/B` compiles on this
branch with no `#4075` present — byte-identical to official, no panic.

On rebase, expect:

1. A conflict in `hoistable_types.rs` — resolve in favour of this branch (the byte slice is gone).
2. Two of #4075's tests to need updating. They deliberately pinned the *unrewritten* output
   ("the two cases must agree with each other … without pinning the wrong output as correct"), and
   that output is now correct. Both inputs measured against official on this branch:

   | #4075 test | new expectation |
   |---|---|
   | `multibyte_comment_between_extends_and_its_heritage_entry` | `type A = /*日本*/B &  { b: string }` |
   | `ascii_comment_between_extends_and_its_heritage_entry` | `type A = /*hi*/B &  { b: string }` |

   Both are byte-equal to official. `no_comment_still_rewrites_extends` is unaffected.

### DIFF
--- a/.changeset/dts-interface-heritage-comment.md
+++ b/.changeset/dts-interface-heritage-comment.md
@@ -1,0 +1,7 @@
+---
+"@rsvelte/svelte2tsx": patch
+---
+
+Emit parseable TypeScript from `--mode dts` when an interface's heritage clause carries a comment.
+
+The `interface X extends Y { … }` → `type X = Y & { … }` rewrite reconstructed three positions by scanning raw text — a backward walk to the `extends` keyword that skipped whitespace but not comments, a `find(',')` for the separator between entries, and a `find('{')` for the body. A comment defeated all three: `extends` survived (`type X extends …`), and a `,` or `{` written inside a comment took the operator into the comment's body. All three now come from spans, as upstream's does. The trailing ` & ` moves with them, so the join now spells `Y &  {` like official instead of `Y  & {`.
--- a/crates/rsvelte_projection/src/svelte2tsx/script/hoistable_types.rs
+++ b/crates/rsvelte_projection/src/svelte2tsx/script/hoistable_types.rs
@@ -759,6 +759,7 @@ fn find_line_comment_start(line: &[u8]) -> Option<usize> {
 pub(super) fn rewrite_interface_to_type_dts(
     iface: &oxc_ast::ast::TSInterfaceDeclaration<'_>,
     raw_content: &str,
+    comments: &[oxc_ast::ast::Comment],
     offset: u32,
     str: &mut MagicString<'_>,
 ) {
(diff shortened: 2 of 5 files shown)


baseballyama:
Re-read the grid as a **set difference** rather than as totals, because 110 → 200 byte-equal is a number that a mix of repairs and regressions can produce:

```
differing before: 92    after: 2
REGRESSED (equal before, differing after):  0
repaired  (differing before, equal after): 90
still differing: no-extends | slot=N block-brace | dts
                 no-extends | slot=L block-brace | dts
```

The after-differing set is a strict subset of the before-differing set, so **no cell that was byte-equal to official stopped being byte-equal** — and the 2 survivors are exactly the two where official is the unparseable side (`type I /*={*/ {`, upstream's `indexOf("{")` landing inside the comment), which were already differing before this change.

That matters most for the trailing ` & ` anchor, which fires on every extends-bearing interface rather than only the comment cases — its blast radius is the widest of the three scans, and the totals alone could not have told a repair there from a swap.

baseballyama:
Rebased onto `f87bcc607` (#4075 merged). One conflict, in `hoistable_types.rs`, resolved in favour of this branch — the `extends_end - 7` slice #4075 made boundary-safe is deleted here, so #4077's fix subsumes that half.

**#4075's two dts rows are updated in a second commit**, because they had become a comment that explains the intent of a fix while describing code that no longer exists — the shape that gets the correct output deleted as a defect by the next reader. Their module doc explained the byte slice; both expectations pinned the un-rewritten `type A extends /*…*/B`, on the stated grounds that the two cases only had to agree with each other. All three sentences are false after this branch.

The new expectations were measured before they were written, on the rebased tree, with a fresh NAPI build:

```
hi:   official = "type A = /*hi*/B &  { b: string }"   rsvelte = same, whole file equal
日本: official = "type A = /*日本*/B &  { b: string }" rsvelte = same, whole file equal
```

The module keeps its own reason to exist: every row in `svelte2tsx_dts_interface_heritage_4077.rs` uses an ASCII comment, so the **multi-byte** one lives only there, and it now says that `/*日本*/` behaves exactly like the `/*hi*/` beside it — which is what goes red if byte arithmetic returns to the site.

Re-measured after the rebase, not carried over:

| check | result |
|---|---|
| 202-cell grid | byte-equal **200** / differ 2 / rsvelte-unparseable **0** / panics 0 |
| grid set difference | REGRESSED **0**, repaired 90, the 2 survivors are the same official-unparseable cells as before |
| `cargo test --no-fail-fast -p rsvelte_projection -p rsvelte_check` | exit 0 — **92** test binaries `ok`, 0 `FAILED` |
| `RSVELTE_REQUIRE_PREREQS=1 … --test svelte2tsx_fixtures` | **256/256**, 0 failed, 0 skipped |
| `cargo fmt --check` | exit 0 |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0, 0 warning/error lines |

Issue #4090, opened 2026-09-01
css: a nested `:global(...)` inside a `:global` block keeps its wrapper in the emitted CSS


baseballyama:
Inside a `:global` block the contents are already unscoped, so official unwraps a nested `:global(...)` to its argument in the emitted CSS. rsvelte keeps the `:global(...)` wrapper.

## Repro

```svelte
<svg class="p q"><path class="a"><i class="b"><u class="c"></u></i></path></svg>

<style>
	.p :global {
		:global(.a) { opacity: 1; }
	}
</style>
```

`generate: 'server'`, `css: 'external'`, official compiler at `submodules/svelte/packages/svelte/src/compiler/index.js`.

| | css output |
|---|---|
| official | `.p.svelte-1lj1c24 { .a { opacity: 1; } }` |
| rsvelte | `.p.svelte-1lj1c24 { :global(.a) { opacity: 1; } }` |

The generated template is byte-identical on both sides (`<svg class="p q svelte-1lj1c24">`, nothing below it scoped), and neither compiler warns. `:global(...)` is not valid CSS a browser understands, so rsvelte's output is not merely differently spelled — it is a selector no browser will match.

## In-file controls

Same markup, same block, agreeing byte-for-byte on both sides:

| style | both compilers |
|---|---|
| `.p :global { .a { opacity: 1; } }` | css `.p.HASH { .a { … } }` |
| `.p :global { & .a { opacity: 1; } }` | css `.p.HASH { & .a { … } }` |

So the divergence is specific to a nested `:global(...)` inside a global block.

## How it was found

Generated grid, 9 parent preludes × 9 child selectors × 2 targets; cell `plain-then-bare / implicit-global`. Present before and after #4059's fix — not a regression from that work.

## What was not measured

Whether the same shape under a bare `:global { … }` (no `.p` prefix) diverges — that cell throws `css_global_block_invalid_modifier_start` on both compilers for the sibling children, so it was not reached here.

baseballyama:
Fixed on `main` — measured, not inferred.

Arm: a NAPI build of `main` at `9171df619`; oracle `submodules/svelte/packages/svelte/src/compiler/index.js`; the issue's repro verbatim, `generate: 'server'`, `css: 'external'`.

```
#4090 repro   css=EQ   html=EQ   warn=EQ
```

[...]

Issue #4089, opened 2026-09-01
css: an `&.a` child of a prefixed `:global` block scopes three elements official leaves bare


baseballyama:
`.p :global { … }` is a *prefixed* global block: the `.p` prefix still scopes, and everything the block contains is emitted unscoped. With an `&.a` child, rsvelte adds the scope class to three elements official leaves bare, while the emitted CSS is byte-identical on both sides.

## Repro

```svelte
<svg class="p q"><path class="a"><i class="b"><u class="c"></u></i></path></svg>

<style>
	.p :global {
		&.a { opacity: 1; }
	}
</style>
```

`generate: 'server'`, `css: 'external'`, official compiler at `submodules/svelte/packages/svelte/src/compiler/index.js`.

| | output |
|---|---|
| official css | `.p.svelte-1lj1c24 { &.a { opacity: 1; } }` |
| rsvelte css | `.p.svelte-1lj1c24 { &.a { opacity: 1; } }` |
| official html | `<svg class="p q svelte-1lj1c24"><path class="a"><i class="b"><u class="c"></u></i></path></svg>` |
| rsvelte html | `<svg class="p q svelte-1lj1c24"><path class="a svelte-1lj1c24"><i class="b svelte-1lj1c24"><u class="c svelte-1lj1c24"></u></i></path></svg>` |

Neither compiler emits a warning. The divergence is in the template alone: `path`, `i` and `u` all gain a scope class they should not have.

## In-file controls

Both of these agree byte-for-byte between the two compilers on the same markup, so the divergence is specific to `&.a` under a prefixed global block, not to the block or the markup:

| style | both compilers |
|---|---|
| `.p :global { .a { opacity: 1; } }` | css `.p.HASH { .a { … } }`, html scopes `svg` only |
| `.p :global { & .a { opacity: 1; } }` | css `.p.HASH { & .a { … } }`, html scopes `svg` only |

## How it was found

`scripts/compat-corpus/matrix`-style grid, 9 parent preludes × 9 child selectors × 2 targets. This cell is `plain-then-bare / explicit-self` and is present both before and after #4059's fix, so it is not a regression from that work.

## What was not measured

Whether any collected-corpus file reaches this shape — the grid is generated, and this cell has no corpus carrier that I looked for.

PR #4093, opened 2026-09-01
fix(css): a nested rule under a fully-global parent needs no ancestor to match

### LINKED ISSUE #4059: css: a nested rule under a :global() parent does not scope its ancestors
## 最小再現（4 行）

```svelte
<div><svg><path class="a"/></svg></div>
<style>:global(.x) { .a { opacity: 1; } }</style>
```

`generate: 'client', dev: false, css: 'external'`、オラクルは
`submodules/svelte/packages/svelte/src/compiler/index.js`（`OFFICIAL_COMPILER_REL`）。

| | scope class が付く要素 |
|---|---|
| official | `<div>`, `<svg>`, `<path class="a">` |
| rsvelte | `<path class="a">` のみ |

CSS 出力は両者一致（ハッシュも `svelte-1lj1c24` で一致）。差は **JS テンプレート側で
どの要素に scope class を付けるか**だけ。

## 効く条件と効かない条件（実測）

| 入力 | 一致 |
|---|---|
| `:global(.x) { .a { … } }`（ネスト） | **不一致** |
| `:global(.x), :global(button:active) { .a { … } }` | **不一致** |
| `:global(.x) .a { … }`（ネストを平坦化した同義形） | 一致 |
| `.y { .a { … } }`（親が `:global` でないネスト） | 一致 |
| `.a { … }`（ネスト無し） | 一致 |
| `.b { :global(.a) { … } }`（内側が `:global`） | 一致 |

つまり **「ネストしていて、かつ親セレクタが `:global(...)` のとき」だけ**、official は
一致した要素の**祖先にも** scope class を付ける。平坦化した同義形では付けない
（= official 自身がネスト形と平坦形で違う答えを出す）。兄弟要素
（`<b class="q"/>`）には付かないので、「全要素に付ける」でもない。

## 影響範囲（実測）

`scripts/dev/partition-known-failures.mjs` を staged binding
(`builtFromCommit: b48a19031`) で全 4 出力 ratchet に流すと、`js-mismatch:scope-class`
= **76 件**（client 15 / client-dev 17 / server 22 / server-dev 22）、
231 エントリ中 **33%**。このラベルは「`svelte-xxxx` トークンを全部消すと一致する」の意で、
ハッシュ違いではなく**付ける場所**の違いを指す。

76 件のうち何件がこの機構かは**未測定**（1 件を縮約して機構を特定しただけ）。
`trakt-web/projects/client/src/lib/components/icons/MediaIcon.svelte` が縮約元。

## なぜ今直さないか

この修正は codegen 出力を変える。フリーズツリー `b48a19031` に対する Corpus Compat run
`33413916328` が全 ratchet の再ベースライン測定になっているので、同じ push に codegen 修正を
入れると測定とツリーが食い違う。**再ベースラインを先に land し、その後で別 push として直す。**

baseballyama:
Fixes #4059

## The rule

Upstream's `get_relative_selectors` (`css-prune.js:171-201`) runs `truncate(node)`
FIRST, then unshifts a `nesting_selector` — with a descendant combinator on
`selectors[0]` — only `if (node.metadata.rule?.metadata.parent_rule && selectors.length > 0)`
and no explicit `&` was written.

That implicit `&` then does **two** things under a fully-global parent, and they
are separate mechanisms:

1. `is_global(selector, rule)` resolves a `NestingSelector` through
   `rule.metadata.parent_rule.prelude` and returns true when some complex
   selector there is entirely global. So `apply_combinator`'s
   `parent_matched || (BACKWARD && (!is_adjacent || parents.length === 0) && every_is_global(...))`
   matches **with no ancestor at all**.
2. `relative_selector_might_apply_to_node` on a bare `&` imposes no requirement,
   so every ancestor that *is* there also matches — and `apply_selector` sets
   `element.metadata.scoped = true` on each one.

rsvelte modelled only (2). That is enough for every shape where the subject sits
under a wrapper, and wrong for every shape where it does not.

## What was wrong

A nested rule under a fully-global parent (`:global(.x) { .a { … } }`) scoped
neither the wrapper above its subject nor — when the subject is at the root of
the template — the subject itself.

## Measurement

Two grids, both against the official compiler at
`submodules/svelte/packages/svelte/src/compiler/index.js` (the entry
`scripts/compat-corpus/oracle.mjs` centralises; the npm build disagrees with it
on output).

**Grid A — 9 parent preludes × 9 child selectors × 2 targets = 162 cells**, every
row carrying a `.p` wrapper element.

| | divergent cells |
|---|---|
| before | 30 |
| after (mechanism 2 only) | 6 |
| after (this PR) | 6 |

Set diff before → after: **repaired 24, REGRESSED 0**.

Those counts are 8 lower than the ones this grid printed while it was being
used, and the 8 were the instrument: it compared official's `e.code` against
rsvelte's `e.message`, so any cell both compilers *reject* could never match.
Measured directly, all four rejecting shapes agree on code **and** message on
both sides. The same 8 cells sit in every arm, so the set diff above is
unaffected; the totals are the corrected ones.

The remaining 6 are three pre-existing classes, each present before this PR and
unmoved by it — 2 cells of #4082 (a trailing `&` under a `:global()` parent is
kept and unwarned), 2 of #4089 (an `&.a` child of a prefixed `:global` block
scopes three elements official leaves bare), and 2 of #4090 (a nested
`:global(...)` inside a `:global` block keeps its wrapper in the emitted CSS).
#4089 and #4090 were filed from this residue.

**Grid B — 12 style shapes × 5 markups = 60 cells**, added because Grid A holds
the wrapper element *fixed* and so cannot reach mechanism 1 at all.

| | divergent cells |
|---|---|
| before this PR | 8 |
| after | 0 |

All 8 were markups with no ancestor for the `&` to match (`root-mid`,
`dyn-mid`). Set diff: repaired 8, **REGRESSED 0**.

Grid A is unchanged by this PR (14 → 14, identical sets), which is the control:
the fix moves exactly the cells the new mechanism reaches.

## Ratchet

`known-failures.{client,server,server-dev,client-dev}.json` each lose the same 5
ids (20 entries total).

## Tests

`crates/rsvelte_core/tests/css_global_parent_implicit_nesting_4059.rs` — every
expectation is the official compiler's output for the same source, measured
before it was written. The `rootless` half is what discriminates the two
mechanisms: it passes with mechanism 2 alone on **none** of its rows, while the
wrapped half passes on all of them.

## Corpus

Full sweep on the rebased tree (`d869bd609`), binary rebuilt from it:
`VERIFY_EXIT=0` — **0 stale entries and 0 new failures**, so the two-sided
ratchet passes in both directions. 34,813 entries × 4 targets; `match` 33,389,
`js-mismatch` 69, `css-mismatch` 1, `output-unparseable` 0.

The 20 removed entries are 5 ids × 4 targets, all js-class — the client
ratchet's single css entry survives, which `report.json`'s `details[].kind`
settles rather than inference. Docs updated: client `52 + 1` → `47 + 1`,
server and server-dev `13 + 2` → `8 + 2`, client-dev `67` → `62`.
`known-failures-md-check.mjs` passes, and because a pass there is spelled as
silence, one addend was perturbed first (`47 + 1` → `46 + 1`) to confirm it goes
red with the predicted message, then restored.


baseballyama:
Rebased onto `7f03d5b68` (#4086 landed after this PR was opened, and it moves the
same two ratchets).

The numbers were **re-derived from the rebased files**, not carried forward
arithmetically:

| ratchet | on `7f03d5b68` | here |
|---|---|---|
| `known-failures.client.json` | 51 | **46** |
| `known-failures.client-dev.json` | 65 | **60** |
| `known-failures.server.json` | 15 | **10** |
| `known-failures.server-dev.json` | 15 | **10** |

Doc partitions resolved to match: client `50 + 1` → `45 + 1`, client-dev `65` →
`60`, server and server-dev `13 + 2` → `8 + 2`. `known-failures-md-check.mjs`
passes; because a pass there is spelled as silence, `60` was perturbed to `59`
first to confirm it goes red with the predicted message, then restored.
`check-pattern-corpus-docs.mjs` passes; the `pattern-corpus/README.md` conflict
with #4086 was resolved by keeping both rows.

The corpus figures in the PR body above (`VERIFY_EXIT=0`, 0 stale / 0 new) were
measured on `d869bd609`. **They do not cover #4086**, which changes client
output — this branch's authoritative verdict is the `Compiler parity` job on
this push.

PR #4096, opened 2026-09-01
fix(compiler): a JSDoc cast leads the parenthesised group, not the field

### LINKED ISSUE #4079: A parenthesized JSDoc type annotation around a rewritten class-field read is placed differently than official
## Summary

A parenthesized JSDoc type annotation around a class-field read — `/** @type {T} */ (this.#x)`, the cast spelling JSDoc-typed code uses everywhere — comes out in a different place than official's. Both outputs parse and mean the same thing, so **only output equality can see this**; the parse gate and any runtime test are blind to it.

Found while measuring #4060 against the real SvelteKit `remote-functions` runtime, which diverges on 3 lines for this reason alone (client and client-dev) and 1 line on server. It is a separate class from #4060 and is **not** a build blocker.

## Reproduction

`m.svelte.js`, compiled with `submodules/svelte/packages/svelte/src/compiler/index.js` (the source entry point, per `OFFICIAL_COMPILER_REL`) and with rsvelte at `9b6d56c07`:

```js
export class C {
	#raw = $state.raw(null);
	#items = $state([]);
	#derived = $derived.by(() => 1);

	read_private() {
		return /** @type {string} */ (this.#raw);
	}

	read_derived() {
		return /** @type {string} */ (this.#derived);
	}

	read_member() {
		return this.#items.reduce((v, r) => r(v), /** @type {string} */ (this.#raw));
	}

	read_plain() {
		const local = 1;
		return /** @type {string} */ (local);
	}
}
```

### client (identical on client-dev) — 3 differing lines

```js
// official
return $.get(/** @type {string} */ this.#raw);
// rsvelte
return (/** @type {string} */ $.get(this.#raw));

// official
return $.get(/** @type {string} */ this.#derived);
// rsvelte
return (/** @type {string} */ $.get(this.#derived));

// official
return $.get(this.#items /** @type {string} */).reduce((v, r) => r(v), $.get(this.#raw));
// rsvelte
return $.get(this.#items).reduce((v, r) => r(v), /** @type {string} */ $.get(this.#raw));
```

### server — 1 differing line

```js
// official
return /** @type {string} */ this.#derived();
// rsvelte
return (/** @type {string} */ this.#derived());
```

## What varies, and what does not

`read_plain` is the control: a `const` local is not rewritten, and both compilers emit `return /** @type {string} */ (local);` byte-for-byte on all three targets. The divergence appears only where the annotated expression is one the transform **rewrites**.

That splits into two distinguishable shapes:

1. **Whose parentheses survive.** Official drops the source parens and prints the comment inside the generated `$.get(...)` call; rsvelte keeps the source parens around the whole rewritten expression and leaves the comment outside. The server row shows the paren half alone — `this.#derived()` is not a `$.get`, and the extra parens still appear — so this is not caused by the read wrap.
2. **Which argument the comment lands on.** In `read_member` official flushes it *inside* the first `$.get(...)`, after its argument; rsvelte moves it onto the *second* argument of `.reduce`. That is a different comment owner, not a spacing difference.

On the server, `read_private` matches (a `$state.raw` field is a plain read there) while `read_derived` does not, so the server row is reached only through the `$derived` call form.

## Verification

Both sides parse under acorn 8.18.0 on all three targets:

```
client     | official: parses | rsvelte: parses
client-dev | official: parses | rsvelte: parses
server     | official: parses | rsvelte: parses
```

## Scale

On the real Kit file (`packages/kit/src/runtime/client/remote-functions/query/instance.svelte.js` at `2eed638`) this is the **only** remaining divergence after #4060: 3 lines on client, 3 on client-dev, 1 on server.

Related: #4048.

baseballyama:
Closes #4079.

## The defect

Upstream wraps the field **node**, so a comment leading it prints inside the generated `$.get(...)`. rsvelte spliced the wrap at the field's own offset, which left the comment outside — and esrap's `ReturnStatement` rule then parenthesised the whole statement, which is where the extra parens in the report come from (they are added by the printer, not kept from the source: `comment-noparen-private`, whose source has none, produced them too).

Two things had to change.

**The wrap opens at the leading comment run.** That alone fixes the un-parenthesised spelling.

**And at the enclosing parenthesised group for the cast spelling.** acorn elides parentheses, so upstream's AST has no node for them and the comment leads the *field*; oxc keeps a `ParenthesizedExpression`, so the comment leads the *group*, one node further out. The wrap therefore has to start at the group — the printer already drops those parens, as `read_plain` shows. This is the same acorn-vs-oxc asymmetry `paren-invisible-comment-bounds.svelte.js` already holds a row for.

A group in **object** position is excluded: in `/** @type {T} */ (this.#derived).toString()` the comment belongs to the deeper chain and official keeps it outside. Widening there is a regression the other rows cannot see — it is what the first version of this patch did, and the grid caught it (2 regressed cells).

## The pass was dead

`wrap_standalone_private_reads` hands the AST pass a bare class **member** (`m() { … }`), and a private field outside a class body is a parse error — so `transform_private_read_wrap_ast`, whose own module doc says it "replaces `class_transforms.rs::wrap_standalone_private_reads`", returned `None` on every call and the text scan did every wrap. None of the above could run before that was fixed. The member is now re-hosted in a synthetic class body declaring the `#name`s it mentions.

Measured on its own before the comment work: **0 repaired, 0 regressed** on the grid below — the AST pass and the text scan agree on every cell of it, so this commit's behaviour change is the comment placement alone.

`private_member_read_wrap_ast.rs` (the sibling pre-pass for `q.foo` / `q[i]`) is called from the same loop with the same text and is dead for the same reason. **Not touched here** — enabling it is a separate blast radius and belongs in its own PR.

## Measurement

Grid: 19 expression shapes × {client, client-dev, server} = 57 cells, expectations taken from `submodules/svelte/packages/svelte/src/compiler/index.js` (`OFFICIAL_COMPILER_REL`). Both arms rebuilt from scratch **after** the rebase onto `6aecb7423`.

```
base 6aecb7423 : 40 equal / 17 differ
this branch    : 52 equal /  5 differ
REPAIRED  12   REGRESSED 0
```

Repaired: `comment-paren-private`, `comment-paren-derived`, `comment-noparen-private`, `comment-paren-arg`, `comment-nested-paren`, `line-comment-paren` — each on client and client-dev.

On the real Kit file the issue cites (`packages/kit/src/runtime/client/remote-functions/query/instance.svelte.js`): **7 differing lines → 3**, 4 repaired, 0 regressed.

### What this does not reach (the 5 remaining cells)

Three shapes, all measured rather than inferred, none of them a regression:

1. **`trailing-comment`** (client, client-dev) — `return (this.#raw /** t */);`. Official prints `$.get(this.#raw /** t */)`; the wrap would have to close *after* a trailing comment, which is esrap's trailing-flush rule rather than a position this splice can carry.
2. **`comment-paren-arg2`** (client, client-dev) — official flushes the comment inside the *first* `$.get(...)` of a `.reduce` chain, a different comment owner. Same trailing-flush rule.
3. **`comment-paren-derived | server`** — upstream replaces the read with an unlocated `b.call`, so its `ReturnStatement` `contains_comment` test (`node.argument.loc &&`) is false and no parens are added. rsvelte's server module path splices text and re-parses, so every node is located and the rule fires. There is no text-level position that avoids it; it needs the module path to build unlocated nodes. The Kit file's own server residue is a fourth shape of the same family (`(/** @type {T} */) => …`, the comment landing inside an arrow's empty parameter list) and is present in **both** arms.

## Gates run

| gate | result |
|---|---|
| `cargo test -p rsvelte_core --lib` + 75 scoped suites (incl. `runtime`, `ssr`, `sourcemaps`, `sourcemaps_gate`, `print`, every `class_*` / `module_*` / `server_*`) | **75 suites ok, 0 failed, 2331 tests** |
| `pnpm run corpus:matrix` | **30,424 comparisons, 0 mismatches** (`js-mismatch` 0, `comment-mismatch` 0, `output-unparseable` 0) |
| `svelte2tsx_fixtures` (`RSVELTE_REQUIRE_PREREQS=1`) | **256/256** |
| `scripts/ci/check-pattern-corpus-docs.mjs` | ✓ |
| `cargo fmt` / `cargo clippy --all-targets --all-features -- -D warnings` | clean |

`svelte2tsx-known-failures.json` and `svelte2tsx-map-known-failures.json` are **not** touched: the change is confined to `3_transform/client`, which svelte2tsx does not run, and its fixture suite is unchanged at 256/256. The collected corpus ratchets were not run locally (they need a full collect) — nothing in this PR edits them.

## Positive controls

Four sites, four ablations, each with an **independently failing** witness in `tests/module_annotated_private_read_4079.rs`:

| ablated site | tests that go red |
|---|---|
| `class_host` (the re-host) | 6 |
| `comment_run_start` (leading run) | 6 |
| `visit_parenthesized_expression` (group widening) | 4 — the two bare/line-comment rows stay green |
| `note_object_paren` (object guard) | **1**, and only that one |

The tree is byte-identical after each (`git status --short` empty). The comment-free controls — `an_unannotated_parenthesised_read_is_wrapped_without_parens` and `an_annotated_local_read_is_untouched` — stay green under all four, which is what says the rows above are about the comment and not about the wrap.

The repro file itself was checked both ways: it diverges on the base binary (client + client-dev, 6 member shapes) and is byte-identical on all three targets on this branch.


baseballyama:
## The `Compiler parity` NEW 4 were the revive, and the two passes cannot be revived separately

Pushed `ae104e64b`, rebased onto `4006f9890`.

### What the 4 were

Three arms, each a rebuilt release binding, run on the two failing corpus sources themselves
(their submodules were not checked out, so the sources were fetched at their pinned SHAs and
TS-stripped with the same esbuild call `compile.mjs` makes):

| arm | `series.svelte.ts` | `LayerManager.svelte.ts` |
|---|---|---|
| before this PR | 45 differing lines / 24 non-comment — pre-existing, already ratcheted | **0** |
| the revive alone, without the comment fix | 46 / 25 | **1** — `this.#activeLayerEventHandlers[type]` |
| this PR as first submitted | 46 / 25 | 1 |
| this PR + the sibling pass revived | **45 / 24** (the pre-PR numbers) | **0** |

The comment fix is not involved: the revive alone reproduces both.

### Mechanism

`class_transforms.rs` handles reads in three steps — the member-chain AST pass, a text
`replace("this.#x.", "$.get(this.#x).")`, then `wrap_standalone_private_reads` (an AST pass with
a text-loop fallback). `this.#x[0]` matches neither of the first two, so **the text fallback in
step 3 was wrapping it**. Reviving step 3's AST pass makes it return `Some`, and that pass skips
a member-chain object on the premise that step 1 owns it — but step 1 is dead for the same
reason step 3 was (a bare class member puts a private field outside a class body, which is a
parse error). So step 3 cannot be revived without step 1.

### Measurement

A generated grid: declaration rune (3) x host (getter / method / method-nested arrow) x
preceding statement (none / a standalone read / a member read / both) x write form (5) x read
shape (16, every position a chain object can take) x target (client / client-dev / server) =
**8,640 cells**, compared after normalizing blank lines and indentation.

| arm | divergences |
|---|---|
| `4006f9890` (this branch's merge base) | 432 |
| this PR as first submitted | 972 |
| this PR + the sibling revive | **72** |

Set differences against the pre-PR arm: **as first submitted, 900 regressed / 360 repaired**;
**with the sibling revive, 0 regressed / 360 repaired**, and every repair is the `jsdoc-cast`
shape this PR is about. The remaining 72 are all `server` and all the server half of #4079
itself (`return /** @type {any} */ this.#x();` against
`return (/** @type {any} */ this.#x());`) — present in the pre-PR arm too, listed under the
`read_annotated_derived` note in the repro.

**900 cells regressed and the corpus reported 4.** That ratio is recorded as blind spot 1h in
`compatibility/GATES.md`: the shape needs a computed read *and* a standalone read of the same
field in the same class, which published code holds rarely, so this fix has to be pinned by a
test rather than by the corpus. `crates/rsvelte_core/tests/private_member_read_object_position.rs`
is that test — 7 cases, one per read shape, each independently failing, with `this.#x.length`
as the control that the `this.#x.` text replace still reaches and that therefore stayed correct
through the regression.

I retract "the revive is behaviour-neutral on the measured population" from #4105. It was
measured on a 57-cell grid with no read shape that puts `this.#x` in the object position of a
member chain; that gap was the whole defect.

Both arms were rebuilt at `4006f9890` after the rebase; the base arm's key set is identical to
the one measured at `2b3a7bdac`, and the fix arm's is identical to the pre-rebase one.

`cargo test --release -p rsvelte_core --lib` plus the 62 private/class/module test binaries:
63 suites, 0 failures. Re-running them on the rebased tree now.

Issue #4135, opened 2026-09-01
fix(client): an `{#await}` arm binding name makes a later unrelated read reactive


baseballyama:
An `{#await}` arm's binding name makes a **later, unrelated** read of that name reactive, so
rsvelte wraps it in `$.template_effect` where the official compiler emits a static
`nodeValue` assignment.

## Reproduction

```svelte
<script>
	let { other } = $props();
	const p = Promise.resolve(1);
</script>

{#await p then code}{code}{/await}
{code}
```

`compile(source, { generate: 'client' })`, official
(`submodules/svelte/packages/svelte/src/compiler/index.js`, 5.56.10):

```js
	var text_1 = $.sibling(node);
	text_1.nodeValue = ` ${code ?? ''}`;
```

rsvelte:

```js
	var text_1 = $.sibling(node);
	$.template_effect(() => $.set_text(text_1, ` ${code ?? ''}`));
```

Every other line of the two outputs is byte-identical, including the expression itself — the
divergence is only whether the read is wrapped.

## The await block is load-bearing

Removing it makes the two agree, so this is not "rsvelte treats an unknown identifier as
reactive":

| input | verdict |
|---|---|
| `{code}` with `let { other } = $props();` and no await block | **match** |
| the same, with `{#await p then code}{code}{/await}` before it | **MISMATCH** |

## Where the decision is made

`phases/3-transform/client/visitors/shared/fragment.js:83` picks between the two forms on one
predicate:

```js
if (has_state && !within_bound_contenteditable) {
	context.state.update.push(update);          // $.template_effect(() => $.set_text(…))
} else {
	context.state.init.push(b.stmt(b.assignment('=', b.member(id, 'nodeValue'), value)));
}
```

and `has_state` comes from phase-2 metadata, not from the phase-3 transform
(`shared/utils.js:142`):

```js
has_state ||= has_await || (node.metadata.expression.has_state && !evaluated.is_known);
```

So the name being bound by an await arm is reaching the **reactivity metadata** of an
expression outside that arm. It is not the phase-3 transform save/restore in
`build_block_with_argument`: this reproduces on `main`, where that function restores
`transform`, `transform_deep_read`, `await_binding_names` and `shadowed_prop_names` for both
arms. The site is unmeasured beyond that.

## Measured population

A 120-cell grid (arm × binding form × what the name already means × target) run against the
official compiler:

| host of the colliding name | diverges |
|---|---|
[...]

Issue #4107, opened 2026-09-01
A `{#await} … then` / `{:catch}` binding does not shadow a prop of the same name


baseballyama:
## Summary

A binding introduced by `{#await … then X}` or `{:catch X}` does **not** shadow a prop
of the same name: every read of `X` inside the block is lowered as the **prop** read
`$$props.X` where the official compiler reads the await block's own binding, `$.get(X)`.

Client and client-dev only; server matches. Every other way of shadowing a prop —
`{#each … as X}`, a keyed each, a `{#snippet}` parameter, `{@const}` — is already
correct, which is what makes this an await-block-specific gap rather than a prop-lowering
one.

## Reproduction

`C.svelte`, compiled with `submodules/svelte/packages/svelte/src/compiler/index.js`
(the source entry point named by `OFFICIAL_COMPILER_REL`):

```svelte
<script>
	let { code, ...rest } = $props();
	const p = Promise.resolve(1);
</script>

{#await p then code}
	{code}
{/await}
```

### client — 1 differing line

```js
// official
$.template_effect(() => $.set_text(text, $.get(code)));
// rsvelte
$.template_effect(() => $.set_text(text, $$props.code));
```

### client-dev — the same line

```js
// official
$.template_effect(() => $.set_text(text, $.get(code)));
// rsvelte
$.template_effect(() => $.set_text(text, $$props.code));
```

### server — 0 differing lines

The prop is read wrongly rather than not at all, so the output parses and runs; it
renders the prop instead of the resolved value. Only output equality can see it.

## Grid: shadow kind × read site

126 cells (7 shadow kinds × 6 read sites × 3 targets), against the submodule oracle.
The four already-correct shadow kinds are the controls and none of them moves:

| shadow kind | non-match / total | failing sites |
|---|---|---|
| `{#await p then code}` | **12 / 18** | expr-tag, attr, comp-prop, spread, member, call-arg |
| `{#await p catch code}` | **12 / 18** | expr-tag, attr, comp-prop, spread, member, call-arg |
| `{#each [1] as code}` | 0 / 18 | — |
| `{#each [1] as code (code)}` | 0 / 18 | — |
| `{#snippet s(code)}` | 0 / 18 | — |
| `{@const code = x}` | 0 / 18 | — |
| no shadow at all | 0 / 18 | — |

All six read sites fail for the two await forms and none fails for anything else, so
the defect is in whether the binding is registered as shadowing, not in any one read
[...]

PR #4139, opened 2026-09-01
fix(compiler): a `{:catch}` binding's transform leaks past its block, as upstream's does (#4111)

### LINKED ISSUE #4111: A `{:catch X}` binding's read transform leaks past its block in the official compiler (rsvelte scopes it)
## Summary

Upstream's `AwaitBlock` visitor clones `state.transform` for the **then** context and **not**
for the **catch** context:

```js
// 3-transform/client/visitors/AwaitBlock.js
const then_context  = { ...context, state: { ...context.state, transform: { ...context.state.transform } } };
…
const catch_context = { ...context, state: { ...context.state } };
```

`create_derived_block_argument` then writes `state.transform[name] = { read: get_value }` into
that object — so a `{:catch X}` binding's read override **mutates the shared transform** and
leaks to every later read of `X` in the component, while the same binding in a `then` block is
correctly scoped.

rsvelte scopes both, so **rsvelte is the more correct side here**; it is filed because byte
equality with the official compiler is the project's goal, so a decision is needed either way.

## Reproduction

```svelte
<script>
	let { code, ...rest } = $props();
	const p = Promise.resolve(1);
</script>

{#await p catch code}{code}{/await}
{code}
```

### client — the second (top-level) read

```js
// official
$.template_effect(() => $.set_text(text, $.get(code)));            // inside the block
$.template_effect(() => $.set_text(text_1, ` ${$.get(code) ?? ''}`)); // AFTER it — leaked
// rsvelte
$.template_effect(() => $.set_text(text, $.get(code)));            // inside the block
$.template_effect(() => $.set_text(text_1, ` ${$$props.code ?? ''}`)); // scoped
```

`$.get(code)` outside the block refers to a binding that does not exist there; the read is the
prop's.

## Grid

| shape | verdict |
|---|---|
| `{#await p then code}{code}{/await}{code}` | match — `then` clones, no leak |
| `{#await p catch code}{code}{/await}{code}` | **MISMATCH** — leaks |
| `{#await p then { code }}{code}{/await}{code}` | match |
| `{#await p catch { code }}{code}{/await}{code}` | **MISMATCH** — leaks |
| `{code}` alone | match |

The `then` rows are the controls, and they are what identify the cause as the missing clone
rather than anything about catch blocks as such: both arms call the same
`create_derived_block_argument`.

## Why it is only visible now

Before #4107 every read of a prop-shadowing name inside an await block was wrong anyway, so
this shape diverged for a different reason and this half was not separable. #4107 fixes the
in-block read; this is what is behind it.

## Decision needed

Either reproduce upstream's leak (byte equality, at the cost of carrying the bug), or leave it
listed as a deliberate divergence with a test pinning rsvelte's behaviour and a report
upstream. I have not made that call here — it is the kind that should not be made silently.

baseballyama:
Closes #4111.

`AwaitBlock.js` gives `then_context` a **copy** of `state.transform` and gives `catch_context`
the parent's own object:

```js
const then_context = {
	...context,
	state: { ...context.state, transform: { ...context.state.transform } }
};
…
const catch_context = { ...context, state: { ...context.state } };
```

so `create_derived_block_argument`'s write survives the catch block and every later read of that
name is rewritten. rsvelte scoped both arms. This makes rsvelte reproduce the leak.

## Why conform rather than stay correct

The leaked read is emitted **outside** the callback that binds the name, so official's own output
throws at render. That is measured, not read off the generated code — the official output was
mounted under jsdom, with three neighbouring inputs as controls so the harness is shown able to
produce a non-exception:

| input | result |
|---|---|
| `{#await p catch code}{code}{/await}` + `{code}` | **`ReferenceError: code is not defined`** |
| `{#await p then code}…` + `{code}` | renders `"OK HELLO"` |
| `{#await p catch err}…` + `{code}` | renders `"Error: boom HELLO"` |
| `{code}` with no block | renders `"HELLO"` |

Byte equality with the official compiler is goal #1/#3 and the documented exception is only for
output no JS parser accepts; this output parses. Deviating would also invert the hazard: code
that renders under rsvelte and throws under the official compiler once it leaves rsvelte. The
population harmed by conforming is already broken under official today; the population harmed by
deviating is people who start on rsvelte, which no collected corpus can measure.

## Measurements

120-cell grid (arm × binding form × what the name already means × target), diffed against the
official compiler:

| tree | divergent cells |
|---|---|
| `main` | 40 |
| #4107 only | 36 |
| this branch | **16** |

`MISMATCH→match: 20`, `match→MISMATCH: 0`.

**Corpus: 0 of 104,439 (entry, target) outputs change.** A zero prints the same as an unmeasured
quantity, so the instrument was controlled two ways: the same sweep reported exactly 4 changed
outputs for #4109 on the same corpus the same day, and the two bindings compared here have
different SHA-256s and demonstrably different output on the reproduction above. Separately, 84 of
34,813 corpus files contain a named `{:catch x}` and **0** of them destructure `x` from
`$props()`.

Eight of the 16 residual cells changed the *shape* of their divergence without changing the
verdict — `const code = 'c'` / legacy `let code = 'l'` colliding with the catch name went from
`text_1.nodeValue = ' c'` to a reactive read. That old output was resolving `{code}` to the outer
binding rather than to the leaked one, so the fold was a symptom of the wrong resolution, not a
lost optimisation; the expression now matches official and only the wrapping differs. A count of
divergent cells cannot see this, since `MISMATCH→MISMATCH` is zero in every total.

## What conforming leaves behind

The 16 residual cells are one shape — `$.template_effect` where official emits a static
`nodeValue` assignment, with the expression already identical — and they are filed as **#4135**.
That is a `has_state` judgement rather than a transform, it reproduces on `main`, and it shares a
decision site with #4126 while having a disjoint input class (#4126's carriers contain no
`{#await}` block and are legacy-mode). The `upstream_issues/` report cross-references it so this
PR is not read as closing the area.

## Test

`crates/rsvelte_core/tests/await_catch_transform_conformance_4111.rs`, 5 tests, with the
"reproducing an upstream defect" note and the `ReferenceError` fact in its header. Ablated:
reverting the one-line change fails exactly the two catch tests, at the predicted assertion lines
with the predicted actual value, while all three controls stay green; restoring gives a
byte-identical patch.

Suites run (release, `--lib` included since a `--test` list skips it): lib 1947, runtime 5, ssr 2,
sourcemaps 2, print 2, compiler_fixtures 3, and the three pin suites — 0 failed.

One process note: mid-run I read six `test result: ok` lines, all green, while `ssr` and
`sourcemaps` had not yet appeared in the `Running` list — the run simply was not finished.
Green result lines and "the population I asked for ran" are separate claims; the numbers above
are from the completed run.

### DIFF
--- a/.changeset/await-catch-transform-conformance.md
+++ b/.changeset/await-catch-transform-conformance.md
@@ -9,3 +9,9 @@ block and every later read of that name is rewritten — including reads of a pr
 or a legacy `export let`. rsvelte scoped both arms and emitted the unrewritten read. The
 divergence and its runtime consequence are recorded in
 `upstream_issues/4111-svelte-await-catch-binding-transform-leaks-out-of-the-block.md`.
+
+Only the read half is conformed. Upstream replaces the whole `transform` entry, so the
+setter is lost too and a later write to the outer binding is emitted as an assignment to a
+call expression — the unparseable class `upstream_unparseable_3306.rs` already pins, where
+this port deliberately diverges. The write halves are therefore restored on the way out of
+the block.
--- a/.changeset/await-catch-transform-conformance.md
+++ b/.changeset/await-catch-transform-conformance.md
@@ -1,0 +1,11 @@
+---
+'@rsvelte/compiler': patch
+---
+
+An `{#await}` catch binding's read transform now leaks past its block, matching the official
+compiler. `AwaitBlock.js` gives `then_context` a copy of `state.transform` and gives
+`catch_context` the parent's own object, so the catch binding's read override outlives the
+block and every later read of that name is rewritten — including reads of a prop, a `$state`
+or a legacy `export let`. rsvelte scoped both arms and emitted the unrewritten read. The
+divergence and its runtime consequence are recorded in
+`upstream_issues/4111-svelte-await-catch-binding-transform-leaks-out-of-the-block.md`.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/await_block.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/await_block.rs
@@ -255,14 +255,14 @@ fn build_block_with_argument(
     context.state.in_control_flow_block = prev_in_control_flow;
     body_statements.extend(fragment_statements);
 
-    // Restore the transform state
-    context.state.transform = saved_transform;
+    // Upstream's `catch_context` spreads `state` without copying `transform`, so a catch
+    // binding's read override outlives its block; reproduced here for byte equality.
+    if block_type != "catch" {
+        context.state.transform = saved_transform;
+        context.state.shadowed_prop_names = saved_shadowed_prop_names;
+    }
     context.state.transform_deep_read = saved_transform_deep_read;
     context.state.await_binding_names = saved_await_binding_names;
-    context.state.shadowed_prop_names = saved_shadowed_prop_names;
-
-    // Log for debugging if needed
-    let _ = block_type;
 
     b::arrow_block(params, body_statements)
 }
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/await_block.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/await_block.rs
@@ -260,6 +260,30 @@ fn build_block_with_argument(
     if block_type != "catch" {
         context.state.transform = saved_transform;
         context.state.shadowed_prop_names = saved_shadowed_prop_names;
+    } else {
+        // The leaked entry is read-only, so a write to the outer binding past the
+        // block loses its setter. Upstream then puts the read on the left of the
+        // assignment and emits output no JS parser accepts (#3306), which is the
+        // one class this port deliberately diverges on — so the write halves are
+        // restored while the read stays leaked.
+        let introduced: Vec<String> = context
+            .state
+            .await_binding_names
+            .keys()
+            .filter(|name| !saved_await_binding_names.contains_key(name.as_str()))
+            .cloned()
+            .collect();
+        for name in introduced {
+            let Some(outer) = saved_transform.get(name.as_str()) else {
+                continue;
+            };
+            let (assign, mutate, update) = (outer.assign, outer.mutate, outer.update);
+            if let Some(current) = context.state.transform.get_mut(name.as_str()) {
+                current.assign = assign;
+                current.mutate = mutate;
+                current.update = update;
+            }
+        }
     }
     context.state.transform_deep_read = saved_transform_deep_read;
     context.state.await_binding_names = saved_await_binding_names;
--- a/crates/rsvelte_core/tests/await_catch_transform_conformance_4111.rs
+++ b/crates/rsvelte_core/tests/await_catch_transform_conformance_4111.rs
@@ -12,6 +12,16 @@
 //! The `then` rows are the control: they must stay scoped, or "we conform" would be
 //! indistinguishable from "we lost the scoping everywhere".
 //!
+//! The leaked entry is READ-ONLY, and that half is not conformed. Upstream replaces the
+//! whole `transform` entry, so a write to the outer binding past the block loses its
+//! setter and upstream puts the read expression on the left of the assignment —
+//! `$.get(v) = "W"`, which acorn rejects. That is the one class this port deliberately
+//! diverges on (`upstream_unparseable_3306.rs`), so the write halves are restored while
+//! the read stays leaked. Measured on the grid this file's issue carries: over the cells
+//! whose tail is a write, the count of cells where the two disagree AND official's output
+//! parses is 0 — so "diverge only where upstream is unparseable" is a measurement here,
+//! not a description.
+//!
 //! Report: `upstream_issues/4111-svelte-await-catch-binding-transform-leaks-out-of-the-block.md`
 
 use rsvelte_core::{CompileOptions, CssMode, GenerateMode, compile};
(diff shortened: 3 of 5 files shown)


baseballyama:
## Follow-up: the first version of this PR was red, and the fix is not a revert

`Test (ubuntu-latest, 1)` failed on an existing pin,
`upstream_unparseable_3306.rs::a_write_past_a_catch_parameter_of_the_same_name_uses_the_setter`.

**Reverting would not have restored it.** Measured on the official compiler
(`submodules/svelte/packages/svelte/src/compiler/index.js`) for that test's own input:

```js
$.delegated('click', button, () => {
	$.get(v) = "W";        // acorn: Assigning to rvalue
});
```

Official's output there is not JavaScript, so that cell was pinning the **deliberate
divergence of #3306**, not upstream's bytes. rsvelte after the first version of this PR
emitted `v = "W"` — a *third* output, neither official's nor #3306's, which overwrites the
`$.mutable_source` object and is silently non-reactive. Worse than the pin it replaced.

### Mechanism

Upstream's `create_derived_block_argument` does `state.transform[name] = { read: get_value }`
— it **replaces the whole entry**, so what leaks past a `{:catch}` block is not only the read
override but the *absence* of `assign` / `mutate` / `update`. Before this PR, leaving the
block restored the outer entry, so a later write picked up the outer binding's setter.

The fix leaks the read and restores the write halves. That combines two decisions this repo
had already made separately, and the join is measured rather than asserted: whenever the
outer entry has an `assign`, upstream puts the read expression on the left of the
assignment, which is exactly the #3306 unparseable class.

### Grid — 360 cells (3 tails × 2 arms × 2 forms × 10 hosts × 3 targets)

The `tail` axis is new: `read` (`{code}`), `write` (`<button onclick={() => { code = 'W' }}>`)
and `read-write`. The 120-cell grid this PR shipped with had only the read tail — it
**reached** the catch-arm leak on every run and could not discriminate the write half of it
(`feedback_generator_reaches_but_cannot_discriminate`).

| arm | non-match |
|---|---|
| `main` (`bddb90da6`) | 116 / 360 |
| this PR, as first pushed | 96 / 360 |
| this PR, fixed | **96 / 360** |

Entry-set diffs, both directions, so a cancellation cannot hide inside a total:

```
main           -> fixed :  match->match 244  match->MISMATCH 0  MISMATCH->match 20  MISMATCH->MISMATCH 96
first-push     -> fixed :  match->match 264  match->MISMATCH 0  MISMATCH->match  0  MISMATCH->MISMATCH 96
```

### Parseability of the oracle's own output

Whole grid, via `scripts/compat-corpus/parseable.mjs` (acorn):

```
official compiled 312 | official-unparseable 64 | official-parseable 248
rsvelte-unparseable 0
```

Counting "write-tail cells that mismatch while official parses" over the **whole** grid gives
**8**, not 0. All eight have host `nothing` (`code` is declared nowhere), and all eight are
**byte-identical to `main`** — they are the pre-existing `then|*|nothing` family, not
something this change created.

Narrowed to the population the claim is about — the cells the write-half restore actually
moved:

```
cells whose output the restore changed         : 56
  of those, still MISMATCH after the fix       : 56
  of those, official PARSES  (falsifiers)      :  0
  of those, official is UNPARSEABLE            : 56
```

**56 / 56 unparseable, 0 falsifiers.** The 56 staying MISMATCH is by design and is not a
backlog: official emits text no JS parser accepts there, so byte equality is unobtainable
by construction. Read the 8 the same way — reporting "0" without the denominator would have
been weaker evidence, not stronger.

### New information the axis produced

`write|catch|ident|nothing|server` and `write|catch|destr|nothing|server` diverge and were
**not reachable from the read-only grid**. They join the `then|*|nothing` cells already
slated for a separate issue.

### Ablation control

The ablation is not a hand-made mutant: this change was uncommitted while it was measured, so
`git checkout HEAD -- await_block.rs` reproduced the tree that CI actually failed on. Run with
`--no-fail-fast`, because cargo stops at the first failing test binary and a check that never
ran is indistinguishable from one that passed:

```
await_catch_transform_conformance_4111:
  a_write_past_a_catch_block_keeps_the_outer_setter               ... FAILED
  (5 others ok)
upstream_unparseable_3306:
  a_write_past_a_catch_parameter_of_the_same_name_uses_the_setter ... FAILED
  (3 others ok)
```

Exactly 2 of 10, and the 2 predicted. The first attempt did not run the second binary; running
it changed nothing, and that could not be known without running it.

### Corpus

**0 of 104,439 (entry, target) outputs change**, measured with **both arms built from
`bddb90da6`** — the base this branch had before it was rebased onto `27328152d`. The delta is
taken at a common base, so it is valid for this change; it is not a measurement at the current
base, and re-measuring there is in progress.

The instrument is controlled: the same sweep returned **4** changed outputs for an arm pair
that differed by more than this change, so a 0 here is not a silent instrument.

Those 4 are worth recording, because they nearly went into this PR as a result. They looked
like a win — two files moving to byte-equality with official, in the expected direction. They
were **#4096's**, not this change's: the two arms had been built from different merge bases,
and all four files are `.svelte.js`, which `compileModule` compiles and a template visitor
cannot reach at all. A direction that matches the hypothesis is the cheapest thing for an
artefact to imitate.

### Binding identity

`buildInfo()` is not evidence of which tree a `.node` was built from, and neither is its
filename — one of the bindings in this session was stored as `main.node` and turned out to
carry a different branch's fix. Every arm here was checked with two discriminating probes
instead:

```
#3306 catch-write : $.set(v, "W");                                  <- pin restored
#4111 leak        : $.set_text(text_1, ` ${$.get(code) ?? ''}`)     <- conformance kept
```

Both are needed: either one alone is satisfied by getting the other wrong, which is how the
first version shipped. The leak reading identical on **both** arms is the load-bearing half —
that is what shows the pair differs by this change rather than by a whole branch.

A probe discriminates only the hypothesis it was given. One of these bindings had already been
probed clean for an unrelated contamination and was *still* mislabeled; the mislabel surfaced
only from a second probe aimed at a different fix.

baseballyama:
## Rebased onto `27328152d` (#4118), and the grid was re-measured there rather than assumed

The measurements above were taken with both arms at `bddb90da6`. Re-running the 360-cell grid
with a binding built from the rebased branch:

```
verdict movement, bddb90da6 -> 27328152d (both arms carrying this fix):
  match->match 264   match->MISMATCH 0   MISMATCH->match 0   MISMATCH->MISMATCH 96
cells whose rsvelte OUTPUT changed across the rebase: 0
```

Not one cell of this grid's population moved, so the grid result carries to the new base as a
measurement rather than as an argument about whether a CSS-scoping change can reach an
await-block transform.

**The corpus figure is still the one measured at `bddb90da6`**, and is labelled as such above.
Re-measuring it at the new base needs a second binding and two full sweeps; it is queued, not
done. What is known is that the arms it compares differed only by this change, so the 0 is a
valid delta for this change at that base.

Binding identity for the rebased arm, three probes rather than a filename:

```
#3306 pin   (must be present) : $.set(v, "W");
#4111 leak  (must be present) : $.set_text(text_1, ` ${$.get(code) ?? ''}`)
#4118 base  (must be present) : <span class="x svelte-1lj1c2j">   <- the new base's own fix
```

The third is what distinguishes "rebased" from "rebuilt at the old base", which neither of the
first two can do.

PR #4140, opened 2026-09-01
fix(server): a self-named `class:` value reaches `$.attributes` untransformed (#4117)

### LINKED ISSUE #4117: fix(server): match upstream on a `class:` shorthand of a $derived beside a spread attribute
`class:name` の**短縮形**の値が `$derived` で、**同じ要素に spread 属性がある**とき、
上流の **server** 出力は `$.attributes` の class オブジェクトに **derived 関数そのもの**を渡します。
関数は truthy なので、**SSR ではそのクラスが常に付きます**。

rsvelte は `active()` と呼び出しており、**rsvelte のほうが正しい**側です。
上流自身の **client** は `$.get(active)` を出しており、rsvelte と一致します。

## 入力

```svelte
<script>
	let { to, icon } = $props();
	let active = $derived(to === "x");
</script>
<a href={to} class:nav-item--icon={icon} class="nav-item" class:active {...icon ? { target: "_blank" } : {}}>t</a>
<style>
	.active { color: red }
	.nav-item { color: blue }
</style>
```

## 両コンパイラの実出力（server）

official（`submodules/svelte/packages/svelte/src/compiler/index.js`）:

```js
let active = $.derived(() => to === "x");

$$renderer.push(`<a${$.attributes(
    { href: to, class: 'nav-item', ...icon ? { target: "_blank" } : {} },
    'svelte-1lj1c24',
    { 'nav-item--icon': icon, active }
)}>t</a>`);
```

rsvelte:

```js
let active = $.derived(() => to === "x");

$$renderer.push(`<a${$.attributes(
    { href: to, class: 'nav-item', ...icon ? { target: "_blank" } : {} },
    'svelte-1lj1c24',
    { 'nav-item--icon': icon, active: active() }
)}>t</a>`);
```

`active` は `$.derived(…)` の**戻り値（関数）**なので、official の `{ …, active }` は
`active: <function>` です。`$.attributes` の class マップは truthy 判定なので、
`to === "x"` が false でも `class="… active"` が出ます。

## client は両者一致（= 上流の 2 港のうち server だけが違う）

official / rsvelte とも:

```js
[$.CLASS]: { 'nav-item--icon': $$props.icon, active: $.get(active) }
```

`dev: false` / `dev: true` の両方で **MATCH**。

## 切り分け（4 セル、server）

| 入力 | 結果 |
|---|---|
| `class:active`（短縮） + `$derived` + **spread 属性あり** | **DIVERGE** |
| `class:active`（短縮） + `$derived` + spread なし | MATCH |
| `class:active` + **plain `let`** + spread なし | MATCH |
| `class:active={active}`（明示） + `$derived` | MATCH |

**spread 属性の有無が効く**のは、spread があると上流が要素を `$.attributes(…)` の
1 回の呼び出しにまとめる経路に入り、そこで短縮形のプロパティが `context.visit` を通らずに
そのまま出るためと思われます（**未確認**）。

## この repo での扱い

`CLAUDE.md` が #2990 で記録している方針に従い、**byte equality が目標なので上流に合わせます**
（プロジェクト目標 #1「100% 互換」と #3「drop-in replacement」は「rsvelte のほうが正しい」を
基準にしません）。したがって:

1. rsvelte を上流に合わせて `active`（短縮）を出す
2. その挙動を**テストで pin** し、コメント 1 行で「上流の欠陥を再現している。合わせているのは
   byte equality が目標だから」と書く
3. `upstream_issues/` に本文書を置く
4. `deliberate-divergences.md` には**入れない**（deliberate conformance であって divergence ではない）

## コーパスでの帰属

`known-failures.{server,server-dev}.json` の
`sveltepress/packages/theme-default/src/components/NavItem.svelte` 1 件。

## 計測の条件

`.corpus-cache/rsvelte.node` の provenance は `builtFromCommit: b48a1903 / stagedAt 2026-08-31` ですが
mtime は 2026-09-01 10:20 UTC で stage を経ずに上書きされています。**`main` のバイナリではありません。**
上記エントリは `main` の ratchet に載っており整合しますが、着手時に取り直してください。

baseballyama:
Closes #4117.

Upstream's `prepare_element_spread` passes a `class:` directive whose value is the identifier of
the same name straight through, with no read transform:

```js
b.init(
	directive.name,
	directive.expression.type === 'Identifier' && directive.expression.name === directive.name
		? b.id(directive.name)
		: transform(context.visit(directive.expression), directive.metadata.expression)
)
```

A `$derived` therefore reaches `$.attributes` as the derived **function** — always truthy — so
SSR renders the class unconditionally. The same compiler's client emits `$.get(active)` and is
right; only the server port diverges. This makes rsvelte reproduce the server output.

## The issue's isolation table held one axis fixed, and the grid found the cell

The issue lists `class:active={active}` (explicit) as MATCH. That row has **no spread**. Crossing
the two axes properly shows the explicit form diverges too once a spread is present:

```
### explicit|derived|spread|server
   official: $.attributes({ ...rest }, 'svelte-1lj1c2j', { active })
   rsvelte : $.attributes({ ...rest }, 'svelte-1lj1c2j', { active: active() })
```

which follows from the mechanism: `class:active={active}` satisfies `expression.type ===
'Identifier' && expression.name === directive.name` exactly as the shorthand does. **A fix keyed
on "was this written shorthand" would have left the explicit form wrong**, and every test written
from the issue's table would have passed. The condition is ported verbatim instead.

## Measurements

168-cell grid (directive form × what the value is × spread present × target) against the official
compiler: **4 divergent → 0**, `match→MISMATCH: 0`, `still-bad: 0`. The four are `$derived` with a
spread, on `server` and `server-dev`, in both the shorthand and the explicit self-named form. The
before-arm binding was built from **this branch's own base** with the patch reverted — reusing a
binding from another branch would not be a control.

**Corpus: exactly 1 of 104,439 (entry, target) outputs changes** —
`sveltepress/.../NavItem.svelte` on the server — and it moves **mismatch → MATCH** under the
gate's own normalization (`flattenTemplateHoles` → oxfmt with `compatibility/.oxfmtrc.json` →
`stripBlankLines`):

```
BEFORE: 1 differing line
   official: …, "svelte-cw6fbm", { "nav-item--icon": icon, active, brand })}>`,
   rsvelte : …, "svelte-cw6fbm", { "nav-item--icon": icon, active: active(), brand })}>`,
AFTER:  MATCH
```

Since that is the only output that changed anywhere in the corpus, `match→mismatch = 0` for every
other entry follows from byte-identity rather than from a second comparison.

## Ratchet

That file is listed on **both** `server` and `server-dev`, and both go to a byte match, so both
entries are retired here (the ratchet is two-sided). `server` 10 → 9, `server-dev` 10 → 9, with
the headings, partitions (`8 + 2` → `7 + 2`) and cluster bullets updated to match.

The sweep above covers `client` / `client-dev` / `server` only, so **`server-dev` for this file was
measured separately** rather than inferred from the `server` result — both arms, both dev settings,
same normalization.

That distinction is the point rather than a detail: retiring a two-sided ratchet entry on the
strength of a sweep that never ran that target would enrol a claim the measurement does not
contain, and it would read exactly like a measured one. It is the same shape as a `--test` list
silently omitting `--lib` — the population that answered is not the population that was asked
about, and nothing in the output says so.

Counts were re-derived from the post-rebase files, not by subtracting from the pre-#4109 numbers.
`known-failures-md-check.mjs` passes, and its positive control fires: writing `6 + 2` gets
"sums to 8 … but that population has 9 entries", and restoring returns the tree byte-identical.

## Two ports, and the one that must not change

rsvelte has **two** ports of `prepare_element_spread` and neither carried the rule; both now call
one helper. The third site, `build_attr_class` — the path an element without a spread takes — has
no such arm upstream and still transforms; widening the helper to it would turn the passing
no-spread cells into divergences. The helper's comment says so, and a test pins it.

## Test

`crates/rsvelte_core/tests/class_directive_shorthand_spread_4117.rs`, 4 tests. All four expected
values were measured against the official compiler before being written down, including the two
controls (`class:on={active}` + spread → `{ on: active() }`; shorthand without a spread →
`$.attr_class(…, { 'active': active() })`).

Ablated: reverting the change fails exactly the two conformance tests, at the predicted assertion
lines with the predicted actual value (`{ active: active() }`), while both controls stay green.

Suites (release, `--lib` included): lib 1951, runtime 5, ssr 2, compiler_fixtures 3, pin 4 — 0
failed, all 5 targets confirmed present in the `Running` list.


baseballyama:
`main` に #4118 が入りました (`27328152d`) — `known-failures.server.json` / `.server-dev.json` / `KNOWN-FAILURES.md` を動かしています。この PR も同じ 3 ファイルを触るので、**マージ前に `main` へリベースしてください**（リベースが先、再ベースラインが後 — #2435）。force-push 後は `gh pr checks` ではなく SHA で確認してください: `gh api repos/baseballyama/rsvelte/commits/<新SHA>/check-runs --jq '.check_runs|length'`。

baseballyama:
@baseballyama これは私（Agent-A / rsvelte-75）ではなく rsvelte-72 の所有です。リベースはブランチ所有者にお願いします — 私が force-push すると、同じブランチでビルド中の測定アームを壊します。

Issue #3026, opened 2026-08-17
codegen (client): assignment inside an inline template arrow double-invokes the prop getter → `TypeError: x(...) is not a function`


MathiasWP:
### Summary

When an arrow function written **in the template** assigns to a member of a destructured prop, every read of that prop on the right-hand side is emitted as `prop()()` instead of `prop()`. Since `prop()` returns the prop value, calling it again throws at runtime.

Versions: `@rsvelte/vite-plugin-svelte@0.5.2`, `@rsvelte/vite-plugin-svelte-native@0.3.7`, svelte 5.56.8, vite 8.2.0.

### Repro

```svelte
<script>
	const { state } = $props();
</script>

<button onclick={() => { state.a = state.b; }}>go</button>
```

`compile(source, { generate: 'client', dev: false })`:

```js
// upstream svelte 5.56.8
state().a = state().b;

// @rsvelte/vite-plugin-svelte-native 0.3.7
state().a = state()().b;   // ← TypeError: state(...) is not a function
```

### Scope

| variant | result |
| --- | --- |
| inline arrow in template, assignment | **wrong** |
| inline arrow in template, read-only (`console.log(state.b)`) | correct |
| named function declared in `<script>`, referenced as `onclick={f}` | correct |
| on a plain element vs on a component | both wrong |
| inside a `{#snippet}` | wrong |
| `state.a = state.b + state.c + state.d` | all three reads become `state()()` |
| `state.a = state` (no member on RHS) | `state()()` |

So the trigger is: *an assignment statement, inside an arrow function expression in the template, whose LHS is a member expression on a destructured prop*. The RHS reads of that same prop are then rewritten twice.

### Runtime confirmation

```ts
// vitest + @testing-library/svelte
const state = { a: '', b: 'HELLO' };
render(Probe, { state });
await fireEvent.click(screen.getByRole('button'));
// TypeError: state(...) is not a function
```

Passes on upstream, fails on rsvelte.

### Why it matters

The production build is green — the output is syntactically valid — and our 9897-test suite does not cover this particular handler, so nothing catches it before the page throws in the browser. In our codebase it hits one real file (a form's `oninput` handler that derives one field from another), which crashes as soon as the user types.

PR #4150, opened 2026-09-01
fix(compiler): a `bind:this` collects each-block ids by declaration scope, not by loop-variable name (#4121)

### LINKED ISSUE #4121: client: an each-block item is read through its signal instead of being passed as a callback parameter
`known-failures.client.json` の **4 件**が同じ形です。

計測条件（全 issue 共通）:

- バインディング `target/release/librsvelte_napi.dylib`、`buildInfo()` = `{"commit":"b308730e520d99c3e53126123984a1885f9b7fcd","dirty":false}`（現在の `main`、クリーン）
- オラクルは `submodules/svelte/packages/svelte/src/compiler/index.js`（`OFFICIAL_COMPILER_REL`。npm ビルドは出力が違うので使っていません）
- オプションは `scripts/compat-corpus/compile.mjs` と同一
- **正規化はゲートと同じ**（`flattenTemplateHoles` → `oxfmt` → `stripBlankLines`）。折り返しだけの差はここで消えています
- 出所: `compatibility/known-failures.client.json` 45 件の機構分割（#4048 のコメント）。分割は網羅・排他を検算済み

### `ha-fusion/src/lib/Main/Views.svelte`
`client-dev` にも載っている: **はい**

正規化後 174 行目から:

```js
// official
($$value, view) => $.mutate(buttons, ($.get(buttons)[view.id || 0] = $$value)),
(view) => $.get(buttons)?.[view.id || 0],
() => [$.get(view)],
);
$.template_effect(
($0) => {
// rsvelte
($$value) => $.mutate(buttons, ($.get(buttons)[$.get(view).id || 0] = $$value)),
() => $.get(buttons)?.[$.get(view).id || 0],
);
$.template_effect(
($0) => {
$.set_attribute(button, "id", $0);
```

### `ha-fusion/src/lib/Modal/VisibilityConfig/Index.svelte`
`client-dev` にも載っている: **はい**

正規化後 350 行目から:

```js
// official
() => ($.get(items)[$$index_1], $.get(items)),
));
},
get items() {
return $.get(items);
},
// rsvelte
() => ($.get(item), $.get(items)),
));
},
get items() {
return $.get(items);
},
```

### `svelte-bits/src/lib/components/library/TextAnimations/RotatingText/RotatingText.svelte`
`client-dev` にも載っている: **はい**

正規化後 226 行目から:

```js
// official
($$value, previousCharsCount, charIndex) =>
(elementsByKey[`${$.get(displayedIndex)}:${previousCharsCount + charIndex}`] =
$$value),
(previousCharsCount, charIndex) =>
elementsByKey?.[`${$.get(displayedIndex)}:${previousCharsCount + charIndex}`],
() => [$.get(previousCharsCount), charIndex],
// rsvelte
($$value) =>
(elementsByKey[`${$.get(displayedIndex)}:${$.get(previousCharsCount) + charIndex}`] =
$$value),
() =>
elementsByKey?.[`${$.get(displayedIndex)}:${$.get(previousCharsCount) + charIndex}`],
);
```

### `svelte-bits/src/lib/components/library/TextAnimations/VariableProximity/VariableProximity.svelte`
`client-dev` にも載っている: **はい**

正規化後 147 行目から:

```js
// official
($$value, idx) => (letterEls[idx] = $$value),
(idx) => letterEls?.[idx],
() => [$.get(idx)],
);
$.template_effect(() => {
styles = $.set_style(span_2, "", styles, {
// rsvelte
($$value) => (letterEls[$.get(idx)] = $$value),
() => letterEls?.[$.get(idx)],
);
$.template_effect(() => {
styles = $.set_style(span_2, "", styles, {
"font-variation-settings": fromFontVariationSettings(),
```

## 修正時に必要なこと

- 直したら `compatibility/known-failures.client.json` と `known-failures.client-dev.json` の
  **エントリ集合を差分**してください。総数の減少だけでは、直した件数と壊した件数の相殺が見えません。
  `match -> mismatch = 0` を明示してください。
- pin テストを `crates/rsvelte_core/tests/` に。**このクラスタの全ファイルが同じ形なら、
  1 本のテストで全部を主張できているか**を確認してください（1 セル 1 ファイルのグリッドは規則間干渉を見られません）。
- 再現は `compatibility/sources/<id>` があれば `expected/` `actual/` ツリー無しで取れます。

baseballyama:
Closes #4121.

`bind:this` decides which identifiers become callback parameters. Upstream's
`build_bind_this` (`3-transform/client/visitors/shared/utils.js`) compares each reference's
binding scope against the scope of every `EachBlock` owner —
`owner.type === 'EachBlock' && scope === binding.scope`. rsvelte compared the loop
variable's **name** (`each_ctx.item_name` / `index_name` / `destructured_update_paths`), so
a `{@const}` declared inside an each block never qualified however it was written.

The fix ports the scope test, and upstream's two exclusions travel with it
(`is_state_source(binding, analysis)` and `binding.kind === 'derived'`) — the scope test
alone over-collects declaration tags.

## Two more divergences were in the same walk, and neither was the reported one

**The walk had no arm for most of `JsExpr`.** `collect_each_block_ids` was a hand-written
match with `_ => {}`, so a reference inside a `||`, a template literal, an object, a `new`,
an optional chain, a tagged template or a unary/update operator was not *looked at* — a
silent miss that reads exactly like "this shape does not qualify". Eleven arms were added.

**Upstream marks a name seen before asking whether the occurrence is a reference.**

```js
if (seen.includes(node.name)) return;
seen.push(node.name);            // <- before the is_reference test
const parent = path.at(-1);
if (!is_reference(node, parent)) return;
```

So an identifier in a **non-reference** position burns the name for every later one:
`els[{ k: k }.k]` collects nothing, while `els[{ kk: k }.kk]` collects `k`. rsvelte inserted
into `seen` only on the success path and never visited non-reference positions at all, so a
faithful port needed both the insert moved to the top and a visit (collect nothing, burn the
name) of a non-computed member property and a non-computed property key.

## Method: reading rsvelte gave a confident wrong cause

The grid was **68/68 green** while three corpus carriers stayed broken, because every one of
its nine declaration rows held the subject expression constant at `els[<name>]`. Declaration
position and subject shape are independent axes and only one was varied — the same shape as
`binding-position` / `write-host` (#3026), reappearing in a different family.

Adding the subject axis turned the object cells red in the **opposite direction** (rsvelte
collecting where official does not), which reading rsvelte explains as "the `JsExpr::Object`
arm is wrong". It is not. Probing **official** with six subject shapes side by side is what
named the axis:

```
shorthand  els[{ k }.k]      -> ($$value) => els[({ k: $.get(k) }).k] = $$value        (not collected)
explicit   els[{ k: k }.k]   -> ($$value) => els[({ k: $.get(k) }).k] = $$value        (not collected)
renamed    els[{ kk: k }.kk] -> ($$value, k) => els[({ kk: k }).kk] = $$value          (collected)
array      els[[k][0]]       -> ($$value, k) => ...                                    (collected)
template   els[`${k}`]       -> ($$value, k) => ...                                    (collected)
logical    els[k || 0]       -> ($$value, k) => ...                                    (collected)
```

The discriminator is not "is it an object", it is "is the key spelled the same as the name".
Nothing on the rsvelte side could have said that.

## Measurement

Grid: declaration position (9) x use site (2) x subject shape (10) x mode (2) x target (2),
before and after on the same base, same grid, cell sets diffed:

```
cells               680
changed (rsv text)  256
toward-official     256
away-from-official  0     <- match -> MISMATCH
still MISMATCH      0
unchanged match     424
```

Denominator: 68 of the 680 cells are the `els[await n]` row, which **both** compilers reject
with `experimental_async` — trivial matches that measure nothing about the collector, so the
informative population is 612. Cells where exactly one side rejects: **0**, so there is no
over- or under-acceptance in either direction. The `els[n!]` row needs `lang="ts"` and
therefore crosses the entry point; those 68 cells compile on both sides and are compared.

Corpus: every `compatibility/manifest.json` entry x {client, client-dev, server}, output
hashed, before and after arms run back to back against the same `sources/` tree.

```
entries hashed: 34813, sources missing: 0, keys: 104439   (both arms)
changed 10 keys / 5 distinct ids | client 5, client-dev 5, server 0
changed-and-listed: client 5/5, client-dev 5/5, server 0
```

`changed-and-listed == changed`, so **no passing entry changed** — the corpus-level
`match -> MISMATCH = 0`. `server` is 0, which is what a client visitor should do.

A changed hash is not a fixed file, so the five were then compared to official through the
gate's own normalization (`flattenTemplateHoles` -> `oxfmtTree(.oxfmtrc.json)` ->
`stripBlankLines`), on both arms:

| | before | after |
|---|---|---|
| pairs | 10 | 10 |
| match | **0** | **6** |
| mismatch | 10 | 4 |

So six ratchet entries retire (3 ids x {client, client-dev}), and the ratchets go
`client` 42 -> 39 and `client-dev` 56 -> 53. The other two ids
(`ha-fusion/.../Sidebar/Navigate.svelte`, `kite-public/.../CategoryNavigation.svelte`)
**changed output and stay listed**: their first divergence is comment placement and is
byte-identical before and after, so this change reached them without deciding them.
"the set of ids whose output moved" and "the set of ids whose verdict moved" are 5 and 3,
and reporting the first as the second is how an earlier draft of this measurement was
wrong.

The sweep's two arms share a base, so their delta is attributable regardless of what has
since landed on `main`. "These three now match official" does not — it is a claim about a
tree. #4139 merged while this was measured, so after taking `main` the binding was rebuilt
and the ten pairs re-measured: **6 match / 4 mismatch, the same three ids**. That the arm
was really the merged tree was checked with a discriminating probe rather than the file
name — the merged binding emits `$.template_effect(() => $.set_text(text_1, ` ${$.get(code) ?? ''}`))`
for a read past a `{:catch}`, where the pre-merge one folds it to `' 1'`.

## Repro and pin

`compatibility/pattern-corpus/issues/4121-bind-this-each-scope-const.svelte` carries ten
cells in one component (a grid of one component per cell cannot see a rule that depends on
sibling blocks). `crates/rsvelte_core/tests/bind_this_each_scope_4121.rs` pins all ten
against expectations measured from the official compiler first, plus three controls stated
separately so a failure names the direction — collecting every identifier satisfies the first
case and breaks the controls; collecting none does the reverse.


baseballyama:
Integrator note (Agent-A). `main` moved to `383cc4433` (#4170). This branch's `Compiler parity` failure is the two-sided ratchet doing its job, not a regression: `ha-fusion/src/lib/Sidebar/Navigate.svelte` and `kite-public/src/lib/components/CategoryNavigation.svelte` now pass on **client** and **client-dev** because of this fix, so they have to leave the ratchet in this PR. `main`'s own Corpus Compat was green at `fd72d98f1`, which is what rules out "they already passed before".

Merge order I am holding, so the `KNOWN-FAILURES.md` count lines are only redone once each:

1. #4157 — client 25 → 23, client-dev 39 → 37 (pending)
2. this PR — five ids per target on top of that
3. #4175 — one more id per target
4. #4165 (CSS, currently CONFLICTING)

When you rebase, take `main`'s count line and subtract your own ids; do not carry the number from the old base. And please list the five retired ids **with the direction** in the PR body (`MISMATCH -> match`), not just the delta — n fixed and n broken is the same delta as nothing.

baseballyama:
Integrator (Agent-A) took this over and force-pushed a rebase onto `4cdc135cb`. The original worktree had not moved in ten hours and CI was red, so the branch was rebuilt rather than left waiting.

**What changed and why**

- The three commits are squashed into one on top of `main`. `each_block.rs` had moved on `main` since this branch's merge base, so it was re-applied as a three-way patch instead of being checked out wholesale — a blind checkout would have reverted `main`'s change with no conflict to warn about.
- The ratchet edits were **recomputed from `main`'s current lists**, not carried over: client 23 → 18, client-dev 37 → 32, the same five ids on both targets.
- `KNOWN-FAILURES.md`'s count and partition lines were redone against those numbers. The branch's prose (including the reconstruction-is-stricter-than-the-gate paragraph) is kept verbatim; only the arithmetic moved.

**The five ids retired, with direction:**

| id | client | client-dev |
|---|---|---|
| `ha-fusion/src/lib/Main/Views.svelte` | `MISMATCH -> match` | `MISMATCH -> match` |
| `ha-fusion/src/lib/Sidebar/Navigate.svelte` | `MISMATCH -> match` | `MISMATCH -> match` |
| `kite-public/src/lib/components/CategoryNavigation.svelte` | `MISMATCH -> match` | `MISMATCH -> match` |
| `svelte-bits/…/RotatingText/RotatingText.svelte` | `MISMATCH -> match` | `MISMATCH -> match` |
| `svelte-bits/…/VariableProximity/VariableProximity.svelte` | `MISMATCH -> match` | `MISMATCH -> match` |

`match -> MISMATCH`: **0**. The last two are the pair the previous run reported as `4 baseline entries already PASS` — they pass because of this fix, and `main`'s own Corpus Compat was green at `fd72d98f1`, which is what rules out "they already passed before it".

`cargo test -p rsvelte_core --test bind_this_each_scope_4121` — 4 passed, 0 failed, on the rebased tree.

Issue #4142, opened 2026-09-01
compiler: a `const` in a `try`/`finally` block inside a template expression does not shadow an outer reactive `let`


baseballyama:
Found while attributing the nine `known-failures.client{,-dev}.json` entries that were grouped as `comment-attachment` (they are not comment divergences — see #4048). Corpus entry: `open-webui/src/lib/components/admin/Settings/Models.svelte`, listed on `client` and `client-dev`.

## What happens

Inside a **template expression**, a `const` declared in a `try` block's body (or a `finally` block's body) does not shadow an outer reactive `let`. Every read of that name is lowered as a signal read, so the generated code reads the component's state where the source reads the local.

```svelte
<script>
	let models = null;
	function reload() { models = []; }
	function send(m) { return m; }
</script>
<button onclick={() => {
	try {
		const models = [1];
		send(models);
	} catch (e) {}
}}>{models}{reload}</button>
```

| | |
|---|---|
| official | `send(models);` |
| rsvelte | `send($.get(models));` |

The output parses and runs, and passes the wrong value — this is a silent wrong-value defect, not a crash.

## The grid — `try`/`finally` against every neighbouring block

Same header in every cell; the only variable is which block the shadowing `const` sits in. `generate: 'client'`, `dev: false`.

| cell | where the shadowing `const` sits | verdict |
|---|---|---|
| a | `try { … }` in a template arrow | **NE** |
| e | `finally { … }` in a template arrow | **NE** |
| b | a bare `{ … }` block in a template arrow | EQ |
| c | an `if (true) { … }` consequent in a template arrow | EQ |
| d | a `catch (models)` **parameter** | EQ |
| g | no shadow at all — the outer `models` read inside `try` | EQ (control) |
| f | the same `try` shadow inside a `<script>` **function** | EQ (control) |

Two controls matter: **f** says the script-side scope walk handles `TryStatement` correctly, so this is the template-expression walk alone; and **d** says the `catch` clause's own parameter binding is handled, so it is the `block` / `finalizer` children specifically.

## Reduction history, because it is evidence about the probe

[...]

PR #4152, opened 2026-09-01
fix(client): a style-directive chunk and a `try` body both lose their scope (#4120, #4142)

baseballyama:
Two client-side defects, each with its own commit and its own pin test. They ship
together because the ratchet is two-sided: an entry that starts passing fails CI
unless the same PR re-baselines it.

## 1. A `style:` directive's chunk metadata was written to the directive (#4120)

Upstream's `StyleDirective` visitor calls `context.next()`, and `ExpressionTag`
replaces `state.expression` with the **tag's own** metadata before walking; the
directive then merges each chunk up (`node.metadata.expression.merge(chunk…)`).
rsvelte wrote straight into the directive's metadata, so the chunk stayed empty —
and phase 3's `build_attribute_value` reads **the chunk**. `has_call` was
therefore always false there, `build_expression` returned early, and the whole
legacy `($.deep_read_state(dep), $.untrack(() => value))` sequence was dropped.

| | |
|---|---|
| official | `background: ($.deep_read_state(DEV), $.untrack(() => DEV && show ? rnd() : null))` |
| rsvelte | `background: DEV && show ? rnd() : null` |

**Why only a call diverged.** `has_member_expression` and `has_assignment` are
re-derived from the expression in phase 3, so `style:x={obj.y}` was correct all
along and hid the shape. An 8-cell grid isolated the axis to *a call in a style
directive*; `a_member_in_a_style_directive_is_unchanged` keeps that cell as a
control.

## 2. A `try` / `finally` body did not open a child scope (#4142)

`apply_transforms_to_statement_with_shadowed`'s `Try` arm walked the `block` and
`finalizer` children with the **outer** scope. Only the `catch` clause built a
child scope, and only for its parameter — every other block form
(`JsStatement::Block`, an `if` consequent, a loop body) calls
`register_block_local_vars` first, which is why those cells were already correct.

| | |
|---|---|
| official | `send(models);` |
| rsvelte | `send($.get(models));` |

The output parses and runs and passes the wrong value: a silent wrong-value
defect, not a crash.

## Pin tests, and the control that makes them pins

`crates/rsvelte_core/tests/style_directive_legacy_untrack.rs` (5) and
`crates/rsvelte_core/tests/try_block_shadowing_scopes.rs` (5). Every expectation
was taken from `submodules/svelte/packages/svelte/src/compiler/index.js`, not from
rsvelte's own output.

Green on its own proves nothing, so both suites were re-run against the tree with
the fix reverted:

```
with the fix      5 passed
without the fix   a_const_in_a_try_body_shadows_an_outer_reactive_let      FAILED
                  a_const_in_a_finally_body_shadows_an_outer_reactive_let  FAILED
                  a_catch_parameter_still_shadows                          ok
                  a_bare_block_still_shadows                               ok
                  an_unshadowed_read_in_a_try_body_still_reads_the_signal  ok
```

The three controls pass in **both** arms, so the pin fixes `try`/`finally` and
nothing wider. `an_unshadowed_read_…` is the one that would catch a fix that
suppressed the transform wholesale.

## Blast radius, two stages

Both arms are the **same tree** modulo this diff — the base arm was built by
reverting these three files, not from an older binary, because `main` moved three
times since the last sweep. Each binary was checked with a discriminating probe on
its own output before use (file size, `buildInfo()` and branch all lie):

```
base      | untrack: false | shadow: false
combined  | untrack: true  | shadow: true
```

**Stage 1** — hash every corpus unit under both arms (`js` + `css`, prod and dev):

```
units 69626 / missing-sources 0
changed 14 units over 7 ids / orphans 0
```

**Stage 2** — compare *each arm* to official through the gate's normalization
(`flattenTemplateHoles` → oxfmt(`compatibility/.oxfmtrc.json`) → `stripBlankLines`),
because a changed hash is not a fixed file:

```
base      match  0 / 14
combined  match 10 / 14
match -> MISMATCH = 0
```

The 10 are 5 ids × {prod, dev} and are removed from
`known-failures.client.json` (42 → 37) and `known-failures.client-dev.json`
(56 → 51) here, since the ratchet is two-sided.

### The four that stay, and why one of them matters

`huly/…/OptimizeSkills.svelte` (2) is a duplicated comment — the esrap
comment-cursor family, unrelated to either fix.

`open-webui/…/Models.svelte` (2) is **#4142's own subject, and it is fixed**. The
two arms' outputs for that file differ on exactly one line:

```
official   const res = await importModels(localStorage.token, models);
base       const res = await importModels(localStorage.token, $.get(models));
combined   const res = await importModels(localStorage.token, models);
```

Its entry does not retire because line 134 of the same file carries an unrelated
comment-placement divergence, which is the first differing line in **both** arms.
A ratchet entry's unit is the file, not the defect — so closing a defect can move
the count by zero, and the same arithmetic in reverse is why a count that moves is
not a count of defects. That is recorded in `KNOWN-FAILURES.md` beside the entry.


baseballyama:
## Control on the sweep's own axis, added after the fact

A teammate retracted a 135,592-pair result today because its negative control did
not move: their harness passed `compile(src, JSON.stringify(opts))`, which the NAPI
binding accepts and silently ignores, so four compile targets were one target
hashed four times and "0 differing pairs" meant nothing.

My sweep passes an object, so it does not have that defect — but **I had not
measured that**, and an unmeasured control is not a control. Measured now, on the
base arm's own stage-1 output:

```
prod != dev   33,311 ids
prod == dev    1,502 ids   (258 modules, 1,244 components)
total          34,813 ids
```

The `dev` axis moves on 96% of the corpus, so the sweep's two targets are two
targets. The 1,244 components where the two agree are not explained here and are
not claimed to be explained — they are simply not the failure mode above, which
would have made the number 34,813.

baseballyama:
## Control on the sweep's own axis, added after the fact

A teammate retracted a 135,592-pair result today because its negative control did
not move: their harness passed `compile(src, JSON.stringify(opts))`, which the NAPI
binding accepts and silently ignores, so four compile targets were one target
hashed four times and "0 differing pairs" meant nothing.

My sweep passes an object, so it does not have that defect — but **I had not
measured that**, and an unmeasured control is not a control. Measured now, on the
base arm's own stage-1 output:

```
prod != dev   33,311 ids
prod == dev    1,502 ids   (258 modules, 1,244 components)
total          34,813 ids
```

The `dev` axis moves on 96% of the corpus, so the sweep's two targets are two
targets. The 1,244 components where the two agree are not explained here and are
not claimed to be explained — they are simply not the failure mode above, which
would have made the number 34,813.

**Second control direction, added after a teammate's point that one direction cannot
separate "the axis moves" from "the axis always differs":** on 392 components under
the base arm, `generate: 'client'` and `generate: 'server'` output agreed **0** times
and differed 392. A sweep whose two targets collapsed onto one would show agreement
here, not just uniformity on the `dev` axis.

baseballyama:
## 追記: 3 件目の修正と、ラチェット 3 件の退役

CI の Compiler parity が **`already PASS` 4 件**で落ちました。名指しされたのは `huly/…/OptimizeSkills.svelte` と `open-webui/…/Settings/Models.svelte` の client / client-dev で、**どちらも私がローカルで「まだ乖離している」と測って残していた 2 件**です。

原因はローカル再構成の側でした。ゲートは oxfmt 正規化後のバイト比較の**あと**に、バイト差のある出力すべてへ **AST 比較**を掛けます（`AST comparison for 3339 byte-different output(s)`）。残差は 2 件ともコメント位置で、AST 比較はそれを表現しません。**再構成が本物より厳しい側に外れると、退役できる entry が「残る」と報告される** — 再構成ハザードの通常の向きとは逆ですが、間違い方は同じです。

### 3 件目: prop デフォルトの入れ子位置での `is_simple_expression`

上流は `is_simple_expression` を**変換後**の初期化子に掛けます。そこでは legacy の store 読みは既に `$s()`、`$:` 変数は既に `$.get(r)` — どちらも呼び出しなので非 simple、したがって `PROPS_IS_LAZY_INITIAL` つきで thunk されます。rsvelte は**変換前のソース**に掛けていたので、どちらも素の識別子でした。

**裸の識別子には専用の分岐が 3 つあった**（`is_store_accessor` / `is_prop_ref` / `LegacyReactive` の thunk）ので、乖離は読みが logical / conditional / binary のオペランドに入っている位置だけに閉じていました。6 binding-kind × 5 position のグリッドで正確に切り分かれます:

```
DIFF 8 セル: store と $: の 非裸 4 位置
EQ  22 セル: 裸の 5 kind、plain let / const / call() の全位置
```

store は**値まで誤り**でした。`$s` を `$s()` に書き換える後段パスはデフォルトが既に `() => …` のときだけ発火するので、出力されるデフォルトは store の値ではなく **getter 関数そのもの**でした:

```js
// official
let c = $.prop($$props, 'c', 24, () => $s() ?? null);
// rsvelte (before)
let c = $.prop($$props, 'c', 8, $s ?? null);
```

### 測定

```
stage 1  全 69,626 client (entry,target) 出力をハッシュ、両アーム
         changed 2 units / 1 id / orphan 0
         対照: prod != dev 33,311 / prod == dev 1,502（後者は throw するファイル）
stage 2  base 0/4 match、fix 2/4 match、match -> MISMATCH = 0
```

server は触れていません（変更は `3_transform/client/props_transforms.rs` の 1 関数、呼び出し元 1 箇所）。

### pin テスト

`crates/rsvelte_core/tests/prop_default_nested_getter_reads.rs` — グリッドの 8 セル + 対照 4 本（裸 store / 裸 `$:` / plain let の 4 位置）。期待値はすべて official の実出力から取っています。

### ラチェット

```
client      37 -> 34
client-dev  51 -> 48
```

PR #4154, opened 2026-09-01
feat(lsp-gate): the ratchet key carries the mechanism, and a label is signed only once it is one

baseballyama:
The LSP differential ratchet held 23,746 entries keyed `aggregate:<file>|<method>`, which is a
file and a method and nothing about **why** the two servers disagree. This branch puts the
mechanism in the key, then works the resulting labels one at a time: a label is signed only once
it has been measured to be **one** mechanism, and it is signed to a terminal — eliminated,
attributed to an `upstream_issues/` report that pastes both compilers' real output, or attributed
to `deliberate-divergences` with a test that pins it.

**The ratchet is deliberately NOT re-baselined here, and the key change has been pulled out of
this PR for the same reason.** A key-format change has no green state between itself and its
re-baseline: it makes every one of the 23,746 committed entries stale at once, where an ordinary
two-sided fix leaves the unaffected side matching. That was measured, not predicted -- on a
7-component bits-ui shard this branch reported `318 NEW / 42 stale`, exit 1, against the committed
baseline; with the key restored to `aggregate:<file>|<method>[|phase]` the same shard is exit 0
with neither. So `mech=` enters the ratchet key in the PR that re-baselines, and the two land
together.

Everything the mechanism work produced stays here: the classifier, its closed vocabulary, the 39
signed labels, the tests, and the attribution table. Only `ratchet.mjs`'s key construction is back
to what `main` has.

## The unit caveat, up front

Six denominators appear in this work and none of them converts into another: **divergent fields**,
**ratchet entries** (`(file, method, mech, phase)` keys), **units** (`(file, method, phase)`
cells), **item occurrences**, **requests**, and **defects**. A number in the prose below always
names which one it is. In particular the entry count is not a burndown curve: 47 of 48 labels can
remove **zero** entries on their own, because a completion unit carries a median of 14 labels and
an entry disappears only when its unit's last label does.

## Reading the attribution table

`scripts/dev/lsp-attribution-table.mjs` prints `STALE-POPULATION`, not `0`, for a signed label the
artifact it is given holds no carrier for. The committed ratchet is only re-keyed by an
`lsp-corpus` baseline run, so every label added after the last artifact was written has no
population yet — and a blank cell, a zero, and a dropped row are three renderings of the same
ambiguity. 19 of the currently signed labels are in that state.

The table's headline is **attribution IDs, not labels**. P3 is complete when every entry is
attributed, and an attribution is a *file* someone writes and someone reads -- so counting labels
overstates the remaining work by the labels-per-ID factor. On the shadcn artifact:

| labels | entries | attribution |
|---|---|---|
| 3 | 86 | `upstream_issues/tsgo-lsp-completion-omits-the-commit-character-inputs.md` |
| 1 | 52 | `upstream_issues/tsgo-lsp-completion-item-omits-the-typescript-kind.md` |
| 1 | 30 | `upstream_issues/svelte-language-server-rune-definition-lands-inside-the-jsdoc.md` |
| 1 | 28 | `deliberate-divergences` |
| 9 | 6 | `upstream_issues/tsgo-lsp-hover-renders-declarations-differently-from-tsc.md` |
| 1 | 2 | `upstream_issues/svelte-language-server-hovers-svelte2tsx-synthesized-render-function.md` |

Three things that table says and a label count does not. Of the 39 signed labels only **16** carry
an attribution ID at all; the other 23 are `R`, whose only end state is zero, and
`attribution-check.mjs` accepts an `upstream_issues/` path or `deliberate-divergences` and nothing
else -- so they are not rows. **The remaining P3 work on this ratchet is 6 attributions + 23 labels
to drive to zero + 20 labels still undecided**, and those three terms are not interchangeable. The
labels-per-ID factor is **not** a constant that could convert between them: it is 9 and 3 on two
IDs and 1 on the other four.

The table also prints `<ID> <- label, label, ...` under any ID with more than one label, because
"all N entries of an N-label defect carry the same attribution" is only a requirement if it is
hand-checkable -- a reader landing on any of those labels has to reach that one report.

## This branch contains four behaviour changes

They are listed explicitly because they invalidate any `--write-current` artifact taken before
them, so **no number measured against the current ratchet can be quoted as a measurement of this
tree**:

| commit | change |
|---|---|
| `333346662` | a definition on `const` / `let` / `var` / `enum` gets an answer |
| `9e5835002` | an open-tag completion is no longer filtered by the name already typed |
| `5ca3b97d9` | a tag completion's `textEdit` range covers the whole name token |
| `0ed798306` | the `style` arm of the trigger-suggest condition escaped its guard |

The last is a real defect with a decisive upstream citation: `vscode-html-languageservice@5.4.0`
`htmlCompletion.js:204-211` guards the `style` name behind `attr.valueSet !== 'v' && value.length`
and only then asks whether it is `style`; rsvelte spelled that inner test as a top-level
alternative, and `&&` binds tighter than `||`, so an attribute already carrying a value still got
`editor.action.triggerSuggest`. The test pins both directions and was confirmed by installing the
old condition and watching it go red.

## What was measured

- **39 labels signed** (U 15 / D 1 / R 23). Each carries `not-MANY (n=…, files=…, pos=…)`; where
  only one position was sampled it is recorded as `insufficient(pos=1)` in a different column,
  because a measurement that did not happen must not share a column with one that did.
- **`ts-symbol-name` and `projection-response-range` are one defect** — a per-line column shift,
  predicted exactly on 3/3 (shifts 11, 6, 6 landing on the neighbouring symbol). The measured
  shifts are 11, 6, 6, 5 and **−1**; the negative one is left unexplained rather than folded in.
  The two labels stay separate: merging them would change the ratchet key and therefore what each
  entry suppresses.
- **`ts-render` was five things.** Its classifier returned the first rewrite that equalized the two
  hover texts, so a hover carrying two renderings took whichever rule sat higher in the table — the
  ratchet key depended on the order the rules were written in. It is now order-free, and the
  residue split into an `import <Name>` line official carries (11 rows), a dropped `(+N overloads)`
  suffix (2), both at once (1), and two rows that are not a rendering difference at all (`any`
  against a resolved type). One row survives unclassified.
- **Two new renderings were re-probed before being attributed.** Official-server against
  rsvelte-server cannot separate a tsgo defect from a svelte2tsx projection artifact, so both were
  re-run through plain TypeScript — `tsc` 6.0.3's `getQuickInfoAtPosition` against `tsgo --lsp`,
  one `.ts` file, no Svelte. Both reproduce, so they join the tsgo report (now six differences);
  the `any` rows were never put to that probe and stay unattributed.
- **A `lib.d.ts` definition lands in the copy that answered it** — pinned as a deliberate
  divergence by `scripts/compat-lsp/test-ts-lib-copy.mjs`, wired into `ci.yml`'s `Language server`
  job. Warm wall-clock 0.89–0.93 s (n=4, debug binary); the single cold run of 142.6 s is
  confounded by load average 6.9 and is not quotable. `scripts/bench-lsp/real-smoke.test.mjs`
  already starts both real servers earlier in the same step, which is a structural argument from
  code rather than a measurement.
- **One upstream report filed** — official hovers svelte2tsx's synthesized `$render`, which exists
  in no document the user has open, with a same-file/same-session control.

## One defect, several labels — on three different axes

Reading the ratchet's entry count as a defect count overcounts, and the measurements say by how
much and why:

- **The field a defect lands on.** `CompletionProvider.ts:894-899` is one `if` body writing
  `textEdit.newText`, `textEdit.range.start` and `completionItem.additionalTextEdits`. rsvelte
  performs none of the three, so the same three `components` units appear under three labels —
  same files, same positions. A per-field ratchet overcounts a skipped block by its field arity.
- **The response shape it produces.** 21 of 23 sampled `rsvelte-empty` hovers are official
  answering with the `import <Name>` line and nothing else — the line tsgo drops. Drop it and
  there is no hover to send, so the null is the *same* rendering difference at the boundary where
  the omitted part was the whole payload.
- **The pointer granularity.** 18 of 21 `projection-origin-range` rows are rsvelte's origin range
  covering the enclosing element's start tag (`<div ` is five characters, `<Example ` nine,
  `<span ` six) — the mechanism already signed under `projection-target-position-workspace`, seen
  through a field-level pointer instead of an element-level one.

## Two negative results kept

**"One side sent nothing" is not an axis.** 48 of 149 sampled `completion-item-set-*` rows are
rsvelte answering zero items, which looks like a key. Splitting on it was written, tested and
reverted: 23 of those 48 are `missing-html-close-tag`, a label already 23/23 empty and already
signed on exactly that reasoning, so the split would have merged a single-mechanism label into a
new many-mechanism one. The defect the count was pointing at was a missing provider *value* —
emmet carries no `kind` and no MDN prose, so the region read it as `html`, and
`detail === "Emmet Abbreviation"` is the one thing on the item that names it.

**A label's population does not tell you whether its key is right.** Ask which already-separated
mechanisms a proposed key would swallow.


baseballyama:
### Why this PR is red, and what its re-baseline means

Posting this from a triage pass rather than as an owner of the branch — the four red checks reduce to **two causes and zero regressions**, and the re-baseline this PR needs is justified differently from how it will read.

**The four reds are two causes.** `Tests` is a rollup of `Language server` (in that run, 27 jobs, the only two non-success are those). `Language server` and `LSP fixture parity current` fail on the *same gate* — step 24 and "Measure fixture and pinned upstream parity" are the same comparison. So: the fixture gate (2 NEW + 2 stale) and real-world shard 7/16 (2 stale, **0 NEW**). All four have logs and completed steps, so none is a vanished runner or a cancellation.

**The fixture gate's four entries are one moving entry.**

```
ratchet: fixtures/html-smoke-lang|textDocument/completion|0:4|/items:missing-rsvelte[count=131,hash=1f0ad94d5109]
run:     fixtures/html-smoke-lang|textDocument/completion|0:4|/items:missing-rsvelte[count=5,  hash=e1a230944816]
```

Same unit, method, position and kind. `count=` is inside the key, so a divergence that gets *smaller* retires one key and enrols another, and the run prints that as a NEW divergence beside a stale entry — the shape a regression produces.

**The cause is this PR's own comparator, not a change in either compiler.** Three readings fit the numbers: rsvelte gained items, official lost items, or the comparator changed how it pairs them. What settles it is that **#4289's `Language server` was SUCCESS on 2026-09-04 against a tree whose ratchet still records `count=131`** — the gate passes elsewhere with the old value, which localises the movement to this branch. This PR's diff adds completion-item pairing logic (`pairing` appears in 12 added lines under `scripts/compat-lsp/`, against 1,553 added lines total), and shard 7's log carries `completion-item-pairing-key-*` labels. So 131 was a mispairing and 5 is the corrected measurement.

Two things that do *not* answer it, recorded so nobody spends the time: the run artifact holds `current` as key strings (2,116 of them) and **carries no values**, so the 126 items that stopped being missing are not recoverable from it; and a `count=`-bearing hash is the doubled form, which has no preimage.

**Consequence for the re-baseline.** There is no degradation to bake in — the (b) reading is excluded. But the justification is not "the implementation advanced" either. It is **"the comparator became more accurate"**, and that is a claim about the whole file, not about one entry: every one of the 23,746 entries is re-measured under a pairing rule that did not exist when they were written.

**That is worth saying in the PR body before this merges.** Without it, the next person to read `lsp-known-failures.json` will assume its entries accumulated under one ruler. It also strengthens rather than weakens the existing constraint that a re-baseline needs the complete 17-artifact run: the job is not to retire one improved entry but to re-measure the ratchet.

The general form is now recorded as gate-coverage rows 27v and 27w in #4292.

baseballyama:
## この PR の 3 つの赤は退行ではなく **re-baseline 債務**です。そして中身の半分は既に `main` にあります。

`main` = `bf6fe3c63` の時点で測りました。

### 1. 失敗している 3 ジョブは全部「ratchet を更新していない」だけ

ログは 3 本とも存在します（runner 消失でも cancel でもない）。失敗ステップの出力そのまま:

**`LSP fixture parity current`** — 同じキーの `count=` が変わっただけです。

```
[lsp-verify] 2 NEW divergence(s):
  differential:fixtures/html-smoke-lang|textDocument/completion|0:4|/items:missing-rsvelte[count=5,hash=e1a230944816]
  differential:fixtures/html-smoke-lang|textDocument/completion|0:4|phase=edit|/items:missing-rsvelte[count=5,hash=e1a230944816]
[lsp-verify] 2 stale ratchet entry/entries:
  differential:fixtures/html-smoke-lang|textDocument/completion|0:4|/items:missing-rsvelte[count=131,hash=1f0ad94d5109]
  differential:fixtures/html-smoke-lang|textDocument/completion|0:4|phase=edit|/items:missing-rsvelte[count=131,hash=1f0ad94d5109]
```

**missing 131 → 5。** file も method も position も同じで、`count` がキーに入っているので二面性 ratchet が「NEW 2 + stale 2」として報告します。**126 件の改善**です。

**`LSP real-world parity (stable shard 7/16)`** — NEW は 0、stale が 2 だけ（純粋な縮小）。

```
[lsp-verify] 249 cases; 257694/257694 compared, 82140 divergent requests / 8418990 divergent fields, 0 skipped
[lsp-verify] corpus divergence mechanisms: 8418990 labelled, 0 unclassified (0.0%), vocabulary 142
[lsp-verify] 2 stale ratchet entry/entries:
  aggregate:corpus/bits-ui/…/is-using-keyboard-test.svelte|textDocument/definition
  aggregate:corpus/bits-ui/…/is-using-keyboard-test.svelte|textDocument/definition|phase=edit
```

**`Tests`** は 5 秒で終わる rollup で、`Verify all test jobs passed` が上の 2 つを読んでいるだけです。

### 2. 4 つの `upstream_issues/` は既に `main` にあり、byte 一致です

`git ls-tree` で blob を突き合わせました（`head` と `main` が同一 = 既に着地済み）:

| path | merge-base | head | main | |
|---|---|---|---|---|
| `upstream_issues/svelte-language-server-hovers-svelte2tsx-synthesized-render-function.md` | 無 | dc2b439cd | dc2b439cd | 着地済み |
| `upstream_issues/svelte-language-server-rune-definition-lands-inside-the-jsdoc.md` | 無 | 4a6bdff23 | 4a6bdff23 | 着地済み |
| `upstream_issues/tsgo-lsp-completion-omits-the-commit-character-inputs.md` | 無 | 68aac73b9 | 68aac73b9 | 着地済み |
| `upstream_issues/tsgo-lsp-hover-renders-declarations-differently-from-tsc.md` | 無 | 514c6209b | 514c6209b | 着地済み |
| `scripts/compat-lsp/mechanism.test.mjs` | 無 | c89ac716d | c89ac716d | 着地済み |
| `scripts/compat-lsp/diff.test.mjs` | 34d7c4457 | d11ec0684 | d11ec0684 | 着地済み |

まだこの PR にしか無いもの:

| path | 状態 |
|---|---|
| `crates/rsvelte_language_server/src/server.rs` | **`main` == merge-base** — この PR の変更は未着地 |
| `crates/rsvelte_language_server/src/completions.rs` | 3 者とも別（`main` は独立に動いた） |
| `scripts/dev/lsp-attribution-table.mjs` | `main` に無い |
| `scripts/compat-lsp/test-ts-lib-copy.mjs` | `main` に無い |
| `.changeset/lsp-completion-and-definition-fixes.md` | `main` に無い |

`131 → 5` を出しているのは `completions.rs` + `server.rs` なので、**この PR の実質は残っています。**

### 3. 上のログの数字を ratchet に手で書き写してはいけません

この測定が読む入力は **5 つ全部が merge-base から動いています**:

```
compatibility/lsp-known-failures.json   778ef30ed → c73073f20
scripts/compat-lsp/verify.mjs           904a6d507 → 27edd8548
scripts/compat-lsp/diff.mjs             3ae182ff7 → eb1a0dedf
scripts/compat-lsp/mechanism.mjs        (merge-base に無し) → 29ad6f1dd   ← main が後から追加
compatibility/lsp-mechanisms.json       (merge-base に無し) → d0c04f872   ← main が後から追加
```

ログは 2026-09-02、`main` はそこから 86 commit 進んでいます。`AGENTS.md` の *An arm and a ratchet from different trees re-detect what was already retired* そのものなので、**rebase → 再実行 → 新しいログから ratchet を直す**の順です。

### 手順

1. `origin/main` (`bf6fe3c63`) に rebase。`completions.rs` と `mechanism.mjs` は conflict する見込み
2. 再実行し、**新しい**ログの NEW / stale を読む
3. `lsp-known-failures.json` を直し、**`lsp-mechanisms.json` の該当レコードも同じ PR で改名**する（sidecar checker は二面性で、ratchet が載せない id を指すレコードも赤になります）
4. CI が飽和しているので、この PR は `lsp-corpus`（16 shard、~1,300 job-minutes）を引きます。キューが空くまで待ちます

### 器械の訂正（この調査で 1 回捏造されました）

最初この blob 表を `git rev-parse "$MB:$p" 2>/dev/null || echo MISSING` で作り、**存在しない 2 つの path が「マージベースの SHA 先頭 9 文字」を blob 欄に返しました**。git 2.33 の `rev-parse` は失敗時に**引数を stdout に echo** するので、`|| echo MISSING` は置換ではなく**追記**になり、`${a:0:9}` がリビジョンを拾います。気づいたのは数ではなく**形**で、別々のファイル 2 つが同じハッシュを持っていました。`git ls-tree <rev> -- <path>` は無い path に対して stdout に**何も出さない**ので、そちらを使ってください。

baseballyama:
Closing as superseded by `main`, measured rather than assumed:

| | this branch | `origin/main` (`6178b9bb9`) |
|---|---|---|
| `scripts/compat-lsp/mechanism.mjs` | 690 lines | **1058 lines** |
| `crates/rsvelte_language_server` version | 0.5.5 | **0.7.3** |
| close-tag completions (`end_tag_bracket`, `tag_name_end`, `close_tag_completions`) | absent | **present** |
| the four `upstream_issues/` reports this branch adds | — | **byte-identical on `main`** (`git diff origin/main HEAD -- <path>` → 0 lines, all four) |

The branch's own last two commits are `revert(lsp-gate): the differential key goes back to main's spelling too` and `revert(lsp-gate): the ratchet key goes back to file|method, without the mechanism` — the headline feature in the title was withdrawn on the branch itself, so what remained was documentation plus the reports, and both have landed by other routes.

`git merge origin/main` produces 7 conflicts, two of them **add/add** (`mechanism.mjs`, `mechanism.test.mjs`), and resolving them would mean choosing `main`'s newer implementation in nearly every hunk — i.e. reconstructing `main`.

Nothing here is lost: if a specific paragraph or cell from this branch is missing on `main`, it is cheaper to lift it into a fresh PR than to merge a branch whose merge resolution is "take the other side".

PR #4169, opened 2026-09-02
fix(svelte2tsx): a binding pattern's `$` key is a store everywhere but the first element

baseballyama:
Stacked on #4162 (base is `fix/svelte2tsx-unparseable-and-events`, so the diff is this
change alone). **After #4162 is squash-merged this branch must be rebased onto `main`, not
merged as-is** — the base commit disappears into the squash.

## Ratchet ids this PR drops

`compatibility/svelte2tsx-known-failures.json`, **8 → 7**, one id:

- `appwrite-console/src/routes/(console)/project-[region]-[project]/storage/bucket-[bucket]/settings/+page.svelte`

No other ratchet moves.

## The defect

`processInstanceScriptContent` decides whether a `$`-prefixed identifier is a **store
reference** or a **declared name** from one boolean, `isDeclaration`. Entering a binding
element's name sets it and pushes an on-leave callback that clears it
(`processInstanceScriptContent.ts:293-296`):

```ts
if (ts.isBindingElement(parent) && parent.name == node) {
    isDeclaration = true;
    onLeaveCallbacks.push(() => (isDeclaration = false));
}
```

The reset is unconditional, so leaving the **first** element of a pattern clears a flag the
enclosing pattern had set, and every element after it is walked with `isDeclaration === false`.
`handleIdentifier` (`:155`) then takes the store branch for that element's *property name*.

rsvelte skipped every object-property key, so
`let { a, $permissions: permissions } = o` emitted no
`let $permissions = __sveltets_2_store_get(permissions);` line at all.

## The axis

An element's index within **its own** pattern, crossed with the pattern's host.
`-` is no `__sveltets_2_store_get` call. Every value is official's own output.

| source | official |
|---|---|
| `let { $permissions: permissions } = o` | `-` |
| `let { a, $permissions: permissions } = o` | `permissions` |
| `let { a, b, $permissions: permissions } = o` | `permissions` |
| `let { a = 1, $permissions: permissions } = o` | `permissions` |
| `let { x: { $permissions: permissions } } = o` | `-` |
| `let { a, x: { $permissions: permissions } } = o` | `-` |
| `let { x: { a, $permissions: permissions } } = o` | `permissions` |
| `let [a, { $permissions: permissions }] = [o, o]` | `-` |
| `function f({ a, $permissions: permissions }: any)` | `permissions` |
| `let { $p: p, $q: q } = o` | `q` |
| `let { a, $p: p, $q: q } = o` | `p`, `q` |
| `({ a, $p: p } = o)` | `-` |

The nested rows are what name the mechanism rather than merely fitting it: `{ a, x: { $p: p } }`
is quiet because entering the nested pattern **re-sets** the flag its sibling had cleared, and
`{ x: { a, $p: p } }` is loud for the same reason one level down. A rule stated as "not the first
key in the statement" fits the flat rows and gets both of those backwards.

## Measurement

Two-stage sweep, distinct binaries per arm, arms identified two-sided (base answers `-` where
after answers `permissions`, and an unrelated `@typedef` fingerprint is 0 differing lines on
both). Stage 1 hashes every corpus component; stage 2 re-verdicts only what moved, through the
gate's own normalization.

```
units: 33901  moved: 1
MISMATCH -> match:    1
match -> MISMATCH:    0
MISMATCH -> MISMATCH: 0
match -> match:       0
```

33,901 is the number of `.svelte` entries in `compatibility/manifest.json` (34,828 total; the
other 927 are `.svelte.(js|ts)` modules svelte2tsx does not convert). svelte2tsx has no compile
targets, so it is not target-multiplied.

Positive control: ablating the fix fails **7 of the 14** grid cells — the branch condition
splitting the grid in half, which a fix reaching every cell or one cell would not do. Restoring
leaves `git diff` empty and the test green.

## Why the entry sat at "output only"

The mechanism table's justification read: *"had a plausible cause — `$permissions` as a
destructuring-rename KEY — and a five-line reduction of that has both tools agreeing, so the
cause is something else."* The cause **was** the destructuring-rename key. The reduction put it
in a pattern of one element, which is the one position where upstream does not resolve it as a
store — so a reduction that reproduces nothing was recorded as a finding about the defect when it
was a fact about the reduction. That is written up in the `Previously:` section.

## Also here

Both are about a pin the previous commit left unheld.

- The `@typedef` offset divergence is now registered in
  `compatibility/GATES.md#deliberate-divergences`. Its pin had to **move out of** `export_decl.rs`:
  the gate's pin predicate accepts only a path under `tests/`, in `compatibility/pattern-corpus/`,
  or a `test-*.mjs` / `*.test.mjs` harness, so a `#[test]` inside `src/` is citable and is not a
  pin. It is now `crates/rsvelte_projection/tests/svelte2tsx_typedef_tag_offset.rs`, with the two
  rows rsvelte *does* reproduce beside it as controls. Gate 42's own prose said "11 of them"
  against a checker that counts 20; corrected.
- That test's doc comment quoted official's corrupted output from a **different input**. The
  deleted slice is offset by `node.pos`, so the quoted bytes change with whatever precedes the
  comment. The value is now the one measured on the test's own source
  (`{\n/**\n * @typedef {i{ a: 1 }}\n */a: a}`).

## Checks run locally

| check | result |
|---|---|
| `cargo test -p rsvelte_projection --lib --tests` | `TEST_EXIT=0` — 84 suites, 702 passed, 0 failed |
| `cargo clippy -p rsvelte_projection --all-targets --all-features -- -D warnings` | `CLIPPY_EXIT=0` |
| `cargo fmt --all` | no diff |
| `known-failures-md-check.mjs` | 50 ratchets across 36 docs, 32 partitions add up (positive control: `5 + 2` → `6 + 2` fails with two independent messages) |
| `deliberate-divergences-check.mjs` | 20 recorded, each pinned (positive control: replacing this section's `**Pinned by**` with prose names it at `GATES.md:8074`) |
| `check-upstream-issues.mjs` | 81 reports, all indexed |
| `check-core-consumer-changesets.mjs` | all three consumer packages named |

### DIFF
--- a/.changeset/svelte2tsx-binding-pattern-store-key.md
+++ b/.changeset/svelte2tsx-binding-pattern-store-key.md
@@ -1,0 +1,18 @@
+---
+'@rsvelte/svelte2tsx': patch
+'@rsvelte/compiler': patch
+'@rsvelte/svelte-check': patch
+---
+
+A `$`-prefixed key of a binding pattern is a store reference everywhere except
+the pattern's first element, and rsvelte emitted no store subscription for any
+of them — `let { a, $permissions: permissions } = o` lost its
+`let $permissions = __sveltets_2_store_get(permissions);` line.
+
+`processInstanceScriptContent` tracks "am I inside a declaration" with a single
+boolean whose on-leave callback clears it unconditionally, so leaving a pattern's
+first element clears a flag the enclosing pattern had set and every element after
+it is walked as an expression. The rule that produces — a key is a name iff it is
+the first element of its own pattern — is reproduced here, including the nested
+cases where entering an inner pattern re-sets the flag. Reported as
+`upstream_issues/svelte2tsx-isdeclaration-is-a-boolean-not-a-stack.md`.
--- a/compatibility/GATES.md
+++ b/compatibility/GATES.md
@@ -5284,7 +5284,7 @@ sample of one tree's output, not an enumeration of the shapes upstream emits.
 
 ## 42. Deliberate-divergence pinning — `scripts/dev/deliberate-divergences-check.mjs`
 
-**Unit.** One `## ` section of `compatibility/deliberate-divergences.md`, 11 of them. The check is
+**Unit.** One `## ` section of `compatibility/deliberate-divergences.md`, 20 of them. The check is
 that each names at least one repository path that (a) exists on disk and (b) is a test — under a
 `tests/` directory, in `compatibility/pattern-corpus/`, or a `scripts/**/test-*.mjs` harness.
 Run by `ci.yml`'s `Corpus verify baseline-flag contract` job and `pnpm run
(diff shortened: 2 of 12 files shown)


baseballyama:
Superseded by #4172 — GitHub auto-closed this PR when `fix/svelte2tsx-unparseable-and-events` was deleted by #4162's squash merge, and it cannot be reopened once the base ref is gone (recreating the ref does not lift the refusal). The branch is unchanged apart from a rebase onto `main`; the ratchet is re-counted on that tree at 7.

PR #4170, opened 2026-09-02
fix(svelte2tsx): a prop widener at the script's last byte is discarded upstream

baseballyama:
Stacked on #4169 (base is `fix/svelte2tsx-store-key-position`, so the diff is this change
alone). **After #4169 is squash-merged this branch must be rebased onto `main`, not merged
as-is.**

## Ratchet ids this PR drops

`compatibility/svelte2tsx-known-failures.json`, **7 → 6**, one id:

- `sveltekit/packages/package/test/watch/expected/Test.svelte`

No other ratchet moves.

## The defect

`preprendStr` (`utils/magic-string.ts:7-17`) does not append. It **overwrites** the single
character at the insertion point with `toAppend + that character`:

```ts
str.overwrite(index, index + 1, toAppend + str.original.charAt(index), { contentOnly: true });
```

`propTypeAssertToUserDefined` uses it to add `;x = __sveltets_2_any(x);` at
`declaration.end`. When the declaration is the last thing in the script, that character is
the `<` of `</script>` — and the script-tag removal overwrites the same chunk afterwards,
discarding the insertion. No error is raised; the widener is simply absent.

The listed entry is the whole file:

```svelte
<script lang="ts">export let answer: number</script>
```

## The axis

One trailing byte of any kind moves the insertion point off that character.

| source | official |
|---|---|
| `export let answer: number</script>` | — |
| `export let answer: number </script>` | widened |
| `export let answer: number\t</script>` | widened |
| `export let answer: number/*c*/</script>` | widened |
| `export let answer: number;</script>` | widened |
| `export let answer: number\n</script>` | widened |
| `…</script>\n<p>x</p>` | — |
| `export let a: number;export let answer: number</script>` | `a` only |
| `export let a: number, answer: number</script>` | `a` only |

Markup after `</script>` does not change the answer: the position that matters is the end of
the script **content**, not of the file. The same position carries the SvelteKit
`import('./$types.js')` annotation when `nameEnd === end`, so `export let data` in a
`+page.svelte` ending at `</script>` loses its `PageData` type too; `export const snapshot`,
whose annotation goes on the name, does not.

## Three sites, one reachable

rsvelte inserts at a declaration end in three places. Rather than guard all three on the
strength of the mechanism, each was measured:

| site | at the script end | verdict |
|---|---|---|
| `export let` (`propTypeAssertToUserDefined`) | reachable | **fixed here** |
| `export { x as y }` re-export of a local `let` | reachable | both tools emit nothing, with and without a trailing byte — the path does not fire when the export precedes the declaration |
| non-exported sibling of that declaration list | **unreachable** — the `export { … }` statement has to follow the list, so a sibling cannot be the script's last token | not guarded |

That measurement found a divergence of its own, unrelated to the script end: for
`export { a as b };let a = 1, c: number` official widens nothing while rsvelte widens the
non-exported sibling `c` — identically with and without a trailing space, and identically on
both arms of this fix. It has no corpus witness (the gate is two-sided and green apart from
the listed entries), and is recorded in `KNOWN-FAILURES.md` rather than fixed here.

## Measurement

Arms identified two-sided: the before arm carries the previous PR's store-key fix and not
this guard, the after arm carries both.

```
units: 33901  moved: 1
MISMATCH -> match:    1
match -> MISMATCH:    0
MISMATCH -> MISMATCH: 0
match -> match:       0
```

Positive control: ablating the guard fails **9 of the 19** cells; the 10 that pass are the
ones a position-blind suppression would break.

## Checks run locally

| check | result |
|---|---|
| `cargo test -p rsvelte_projection --lib --tests` | `TEST_EXIT=0` — 85 suites, 703 passed, 0 failed |
| `cargo clippy -p rsvelte_projection --all-targets --all-features -- -D warnings` | `CLIPPY_EXIT=0` |
| `cargo fmt --all` | no diff |
| `known-failures-md-check.mjs` | 50 ratchets across 36 docs, 32 partitions add up |
| `check-upstream-issues.mjs` | 82 reports, all indexed |
| `check-core-consumer-changesets.mjs` | all three consumer packages named |

### DIFF
--- a/.changeset/svelte2tsx-script-end-insertion.md
+++ b/.changeset/svelte2tsx-script-end-insertion.md
@@ -1,0 +1,16 @@
+---
+'@rsvelte/svelte2tsx': patch
+'@rsvelte/compiler': patch
+'@rsvelte/svelte-check': patch
+---
+
+A prop widener that would land on the script's last byte is no longer emitted, because
+upstream discards it there.
+
+`preprendStr` overwrites the single character at its insertion point rather than appending,
+so `propTypeAssertToUserDefined`'s `;x = __sveltets_2_any(x);` at `declaration.end` is
+overwritten by the `</script>` removal when the declaration is the last thing in the script.
+Any trailing byte — a space, a tab, a comment, a `;`, a newline — moves the insertion point
+and the widener survives. The same position carries the SvelteKit `./$types.js` annotation
+when the declaration ends at its name, so that is lost with it. Reported as
+`upstream_issues/svelte2tsx-preprendstr-insertion-at-the-script-end-is-overwritten.md`.
--- a/compatibility/KNOWN-FAILURES.md
+++ b/compatibility/KNOWN-FAILURES.md
@@ -6216,11 +6216,11 @@ The svelte2tsx output-parity corpus (`scripts/compat-corpus/svelte2tsx-*`) compa
 rsvelte's svelte2tsx port against **official `svelte2tsx`** byte-for-byte (after
 oxfmt normalization). The ratchet may only shrink.
 
-**Current baseline: `svelte2tsx-known-failures.json`, 7 entries.**
+**Current baseline: `svelte2tsx-known-failures.json`, 6 entries.**
 
-Partition of `svelte2tsx-known-failures.json` by verdict: `5 + 2`
+Partition of `svelte2tsx-known-failures.json` by verdict: `4 + 2`
 
-- **5 — the emitted TSX differs** (`ts-mismatch`).
+- **4 — the emitted TSX differs** (`ts-mismatch`).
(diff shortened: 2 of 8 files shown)


baseballyama:
Base を `fix/svelte2tsx-store-key-position` から `main` に付け替えました。#4172 の squash merge が base ブランチを消すと、GitHub はこの PR を **auto-close** し、`gh pr reopen` も `gh pr edit --base` も受け付けなくなります（base ref を `git push <sha>:refs/heads/<name>` で作り直しても `reopenPullRequest` の拒否は解けません）。#4169 がそれで失われ、#4172 として作り直しました。

#4172 がマージされるまで、この PR の差分には #4172 の commit も含まれて見えます。マージ後に解消します。

PR #4184, opened 2026-09-02
fix(client): a member assignment rooted at a global is not wrapped in `$.assign` (both ports)

baseballyama:
## What

Upstream's `build_assignment` walks an assignment target down to its root identifier and stops at
`const binding = context.state.scope.get(object.name); if (!binding) return null`
(`phases/3-transform/client/visitors/AssignmentExpression.js:104-118`), so a chain rooted at a
**global** is never instrumented. rsvelte's settled-script pass (`client/assign_dev_ast.rs`) had no
binding test at all and wrapped every member assignment, so dev-mode client output carried a
`$.assign(...)` around `document.body.q = v` that official does not emit.

## Why the guard is a union

Neither half can answer the question alone, and the grid has one row for each:

| half | the row that needs it |
|---|---|
| the fragment's own `SemanticBuilder` resolution | **shadowed global** — a `let document` *is* a binding, so a name test would reject it |
| the component's binding set | **import** — the instance body this pass walks has had its imports hoisted out, so the fragment cannot resolve the name |

## Evidence

**Expectations are generated, not inferred.** All 17 cells were compiled through
`submodules/svelte/packages/svelte/src/compiler/index.js` (the source path, per `oracle.mjs`'s
`OFFICIAL_COMPILER_REL` — the npm build and the built submodule disagree with it on output) and the
counts pasted in.

**Blast radius: 0.** A two-arm sweep hashed `js.code + css.code` for every `.svelte` manifest id x
4 targets = **135,560 pairs** (135,560 rows / 135,560 distinct keys, so no key collapse) and
reported **MOVED=0**. Stage 2 therefore had an empty input and there is no direction to report.

**The arms were separated by a discriminating probe before that zero was believed**, because a
MOVED=0 is exactly what a mislabelled pair of arms also produces:

```
napi-base920.node        global-root=1  local-root=1  import-root=0
napi-fix-globalroot.node global-root=0  local-root=1  import-root=0
```

`local-root=1` in both arms is the half of the change that nets to zero — a probe on that row alone
would have been powerless.

So the shape has **no witness in the collected corpus**, and the 17-cell grid is the entire
detector. That is a statement about the population, not about the change.

## Both ports, because the rule is ported twice

`build_assignment` is ported twice — the settled-script pass (`client/assign_dev_ast.rs`) and the
template-expression converter (`visitors/expression_converter.rs::try_dev_assign_wrap_typed`). The
17 cells above put every assignment in a `<script>` function, so **all 17 reach the first port and
none reaches the second**: host and expression shape were confounded, the same shape this repo
records for `binding-position`, whose seven rows bake one host into each `wrap`.

The second commit (`b6531e5cc`) gives the same guard to the template port, with the host declared
as its own axis — 27 cells, three hosts, oracle-generated:

```
                   global  window  undecl  this  call  local  import  state  shadowed
script fn             EQ      EQ      EQ    EQ    EQ    EQ      EQ     EQ      EQ
template onclick      EQ      EQ      EQ    EQ    EQ    EQ      EQ     EQ      EQ
template on:click   DIFF    DIFF    DIFF  DIFF  DIFF    EQ      EQ     EQ      EQ
```

`template onclick` is all-zero on **both** sides — a modern inline handler is never instrumented —
so the two template hosts are not interchangeable, and a grid that had picked the modern spelling
would have reported this port as correct.

**Positive control for the second commit.** With that guard reverted in the same tree, the grid
reports exactly the five `template on:click` cells and nothing else; the other two hosts stay green
in both arms.

```
template on:click/global document: official 0, rsvelte 1
template on:click/global window:   official 0, rsvelte 1
template on:click/undeclared:      official 0, rsvelte 1
template on:click/this member:     official 0, rsvelte 1
template on:click/call root:       official 0, rsvelte 1
```

`this member` and `call root` are upstream's **first** stop (`object.type !== 'Identifier'`), not
the reported defect's — controls added because the rule has two stops, and they diverged.

**Blast radius of the second commit, measured.** A two-arm sweep hashed `js.code + css.code` for
every manifest id x 4 targets — **139,252 pairs**, 139,252 rows / 139,252 distinct keys, so no key
collapse — and reported **MOVED=1**. The arms differ by this commit alone: the base was built by
reverting `expression_converter.rs` to its `7a298d5ce` content in the same tree.

The one moved unit is
`client-dev huly/plugins/contact-resources/src/components/SelectAvatarPopup.svelte`, whose write is
in an `on:click={() => (document.body.onfocus = …)}`, and stage 2 scores it **`MISMATCH` before and
`match` after** — it retires from the ratchet. That reconstruction has no `ast_equiv_batch` rescue,
so it is stricter than the gate, and a `match` under a stricter comparison is a match under the
gate's.

The arms were separated by a discriminating probe before the count was believed, one process per
arm, probing for what each should contain and what it should lack:

```
                       on: global   on: call root   on: local   on: state
napi-port2-base.node            1               1           1           1
napi-port2-fix.node             0               0           1           1
```

`on: local` and `on: state` are the half that must not move.

The `Blast radius: 0` for the first commit is a statement about the settled-script port only: the
corpus holds no `.svelte` file that writes to a global-rooted member from a settled script, and
this entry is the one that does it from a template. **The two ports' sweeps moved one unit each,
and it is a different unit each time.**

## Not in this PR

Two adjacent mechanisms were measured and deliberately left alone, recorded here so the next person
does not rediscover them:

- `compileModule` emits **no** `$.assign` at all, on any shape. That is a **third port** of the
  same rule, erring in the opposite direction, and it carries three of the client-dev ratchet's
  entries (`photon/.../profile.svelte.ts`, `photon/.../actions.svelte.ts`,
  `threlte/.../useGamepad.svelte.ts`). Its own PR.
- the `console.*` dev wrap fires on a **name** test here, where upstream asks
  `node.arguments.some(arg => arg.type === 'SpreadElement' || scope.evaluate(arg).has_unknown)`
  (`CallExpression.js:91-104`) — a value test.

### DIFF
--- a/.changeset/assign-dev-global-root.md
+++ b/.changeset/assign-dev-global-root.md
@@ -1,0 +1,15 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Dev-mode `$.assign` is not emitted for a member chain rooted at a global.
+
+Upstream's `build_assignment` walks the assignment target down to its root identifier and
+stops at `if (!binding) return null`, so `document.body.onfocus = handler` is left alone.
+rsvelte's settled-script pass had no binding test and wrapped every member assignment.
+
+The guard reads two things, because neither half is sufficient on its own: the pass walks
+the instance body, whose imports have been hoisted out of it, so an imported root resolves
+nowhere in the fragment and is known only to the component's binding set — while a name
+declared inside a function here is not a component binding and is known only to the
+fragment. `shadowed global` and `import` are the two rows that separate them.
--- a/.changeset/assign-dev-template-port.md
+++ b/.changeset/assign-dev-template-port.md
@@ -1,0 +1,9 @@
+---
+'@rsvelte/compiler': patch
+---
+
+dev `$.assign` is not emitted for a member chain rooted at a global in a template expression
+
+`build_assignment` is ported twice; the settled-script port already stopped at an
+unresolvable root, and the template-expression converter did not — so an assignment
+written in a legacy `on:` handler was still instrumented.
--- a/compatibility/KNOWN-FAILURES.md
+++ b/compatibility/KNOWN-FAILURES.md
@@ -3310,17 +3310,29 @@ that became unparseable only with `dev: true`; #3877 corrected the component
 callback tail-comment insertion point, so both its parse and output entries have
 been retired.
 
-### Client dev (`known-failures.client-dev.json`, 25 entries)
+### Client dev (`known-failures.client-dev.json`, 24 entries)
 
-Partition of `known-failures.client-dev.json` by verdict: `25`
+Partition of `known-failures.client-dev.json` by verdict: `24`
 
-- **25 — the generated JS differs.**
+- **24 — the generated JS differs.**
 
 Unlike `client`, no CSS entry survives on this target.
 
-All remaining 25 arrived with the wave-2 enrolment (#3176); this target was at 0 before
+All remaining 24 arrived with the wave-2 enrolment (#3176); this target was at 0 before
 it, and it is the largest of the four — 15 JS entries that `client` does not
 carry, which is the reason it is ratcheted separately.
+
+`huly/…/SelectAvatarPopup.svelte` left it when a member assignment whose root resolves to
+no binding stopped being wrapped. Upstream's `build_assignment` opens with `if (!binding)
+return null` (`AssignmentExpression.js:117`), so `document.body.style.overflow = 'hidden'`
+gets no `$.assign` at all; rsvelte wrapped it, and did so from two ports of that function
+which had to be corrected separately. Both compilers' output for this file is now
+byte-identical on `client-dev` before any normalization — and the two-sided ratchet named
+it independently, on Linux, in the PR's own `Compiler parity` job (`1 baseline entries
+already PASS … client-dev 1`). The local measurement carried its own control (two entries
+still listed on this ratchet, run through the identical script, both `DIFF`), but the CI
+naming is the stronger citation because it is the gate's own verdict rather than a
+reconstruction of it.
 
 `musicat/…/InternetArchiveView.svelte` left this target when the dev `console.*` wrap started
 asking whether an argument's **value** is known rather than whether its name is a state binding.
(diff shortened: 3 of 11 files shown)


baseballyama:
## Correction: this fixes one of two ports, and the carrier is in the other

The PR body above says the shape has no witness in the collected corpus. That is true and it is
also, on its own, misleading — I read it as "the class is closed" and it is not.

`known-failures.client-dev.json` lists `huly/.../SelectAvatarPopup.svelte`, whose first material
difference is exactly this defect:

```
official   return document.body.onfocus = handleFileSelectionCancel;
rsvelte    return $.assign(document.body, 'onfocus', '=', handleFileSelectionCancel, '…:266:23');
```

It does **not** move under this branch, and the sweep's `MOVED=0` is consistent with that rather
than contradicted by it. The assignment is written at line 266 as
`on:click={() => (document.body.onfocus = handleFileSelectionCancel)}` — a **template
expression**, which reaches `visitors/expression_converter.rs::try_dev_assign_wrap_typed`, not the
settled-script pass this branch changes. `build_assignment` is ported twice
(`compatibility/GATES.md#two-ports-inventory`), and the 17-cell grid held the **host** fixed at
`<script>`.

Crossing the host and re-measuring, with the arm that has this branch:

```
                   script fn   template onclick   template on:click
global document        EQ            EQ                 DIFF
global window          EQ            EQ                 DIFF
undeclared             EQ            EQ                 DIFF
this member            EQ            EQ                 DIFF   <- control, added
call root              EQ            EQ                 DIFF   <- control, added
script local           EQ            EQ                 EQ
import                 EQ            EQ                 EQ
state member           EQ            EQ                 EQ
shadowed global        EQ            EQ                 EQ
                                            36 cells, 5 DIFF
```

Two things this says that the original grid could not.

**The discriminating host is a legacy `on:` directive**, not "the template" — a modern
`onclick={...}` inline handler is never instrumented on either side, so it agrees for a reason
that has nothing to do with this rule.

**Two of the five divergences came from control cells I added rather than from the reported
defect.** Upstream stops twice on the way down — `if (object.type !== 'Identifier') return null`
and then `if (!binding) return null` — and the second port has neither. `this.q` and `g().q` are
the first stop; nobody reported them.

I am extending this branch rather than merging it and filing a follow-up, because a PR whose body
names the class while fixing one of its two ports is the failure mode this repository already
records: a note that argues the enumeration is closed.

**Method note.** `MOVED=0` over 135,560 pairs is a statement about the population, not about the
mechanism. It was measured correctly (the arms were separated by a discriminating probe first) and
then read as one claim too strong.

PR #4192, opened 2026-09-02
fix(client): `$$ownership_validator` is declared for a mutation that was seen, not one that was wrapped

### LINKED ISSUE #4130: client-dev: $.create_ownership_validator is never emitted
`known-failures.client-dev.json` の **2 件**が同じ形です。これは `client` には出ず、**dev 計装だけに現れます**。

計測条件:

- バインディング `target/release/librsvelte_napi.dylib`、`buildInfo()` = `{"commit":"b308730e520d99c3e53126123984a1885f9b7fcd","dirty":false}`（現在の `main`、クリーン）
- オラクルは `submodules/svelte/packages/svelte/src/compiler/index.js`（`OFFICIAL_COMPILER_REL`）
- オプションは `scripts/compat-corpus/compile.mjs` と同一（`generate: "client"`, `dev: true`）
- **正規化はゲートと同じ**（`flattenTemplateHoles` → `oxfmt` → `stripBlankLines`）
- 出所: `compatibility/known-failures.client-dev.json` 59 件のうち、`known-failures.client.json` に**載っていない 15 件**の機構分割（#4048）。15 件の割り当ては網羅・排他を検算済み

### `huly/plugins/process-resources/src/components/settings/ProcessAttributeEditor.svelte`

正規化後 20 行目から:

```js
// official
var $$ownership_validator = $.create_ownership_validator($$props);
const value = $.mutable_source();
const slot = $.mutable_source();
const presenterClass = $.mutable_source();
const context = $.mutable_source();
let process = $.prop($$props, "process", 8);
// rsvelte
const value = $.mutable_source();
const slot = $.mutable_source();
const presenterClass = $.mutable_source();
const context = $.mutable_source();
let process = $.prop($$props, "process", 8);
let _class = $.prop($$props, "_class", 8);
```

### `huly/plugins/tracker-resources/src/components/issues/move/SelectReplacement.svelte`

正規化後 56 行目から:

```js
// official
var $$ownership_validator = $.create_ownership_validator($$props);
const missingComponents = $.mutable_source();
let targetProject = $.prop($$props, "targetProject", 8);
let issues = $.prop($$props, "issues", 8);
let components = $.prop($$props, "components", 24, () => []);
let componentToUpdate = $.prop($$props, "componentToUpdate", 12);
// rsvelte
const missingComponents = $.mutable_source();
let targetProject = $.prop($$props, "targetProject", 8);
let issues = $.prop($$props, "issues", 8);
let components = $.prop($$props, "components", 24, () => []);
let componentToUpdate = $.prop($$props, "componentToUpdate", 12);
$.legacy_pre_effect(
```

## 修正時に必要なこと

- 直したら `known-failures.client-dev.json` の**エントリ集合を差分**してください。総数の減少だけでは、
  直した件数と壊した件数の相殺が見えません（`match -> mismatch = 0` を明示）。
- pin テストを `crates/rsvelte_core/tests/` に。**このクラスタの全ファイルが同じ形なら、1 本で
  全部を主張できているか**を確認してください。

baseballyama:
Closes #4130.

Upstream latches `analysis.needs_mutation_validation` **before** it builds the mutation's
property path (`shared/utils.js:406`), so a computed key it cannot spell — anything but an
identifier or a literal — leaves the mutation unwrapped and **still emits the preamble**.
rsvelte derived the flag from a text scan for `$$ownership_validator.mutation`, which by
construction can only find a mutation that *was* wrapped, so
`object[objectKey ?? key] = v` emitted neither the declaration nor the wrap where official
emits the declaration alone.

A second divergence sat one level further out. `scan_member_chain_names` bailed on a root
wrapped in plain parentheses, because the helper that steps over a parenthesised root
accepted only an `as` / `satisfies` assertion inside them — and acorn erases an empty pair,
so `(object)[key] = v` is the same tree as `object[key] = v`.

## The two halves are NOT orthogonal, and that was predicted wrong

The prediction was that ablating the paren arm would leave the latch intact and cost 3 cells
(`decl=1 wrap=0`). Measured: **8 cells, all `decl=0`**. `props_transforms`'s scan builds
`PropMutationSites`, `source_has_member_write` reads it, and that gates the latch — so the
root arm sits *upstream* of the index rule rather than beside it, and a fix for the index
alone is dead on a parenthesised root. The wrong prediction is what named the dependency;
an ablation that only asked "does it go red" would not have.

## Grid — 24 cells, every expectation generated from the oracle

`crates/rsvelte_core/tests/ownership_validator_unbuildable_path.rs` crosses 3 roots
(`object`, `(object as any)`, `(object)`) with 8 computed-key shapes, and counts
`create_ownership_validator` **and** `$$ownership_validator.mutation(` separately per cell,
so "flag set but wrap missing" is observable rather than folded into one verdict. The five
non-nameable indices are the controls that must keep `wrap == 0` while `decl == 1` — they
reject a fix that simply wraps everything.

## Positive control

| arm | red cells | reading |
|---|---|---|
| latch ablated (`if false && …`) | **15** — the 5 unspellable indices at every root | `decl=0 wrap=0` |
| paren arm ablated (empty-assertion return removed) | **8** — every `paren` cell | `decl=0`, incl. the 3 whose wrap the latch never touches |
| both restored | **0** | 24/24, `git diff` against the kept copies empty |

## Population of the local checks

- `cargo clippy -p rsvelte_core --lib --tests --all-features -- -D warnings` — exit 0
- `cargo test -p rsvelte_core --lib` — `running 1959 tests`, **1958 passed, 0 failed, 1 ignored**
  (the four-digit count is the fingerprint that the `--lib` half ran; the first pass of this
  work compiled only the integration target and missed 13 broken in-lib call sites)
- `cargo fmt --all --check` — exit 0

## Not measured here

Which corpus ratchet entries this retires. Three `huly` client-dev entries are the census's
candidates for this mechanism; they are measured in a follow-up rather than asserted, and no
ratchet JSON is touched in this PR.

### DIFF
--- a/.changeset/ownership-validator-unspellable-path.md
+++ b/.changeset/ownership-validator-unspellable-path.md
@@ -1,0 +1,21 @@
+---
+"@rsvelte/compiler": patch
+---
+
+`$$ownership_validator` is declared for a prop mutation that was SEEN, not only for one that could be wrapped
+
+Upstream latches `analysis.needs_mutation_validation` before it builds the mutation's
+property path (`shared/utils.js:406`), so a computed key it cannot spell — anything but an
+identifier or a literal — leaves the mutation unwrapped and still emits the preamble.
+rsvelte derived the flag from a text scan for `$$ownership_validator.mutation`, which by
+construction can only find a mutation that *was* wrapped, so
+`object[objectKey ?? key] = v` emitted neither.
+
+A second divergence sat one level further out: `scan_member_chain_names` bailed on a root
+wrapped in plain parentheses, because the helper that steps over a parenthesised root only
+accepted an `as` / `satisfies` assertion inside them and acorn erases an empty pair. The
+two are not orthogonal — the scan builds `PropMutationSites`, which `source_has_member_write`
+reads, which gates the latch — so a fix for the index alone is dead on a parenthesised root.
+Ablating each half separately measures it: the latch alone falls on 15 of the grid's 24
+cells, the root arm alone on all 8 parenthesised ones including the three whose wrap the
+latch never touches.
--- a/.changeset/ownership-validator-unspellable-path.md
+++ b/.changeset/ownership-validator-unspellable-path.md
@@ -19,3 +19,13 @@ reads, which gates the latch — so a fix for the index alone is dead on a paren
 Ablating each half separately measures it: the latch alone falls on 15 of the grid's 24
 cells, the root arm alone on all 8 parenthesised ones including the three whose wrap the
 latch never touches.
+
+The latch itself then had to learn the other half of upstream's condition. Upstream reaches
+the validator through `scope.get(name)` (`shared/utils.js:396`), so a name that is a prop
+but resolves to something else declares nothing; this pass asked only whether the name
+matched a prop, so `list.forEach((p) => { p.x = 1 })` latched on the parameter. The
+generated instance body parses as a top-level statement list, which makes the props the
+root scope's bindings and a shadow any other scope — the test `state_pipeline_ast` already
+makes. Deriving the answer from the binder rather than from a list of shadowing syntaxes is
+what closes it: `for (const p of …)`, `catch (p)` and a block-scoped `let p` are not
+parameters, and `is_shadowed_by_function_param` declines all three.
--- a/compatibility/KNOWN-FAILURES.md
+++ b/compatibility/KNOWN-FAILURES.md
@@ -3332,16 +3332,16 @@ that became unparseable only with `dev: true`; #3877 corrected the component
 callback tail-comment insertion point, so both its parse and output entries have
 been retired.
 
-### Client dev (`known-failures.client-dev.json`, 24 entries)
+### Client dev (`known-failures.client-dev.json`, 22 entries)
 
-Partition of `known-failures.client-dev.json` by verdict: `24`
+Partition of `known-failures.client-dev.json` by verdict: `22`
 
-- **24 — the generated JS differs.**
+- **22 — the generated JS differs.**
 
 Unlike `client`, no CSS entry survives on this target.
 
-All remaining 24 arrived with the wave-2 enrolment (#3176); this target was at 0 before
-it, and it is the largest of the four — 15 JS entries that `client` does not
+All remaining 22 arrived with the wave-2 enrolment (#3176); this target was at 0 before
+it, and it is the largest of the four — 11 JS entries that `client` does not
 carry, which is the reason it is ratcheted separately.
 
 `huly/…/SelectAvatarPopup.svelte` left it when a member assignment whose root resolves to
@@ -3361,6 +3361,26 @@ asking whether an argument's **value** is known rather than whether its name is 
 The pass reads lowered text, so `$state(0)` had reached it as an opaque `$.state(0)` call; upstream
 evaluates the rune's argument (`scope.js:465-500`). A two-arm sweep over 139,252 `(id, target)`
 pairs moved this unit and no other.
+
+#4192 retired two `huly` entries (`process-resources/.../ProcessAttributeEditor.svelte`,
+`tracker-resources/.../move/SelectReplacement.svelte`). Upstream latches
+`analysis.needs_mutation_validation` before it builds the mutation's property path
+(`shared/utils.js:406`), so a computed key it cannot spell still emits
+`$$ownership_validator`; rsvelte derived the flag from a text scan for the wrap, which by
+construction only finds a mutation that was wrapped. Measured on two arms sharing the merge
+base `c1af73536` (base sha256 `3c7a3ccb…`, fixed `ec564195…`): both entries read `PASSES`
+on the fixed arm and the base arm reports `PASSES on base: 0`, so the retirement is
+attributed rather than inherited.
+
+The same PR then narrowed the latch again — upstream reaches the validator through
+`scope.get(name)`, so a name that spells a prop but resolves to a parameter, a `for`
+binding or a `catch` binding declares nothing — and **that half was not re-measured
+locally**: the `PASSES` above was read on the pre-narrowing arm. The two-sided ratchet is
+the complete check for these two entries, because a listed-but-passing entry and a new
+failure both fail, so the verdict here is CI's rather than a local reconstruction's. The
+count of entries `client` does not carry was read as 15 while the file held 13; it is
+recomputed above rather than decremented, because the partition line is gated and this
+sentence is not.
 
 The `client-dev` target is the `client` target with `dev: true`. It is a
 separate ratchet because `dev` gates 18 client codegen files plus the CSS
--- a/compatibility/known-failures.client-dev.json
+++ b/compatibility/known-failures.client-dev.json
@@ -2,8 +2,6 @@
 	"chatgpt-web/src/lib/ChatCompletionResponse.svelte",
 	"ha-fusion/src/lib/Modal/VisibilityConfig/Index.svelte",
 	"headscale-ui/src/lib/settings/ServerSettings.svelte",
-	"huly/plugins/process-resources/src/components/settings/ProcessAttributeEditor.svelte",
-	"huly/plugins/tracker-resources/src/components/issues/move/SelectReplacement.svelte",
 	"huly/plugins/workbench-resources/src/components/HelpAndSupport.svelte",
 	"immich/web/src/lib/components/asset-viewer/ActivityViewer.svelte",
 	"immich/web/src/lib/components/timeline/Timeline.svelte",
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -521,6 +521,10 @@ pub(crate) fn transform_client(
 ) -> Result<CodegenResult, TransformError> {
     use crate::compiler::phases::phase3_transform::client::visitors::fragment::fragment;
 
+    // Set where the instance pass SAW a prop member write, which is a different
+    // question from whether it could wrap one.
+    let mut saw_prop_member_mutation = false;
+
     // Create initial node (anchor) for the transformation
     let initial_node = b::id("$$anchor");
 
@@ -756,6 +760,7 @@ pub(crate) fn transform_client(
             split_top_level_declarations,
             retained_instance,
             body_projection,
+            &mut saw_prop_member_mutation,
         );
         if let Some(comment) = repeated_projection_comment
             && let Some(start) = transformed.find(&comment)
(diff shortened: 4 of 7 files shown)


baseballyama:
## Blast radius on the 38 client / client-dev ratchet entries — one arm measured, attribution pending

Arm identity, before the numbers: napi built at `b2634d43b`, tree confirmed from the build's own
`Compiling rsvelte_core (/Users/.../wt-main/crates/rsvelte_core)` line, sha256 `ec564195…` —
**distinct** from the #4184 arm `be9ab702…`, which is the check that separates "two arms" from
"two labels on one binary".

Comparison is JS **and** CSS. `targets.mjs` sets `css: true` on both `client` and `client-dev`, so
a JS-only comparator would have been *looser* than the gate — the direction that manufactures a
false retire.

| verdict | n |
|---|---|
| `PASSES` (js equal after oxfmt, css equal) | **2** |
| `js=MISMATCH css=EQUAL` | 32 |
| `ORACLE-THREW` | 4 — **unmeasured**, see below |

The two:

```
client-dev  huly/plugins/process-resources/src/components/settings/ProcessAttributeEditor.svelte
client-dev  huly/plugins/tracker-resources/src/components/issues/move/SelectReplacement.svelte
```

### Two things this does NOT yet say

**Attribution.** Whether those two pass *because of this fix* or were already stale on `main`
needs the base arm, which is building. A pass on this branch is sufficient for the two-sided
ratchet either way, but "this PR retires 2" and "2 were stale" are different claims and only the
second arm separates them. No ratchet JSON is touched here until it does.

**The four `ORACLE-THREW`.** All four are `.svelte.(js|ts)` and my harness called `compile` where
they need `compileModule`. That is a defect in the instrument, not a property of the entries —
they are `UNMEASURED`, not zero. Fixed and re-run with the base arm.

### A cross-check that fell out of it

`huly/…/SelectAvatarPopup.svelte` reads `js=MISMATCH` on this arm and **byte-equal** on #4184's.
That is the right answer in both directions — this branch is cut from `main` and carries none of
#4184's `$.assign` work — so the two measurements corroborate rather than contradict each other.

### An instrument bug caught before it produced a number

The first base-arm run staged the wrong binary. The waiter polled `pgrep` for cargo, and `pgrep`
ran *before* cargo launched (the command begins with a `git checkout`), so the loop fell straight
through and copied the **previous** build as "base". Had it run, both arms would have been the
same binary, every entry would have read identically, and the honest-looking conclusion would
have been "already stale". It now blocks on the build's own `Finished` line rather than on a
process being visible — a process not being there yet and a process being done look the same to
`pgrep`.

baseballyama:
## Two corrections and one closed gap

**The base arm is `c1af73536`, and that is NOT `main`.** `main` is `295d9621b` — #4185 merged
first and #4183 after it, and I read the pair in the wrong order. The comparison is unaffected
and the reason is measurable rather than asserted: `git diff --name-only c1af73536 295d9621b --
'crates/**/src/**'` returns **0**. #4183 touches `scripts/compat-lsp/`, docs and tests only, so
it cannot reach compiler output. Both arms still share `c1af73536` as their merge base, which is
what the commit message and the ratchet prose claim; only the label "main" was wrong.

**The four `ORACLE-THREW` entries are now measured, and they change nothing.** They are
`.svelte.(js|ts)`, and `compile.mjs:94` type-strips those with esbuild before either compiler
sees them — a raw-TS `Unexpected token` was my harness, not the entry. With the strip in place:

```
live-units 38 / dead-units 0 / total 38
client-dev  huly/…/settings/ProcessAttributeEditor.svelte   PASSES
client-dev  huly/…/issues/move/SelectReplacement.svelte     PASSES
```

Same two, and the four newly-live entries do not pass on this arm. Recording it because a closed
gap that moves the answer by zero is still a result — the alternative is a blank, and a blank and
a zero are the same pixel.

The instrument now prints `live-units / dead-units / total` beside every run. That line is
rsvelte-72's, from the same class landing on their side of the same day: their sweep folded an
oracle error into a hash and counted it as agreement, so it reported `MOVED = 0` on a population
where four carrier repositories had **zero files checked out**. The two halves are worth stating
together — **the same defect fabricated a zero on one instrument and produced an honest blank on
the other**, and only the blank is recoverable, because nothing about a fabricated zero tells you
to look again.

baseballyama:
## Diagnosis: the latch asks a name question where upstream asks a scope question

CI is right and this PR is wrong. The five NEW failures plus the mutation-fuzz NEW are **one cause**, and I have a witness for each.

The latch I added reads:

```rust
if let Some(root) = self.target_root_name(&assignment.left)
    && self.prop(&root).is_some()            // a NAME lookup against a prop-name list
    && self.source_has_member_write(&root)
{ self.saw_prop_member_mutation.set(true); }
```

Upstream (`3-transform/client/visitors/shared/utils.js:390-406`) asks something else:

```js
const name = object(left);
const binding = context.state.scope.get(name.name);            // a SCOPE lookup
if (binding?.kind !== 'prop' && binding?.kind !== 'bindable_prop') return expression;
state.analysis.needs_mutation_validation = true;
```

So a member write to a **local that merely shares a prop's name** sets my latch and sets nothing upstream. Confirmed against the oracle: official declares `$.create_ownership_validator` in **none** of the five files, and emits **zero** `$$ownership_validator.mutation` calls in any of them.

## The five witnesses, and the shadowing construct in each

| file | mode | shadowing construct |
|---|---|---|
| `huly/…/filter/FilterTypePopup.svelte` | legacy | `for (const filter of filters)`, and `const filter: Filter = {…}` inside `setNestedFilter` — prop is `export let filter` |
| `kener/…/alerts/[alert_config_id]/+page.svelte` | **runes** | `const data: Record<string, unknown> = {…}` inside a handler — prop is `let { data } = $props()` |
| `open-webui/…/common/RichTextInput.svelte` | legacy | `({ editor }) => …` destructured parameter — prop is `export let editor` |
| `pattern/issues/prop-shadowed-by-local-in-instance-script.svelte` | legacy | both of the above, deliberately |
| `svelte.dev/…/Output/Viewer.svelte` | **runes** | `let error = event.value;` in an arrow body — prop is `error = $bindable()` |

Four distinct constructs — a `for…of` binding, a plain `const` in a function body, a destructured arrow parameter, a `let` in an arrow body — across **both** legacy and runes, and in runes it shadows a `$bindable()` prop, where upstream's `binding.kind` guard would otherwise have passed.

I checked the alternatives before settling on this: it is not a positional difference (official emits the declaration nowhere, so there is nothing to be out of position with), and it is not `svelte-ignore ownership_invalid_mutation` (zero occurrences in either runes file). The upstream call sites in `AssignmentExpression.js:27` and `UpdateExpression.js:53` are unguarded on runes, so runes-vs-legacy is not the discriminator either — the scope lookup is.

## Why this PR's own test could not catch it

`ownership_validator_unbuildable_path.rs` has 24 cells and **every one expects `decl=1`**. There is no cell in which the validator must be *absent*, so no cell can report an over-declaration. That is the shape this repo already records — a grid assembled from the cells a defect breaks has none left to regress — and it applies to the grid I wrote to demonstrate the fix.

The fix therefore has two parts, and the test half comes first: add cells for the four constructs above with `decl=0` expected. Those are correct on `main` as well as on the fix, so they are meaningful independently of this change rather than being a restatement of it.

baseballyama:
## Refinement: it is not a scope lookup this code is missing — an absent key spells two things

My previous comment named the divergence from upstream correctly (upstream asks a scope question) but proposed the wrong repair, because rsvelte's collector walks the **generated** program while the shadow knowledge lives in the **source**-side scan. `props_transforms.rs` already has it — `is_shadowed_by_function_param`, and the comment at `props_transforms.rs:26` about positions that "declare a new local that shadows a like-named prop". So those five names are correctly **absent** from `PropMutationSites`.

The actual defect is the predicate I wrote on top of that:

```rust
fn source_has_member_write(&self, name: &str) -> bool {
    self.sites.iter().find(|(candidate, _, _)| candidate == name)
        .is_none_or(|(_, _, sites)| sites.has_member_write())
}
```

`is_none_or` means **absent ⇒ true**, and I chose that default deliberately, to cover the case the PR exists for: a member write whose computed key the pass declines to spell is not recorded as a site, and upstream still latches. But "absent from `sites`" also means *shadowed*, and it also means *no prop member write at all*. Three conditions, one spelling, and I picked the default that serves one of them.

This is the failure mode this repo already records under **"A missing key can spell two things, and the conservative default reads the wrong one"** — where `None` meant either "nothing renders this snippet" (an empty ancestor set, which is knowledge) or "a callee that could not be read" (ignorance), and conservatism applied to the first is a silent over-match. Same shape, and the same direction: an empty set is knowledge, not ignorance.

So neither `is_none_or` nor `is_some_and` is right — swapping it would restore these five and re-break the unspellable-path case the PR is for. The two have to stop sharing a spelling: the source scan must record that a prop member write **occurred** even when the path is unspellable, as a state distinct from "no site", after which the latch can require presence. That mirrors upstream, which latches *before* it builds the path (`shared/utils.js:406`) and so never has to encode "I saw one but could not spell it" as an absence.

The test cells stand as described — the four shadowing constructs at `decl=0`, which are correct on `main` independently of this fix — and they need a fifth companion: an unspellable computed key at `decl=1`, which is the cell that fails if the repair is done by flipping the predicate instead of splitting the state.

baseballyama:
Rebased onto `9693010f2` and the latch narrowed to match upstream's second condition.

## The defect the first version left behind

Upstream reaches the mutation validator through `scope.get(name)` (`shared/utils.js:396`), so a name that **spells** a prop but resolves to something else declares nothing. This pass asked only whether the name matched a prop, so `list.forEach((p) => { p.x = 1 })` latched on the *parameter* and emitted a `$$ownership_validator` preamble official does not — five NEW corpus failures, across four syntactic forms, in both legacy and runes.

## Expectations generated from the oracle, not inferred from its output

18 new cells: 7 shadowing forms × legacy/runes, plus two controls per mode.

```
before:  14 of 18 DIFF   (official decl=0 wrap=0, rsvelte decl=1 wrap=0)
         4 controls pass  (CONTROL-write decl=1, CONTROL-none decl=0, both modes)
         existing 24-cell grid passes

after:   42 of 42 pass
```

`wrap=0` on both sides in every failing cell, so the latch was firing while the wrap never happened — the preamble was the whole difference. Two of the eighteen must stay at `decl=1`, which is what lets the grid report an over-correction; the existing 24 cells could not, because every one of them expects `decl=1`.

## The fix asks the binder

The generated instance body parses as a top-level statement list, so the props are the root scope's bindings and a shadow is any other scope — the same test `state_pipeline_ast.rs:267` already makes. Adding `for (const p of …)`, `catch (p)` and a block-scoped `let p` to `is_shadowed_by_function_param` was the other option and was not taken: that helper explicitly declines `for`, `catch`, `if`, `while` and `switch`, and extending it from the shapes that happened to break is a work log rather than a partition.

## What is measured and what is not

| | |
|---|---|
| `--lib` + 13 ownership / prop-mutation targets | 14 suites, 1958 lib tests, **0 failed** (`Running unittests src/lib.rs` present) |
| `cargo clippy --release -p rsvelte_core --lib --tests --all-features -D warnings` | EXIT=0 |
| `corpus:known-failures-check` / `test:known-failures-md-check` | EXIT=0, EXIT=0 |
| the two `huly` entries this PR retires, **after** the narrowing | **UNMEASURED locally** |

The `PASSES` reading for those two was taken on the pre-narrowing arm. The two-sided ratchet is the complete check for them — a listed-but-passing entry and a new failure both fail CI — so the verdict is the gate's rather than a local reconstruction's, and the doc says so.

One correction landed with the rebase: the prose read `15 JS entries that client does not carry` while the file held 13. It is now recomputed (11 against 22) rather than decremented — the partition line is gated by `known-failures-md-check`, that sentence is not, and the two rot separately.

PR #4243, opened 2026-09-03
chore(lsp-gate): fill the mechanism sidecar, and wire the checker that reads it

baseballyama:
The mechanism sidecar shipped complete and empty in #4183: 23,746 ratchet
entries, **0** with a mechanism set, 142 labels declared, `lsp-mechanisms-check.mjs`
exiting 1. Its own header says it cannot be filled by hand — only
`merge-current.mjs` over one run's complete 17-artifact set can write it — so it
was waiting on an input, not on a decision.

## The input, and why it is the right tree

`Corpus Compat` run **33681523655** (`2026-09-02T20:48Z`, head `415173551`,
conclusion `success`), 17 unexpired LSP artifacts: `lsp-current-corpus-0..15`
plus `lsp-current-fixtures`.

Its tree carries the same blobs as `main` for all three files that decide the
measurement, checked against the *current* `main` (`46f07b412`), not the one
that was current when the run started:

```
                                     415173551                                 46f07b412
lsp-known-failures.json   778ef30ed67306e9b81d2354c819106b17ad48e4   ...same
mechanism.mjs             b0a78b26cf220dc8da9a892b155ebb087e15906f   ...same
merge-current.mjs         4dfaf257e488e6ff7d66e39dc6c020f808bd9a0f   ...same
git merge-base --is-ancestor 415173551 46f07b412 → YES
```

So the id set cannot have moved and re-baselining cannot FALSE-SHRINK.

**Positive control before the write**, `merge-current.mjs` with no `--update-baseline`:

```
[lsp-merge] 23746 current, 0 new, 0 stale
```

**And after the write, the ratchet is byte-identical to `main`:**

```
git diff --stat 46f07b412 -- compatibility/lsp-known-failures.json   → empty
```

One note for whoever runs this next: `readArtifacts` walks the directory for
**every** `.json`, so a plain `gh run download` (19 artifacts — the two corpus
reports carry 6 more JSONs) fails the `expected exactly 17` assertion. The
directory has to hold the `lsp-current-*` artifacts alone.

## What the fill measures

23,746 entries, all with a set, mean **7.35** labels each (1–28; 3,610 entries
carry exactly one). The structural claim in `KNOWN-FAILURES.md` was established
on 151 entries of one shard and survives the change of population:

| label | appears on | sole label on |
|---|---|---|
| `rsvelte-empty` | 10602 | 148 |
| `completion-item-set-extra-ts` | 7262 | **0** |
| `completion-text-edit-range-end` | 7262 | **0** |
| `completion-command-presence-rsvelte-only` | 7260 | **0** |
| `rsvelte-empty-import-only` | 6720 | 982 |
| `official-empty` | 6682 | 4 |

10 of the 72 used labels are ever an entry's sole label (the shard read 9 of 62),
and repairing any of the large completion labels still removes **no** entry.

**The sidecar declares 142 labels and the artifacts use 72.** The merge only ever
adds a label, so the other **70** are carried by zero entries — there is nothing
behind them to establish a terminal for. Sizing the remaining terminal work at
142 counts a vocabulary rather than a population. Neither 72 nor 142 is the
number of decisions that would move the ratchet either: a greedy union over the
used labels touches all 23,746 entries with **12** labels, and two of them reach
75.2%.

`unclassified` is 2,120 entries and the sole label on 2,116 — the `differential:`
and `expected:` half, spelled rather than blank, because the classifier runs on
the corpus branch only.

Every `terminal` is still `null`, which the checker reports and does not pass:

```
[lsp-mechanisms-check] 23746 ratchet entries, 23746 with a mechanism set,
142 labels declared, 0 entries attributable today
(23746 blocked by an unclassified or unestablished terminal)   EXIT 0
```

Establishing those terminals is the next step and is not in this PR.

## The checker was wired to nothing

`scripts/ci/lsp-mechanisms-check.mjs` had a `package.json` script and appeared in
**no** workflow, and `known-failures-md-check` does not cover it — that checker
reads a sidecar as one slug per entry (`for (const [id, slug] of ...)`), where
this one holds a set, so `lsp-mechanisms.json` is deliberately outside its
`MECHANISMS` list. Nothing would have noticed the file emptying again. It now
runs beside `attribution-check` in `ci.yml`, which needs no submodules and no
build.

Wiring it is a judgement call rather than part of the fill — drop that hunk if
you would rather it landed separately. Before this PR the checker exits 1, so it
could not have been wired without the fill in the same commit.

## Cost, stated rather than buried

`compatibility/lsp-mechanisms.json` goes from **11 KB to 9.9 MB** (+222,139
lines). That is the design merged in #4183 — one line per ratchet entry — and it
is what a per-entry map of 23,746 entries costs; it is not a choice this PR
makes, but it is a number that belongs in the review rather than in the diff
stat alone.

## Gates

```
node scripts/compat-lsp/merge-current.mjs <17 artifacts>     23746 current, 0 new, 0 stale
node scripts/ci/lsp-mechanisms-check.mjs                     EXIT 0 (was 1)
node scripts/dev/test-lsp-mechanisms-check.mjs               all cases pass
node scripts/compat-corpus/known-failures-md-check.mjs       EXIT 0
node scripts/ci/workflow-trigger-guard.mjs                   20 workflows, no violations
node scripts/ci/step-environment-guard.mjs                   EXIT 0
node scripts/release/check-core-consumer-changesets.mjs      EXIT 0
```

`lsp-ratchet=false` for this change set, so the 950-job-minute `lsp-corpus` job
does not run on this PR; the other Corpus Compat jobs read `true`, which is the
filter's deliberate over-approximation for a non-`crates/` path.

No ratchet entry is added or dropped.


Kimenzo2:
God I love Svelte.. Can't wait for V1

PR #4269, opened 2026-09-03
perf(compiler): a one-byte needle does not need CharSearcher

baseballyama:
Stacked on #4260 (`Substring`) and #4268 (`json_field`) — merge those first.

## What

`str::find`/`contains` with a `char` needle routes through `CharSearcher`, which
a `sample` profile of a client compile puts at **2.37% self time inside
`compile()`** — the same "the setup dominates the search" shape #4260 answers for
string needles, one needle-type over.

`Substring` gains `find_byte` / `rfind_byte` / `has_byte` over `memchr`, and the
conversion covers:

- **212** single-ASCII-char call sites (`.find('\n')` → `.find_byte(b'\n')`)
- **43** string-literal needles in the parts of the tree #4260 did not reach
  (phase 1, phase 2, `print/`, `server/`)

across 62 files. All 33 distinct needles are ASCII; a byte-level match of valid
UTF-8 cannot land inside a multi-byte sequence, so every offset is unchanged.

## Why this is safe to do mechanically

The trait is implemented for `str` alone, so a receiver that is not a string
fails to compile rather than silently changing meaning. That is not a
hypothetical: the regex hit **15 set-membership sites** (`IndexSet`,
`HashSet<String>`) where `.contains("click")` asks a different question, and
every one of them was an `E0599` rather than a shipped defect. They are reverted.

`#[cfg(test)]` regions are excluded — test code is cold, and converting it only
adds diff noise.

## Verification

Both arms built from one tree, artifact hashes differ
(`c31e2fe9…` / `54d39e14…`).

**Content equality** — `corpus_hash --map` over `compatibility/sources`, all four
targets:

| target | rows | distinct keys | live units | DIFF |
|---|---|---|---|---|
| client | 33,897 | 33,897 | 32,654 | **0** |
| server | 33,897 | 33,897 | 32,654 | **0** |
| client-dev | 33,897 | 33,897 | 32,654 | **0** |
| server-dev | 33,897 | 33,897 | 32,654 | **0** |

`rows == distinct keys` on every target, so no key collapsed (corpus ids contain
spaces; the reader splits from the right). The 1,243 units both arms reject are
excluded from `live`. Positive control: the same comparator run **client against
server** reports 32,655 of 33,897 differing, so a non-zero DIFF is reachable —
the zeros above are a measurement, not a dead instrument.

Gate results and the paired CPU measurement follow in a comment.

### DIFF
--- a/.changeset/fast-substring.md
+++ b/.changeset/fast-substring.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Search phase-3's generated-code probes with `memmem` through a shared `Substring` trait
--- a/.changeset/json-field-linear-scan.md
+++ b/.changeset/json-field-linear-scan.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Read CSS AST fields by linear scan instead of hashing the key, 3.0-4.3% on every surface
--- a/.changeset/json-field-wide.md
+++ b/.changeset/json-field-wide.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Read every compiler-side JSON AST field by linear scan, a further 1.2-1.5% on each surface
--- a/.changeset/substring-byte-needles.md
+++ b/.changeset/substring-byte-needles.md
@@ -1,0 +1,12 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Search for a single ASCII byte with `memchr` rather than `CharSearcher`
+
+`str::find`/`contains` with a `char` needle routes through `CharSearcher`, which
+a profile puts at 2.37% of a client compile — the same shape as the `StrSearcher`
+setup cost `find_sub` already avoids. 212 single-ASCII-char call sites now take
+`find_byte`/`has_byte`, and the 43 remaining string-literal needles outside
+`3_transform/{client,shared}` take `find_sub`/`has_sub`. Offsets are unchanged:
+a byte-level match of valid UTF-8 cannot land inside a multi-byte sequence.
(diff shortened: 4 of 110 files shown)


baseballyama:
Closing: **measured, and it does nothing.** Recording why, because the profile line that motivated it is real and the next person will read it the same way.

## The measurement

12 ABBA pairs per target, both arms from one tree with differing artifact hashes, single-threaded, indexers stopped, `sink` identical across every pair:

| target | A median | B median | ratio | B faster | sign p |
|---|---|---|---|---|---|
| client | 933.8 ms | 944.8 ms | 0.9926 | 5/12 | 0.774 |
| server | 794.0 ms | 773.7 ms | 1.0200 | 8/12 | 0.388 |
| client-dev | 1095.0 ms | 1113.8 ms | 0.9903 | 3/12 | 0.146 |
| server-dev | 804.8 ms | 797.0 ms | 1.0055 | 7/12 | 0.774 |

Two of four point the wrong way and no target reaches significance. Correctness was fine — 0 divergences over 32,654 live units on all four targets, all gates green — so this is a clean "no effect", not a broken arm.

## Why

`CharSearcher::next_match` **already calls memchr**:

```rust
// library/core/src/str/pattern.rs:439
if let Some(index) = memchr::memchr(last_byte, bytes) {
```

So the 2.37% the profile attributes to `CharSearcher::next_match` is *already* memchr time, and `.find('\n')` → `.find_byte(b'\n')` swaps `core::slice::memchr` for the `memchr` crate — the same algorithm, differing only in SIMD width, which measures null at these haystack lengths.

This is not the same situation as #4260. There the symbol was `StrSearcher::new`, which is *setup* std does not avoid: a two-way searcher with a critical-factorization precompute, rebuilt per call. A symbol named after a search is worth converting only when the time in it is setup you can delete, and `CharSearcher` has none.

**The generalisable form: a profile symbol can already be the optimization you are about to apply.** The check costs one `grep` of `library/core` and settles it before an arm is built.

PR #4273, opened 2026-09-03
perf(compiler): the other 1,813 JSON field reads, 1.2-1.5% on every surface

baseballyama:
Re-opening #4268, which GitHub closed when its base branch `perf/json-field` was deleted on #4266's merge. Same commit, rebased onto `main`; #4268's own CI was **green on all 44 checks** before it was closed.

## What

#4266 converted `3_transform/css.rs`'s 543 `Value::get("…")` calls to the linear `field()` lookup. This is the other 1,813, across 53 files under `crates/rsvelte_core/src/compiler/` and `src/ast/` — largest are `client/visitors/expression_converter.rs` (351), `print/helpers.rs` (177), `2_analyze/css/analyze.rs` (146), `client/visitors/shared/utils.rs` (134), `each_block.rs` (97), `await_block.rs` (88).

`preserve_order` makes `Value::Object` an `IndexMap<String, Value, RandomState>`, so `get("type")` costs a SipHash of the key plus a table probe. AST nodes carry 4-6 keys, and a linear scan over `Map::iter()` beats hashing at that width.

## Rebase note

Four files conflicted with #4260, which converted the same lines to `has_sub`/`find_sub`. Resolved by taking this commit's side and re-applying #4260's conversion; the resulting set of converted call sites is **byte-identical to `main`'s** in all four files (checked per file, not per line count).

### DIFF
--- a/.changeset/json-field-wide.md
+++ b/.changeset/json-field-wide.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Read every compiler-side JSON AST field by linear scan, a further 1.2-1.5% on each surface
--- a/crates/rsvelte_core/src/ast/js.rs
+++ b/crates/rsvelte_core/src/ast/js.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use super::arena::{IdRange, JsNodeId};
 use super::span::SourceLocation;
 use super::typed_expr::{JsNode, Loc, SourcePosition};
+use crate::compiler::phases::phase3_transform::shared::json_field::Field;
 
 /// Wrapper for a typed `JsNode` with lazy JSON cache.
 /// The cache is only populated when `as_json()` is first called (during Phase 2/3),
@@ -673,7 +674,7 @@ impl<'de> Deserialize<'de> for Expression<'_> {
         // must return a deserialize Error here rather than letting
         // `JsNode::from_value` degrade a typeless object to `Null`.
         let is_node = value
-            .get("type")
+            .field("type")
             .and_then(|t| t.as_str())
             .is_some_and(|t| !t.is_empty());
         if !is_node {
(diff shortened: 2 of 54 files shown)


baseballyama:
## Measured, both arms

The fix restores `identifier_name` as a parameter of `is_non_proxy_node_type` (with the
`matches!` kept inline, so the cross-module call this branch removed stays removed) and
deletes the two caller-side `|| (Identifier && "undefined")` arms. Plus `cargo fmt`:
five import-ordering diffs across four files.

Two NAPI arms, built from this worktree (`Compiling` lines confirm the tree), distinct
by `sha256`:

| arm | sha256 (first 16) |
|---|---|
| `68e12b178` (before) | `b14000d5e9d18dc5` |
| `d90fdf31f` (after)  | `fc99d1585ff5da69` |

**Grid** — 12 initial-value kinds × 3 hosts × `dev ∈ {false, true}` = 72 cells, against
`submodules/svelte/…/src/compiler/index.js`. Live/dead is printed so a both-throw pair
cannot read as agreement.

```
68e12b178   live 72 (EQ 68 / DIFF 4), dead 0
d90fdf31f   live 72 (EQ 72 / DIFF 0), dead 0
```

All four DIFF cells are the `undefined` value, in both `dev` modes:

```
O| $.set(s, a());
R| $.set(s, a(), true);
```

**Corpus carrier**, through the gate's own normalization (`flattenTemplateHoles` + oxfmt +
`stripBlankLines`) — a strict reconstruction, so a `match` here is a match under the gate:

```
68e12b178   client MISMATCH / client-dev MISMATCH   ($.set(previousAssetId, assetId(), true))
d90fdf31f   client match    / client-dev match
```

The instrument is live in the same run: it reports `MISMATCH` for `Point.svelte` and the
other listed ids.

### What is measured and what is argued

Which of the four call sites carries each red is **not** measured, and is not decidable
from the source: all four gate on `BindingKind::Prop`. That the two unrestored ones were
the two that regressed is a **structural argument from the patch**, read independently by
two of us; that restoring the parameter reaches all four is likewise structural. The
measurement above is that the defect is gone on every cell the grid can express — and the
grid is shown able to express it, because the before-arm reports four.

`cargo clippy -p rsvelte_core --lib -- -D warnings` exits 0 (`Compiling rsvelte_core` present
in the log; the crate was really checked).

PR #4274, opened 2026-09-03
Please release a new version of C API

alanjds:
Per https://github.com/baseballyama/rsvelte/blob/main/crates/rsvelte_capi/README.md, to release the C API binaries the crate version should be bumped & merged.

As of today, only 2 CAPI releases occurred, and they are very old. Could you please release newer?

Context: I am developing a package that compiles `.svelte` without needing Node.js tooling. The cleanest option I got is fetching the `.so` from here. However the released version have bugs that I cannot sidestep without self-compiling the `.so`, and I want not to make the developer using the package to have the Rust toolchain installed.

And also as a question: Why is it not being released automatically, as it surely compiles?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2517,7 +2517,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_capi"
-version = "0.1.1"
+version = "0.2.0"
 dependencies = [
  "cbindgen",
  "rsvelte_core",
--- a/crates/rsvelte_capi/Cargo.toml
+++ b/crates/rsvelte_capi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsvelte_capi"
-version = "0.11.1"
+version = "0.2.0"
 edition = "2024"
 rust-version = "1.95"
 description = "C ABI bindings for the rsvelte Svelte 5 compiler — drive the compiler from C, Go, Python, Ruby, PHP, Zig, Java (JDK 22+ FFM), or any language with a C FFI"
--- a/crates/rsvelte_capi/Cargo.toml
+++ b/crates/rsvelte_capi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsvelte_capi"
-version = "0.1.1"
+version = "0.11.1"
 edition = "2024"
 rust-version = "1.95"
 description = "C ABI bindings for the rsvelte Svelte 5 compiler — drive the compiler from C, Go, Python, Ruby, PHP, Zig, Java (JDK 22+ FFM), or any language with a C FFI"
--- a/crates/rsvelte_capi/README.md
+++ b/crates/rsvelte_capi/README.md
@@ -41,7 +41,7 @@ library, the static archive (where applicable), and `include/rsvelte.h`.
 #   darwin-arm64, darwin-x64,
 #   linux-x64-gnu, linux-arm64-gnu,
 #   win32-x64-msvc
-VERSION=0.1.1
+VERSION=0.2.0
 TRIPLE=darwin-arm64
 
 curl -L -o rsvelte_capi.tar.gz \
@@ -90,7 +90,7 @@ git dependency in the meantime:
 
 ```toml
 [dependencies]
-rsvelte_capi = { git = "https://github.com/baseballyama/rsvelte", tag = "capi-v0.1.1" }
+rsvelte_capi = { git = "https://github.com/baseballyama/rsvelte", tag = "capi-v0.2.0" }
 ```
 
 ## API at a glance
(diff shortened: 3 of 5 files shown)


baseballyama:
Thanks for the report — and you're right that the released binaries are stale.

I've pushed one commit to this branch adjusting the version, with the reasoning
measured rather than assumed.

**The C ABI does not track `@rsvelte/compiler`.** `crates/rsvelte_capi/README.md`
says the tag scheme (`capi-vMAJOR.MINOR.PATCH`) is *intentionally independent*
from the npm/changeset pipeline, so `0.11.1` would tie it to a version line it is
deliberately not on.

**What actually changed since `capi-v0.1.1`**, from the public header rather than
from the commit log:

```
old exported rsvelte_* functions   7
new                                9
shared                             7
ADDED    rsvelte_compile_with_callbacks
         rsvelte_compile_module_with_callbacks
REMOVED  (none)
```

Two additions, nothing removed — a backwards-compatible surface change, so
**`0.2.0`**. `Cargo.lock` is re-resolved in the same commit so `--locked` builds
agree; the version bump alone would have left it at `0.1.1`.

Once this merges I'll push the `capi-v0.2.0` tag, which is what actually builds
and attaches the five-triple archives.

One thing worth flagging for your use case: the archives are built from `main` at
tag time, so they carry whatever the compiler is at that moment. If you hit a
specific miscompile, an issue with the `.svelte` source that reproduces it is
more useful than a version bump — the compiler is under active work and the bug
you are sidestepping may already be fixed, or may not be.

baseballyama:
@alanjds Sorry! I currently have a list of compiler bugs that I'm having AI work through, and it ended up posting a comment automatically.

I'll take a look at this. I don't normally use the C ABI, so I may not have noticed that the release process was broken.

baseballyama:
> I am developing a package that compiles .svelte without needing Node.js tooling. 

Wow, this is a really interesting project! There are definitely times when I don't want to use Node.js on the backend. Being able to use Svelte on the frontend without having to use Node.js is awesome!!!

baseballyama:
Thanks for the bump — and sorry the prebuilt C ABI has been sitting at 0.1.1 since May.

**On the version number.** I changed 0.11.1 to 0.2.0: the C ABI has its own `capi-vX.Y.Z` scheme, deliberately independent of `rsvelte_core` / the npm `@rsvelte/compiler` (documented under "Cutting a release" in the crate README). Its only releases so far are `capi-v0.1.0` and `capi-v0.1.1`, so 0.2.0 is the next one.

Your reading was reasonable, though, because `rsvelte.h` said the wrong thing: the doc on `rsvelte_version()` claimed it "matches the `rsvelte_core` crate version", while the implementation returns this crate's own `CARGO_PKG_VERSION`. I've pushed a commit to this branch fixing that doc (and regenerating the header), plus the README references that still pointed at 0.1.1.

**What 0.2.0 gives you over 0.1.1**, since the release notes are generated and won't say:

- Added: `rsvelte_compile_with_callbacks` and `rsvelte_compile_module_with_callbacks`, with `RsvelteStr` / `RsvelteCssHashInput` / `RsvelteCallbacks`, which resolve the `cssHash` and `warningFilter` options that can't cross the JSON boundary. Nothing was removed and `RsvelteBuf`'s layout is unchanged, so existing callers keep working.
- Behaviour change: an unrecognised key in `opts_json` is now rejected with `Unrecognised compiler option <key>` (matching the official compiler). 0.1.1 ignored unknown keys silently, so if you pass extra fields, drop them.
- `rust-version` moved 1.90 → 1.95, which only matters if you depend on the crate as a git dependency rather than using the prebuilt archives.
- Everything else is ~3 months of compiler fixes in `rsvelte_core`.

**Why it isn't automatic.** `release-capi.yml` triggers on a pushed `capi-v*.*.*` tag, not on a merge — a version bump landing on `main` publishes nothing by itself. The tag is pushed by hand after merge (the workflow then refuses to build if the tag and `Cargo.toml` disagree). So the flow is: merge this, then `git tag capi-v0.2.0 && git push origin capi-v0.2.0`, and the five-triple matrix builds the Release. Automating the tag on a version change is worth doing — I'll open an issue for it.

If you hit anything in 0.2.0 that still blocks the Node-free path you're building, please open an issue; the C ABI surface is easier to extend than to guess at.

baseballyama:
Filed the tag-automation follow-up as #4285.

alanjds:
Thanks a lot. I will revisit this soon. Planning to publish to PyPI when stabilized.

> On the version number. I changed 0.11.1 to 0.2.0: the C ABI has its own capi-vX.Y.Z scheme, deliberately independent of rsvelte_core / the npm @rsvelte/compiler 

Out of curiosity, why is it segregated? Doesn't it gets confusing this way?

At least my LLM got confused, thinking that npm package was a lot newer just by the version numbers. This time it has correct, but next time it will see 0.2 vs 0.11.1 and got to the same conclusion.

PR #4313, opened 2026-09-04
docs(agents): five rows from one integration window — a stacked PR's silent close, and four instrument failures

baseballyama:
Five rows, all from one integration window on 2026-09-04/05, all measured. Each was found by a mechanism other than the one it warns about, which is the property that made them worth writing down.

## 1. A stacked PR is CLOSED by its parent's merge, and its stale checks still read green

Merging `#4306` at `14:07:09Z` deleted its head branch; `#4308`, based on that branch, has `closedAt` `14:07:13Z`. Four seconds. GitHub does not conflict such a PR and does not annotate it — it closes it.

Twenty minutes later its author reported it open and green from `total=39 received=39 completed=11 bad=[]`. That is a **correct** measurement of check-runs still attached to a closed PR. Every check-shaped predicate passes; `state` is the only field that separates them, and a rollup census does not read it.

Recovery has an order, because `gh pr reopen` refuses while the base is missing — the row gives it, including that the parent's pre-merge head sits on no local ref after a squash-merge, and that the refspec needs braces (`"${sha}:refs/heads/x"`; zsh applies `:r` inside double quotes too — the recorded `${MB}:AGENTS.md` hazard reached from the push side, by someone who had cited it that morning).

## 2. Two arms answering `false` is two absences agreeing, not an identity

`assign_present` read `false` on both arms and was reported as "the two arms agree". True of the two values, and satisfied by any pair of arms including two copies of one file. Same shape as a filter that discards on both sides reading as agreement, one level out. The replacement needed no binary: `git merge-base --is-ancestor` per fix per arm, with the negative control that HEAD is not an ancestor of the merge base.

## 3. A control needle written from memory fails the control, and the needle is the fault

A corpus census used `$.template(` as its positive control and got 0; the generated client output spells it `$.from_html(`. When a positive control fails the standing habit is to suspect the instrument — and here that produces a rewrite of working code. The other half: the person declined to report the numbers from the run whose control had failed.

## 4. The strongest carrier is the one most likely to belong to a different mechanism

A mechanism was established by reading code and its input error measured at 19 of 1,310 sites, one direction, 10–64 columns. The carrier chosen to demonstrate an output consequence turned out to be a **third** quantity in the same branch of the same function. A carrier is selected for the *size* of its divergence, so the selection is biased toward whichever mechanism in that region is loudest.

## 5. A sweep that stores hashes cannot be grepped for code, and the ids answer instead

The prescription "grep the stage-1 output" was mine and was wrong: stage 1 stores `id` plus a digest. Run anyway, `grep -c 'assign'` returns **124**, every hit a *path*. Reported as a denominator, `n=124` is the right shape and about filenames. **A zero is suspected and a plausible number is not.** The fix is a second pass over the population, which cost a quarter of one sweep arm and turned the `0` into a real absence (`n=64` carriers, none moved).

## Credit

Rows 2, 3 and 5 come from rsvelte-98's measurements; row 4 from rsvelte-72's. Row 1 is mine, found while recovering #4308.

## Verification

- `git diff --numstat` reports `88 0` — insertions only, the shape that would have caught the `$'` expansion that doubled this file in a previous batch.
- Headings 189 → 194; collision check against `origin/main` returns 0, with an injection control that returns 1 when a known-present heading is added to the candidate set.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` pass (pre-commit hook, 17 crates).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5306,6 +5306,94 @@ one line comment with a trailing attachment kills the instrument instantly, and 
 ablating the code finds it, because ablation moves the code and holds the cells. Where an
 instrument excludes by name, write the control that carries each *sibling* of the name.
 
+
+### A stacked PR is CLOSED by its parent's merge, and its stale checks still read green
+
+Merging a PR deletes its head branch, and GitHub **closes** every PR whose base was that
+branch — it does not conflict them, and it does not annotate them. Measured to the second:
+`#4306` merged at `14:07:09Z`; `#4308`, based on `#4306`'s branch, has `closedAt`
+`14:07:13Z`. Twenty minutes later its author reported it open and green from a check census
+reading `total=39 received=39 completed=11 bad=[]` — a correct measurement of check-runs that
+are still attached to a **closed** PR. Every check-shaped predicate passes: the count is
+plausible, nothing is failing, the head SHA matches the local branch. `state` is the only
+field that separates them, and a rollup census is exactly the instrument that does not read it.
+
+Recovery has an order, because `gh pr reopen` refuses while the base is missing: push the
+parent's pre-merge head back to the deleted branch name, `reopen`, `--base main`, then delete
+the branch again (the PR stays open once its base is `main` — verified). Two cautions from
+doing it. The parent's pre-merge head is on **no** local ref after a squash-merge, so read it
+from `gh pr view <parent> --json headRefOid` and `git fetch origin <sha>`. And brace the
+refspec: `"$sha:refs/heads/x"` in zsh applies the `:r` modifier to `$sha` and pushes to
+`…efs/heads/x`, inside double quotes too — the recorded `${MB}:AGENTS.md` hazard, reached
+from the push side by someone who had cited it that morning.
+
+So a settle predicate over a PR needs `state == "OPEN"` beside its denominator floor, and a
+stacked branch needs its base checked after every parent merge, not only when a check goes red.
+
+### Two arms answering `false` is two absences agreeing, not an identity
+
+An arm probe must discriminate, and the recorded failure is a probe whose answer predates the
+change. A quieter one: a probe that returns the **same negative** from both arms. `assign_present`
+read `false` on the base arm and `false` on the head arm, and was reported as "the two arms
+agree" — which is true of the two values and says nothing about the two binaries, because a
+probe that finds nothing in either is satisfied by any pair of arms, including two copies of one
+file and two unrelated trees.
+
+It has the same shape as a filter that discards on both sides reading as agreement, one level
+out: there the comparison drops the carrier symmetrically, here the probe never had one. The
+replacement was stronger and needed no binary at all — `git merge-base --is-ancestor` for each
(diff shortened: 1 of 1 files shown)


baseballyama:
## 追記コミット `86e23c40a` — 4 節と、この文書自身の `142` の扱い

### 新しい 4 節

- **The same number spelled twice, and only one spelling is gated** — rsvelte-2f が `parse-ast` の節で 1 つずつ注入して測りました。6 つの数字のうちゲートされているのは 2 つ（宣言値と partition 行）だけで、**表の `keys` 列は partition 行と同じ量の 2 つ目の綴りなのにゲートされていません**。既存の「gated な半分と ungated な半分は別々に腐る」は 2 つの*文*の話でしたが、これは 2 つの*綴り*で、しかも隣接した表の中にあります。
- **Move a constant, then grep its value** — ワイヤ形式のバージョンが 3 港（Rust writer / JS reader / エンベロープを手書きするテスト）にあり、3 つ目は grep でしか見つかりません。教訓は「港を数えろ」ではなく「定数を動かしたらその値を grep しろ」— 港は概念で、定数は文字列です。**その grep でも届かない半分**（同じテストがレコードのレイアウトも写しているが、今回は別の型を写していたので当たらなかった）込みで書いてあります。
- **An old arm can fail for a reason that is not the arm** — `unsupported version 9 (expected 10)` で古いアームが陰性対照になりませんでした。ただし壊れた対照が観測としては有用で、そこから「この変更もバージョンを上げる必要がある」が分かっています。
- **A branch checkout is a silent parameter of every edit that follows it** — 私自身の失敗です。`git checkout main` のあと 3 回の編集が `main` に落ち、**すべて成功しました**（アサーションは通り、`numstat` の形も正しい）。捕まえたのは見出し数が `189+5+3=197` の予測に対して `192` だったことで、`192` はこのファイルの数として完全にもっともらしい値です。**予測した数だけが間違いになれる。**

### `142` を数字ごと直さなかった理由

`AGENTS.md` には `142 declared labels` が 2 か所あり、rsvelte-2f が `d168231700` で **141** を実測しました。ただしその 2 か所は**「同じ日に 4 つの数が作業量と読まれた」という出来事の記録**なので、`141` に置き換えるとその日に読まれた数そのものが消えます。

なので数字は残し、**それが読み取りであってツリーの事実ではないこと**、JSON が primary であること、再測定値と revision、そして「4 つのうち 3 つは再導出しても成り立ち 1 つが 1 ずれていた、どれかは散文からは見えない」を足しました。

`71` と `72` はどちらも正しく腐っていません（`labels_used=72` は `unclassified` を含み、terminal が動かすのは 71）。これも rsvelte-2f の実測です。

### zsh の行

`:A` を名指ししていた行を**修飾子の集合**に一般化しました。同じ日に `:r`（push refspec が `…efs/heads/x` になる）と `:s`（`git show` が `…suiteshanism.mjs` を読む）を別々の人が踏んでいます。文字ではなく波括弧を覚える、という形に。

PR #4321, opened 2026-09-04
chore(reports): regenerate the site's compatibility and performance artifacts on main

baseballyama:
Both site artifacts were measured on 2026-09-01/02; the compatibility half predates four days of fixes. Regenerated on `main` (d4c79bb04) from `compatibility/`, which is what `publish-compatibility.mjs` already reads: collect (104/104 sources, 34,839 entries) → compile → verify, plus the fixture suite and the fmt / svelte2tsx / lint sub-reports.

## Compatibility — 34,836 / 34,839 exact (99.99%)

| surface | matched / total | differences |
|---|---|---|
| Parser | 110 / 110 | 0 |
| CSR | 34,839 / 34,839 | 0 |
| SSR | 34,837 / 34,839 | 2 |
| CSR dev | 34,838 / 34,839 | 1 |
| SSR dev | 34,837 / 34,839 | 2 |
| svelte2tsx | 33,815 / 33,817 | 2 |
| lint | 6,765 / 6,765 | 0 |
| fmt | 33,092 / 33,642 | 550 |

Warning parity, compile-error parity and output parseability are **identical to official on all 34,839 entries × 4 targets** (133,944 generated modules all parse). Fixtures 3,592 / 3,592. The residual divergences are the entries already listed in `KNOWN-FAILURES.md`.

### The first run's 8 "regressions" were the instrument

It reported 8 new `css_unused_selector` warning failures on two `4115-*` pattern files. `.corpus-cache/rsvelte.node` was staged **2026-09-01 23:00** — a compiler four days old that predates #4165, the fix those files were added with. Rebuilt from `main`, then confirmed by a discriminating probe on the carrier (official `["css_unused_selector","css_unused_selector"]`, rsvelte the same) *before* re-running. The re-run is green.

## Performance

| surface | multi | single |
|---|---|---|
| client | 16.10x (cv 13.7%) | 3.81x (cv 0.13%) |
| server | 22.56x (cv 10.3%) | 3.99x (cv 0.17%) |
| client-dev | 15.20x (cv 12.2%) | 3.32x (cv 0.73%) |
| server-dev | 20.09x (cv 12.3%) | 3.89x (cv 0.32%) |

Measured on a quiesced box, with the quiet-box reader positive-controlled (a planted `yes` must be the top process) — the first attempt was discarded because Spotlight resumed at up to 313% while indexing the 1 GiB of corpus output the compatibility run had just written.

**These are lower than the same box produced earlier the same day**, and the report's own control says that is a real regression rather than drift: `official` — unchanged code — moves 0.982-1.000 across the four surfaces, while `rsvelte-single` is **5.7-9.2% slower**, at a single-threaded cv of 0.13-0.73%. Both server and client regress, so the cause is on a shared path rather than in the client-only fixes. Being tracked separately; these artifacts report `main` as it is.


baseballyama:
`main` has moved three times since this was measured (`d4c79bb04` → `b860ab511`: #4312, #4316, #4315, #4318, #4317, #4320, #4299). The report is self-describing about it — `commit.rsvelte = d4c79bb0402a98f40790cb9facd57ba77f657480` — so this is a note about which rows are now stale, not a criticism of the artifact.

Diffed every ratchet by **blob id** across the two trees, with the entry counts beside them:

```
ratchets compared                              64
blob-identical                                 63
MOVED                                           1
  known-failures.client-dev.json           1 -> 0
```

The one entry is `svelte/packages/svelte/tests/snapshot/samples/delegated-locally-declared-shadowed/index.svelte`, retired by **#4312** (`3564c1876`, `fix(analyze): a template-local declaration produces one binding record, not two`). So the report's

```
| CSR dev | 34,838 / 34,839 | 1 |
```

row reads **0** on `b860ab511`, and the headline becomes 34,837 / 34,839 → 34,838 / 34,839. Every other surface's residual count is unchanged, because every other ratchet's blob is identical.

## One row is stale in a way the ratchets cannot show

`b860ab511` is #4299, a formatter behaviour fix (`crates/rsvelte_formatter/src/style.rs`, +48). Its ratchet did **not** move — `fmt-known-failures.json` is blob `a2216a65c…` at `d4c79bb04`, at `91ea9e741`, at `702b0095e` and at `b860ab511`, verified against a negative control (`style.rs` reads `c8415d706` vs `ff3031386` across the same pair, so the comparison does distinguish).

But the report's `fmt 33,092 / 33,642 = 550` counts a **different population** from the ratchet's 520 listed keys, so an unchanged ratchet does not entail an unchanged report row. Whether #4299 moves that 550 is **UNMEASURED** here — I did not run the fmt gate. It is the one row I would re-derive before merging; the other seven can be patched from the ratchet diff above.

Related, from rsvelte-72's independent measurement on this same pair of trees: of the 520 fmt ratchet entries, **4 are already byte-equal** (520 staged, 520 listed, 0 missing, byte-differing 4 short of the count), which is a second reason the 550 and the 520 are not the same quantity and should not be reconciled by arithmetic.

## Two things worth keeping from the body

The `.corpus-cache/rsvelte.node` staleness you caught and named — a staged arm four days old producing 8 plausible `css_unused_selector` regressions, killed by a discriminating probe rather than by re-reading — is the arm-identity row in `AGENTS.md` reached from the report side. And recording `commit.rsvelte` in the artifact is what makes this comment cheap: without it, "which tree is this a measurement of" would not be answerable from the file at all.

baseballyama:
Correction to my previous comment — the last paragraph before the closing section is wrong twice, and the ratchet-diff measurement above it is unaffected.

I wrote:

> Related, from rsvelte-72's independent measurement on this same pair of trees: of the 520 fmt ratchet entries, **4 are already byte-equal** (520 staged, 520 listed, 0 missing, byte-differing 4 short of the count)

**The attribution is wrong.** rsvelte-72 reported `staged=520 / byte-differing=0 / MISSING=0` — that is a *control* showing their staged copies are byte-identical to the corpus sources, and it says nothing about entries being byte-equal to the oracle. They did not report a byte-equal count at all.

**The number is wrong too.** The `524 listed / 520 live / 4 already byte-equal` figure is `AGENTS.md`'s, in the region-granularity row (line 3122), and it is a measurement of a tree whose ratchet held **524**. Today's ratchet holds **520** (`git show origin/main:compatibility/fmt-known-failures.json` → array, 520). So the live count on `b860ab511` is not 520 and is not 516 either — it is 520 minus however many are byte-equal today, and the four that were byte-equal then may be among the four entries that have since left the file. **UNMEASURED.**

What survives is the claim the paragraph was supporting, which needed neither: the report's `550` counts fmt-eligible corpus entries and the ratchet's `520` counts listed keys, so an unchanged ratchet blob does not entail an unchanged report row. That still stands and is still UNMEASURED here.

Recording the failure because it is a shape I had not seen named. The figure came out of `AGENTS.md`, and I attached it to the peer who happened to be discussing the same subject in the same hour — **a number's provenance gets overwritten by the most recent conversation about its subject**, and the citation then reads as freshly measured rather than as recalled. `AGENTS.md` already says a citation feels verified because it is offered as the verification; this adds that the *source* can be swapped without the number changing, which leaves nothing in the sentence to check against.

PR #4322, opened 2026-09-04
ci: a concurrency group is keyed by the EVENT too, or the nightly cancels the merge

baseballyama:
`github.head_ref` is empty for every event but a pull request, so
`${{ github.head_ref || github.sha }}` resolves to the same commit for a `push`, a
`schedule` and a `workflow_dispatch` on one ref — **one group**, and with
`cancel-in-progress: true` the later event kills the earlier run.

## Measured

`main = 64fa49231`, 2026-09-04:

```
2026-09-04T20:15:25Z  push      Corpus Compat  completed/cancelled  run=33915315776
2026-09-04T20:33:28Z  schedule  Corpus Compat  queued               run=33916817559
                                push run updated_at = 20:33:47Z     (19s after the schedule was created)
```

The push run's jobs at the moment it was killed:

```
completed/cancelled  started 20:16:52Z  Compiler parity        ← 17 minutes in
completed/success    ×9   (Affected gates, Shape matrix, SCSS, Formatter, Lint,
                           svelte-check ×3, LSP fixture parity current)
completed/skipped    ×2   (LSP real-world shards, LSP full-population ratchet)
```

Eleven of twelve jobs were already green. `main` is left carrying a `cancelled`
verdict — which under this repository's own recorded rule reads exactly like a red
one, and is not a success under the campaign's P0.

## The comment asserted the opposite

`corpus-compat.yml` said, above the very line that produced this:

> Keyed by the head branch on a PR … and by the commit otherwise, **so a push to
> `main` shares a group with nothing and cannot be cancelled.**

That was the claim, not the behaviour. It is replaced with what was measured, in
the same commit, because a comment asserting fidelity is where this class hides.

## Scope: 7 workflows, chosen by measurement

A collision needs two or more **non-`pull_request`** triggers plus a cancelling
group — a `pull_request` keys on `github.head_ref`, which no other event sets.
Censused over all 21 workflow files (the denominator is printed because a
`grep -A3` over `^concurrency:` silently dropped 8 of them on the first pass —
`corpus-compat.yml`'s `group:` is 10 lines below its `concurrency:`):

| | triggers |
|---|---|
| **benchmark.yml** | push, pull_request, workflow_dispatch |
| **capi.yml** | push, pull_request, workflow_dispatch |
| **codspeed.yml** | push, pull_request, workflow_dispatch |
| **corpus-compat.yml** | workflow_dispatch, schedule, push, pull_request |
| **coverage.yml** | push, pull_request, workflow_dispatch |
| **crates-io-packages.yml** | pull_request, push, workflow_dispatch |
| **type-aware-lint.yml** | workflow_dispatch, push, pull_request, schedule |
| ci.yml / changeset.yml / differential-corpus.yml / site-benchmark.yml | ≤1 non-PR trigger — cannot collide |

The last row is a **negative control**, not an exemption: those four keep the old
shape and the guard stays silent for them, while the seven above each make the
guard fire when reverted.

## The fix is in the guard, not only in the files

`workflow-trigger-guard` already asked *can one event collide with itself* and
required `github.sha`/`github.run_id`. That rule is satisfied by every one of
these seven. The new predicate asks *can two **different** events collide*, which
no amount of per-push keying answers.

Controls, run and restored byte-identically:

```
POSITIVE  revert each of the 7 in turn      → exit 2, and the message names that file   (7/7)
NEGATIVE  the 4 with ≤1 non-PR trigger      → silent
ABLATION  force the predicate false         → exactly the 2 must-fire self-tests go red
SELF-TEST 34 ok / 0 FAIL, exit 0
```

Five new self-test cases pin both directions, including a workflow with
`schedule` + `workflow_dispatch` and **no push at all** — the existing rule's
`pushes &&` guard cannot reach that case.

## Cost

`corpus-compat-job-filter` on this diff emits `lsp-corpus=true, lsp-ratchet=false`,
and `lsp-corpus`'s `if` is a conjunction whose first disjunct requires a schedule,
a dispatch, or `lsp-ratchet` — so the 950-job-minute job does **not** run here.

The change means a `push` and a `schedule` at one commit now both run instead of
one cancelling the other. That is the intended cost: a run with no verdict is
indistinguishable from a green one, which is how a stale ratchet reached `main`
unnoticed (#2435/#2593).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -37,7 +37,7 @@ concurrency:
   # never produced a verdict. The cost is one run per merge instead of one per
   # burst; a branch with no verdict reads exactly like a green one, which is
   # how a stale ratchet reached `main` unnoticed (#2435/#2593).
-  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.sha }}
   # CodSpeed compares a PR against `main`'s run for the merge base, so a `main`
   # branch with no run makes it fall back to an older commit and report every
   # PR as a large regression.
--- a/.github/workflows/capi.yml
+++ b/.github/workflows/capi.yml
@@ -40,7 +40,7 @@ concurrency:
   # `github.sha`, not `github.ref`: every push to `main` shares one ref, so a
   # ref-keyed group cancels each merge's run as the next merge starts and `main`
   # carries no verdict — which reads exactly like a green one (#2435/#2593).
-  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.sha }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -16,7 +16,7 @@ concurrency:
   # never produced a verdict. The cost is one run per merge instead of one per
   # burst; a branch with no verdict reads exactly like a green one, which is
   # how a stale ratchet reached `main` unnoticed (#2435/#2593).
-  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.sha }}
   cancel-in-progress: true
 
 permissions:
--- a/.github/workflows/corpus-compat.yml
+++ b/.github/workflows/corpus-compat.yml
@@ -75,15 +75,24 @@ on:
       - '.github/actions/**'
 
 concurrency:
-  # Keyed by the head branch on a PR (a new push supersedes the old run) and by
-  # the commit otherwise, so a push to `main` shares a group with nothing and
-  # cannot be cancelled. The previous guard — `cancel-in-progress:
-  # ${{ github.ref != 'refs/heads/main' }}` — did not hold: every `main` run
-  # under it was still cancelled the instant the next merge started, so `main`
-  # never produced a verdict. The cost is one run per merge instead of one per
-  # burst; a branch with no verdict reads exactly like a green one, which is
-  # how a stale ratchet reached `main` unnoticed (#2435/#2593).
-  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  # Keyed by the event, the head branch on a PR (a new push supersedes the old
+  # run) and the commit otherwise. Two runs should supersede each other only
+  # when they answer the same question about the same target, and the event is
+  # part of that: without it a `schedule` and a `push` on `main` both resolve
+  # `github.head_ref || github.sha` to the same commit and land in one group.
+  # Measured 2026-09-04 — the nightly firing at 20:33:28Z killed the merge's own
+  # run at 20:33:47Z with 11 of its 12 jobs already green and `Compiler parity`
+  # 17 minutes in. The comment this replaces asserted that a push to `main`
+  # "shares a group with nothing and cannot be cancelled"; that was the claim,
+  # not the behaviour.
+  #
+  # The earlier guard — `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`
+  # — did not hold either: every `main` run under it was still cancelled the
+  # instant the next merge started, so `main` never produced a verdict. The cost
+  # is one run per event instead of one per burst; a run with no verdict reads
+  # exactly like a green one, which is how a stale ratchet reached `main`
+  # unnoticed (#2435/#2593).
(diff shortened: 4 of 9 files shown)


baseballyama:
Rebased onto `b860ab511` (post-#4320, post-#4299) and force-pushed; `test-workflow-trigger-guard.mjs` re-run after the rebase reads **34 ok / 0 FAIL, exit 0**, and `workflow-trigger-guard.mjs` itself exits 0 (`21 workflows checked … no violations`).

## An independent re-derivation of the carrier set — and my first attempt at it was wrong

The carrier set here was originally derived from the guard. I re-derived it from the workflow files with a second instrument, which is worth recording because the **first version of that instrument reported an eighth carrier that does not exist**.

That version tested `cancel-in-progress: true` and `group:` *anywhere in the file*, so it folded a **job-level** concurrency block into the workflow-level question and flagged `deploy-docs.yml`. `deploy-docs.yml` has no top-level `concurrency:` at all — its block is on the `build` job, with the constant group `pages-build` and a comment explaining the intent ("Supersede stale builds without interrupting the atomic Pages deployment"), and it is one of the guard's three allowlisted converging job groups. Two instruments disagreed and the guard was the right one.

Corrected to distinguish the two levels, with the denominator printed so a partition error cannot hide:

**`main` (`b860ab511`)**
```
files=21 workflow-level=17 job-level-only=3 none=1 sum=21
  !! benchmark.yml            ${{ github.workflow }}-${{ github.head_ref || github.sha }}
  !! capi.yml                 ${{ github.workflow }}-${{ github.head_ref || github.sha }}
  !! codspeed.yml             ${{ github.workflow }}-${{ github.head_ref || github.sha }}
  !! corpus-compat.yml        ${{ github.workflow }}-${{ github.head_ref || github.sha }}
  !! coverage.yml             ${{ github.workflow }}-${{ github.head_ref || github.sha }}
  !! crates-io-packages.yml   crates-io-packages-${{ github.head_ref || github.sha }}
  !! type-aware-lint.yml      type-aware-lint-${{ github.head_ref || github.sha }}
carriers=7 guarded=0 unguarded=7
```

**this branch (`023bef836`)**
```
files=21 workflow-level=17 job-level-only=3 none=1 sum=21
  … same seven, each now carrying ${{ github.event_name }} …
carriers=7 guarded=7 unguarded=0
```

Selector: ≥2 non-`pull_request` triggers **and** a workflow-level cancelling group. Seven on both sides, so the change is exactly the guarded/unguarded flip and adds no carrier and drops none.

## The live window this closes

`corpus-compat.yml` is the only one of the seven whose schedule is daily (`0 18 * * *`); `type-aware-lint.yml` is weekly (`41 4 * * 2`). The other five have no `schedule` and collide only between a `push` and a `workflow_dispatch` on one ref. So the recurring loss is one workflow, once a night, and it lands on whichever merge is in flight — which is what produced the `cancelled` verdict at the merge commit measured 2026-09-04.

Note that the nightly currently in flight (run `33916817559`, `schedule` at `64fa49231`) and the push runs at `b860ab511` are in **different** groups today only because `main` moved between them; had the merge landed while `main` still sat at `64fa49231`, they would have been one group.

baseballyama:
## What this change lets happen that could not happen before

Adding `github.event_name` to a cancelling group means two events at one commit now **coexist** where one used to kill the other. That is the point, and it is worth asking what the surviving run could collide with — a `workflow_dispatch` racing a `push` on a publishing workflow would be a real regression, and it would be caused by this PR rather than fixed by it.

Measured over all seven changed files, counting steps that mutate anything outside the run (`npm publish`, `cargo publish`, `vsce publish`, `ovsx publish`, `gh release`, `softprops/action-gh-release`, `changesets/action`, `actions/deploy-pages`, `git push`):

```
benchmark.yml            0        publish-crates.yml    8   <- control
capi.yml                 0        release.yml          11   <- control
codspeed.yml             0        release-capi.yml      5   <- control
corpus-compat.yml        0        deploy-docs.yml       1   <- control
coverage.yml             0
crates-io-packages.yml   0
type-aware-lint.yml      0
```

All seven carriers are measurement gates with no external side effect, so two of them running concurrently produces two verdicts and nothing else. `crates-io-packages.yml` is the one whose name suggests otherwise and it does not publish — it runs `check-crates-io-packages.mjs`, the registry dependency-surface check and MSRV `cargo check` / `cargo test`; `publish-crates.yml` is the workflow that publishes, and it is not in this change set (it has no cancelling group with two non-`pull_request` triggers).

**One control returned a false zero and it should be said rather than left in the table.** `publish-vscode.yml` reads 0 because it publishes through `node scripts/release/publish-vscode.mjs` rather than by invoking `vsce` from the YAML, so this pattern cannot see it. That does not weaken the carrier result — four of five controls fired, so the pattern demonstrably selects — but it does mean the zeros above are "no *literal* publish invocation", and the argument for the seven rests on reading them as gates, which their steps confirm.

The other direction is unchanged: within one event, `github.head_ref || github.sha` still supersedes a stale run exactly as before. This PR only stops a `schedule` and a `push` from sharing a group when `head_ref` is empty for both.

Issue #4248, opened 2026-09-03
LSP: the 2,112 `unclassified` ratchet entries decompose, and 56 of them are `null` vs `[]`


baseballyama:
### What

`lsp-known-failures.json` holds 23,742 entries; **2,112 carry the single label `unclassified`** and are
therefore un-attributable under P3 no matter how many of the other 70 labels get a terminal.
This issue decomposes the readable part of that population **with zero server runs**, by
re-implementing `diff.mjs`'s `digest()` and preimaging the values the keys already carry.

All numbers below are from the sidecar on `chore/lsp-mechanism-sidecar-fill` (#4243, `101542627`).

### The population, and why 1,886 of it cannot be read this way

```
only-unclassified entries          2112
  both-value digests (readable)     226   ← [official=<12hex>,rsvelte=<12hex>]
  set digest only (not readable)   1886   ← [count=N,hash=<12hex>] over identity keys
```

`verify.mjs` stores a **value** digest for a `:value-mismatch` and a digest of the *identity key
set* for `:extra-rsvelte` / `:missing-rsvelte`. The identity keys are themselves
`item-<digest(value)>`, so the set digests are doubled and have no preimage. That is the whole
reason for the 226 / 1,886 split — not a property of the methods.

### Why the classifier produces `unclassified` here at all

Two independent causes, and only the first was previously named:

1. `classifyDivergence` is called **only** on the corpus branch (`verify.mjs:441`, inside
   `if (entry.suite === "corpus")`); the `else` at `:468` records `[UNCLASSIFIED]`
   unconditionally.
2. `classifyDivergence` (`mechanism.mjs:675-685`) dispatches on **hover / definition /
   completion** and returns `unclassified` for everything else.

Measured over the 2,112:

```
   802  textDocument/diagnostic          ← no classifier arm
   550  textDocument/documentSymbol      ← no classifier arm
   314  textDocument/inlayHint           ← no classifier arm
   260  textDocument/foldingRange        ← no classifier arm
   100  textDocument/completion          ← HAS an arm, never reached (cause 1)
    20  textDocument/formatting          ← no arm
    16  textDocument/codeAction          ← no arm
    16  textDocument/hover               ← HAS an arm, never reached (cause 1)
    14  textDocument/selectionRange      ← no arm
    10  initialize                       ← no arm
    10  textDocument/documentHighlight   ← no arm

arm exists   116 / 2112   (5.5%)
no arm      1996 / 2112  (94.5%)
```

[...]

PR #4332, opened 2026-09-05
chore(lsp-gate): the empty-result label carries its method, because the two spellings come from different places

baseballyama:
`empty-result-spelling` is one label over 56 ratchet entries. A terminal written on it attributes two mechanisms at once, so it is split by method first.

## What the 56 entries are

Measured on the ratchet's own digests. `digest(v) = sha256(v).hex.slice(0,12)`, identified against five candidates so the two that appear are pinned rather than guessed:

```
74234e98afe7  <- JSON null       (appears)
4f53cda18c2b  <- JSON []         (appears)
12ae32cb1ec0  <- JSON ""         (does not)
44136fa355b3  <- JSON {}         (does not)
eb045d78d273  <- JSON undefined  (does not)
```

```
entries 56 / bases 28   (phase=edit doubles each)

  34  textDocument/inlayHint          official=[]    rsvelte=null
   6  textDocument/inlayHint          official=null  rsvelte=[]
   6  textDocument/inlayHint          official=[]    rsvelte=null   [expected suite]
  10  textDocument/documentHighlight  official=null  rsvelte=[]
```

The two methods are not the same mechanism. `textDocument/inlayHint` has **no native handler** — it is in `forward_tsgo_request`'s arm at `server.rs:711`, and `fn on_inlay_hint` does not exist (grep returns 0; positive control: `document_highlight` returns 4 hits in the same file). `textDocument/documentHighlight` is answered by rsvelte's own provider at `server.rs:1245`.

## `EMPTY_SPELLING_METHODS` is measured, not listed

Two wrong lists were considered and rejected by measurement:

- **the two methods that carry the label today** — a third method answering "nothing" on both sides would throw, and a gate that crashes on a real measurement is worse than one emitting an unexpected label;
- **every method the ratchet holds (12)** — driving all twelve through `classifyDivergence` shows `hover`, `definition`, `completion`, `diagnostic`, `documentSymbol` and `foldingRange` have classifiers of their own that never fall through to `classifyGeneric`, so those six labels would be in the vocabulary with no input able to produce them.

The list is the **six that reach it**, and `mechanism.test.mjs`'s cross-product test now drives every member in both directions, so adding an unreachable one fails there.

## The sidecar edit is machine-checked

`lsp-mechanisms.json` is generated from the complete 17-artifact set of one `lsp-corpus` run, and this re-label was applied to it by hand — which is safe only because the new label is a pure function of the method already in the entry's key. `lsp-mechanisms-check.mjs` now verifies exactly that, so a hand edit the generator would not reproduce cannot sit there silently.

## Controls

| control | result |
|---|---|
| inject one mismatched label into the real sidecar | `rc=1`, names the exact entry |
| restore | `rc=0`, `cmp` byte-identical |
| add an unreachable method (`hover`) to the vocabulary | that test alone reddens |
| restore | 35/35, `cmp` byte-identical |
| `test:lsp-mechanisms-check` | 11 cases, both directions of the new check |

## Both terminals stay `null`

The inlay-hint half is **not** settled as upstream, and that was the first reading. `InlayHintProvider.getInlayHints` (language-tools) returns `null` when inlay hints are **disabled by preference** and `[]` when they are on with nothing to report; `ExecuteMode.FirstNonNull` passes `[]` through unchanged (`PluginHost.ts:970-977`). So official's two spellings carry a distinction rsvelte does not make, because it forwards tsgo's answer verbatim. Which of the two produced each entry is **unmeasured** — it needs both servers running.

`empty-result-spelling-document-highlight` additionally contains #4331, where rsvelte's own two ports disagree on one fixture. That is a defect, not a spelling, and it must not be closed by a terminal written for the other nine entries.

Refs #4048, #4248, #4331

### DIFF
--- a/compatibility/lsp-mechanisms.json
+++ b/compatibility/lsp-mechanisms.json
@@ -613,7 +613,10 @@
 		"inlay-hint-mixed": {
 			"terminal": null
 		},
-		"empty-result-spelling": {
+		"empty-result-spelling-document-highlight": {
+			"terminal": null
+		},
+		"empty-result-spelling-inlay-hint": {
 			"terminal": null
 		},
 		"diagnostic-mixed": {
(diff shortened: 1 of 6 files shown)


baseballyama:
## 訂正 — この PR は **DoD の数を動かしません**。動くのは sidecar 側だけです。

2 つ目の commit（`aac0da4ad`）で `initialize` の capability 3 件に `deliberate-divergences` の terminal を配線しました。そのとき **`lsp-mechanisms-check` の数だけを見て「3 件 attributed」と読みかけました。** 両方を回すと:

```
lsp-mechanisms-check   1222 -> 1225 attributable   （sidecar の terminal を読む）
attribution-check      1224 -> 1224                （KNOWN-FAILURES.md の表を読む）
```

`attribution-check` が動かないのは、**表には既に載っていたから**です（`KNOWN-FAILURES.md:5388` の
`| 5 | deliberate-divergences | … workspace.workspaceFolders, positionEncoding,
diagnosticProvider.identifier … |`）。私が閉じたのは DoD の残りではなく、
**表と sidecar が同じ決定について食い違っていた**という不整合の方です。

`AGENTS.md` が記録している「1 つの ratchet を 2 つのゲートが採点すると『残り』が 2 通りになる」の
実例で、しかも**小さい方の差が危ない**という指摘どおりでした — 1225 と 1224 は
「片方がもう片方の言い換え」に見える距離です。**無修飾の「残り」を書かない**が正しく、
この PR の主張は次のとおりです:

- `attribution-check`（DoD）: **変化なし**、1224 のまま
- `lsp-mechanisms-check`: 1222 → 1225、`unclassified`/未確立 terminal でブロックされる entries が 3 減
- 実質: **決定は既に下されて文書化され pin もされていたのに、sidecar の `terminal` だけが繋がっていなかった**

ついでに、この訂正が出たのは私が最初 `node scripts/ci/attribution-check.mjs 2>&1 | tail -20; echo "rc=$?"` と書いて **`tail` の終了コードを読んだ**からです（`rc=0` と表示され、実際は 1）。ファイルに落として読み直しました:

```
node scripts/ci/attribution-check.mjs > /tmp/attr.log 2>&1
ATTR_EXIT=1
```

PR #4314, opened 2026-09-04
fix(parse): a `this` parameter and a rest parameter are both `params`

baseballyama:
`FunctionDeclaration.params` came out short **two independent ways**, and one ratchet key
counted both — which is why `FunctionDeclaration.params[]#length` stayed listed after the first
half was fixed. A key fine enough to separate them would have gone green on the first fix and
the second cause would have shipped as closed.

| loss | where it lives |
|---|---|
| a TypeScript `this` parameter — TSESTree models it as an ordinary leading `params[0]` | `convert_function_declaration_as_node` never consulted `this_param` |
| a rest parameter — oxc keeps it in `FormalParameters::rest`, outside `items` | the same converter never emitted it, so `function f(...a)` and `export function f(...a)` disagreed with each other (only the export form reached the Value form) |

**A rest parameter's type annotation belongs to the `RestElement`, not to its `argument`.**
Official leaves `RestElement.argument.typeAnnotation` undefined, so nothing downstream can
recover it from the argument — `JsNode::RestElement` grows a `type_annotation` and the binary
parse envelope carries it.

That makes this a **wire-format change**. `VERSION` moves `10 -> 11` in all three places that
spell it: the Rust writer (`napi_raw_parse.rs`), the JS decoder (`parse-envelope.js`), and
`scripts/dev/test-parse-envelope-validation.mjs`, which hand-builds an envelope. The third was
found by grepping the *constant's value* after moving it, not by counting ports — a port is a
concept and cannot be searched, a constant is a string and can. Note the half that grep would
still have missed: that third file also hand-copies a record's **layout**, and it happens to
copy `JS_IDENTIFIER` rather than the record this PR changed.

## Measurement

Built from `d1682317` and staged clean (`git status --porcelain --untracked-files=no` empty —
the predicate `crates/rsvelte_napi/build.rs:22` stamps the binary with), sha256
`3879dc6c71c53a2b744429c072216e72f2bf9b581ff57fb9e9c539d2eae53f5d`.

```
33,908 .svelte manifest entries x 2 axes + 7 loose sources; 66,901 compared pairs
listed key no longer diverges: 8        NEW divergence: 0
confirm run: OK — 211 listed divergence keys, none new, none stale
```

The eight deleted ratchet keys are exactly the eight the pre-baseline run named — set-compared
both ways (`deleted_not_named=0`, `named_not_deleted=0`), with a fabricated key injected to show
the comparison can report a difference.

[...]

PR #4338, opened 2026-09-05
fix(parse): a rest parameter carries its TypeScript annotation, in all three ports that build one

baseballyama:
Closes the `RestElement.typeAnnotation#missing` half of #4335 (2 of 219 `parse-ast` entries).

## The oracle decides where the field lives, and it is narrower than the node type suggests

Probed against `submodules/svelte`'s compiler, both directions:

| source | official `RestElement.typeAnnotation` |
|---|---|
| `function f(...a: string[]) {}` | `TSTypeAnnotation` |
| `const g = (...a: number[]) => a` | `TSTypeAnnotation` |
| `const { x, ...r }: T = o` | **absent** — the annotation is on the declarator id |
| `const [ x, ...r ]: T[] = o` | **absent** |
| `let r; ({ ...r } = o)` | **absent** |

So only a function/arrow **parameter** rest carries one. Eight of the patched sites therefore get
a *measured* `None` rather than a placeholder.

## The named carrier pointed at one of three ports

The ratchet names `animotion/.../code.svelte`. Its three rest params are all `export function`.

Patching every typed `JsNode::RestElement` site first, the gate reported:

```
[parse-ast] OK — 219 listed divergence keys, none new, none stale
```

**Nothing retired.** A seven-host grid says why:

| host | official | rsvelte after the first patch |
|---|---|---|
| plain `function` | annotation | **no `RestElement` at all** |
| `export function` | annotation | **ABSENT** ← the surviving carrier |
| function expression / class method / object method / arrow | annotation | present |

Re-taking the population **from the code rather than from the carrier** — every
`set_field("type", …, "RestElement")` builder — gives five Value-path sites: one already correct
(the exemplar this copies, ~6,600 lines above the omission and carrying the rule in full), one
array-pattern rest that is supposed to have no annotation, and **three missing it**.

Plain `function` emitting no `RestElement` is the separate `params[]#length` defect AGENTS.md
already records, and stays out of scope.

## What is not claimed

The `_with_adjustment` template-expression port measured **4/4 EQ without the field** — missing but
not diverging on `parse()`. It is filled for consistency; its **program surface** (the JSON
`rsvelte_lint` and svelte2tsx read) is **UNMEASURED**.

The field is output-only by mechanism. Locally I measured only that it does not appear in
generated code — `parse()` emits `TSTypeAnnotation` for a source whose four `compile()` targets do
not (n=1 file, 4 targets). That is necessary and not sufficient: it shows the annotation text does
not leak, not that no byte moved.

The corpus output ratchets run on this PR, and they narrow it further — but only to **verdict**
movement, not byte movement: an entry already listed can change its output bytes and keep its
verdict, and the gate stays green. So "no `compile()` byte moved" is **not** established by
anything here, and is not claimed.

## Wire format

A new field moves the parse envelope `VERSION` 10 → 11. It lives in three places — the Rust
writer, the JS reader, and `scripts/dev/test-parse-envelope-validation.mjs`, which hand-assembles
an envelope and is findable only by grepping the value.

## Ratchet

`parse-ast-known-failures.json` 219 → 217, re-baselined **after** rebasing onto `main`. Both
removals are stale-side; the run reported no new keys.


baseballyama:
## This PR duplicates #4314, and I am folding it in rather than merging both

I opened this without first asking which of my own open PRs touch these files — the check
AGENTS.md prescribes and I skipped. Measured after the fact:

**8 of this PR's 9 files are also in #4314**, and the envelope-writer diffs are *byte-identical*:

```
#4314 : JsNode::RestElement { …, type_annotation }  +  write_opt_type_annotation(…)  +  VERSION 10 → 11
#4338 : JsNode::RestElement { …, type_annotation }  +  write_opt_type_annotation(…)  +  VERSION 10 → 11
```

Merging both is not a conflict-free no-op and not a duplicate-free win either. Actually merging
them (`git merge --no-ff`, not `merge-tree`, whose `128` on git 2.33 is *flag unsupported* and not
*conflict*) gives:

```
CONFLICT  crates/rsvelte_core/src/ast/typed_expr.rs
CONFLICT  crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
CONFLICT  compatibility/KNOWN-FAILURES.md
auto-merged: parse-ast-known-failures.json → 209 entries
VERSION: stays 11
```

That last line is the one that matters: both sides make the *same* edit to `VERSION`, so git keeps
one copy silently and two independent layout additions would ship under one version number — which
is precisely what the version exists to prevent.

## The two are complementary, not redundant

#4314 adds the field to the typed variant and the wire format, and retires
`FunctionDeclaration.params[]#length`. It does **not** retire
`RestElement.typeAnnotation#missing` — because the typed sites alone retire nothing, which is this
PR's finding: the ratchet's named carrier (`export function`) is built by a **Value-path** builder
the typed sites never reach, and re-taking the population from the code found three such builders
missing the field where the carrier named one.

So the unique content here is the three Value-path sites, the measured `None` on the eight
pattern-rest sites, and the 2 ratchet removals. Folded onto #4314 they compose: 219 → **209**.

And after the fold there is no second wire change at all — #4314 already carries it — so no second
`VERSION` bump is needed.

## Plan

1. Rebase #4314 onto current `main`. Its verdicts are from `2026-09-04T16:49:10Z` against a
   `main` tip of `2026-09-05T05:05:12Z`, and `main` gained real `2_analyze` source in that window
   (`scope.rs` +58, `shared/utils.rs` +53/−2, `variable_declarator.rs` +6/−3). A 12-hour-old green
   describes a superseded tree, and merging it as-is risks `main` — which is a DoD item.
2. Apply this PR's unique part on top.
3. Re-baseline the combined ratchet **by measurement**, not by hand-merging to 209.
4. Close this PR.

Closing is deferred until step 2 lands so nothing is lost in between.

PR #4381, opened 2026-09-06
fix(compiler): place an erased-TS comment by the source declarator count

### LINKED ISSUE #4279: A comment inside an erased TS type alias is spliced between `let` and the declarator name
### What

A comment written inside a TypeScript **type alias** or **interface** body survives erasure —
correctly — but is spliced between the `let` keyword and the declarator name, so official's

```js
	/** keep me */
	let v = $.prop($$props, 'v', 24, () => ({ a: 1 }));
```

comes out as

```js
	let /** keep me */
	v = $.prop($$props, 'v', 24, () => ({ a: 1 }));
```

Both parse and mean the same thing, so nothing below output equality sees it.

### Repro

```svelte
<script lang="ts">
	type P = {
		/** keep me */
		a?: number;
	};
	export let v: P = { a: 1 };
</script>
<p>{v}</p>
```

`generate: 'client'`, `dev: false`, oracle `submodules/svelte/packages/svelte/src/compiler/index.js`.

### Grid

| cell | verdict | note |
|---|---|---|
| `type P = { /** c */ a?: number }` + `export let v: P` | **DIFF** | `let /** c */` |
| `type P<U> = U extends string ? { /** c */ a?: number } : unknown` | **DIFF** | same splice, nested conditional |
| `interface I { /** c */ a?: number }` + `export let v: I` | **DIFF** | same splice **and** the comment is emitted **twice** — official×1, rsvelte×2 |
| `type P = { a?: number }` (no comment) | EQ | the erasure itself is correct |
| `/** c */ export let v = 1` in plain `<script>` | EQ | the ordinary leading-JSDoc path is right |

The last two rows are the controls that matter. The no-comment row says the type alias is erased
correctly and only the comment is misplaced; the plain-JS row says the general
"comment before a prop declaration" path is right, so the carrier is the *erased* declaration
specifically.

The `interface` row carries a second, independent defect — a duplicate — which is #4239. It is
listed here because the same cell shows both, not because they are one mechanism: the splice is
present on all three DIFF rows and the duplicate only on the interface row.

### Not the same as #4244

#4244 is a comment inside an **inline type annotation** on a local (`let a: { /* c */ b: number }`),
where rsvelte emits **zero** copies and official emits two. Here the comment survives and only its
position is wrong. Different host, different sign.

### Carrier — corrected

**The `Carrier` paragraph this replaces was wrong, and it is worth saying how.** It asserted that
`svelte-tweakpane-ui/src/lib/control/Point.svelte` — the remaining `known-failures.client.json` /
`known-failures.client-dev.json` entry — is this mechanism, on the strength of the file having a
`$$Props` chain of the right *shape*. It was never measured. Measured, that file's output contains
**no `let /**` splice at all**, and its whole divergence is two other mechanisms:

```
official blocks 18  distinct 12
rsvelte  blocks 13  distinct 12
```

The **distinct block sets are identical**; official emits five of them twice. That is the
acorn-typescript 1.0.10 `tsLookAhead` duplication documented in #4244 (fixed upstream in 1.0.13,
which svelte 5.56.10 does not bundle), so on that axis **rsvelte is the more faithful side** and the
#2990 rule applies: byte equality is the goal, so rsvelte has to reproduce it.

The residue is closed by arithmetic rather than by a line diff:

| | measured |
|---|---|
| lines only official has | **51** |
| of those, non-comment | **1** — `let expanded = …; //  $$Props['expanded']; not working here?` |
| lines only rsvelte has | **1** — the same line without the trailing `//` |
| excess official blocks | 5, accounting for **50** lines |
| predicted official-only total (50 + the trailing `//`) | **51** ✓ |

So Point.svelte is fully explained by the #4244 duplication plus #4278's dropped trailing `//`,
with **nothing left over**, and it is not a carrier for this issue.

**A line-by-line reading says the opposite, and that is the reusable part.** Collapsing official's
duplicates with a regex and diffing by line index reported `differing lines: 263 of 327` with the
first differing pair being `Y dimension` against `X dimension` — one block apart, i.e. the
signature of an alignment shift, not of a third mechanism. A *multiset* comparison of lines and of
comment blocks has no alignment to lose, and it reduced the same pair of outputs to `51` and `1`.
Where a candidate mechanism inserts or deletes whole runs of lines, compare bags, not indices.

This issue therefore has **no known real-world carrier**. Its repro is constructed, and the honest
statement is that the shape's frequency in the collected corpus is **unmeasured**.

baseballyama:
Closes #4279.

## What was wrong

A comment left behind by TypeScript erasure ahead of an `export let` was
unconditionally flushed **after** the `let` keyword. Upstream does that only when
the source declaration has **more than one declarator**: with a single declarator
the declaration keeps its `loc`, esrap's comment cursor reaches it before the
identifier, and the comment prints ahead of `let`.

```svelte
<script lang="ts">
	interface I { /* c */ a?: number }
	export let one: I = 1;
</script>
```

| | official | rsvelte (before) |
|---|---|---|
| one declarator | `/* c */` then `let one = $.prop(...)` | `let /* c */` then `one = $.prop(...)` |
| two declarators | `let /* c */` then `two = $.prop(...)` | same (already correct) |

## The axis is the declarator count, not the erased host

Four hypotheses were falsified before this one held — the erased host
(`interface` vs `type` alias), where in the host the comment sits, the comment
kind, and the `dev` flag all leave the verdict unchanged. Printing official's
**full** output rather than the diff is what named it: upstream itself prints
`let // c` in the `>= 2` case, so the two shapes are both upstream's and the
choice between them is made by the source.

## What is deleted

`unlocate_prop_declarations_after_erased_comments` forced the multi-declarator
shape in both cases, together with the flag it set
(`RawMappedSpan::erased_comment_before_export_prop`), the projection field that
fed it (`Analysis::erased_leading_comments_before_export_props`) and its builder
— 4 call sites, net −149/+51 under `crates/`.

Three things say the whole chain was dead rather than merely wrong:

- Its doc comment asserted the opposite of what upstream does ("upstream reaches
  the identifier first and therefore emits `let // comment`"), and that is false
  on all six erased-host × comment-kind cells, **including the exact spelling it
  names**.
- Its unit test asserted the same wrong string. The svelte pin has not moved
  since that function landed, so the expectation was never oracle-generated.
- It is not load-bearing for the `>= 2` case either. Under a two-sided
  `ABLATE_UNLOCATE` switch on one binary, all 24 grid cells are EQ with it
  disabled and the base arm reproduces `main`'s 6 DIFF with it enabled — and the
  two real corpus files it was originally added for still match on all four
  targets without it.

The one control never taken when it landed was *are the `>= 2` cells already EQ
without this*. They are.

## Measurement

Two arms, one process each, `sha256`-distinct, separated by a live two-sided
probe (base 6/24 DIFF, head 0/24). The arm the sweep ran predates a clippy
`collapsible_if` collapse, so the final tree was swept too and the two agree by
measurement rather than by argument: **MOVED 0** over all 135,560 units, with
base-vs-final reproducing the same ten keys and the same hashes.

```
135,560 units compared (130,586 live, 4,974 dead in both)
MOVED 10
  MISMATCH -> match      8
  MISMATCH -> MISMATCH   2
  match -> MISMATCH      0
```

The two that stay are one file (`svelte-tweakpane-ui/.../AutoValue.svelte`) whose
residual is block-comment continuation indentation. Its head differing-line set
is a **strict subset** of its base one (26 ⊂ 28, 0 lines present in head and
absent from base), and the two non-comment lines it carried on base — the `let //`
splice and the `let`-less `value = $.prop(...)` under it — are gone.

Local: `cargo fmt --all --check` 0, `cargo clippy --workspace --all-targets
--all-features -- -D warnings` 0, and `cargo test --release -p rsvelte_core --lib`
1969 passed / 0 failed (`Running unittests src/lib.rs` present), all re-run after
the rebase.

## Why no gate saw it

The corpus output gate scores all ten units `match` on **both** arms.
`ast_equiv_batch` runs under `CommentPolicy::Ignore` (`GATES.md` blind spot 1a,
already documented with a positive control), and the two shapes are one
`let v = ...` declaration either way — so a comment-placement-only byte
difference is rescued, and retirement and regression are equally invisible.

That is the interesting part rather than a missing gate row: the blind spot **was**
measured, cited by file and line, and carried its own control, and the defect went
in through it anyway. Documenting a blind spot is not closing one.

It also decides what the pattern-corpus file is for. It is the **reproduction**;
the guard is `erased_typescript_interface_comment_follows_the_declarator_count`
in `3_transform/client/tests.rs`, whose two cells are generated from the oracle:

```
one declarator   ->  `// c` then `let use = $.prop(...)`
two declarators  ->  `let // c` then `use = $.prop(...)`
```

## Two pattern-corpus conventions

The first repro I wrote was green on **both** arms. It carried a `//` line
explaining each cell, and a real leading comment ahead of an `export let` changes
the very attachment the file exists to pin — so the convention against provenance
comments is widened from HTML to JavaScript, and a second one is added:

> **Measure a repro on both arms.** A file that matches on the fixed tree is only a
> repro if it *diverges* on the tree without the fix; a cell that is green either
> way pins nothing and cannot regress.

Rule 4 would not have caught it (I was not writing provenance, I was labelling
cells), so they are two rules rather than one.

## Base

Rebased onto `9171df619` (#4376), which touches `2_analyze/scope_builder.rs` —
disjoint from this change's files, no conflict. The sweep above was taken at
merge-base `1512bcdcf`; the merge ref's verdict is CI's corpus gate.

### DIFF
--- a/.changeset/erased-ts-comment-declarator-count.md
+++ b/.changeset/erased-ts-comment-declarator-count.md
@@ -1,0 +1,9 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Place a comment left behind by TypeScript erasure where upstream places it. A
+comment inside an erased `interface` or type alias preceding an `export let` was
+unconditionally flushed after the `let` keyword; upstream only does that when the
+source declaration has more than one declarator, and with a single declarator it
+prints the comment ahead of `let`.
--- a/compatibility/pattern-corpus/README.md
+++ b/compatibility/pattern-corpus/README.md
@@ -31,15 +31,21 @@ Ids are `pattern/issues/<file>`, `pattern/matrix/<axis>/<file>` and
    message or position belongs here too; say so in its row.
 3. **One behaviour per file, minimal.** Delete everything the shape does not
    need.
-4. **Never write provenance in an HTML comment.** Removed comments are
-   themselves a whitespace-sensitive compiler input (see #1975) — a `<!-- from
-   issue N -->` line silently changes what the file tests. Provenance belongs in
-   the table below and nowhere else. (The comments in
-   `matrix/whitespace-comments/` are the *payload*, not provenance.)
+4. **Never write provenance in a comment — HTML or JavaScript.** Removed
+   comments are themselves a whitespace-sensitive compiler input (see #1975), and
+   a comment inside `<script>` is the same input one level down: #4279's first
+   repro carried a `//` line explaining each cell and was green on *both* arms,
+   because a real leading comment ahead of an `export let` changes the very
+   attachment the file exists to pin. Provenance belongs in the table below and
+   nowhere else. (The comments in `matrix/whitespace-comments/` are the
+   *payload*, not provenance.)
 5. **Commit formatted files.** They flow through the fmt gate, so keep them in
    the shape prettier-plugin-svelte would produce; an unformatted file is a
    needless new formatter case, not a compiler case.
-6. **A repro lands with its fix, not before.** Adding a file for a still-open
+6. **Measure a repro on both arms.** A file that matches on the fixed tree is
+   only a repro if it *diverges* on the tree without the fix; a cell that is green
+   either way pins nothing and cannot regress.
+7. **A repro lands with its fix, not before.** Adding a file for a still-open
    divergence would mean seeding a `known-failures` entry, and the seed then has
    to be tracked and burned down separately from the fix. Add the repro in the
    fix PR (or immediately after it merges) so it lands green.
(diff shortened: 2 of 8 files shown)


baseballyama:
Rebased onto `db0486018` and re-run. The previous green was **stale, and conflicting**.

```
#4377 merged at                     2026-09-06T14:39:47Z
this PR's newest check started at   2026-09-06T14:32:28Z
```

Every verdict on the old head predates #4377, so the 48/48 green was a measurement of a tree that does not contain it — and #4377 touches the same subsystem (client comment placement). The rebase also turned up a real `compatibility/pattern-corpus/README.md` add/add conflict: both branches append a row at the same anchor, so the branch was not merely stale but not mergeable, which `mergeable=UNKNOWN` was hiding while the checks read green. Resolved by keeping both rows.

Re-verified on the rebased tree, with #4377's own guards run explicitly because a rebase does not run the pre-commit hooks and the conflict was resolved by hand:

```
cargo fmt --all --check                                            clean
cargo clippy --workspace --all-targets --all-features -D warnings  exit 0
rsvelte_core --lib                                                 1969 passed / 1 ignored
--test export_let_trailing_comment                                 1 passed   (#4377's guard)
--test prop_trailing_comment_hosts                                 3 passed   (#4377's guard)
--test prop_default_leading_comment                                2 passed
--test props_initializer_comment_3515                              5 passed
--test type_literal_head_comment_repeat                            2 passed
node scripts/ci/check-pattern-corpus-docs.mjs                      exit 0
```

That last group is the point of the re-run rather than a formality: this PR deletes `unlocate_prop_declarations_after_erased_comments`, and #4377 rewrote `props_transforms.rs` — the two are textually disjoint and both decide where a client comment lands, so nothing but running one against the other settles it.

PR #4390, opened 2026-09-06
fix(parse): a TS grammar rule acorn does not implement is not a parse error outside a script

### LINKED ISSUE #4083: rsvelte parse rejects `{accessor satisfies T}` / `{declare satisfies T}` in a template expression, which official accepts
## Summary

`rsvelte_core::parse` rejects a template expression whose head identifier is named `accessor` or `declare` and whose operator is `satisfies`. The official Svelte compiler accepts it, and so does the `oxfmt --svelte` (prettier-plugin-svelte) oracle. `rsvelte-fmt` inherits the rejection, so the whole file fails to format.

The reported error is about a **parameter modifier**, which is the tell: the expression is being parsed inside a parenthesized wrapper that oxc reinterprets as an arrow-function parameter list, so `accessor` / `declare` are read as TS parameter modifiers rather than as identifier references.

## Reproduction

```svelte
<script lang="ts">let accessor: any;</script>
<p>{accessor satisfies string}</p>
```

```svelte
<script lang="ts">let declare: any;</script>
<p>{declare satisfies string}</p>
```

## Measured, both sides

| entry point | `accessor satisfies string` | `declare satisfies string` |
|---|---|---|
| official `submodules/svelte/packages/svelte/src/compiler/index.js` `compile()` | **accepts** (no error) | **accepts** (no error) |
| `oxfmt` 0.64.0 with `scripts/fixtures/fmt-corpus.oxfmtrc.json` (`svelte: true`) | **accepts**, emits `<p>{accessor satisfies string}</p>` | **accepts**, emits `<p>{declare satisfies string}</p>` |
| `rsvelte_core::parse` | **rejects** `js_parse_error` — `'accessor' modifier cannot appear on a parameter.` | **rejects** `js_parse_error` — `'declare' modifier cannot appear on a parameter.` |
| `rsvelte_formatter::format` | **rejects** `ScriptParse` — TS1090 `'accessor' modifier cannot appear on a parameter.` + `Expected ',' or ')' but found 'string'` | same with `'declare'` |

The formatter failure is downstream of the parser one: both report the identical message, so this is one defect at the parse entry point, not two.

## Where it reproduces and where it does not

Measured on the #4062 grid (`name` × expression shape × 5 markup host slots — attribute, mustache, `{#if}` header, quoted attribute, spread):

- **reproduces**: `accessor satisfies T` and `declare satisfies T`, in all 5 host slots.
- **does not reproduce**: the same two names with `as` (`{accessor as string}`, `{declare as string}` both parse and format fine), and with every other shape in the grid (`n`, `n as A as B`, `n!`, `n! as T`, `(n as T)!`, `n.p as T`, `(n as any).p`, `f(n as T)`, `[n as T]`, `n as any ? 1 : 2`).
- **does not reproduce** for any other name: `type` / `module` / `using` / `component` / `hook` / `foo` / `namespace` / `satisfies` / `of` / `global` all accept `satisfies`.

So the axis is (name ∈ {`accessor`, `declare`}) × (operator = `satisfies`), not the name alone and not `satisfies` alone.

## Note on what "the oracle accepts it" does and does not mean

`svelte-eslint-parser` being more permissive than rsvelte's parser is *not* by itself a reason to loosen rsvelte's parser — that is called out in `CLAUDE.md`. What makes this one actionable is the first row of the table: **the official Svelte compiler accepts these two inputs**, so rsvelte's parser is stricter than the thing it is a port of.

## Context

Found while fixing #4062 (`{type as X}` → `{(type) as X}`). These two cells are excluded from that PR's grid test with an in-file comment pointing here; the #4062 fix does not move them in either direction.

### LINKED ISSUE #4084: rsvelte parse rejects `{<T>() => x}` in a template expression, reading `<T>(` as a generic arrow type parameter list
## Summary

`rsvelte_core::parse` rejects an angle-bracket type assertion in a template expression when the asserted operand is an arrow function: `{<string>() => a}`. The official Svelte compiler accepts it, and so does the `oxfmt --svelte` oracle. `rsvelte-fmt` inherits the rejection.

The message — `Type parameter name cannot be 'string'` — names the mechanism: `<string>(` is being read as a **generic arrow function's type parameter list** (`<T>(x) => …`) rather than as a cast applied to a parenthesized arrow.

## Reproduction

```svelte
<script lang="ts">let a: any;</script>
<p>{<string>() => a}</p>
```

## Measured, both sides

| entry point | verdict |
|---|---|
| official `submodules/svelte/packages/svelte/src/compiler/index.js` `compile()` | **accepts** (no error) |
| `oxfmt` 0.64.0 with `scripts/fixtures/fmt-corpus.oxfmtrc.json` (`svelte: true`) | **accepts**, emits `<p>{<string>() => a}</p>` |
| `rsvelte_core::parse` | **rejects** `js_parse_error` — `Type parameter name cannot be 'string'` |
| `rsvelte_formatter::format` | **rejects** `ScriptParse` — TS2368 `Type parameter name cannot be 'string'` |

## Where it reproduces and where it does not

Measured on the #4062 grid (operand kind × type operator × 5 markup host slots):

- **reproduces**: `<T>` applied to an arrow function (`<string>() => a`), in all 5 host slots.
- **does not reproduce** for `<T>` applied to any other operand in the grid: `<string>foo`, `<string>foo.p`, `<string>foo['p']`, `<string>f()`, `<string>'a'`, `<string>1`, `<string>{ a: 1 }`, `<string>[1, 2]`, `<string>a + b`, `` <string>`x${a}y` ``, `<string>!a`, `<string>a?.p`, `<string>new Foo()` — all parse and format, and all match the oracle byte-for-byte.

So the axis is the **operand being an arrow function**, not the `<T>` cast form in general.

## Relationship to #4083

Same entry point (`rsvelte_core::parse` of a template expression), **different mechanism** — #4083 is oxc reading a name as a TS parameter modifier inside a parenthesized wrapper, this one is oxc reading `<string>(` as a generic arrow's type parameter list. Filed separately so the mechanism is in the key rather than folded into one "template expression over-rejection" bucket.

## Note on what "the oracle accepts it" does and does not mean

Matching `svelte-eslint-parser`'s permissiveness is not a goal here. What makes this actionable is the first table row: the **official Svelte compiler** accepts it.

## Context

Found while fixing #4062. The #4062 fix does not move this cell in either direction (it is a parse rejection, reached before the paren decision that #4062 changes).

baseballyama:
Closes #4084.

## The cause is not OXC, and it is not one site

OXC ports TypeScript **grammar** checks that `acorn-typescript` does not implement, so it reports on programs upstream compiles. `expression.rs` already knows this — `ACORN_UNCHECKED_TS_GRAMMAR_RULES` listed 17 such codes, each confirmed against `svelte.compile`. Two things were wrong with it:

1. **TS2368 (`Type parameter name cannot be 'X'`) was missing from the table.**
2. **The table was consulted by two sites**, both script-side (`repair_ts_newline_import_assert`, `convert_parsed_program`). The template-expression, attribute, statement and parameter probes each called `result.diagnostics.first()` directly, and four expression-conversion gates each tested `result.diagnostics.is_empty()`.

So the same source compiled in a `<script>` and was a `js_parse_error` in `{…}`, an attribute value, a `{@const}`, an `{#each}` collection, a snippet parameter and a `{@render}` argument. One upstream rule, nine ports.

`tsc` 7.0.2 agrees with OXC that TS2368 is reportable **and still emits** `const x = () => a;` with exit 0 — so the divergence was rsvelte's fatality policy, not OXC being stricter than TypeScript.

## Fixing either half alone reads as done

Adding `"2368"` and leaving the probes as they are makes the template accept the source and then emit `p.textContent = ;` — text no JS parser accepts, because the conversion gate still bails and the expression is dropped. That is why the repro carries every host in one file rather than only the reported one: a repro at the reported position cannot separate a fix that reaches one probe from one that reaches all of them.

## Measured

15 hosts × 4 targets (`client`, `client-dev`, `server`, `server-dev`), against the source oracle `submodules/svelte/packages/svelte/src/compiler/index.js`:

| arm | EQ | OVER-REJECT | DIFF |
|---|---|---|---|
| base (`93126f69…`) | 2 | 56 | 4 |
| fix (`13c79888…`) | **58** | **0** | 2 |

The 2 residual DIFF cells are `snippet-param` on `server` / `server-dev` and are **byte-identical on `main`**. They are a different mechanism, not a ninth port of this one: `extract_snippet_param` strips TS *annotations* and not `<T>` *casts*, so `reparse_params` — which parses as plain `mjs`, where `<string>() => 1` is a real JS syntax error — falls back and drops the **whole** parameter list. Measured across 5 default shapes, only the cast-on-an-arrow row diverges (`p = 1`, `p = () => 1`, `p: any = 1`, `p` are all EQ), and a second parameter beside it is dropped too. Filed separately.

Corpus, base vs fix over every `.svelte` and `.svelte.(js|ts)` in the collected sources × 4 targets:

```
rows 142900   distinct keys 142900
live-in-both 134974  dead-in-both 7926  live->dead 0  dead->live 0
MOVED 0
```

`dead-in-both` is dominated by `.svelte.(js|ts)` sources this harness feeds to `compileModule` without the gate's esbuild type-strip (723 of 843 module files, identically in both arms) — they are dead for a harness reason, not a compiler one, and the comparison is per key so they contribute nothing either way. The live component population is 33,623 files x 4 targets.

The two arms hash differently and the 4-target grid is the two-sided discriminating probe; the arm was rebuilt from this tree and reproduced `13c79888…` bit for bit. The sweep was measured on a tree that differs from this branch by two comments only.

**Ablation**, each half separately: removing `"2368"` from the table reddens the `script` row; reverting the server `reparse_expression` port reddens the `const_tag` server row. Restored, the tree is byte-identical and all four tests pass.

## What this does not fix

#4083 (`{accessor satisfies string}` / `{declare satisfies string}`) is the sibling over-rejection at the same entry point and is **unmoved** by this change — measured on both arms, DIFF before and DIFF after, with `{accessor as string}` and `{foo satisfies string}` as live EQ controls. Its diagnostic is not a TypeScript grammar rule acorn skips: the `(…)` this probe wraps the body in makes OXC read `accessor` as an arrow-parameter modifier, so the fix belongs in the wrapper's own `retry_bare` path, not in the rule table.

## Controls

`crates/rsvelte_core/tests/acorn_unchecked_ts_grammar_hosts_4084.rs` pins the 14 host rows with expectations generated from the oracle, plus:

- `STILL_REJECTED` — TS in a plain `<script>`, an invalid template expression, an invalid script and an invalid snippet parameter must all still be rejected, so the fix is not "stop reporting".
- `a_legal_generic_arrow_is_untouched` — `{(<T,>(t: T) => t)(1)}` reaches the same probes and must be unchanged.
- `the_server_port_reaches_the_same_decision` — three rows the client-only fix leaves diverging.

## The repro's script-side control is in the test, not in the corpus file

Adding `<string>() => a` inside a `<script lang="ts">` to a corpus file makes the
**svelte2tsx** gate report `1 NEW ts-mismatch`: rsvelte's TSX shadow writes
`<string,>() => a` where upstream copies the source verbatim, which turns a type
assertion into a generic arrow named `string`. Measured — with that one line
removed the shadow is byte-identical and every compiler cell is unmoved, and with
it present only that line differs. It is a separate defect with no other carrier,
filed as #4399, so the `script` and `script_module` controls live in
`crates/rsvelte_core/tests/acorn_unchecked_ts_grammar_hosts_4084.rs` and the corpus
file carries the seven template hosts plus the legal-generic-arrow control.


baseballyama:
CI 赤の 2 件を潰しました。

**1. `Compiler parity` → parse() AST ratchet (6 件が listed-but-passing)**

このブランチの parse 修正で 6 キーが実際に一致するようになったため削除しました。NEW divergence は 0 件です。

- `{legacy,modern}::Identifier.optional#missing` → cluster `estree-fields`
- `{legacy,modern}::TSNumberKeyword#span` → cluster `span`
- `{legacy,modern}::TSTypeAnnotation#span` → cluster `span`

`parse-ast-known-failures.json` は 206 → 200。`KNOWN-FAILURES.md` の gated な 2 箇所（宣言 `**200 entries**`、partition 行 `55 + 44 + 30 + 28 + 14 + 14 + 6 + 6 + 2 + 1`）と、ungated な派生値（cluster 表の `span` 55/29・`estree-fields` 28/14、`112 distinct bases`・`88x2 + 24 = 200`・`0 of 88`）も JSON から再計算して更新しています。`known-failures-md-check.mjs` は EXIT=0、宣言を 201 に書き換える positive control でちゃんと赤くなることも確認済みです。

**2. `Formatter parity` → 新 repro が 1 件 NEW failure**

repro の `<string>() => a` は型引数名が `string` の generic arrow で、**TypeScript としては不正**（TS2368: type parameter name cannot be 'string'）です。oxc はこれを拒否するので `rsvelte-fmt` が `script parse failed` で落ち、ファイルが未整形のまま残って oracle と差分になります。

```
rsvelte-fmt: error: script parse failed:
  "Type parameter name cannot be 'string'" ... OxcCode { scope: Some("TS"), number: Some("2368") }
```

この PR が動かしたのは**コンパイラ側の** oxc 境界であって、フォーマッタの境界ではありません。フォーマッタは Svelte コンパイラの再現ではなく TypeScript を扱う整形器なので、不正な TS を拒否するのは正しい挙動です。

既に同じ形が `fmt-oracle-excluded.json` に `invalid-input` として入っています — `snippet-typescript` の `c?: number = 5`（TS1015）。TS1015 も TS2368 も同じ `ACORN_UNCHECKED_TS_GRAMMAR_RULES` の一員で、機構も同一です。そこで同じ分類で 1 件追加しました（24 → 25 entries）。DoD-4 の attribution 表（`invalid-input` and `migrate` 行 4 → 5）と `### invalid-input` 節の箇条書きも更新しています。

なお、フォーマッタ側の 23 箇所の oxc 境界を同じルールで揃えるのは**この PR の範囲外**として意図的に見送りました。受理面が corpus 全体で変わるため、`fmt-known-failures.json` の再ベースライン（オラクル木の再生成）を伴う独立した計測が要ります。

`attribution-check.mjs` の default mode は `main` でも赤（3 problems）で、この変更は未 attribution 件数を 188 → 182 に**減らす**方向にのみ動かしています。

PR #4401, opened 2026-09-06
feat(compiler): track Svelte 5.57.0

baseballyama:
Ports the 40 upstream compiler commits between `5.56.10` and `5.57.0`. Supersedes #4058 and #4091.

## What moved

**`f2648b3` — "end a restored reaction context at the end of its synchronous segment"** reached the most code, on two axes.

*Pickling.* An `await` that is not the last evaluated expression restores the reaction context, and **once an expression pickles one `await`, every later one in it pickles too**. rsvelte carries that decision in **three ports** — the typed script walk (`2_analyze/visitors/await_expression.rs`), the template collector (`2_analyze/visitors/await_block.rs`) and the instance-script text scanner (`client/expression_utils.rs`) — and all three needed the sticky flag. Two of them also disagreed with upstream on the shape rules: `is_last_evaluated_expression` returns `false` for **either half of a member expression** and for a **callee** (a callee is never `arguments.at(-1)`), where rsvelte returned `is_last` for a non-computed object and for an argument-less callee.

*Thunks.* Upstream's new `async_thunk` **never unthunks**, so `$.async(node, [], [() => firstTest()], …)` replaces `[firstTest]`. The client's `{#if}`, `{#each}`, `{#await}`, `{@html}` and `{#key}` thunks and all four `$.async_derived` emission sites keep their arrow, and a thunk whose body restored the context exits through `$.unsave()`.

**`9d0062d` — store subscriptions wait for the promise that assigns the store.** A `$name` binding adopts the store's blocker before function tracing runs, so a function reading `$store` inherits it. That needed its own whole-token scan: `extract_all_identifiers_from_statement` drops every `$`-prefixed token, so a store subscription is invisible to the identifier-based tracing. Upstream's `Memoizer.#blockers` is a `Set<Expression>` deduped by **object identity**, and `store`/`$store` share one `binding.blocker` while a traced function gets a fresh one — hence `[$$promises[0], $$promises[0]]`. On the server the unsubscribe moves into `$$renderer.on_destroy` when any subscription has a blocker.

The rest: `63b4c36` (no scoping class inside `<svelte:head>`), `b20b2ee` (an exported snippet keeps the CSS from being tree-shaken), `05b6916` (`bind:focused` omitted from SSR, in **both** ports of the `omit_in_ssr` table), `8299cbf` / `34489c1` (`$.comment()` for a lone anchor, `$.only_child` for a single child), `3405b5e` (one memoizer entry per class directive), `2ac7366` (a simple object destructure in `{@const}` returns the object), `a6153f1` (whitespace after an inline element when printing), `1be496f` (`onfocusin`/`onfocusout` satisfy `a11y_mouse_events_have_key_events`), `09a67ef` (warn on an undeclared shorthand handler on `svelte:window`/`document`/`body`), `0f458d5` (`derived_invalid_export` wins over `legacy_export_invalid`), `14ec75c` (a block-open marker folds into the branch's first push).

## Two collector gaps the store-blocker work surfaced

`collect_ids_from_expr` and its two siblings had no arm for `JsExpr::OpaqueIdentifier` or `JsExpr::SourceAnchored`. A store read is lowered to `Spanned(OpaqueIdentifier("$store"))`, so `$store` was **invisible to blocker collection** — the symptom was one array entry where upstream emits two, and no gate keyed on the generated call could see the cause.

## Verification

All ten fixture suites at 100%, measured on the rebased tree:

| Suite | Result |
|---|---|
| runtime-runes | 1046/1046 |
| runtime-legacy | 1207/1207 |
| hydration | 81/81 |
| SSR | 104/104 |
| validator | 334/334 |
| CSS | 183/183 |
| compiler-errors | 146/146 |
| Compiler Snapshot | 30/30 |
| print | 50/50 |
| sourcemaps | 29/29 |
| svelte2tsx | 256/256 |

Plus `cargo test --workspace --exclude rsvelte_core --release` (285 suites, 0 failures — `rsvelte_core` is the nine suites above) and `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean. The `--workspace` denominator is deliberate: the change touches `JsExpr`, which `rsvelte_bindings_support` matches exhaustively.


baseballyama:
Compiler parity の赤は本物のバグでした（`js-mismatch` 1件 + そのミュータント1件）。5.57.0 の 40 コミットのうち `c7d8233`（svelte#18718, "apply ownership mutation ignores to bindings"）を取りこぼしていたためです。

**修正**

- `BindDirective` はセッター用に合成した assignment へディレクティブの ignore をコピーするので、`validate_mutation` がそこで抜ける。rsvelte 側にその経路がなく、`<!-- svelte-ignore ownership_invalid_mutation -->` が `bind:value` に効いていませんでした。
- ノードが持つ ignore は**囲っているスタック全体**です（upstream の `ignore_map.set(node, get_ignore_snapshot())` は各ノードの push 後に走る）。rsvelte はそのノード直前のコメントだけを保存していたため、報告された形だけが直り隣接形が直らない状態でした。`metadata.ignored_codes` をスナップショットに変更しています。
- `build_bind_this` は自前の assignment を作り ignore を継承しないので、`bind:this` は同じコメントがあっても検証されたままです。共有ビルダー内でフラグを読むのではなく引数で渡し、`bind:this` の呼び出し側は `false` を渡しています。この行はコメントではなくテストで固定しました。

**検証**

`submodules/svelte/.../compiler/index.js`（`7bc0a70fe64d` = `VERSION === '5.57.0'`）を oracle として、`decl` = `$$ownership_validator = $.create_ownership_validator`、`call` = `$$ownership_validator.mutation(` を全セルで突き合わせました。

| markup | official | rsvelte (修正後) |
|---|---|---|
| `<input bind:value={item.heading} />` | no | no |
| `{#each [1] as i}<input bind:value={item.heading} />{/each}` | no | no |
| `{#if true}<input bind:value={item.heading} />{/if}` | no | no |
| `<div bind:this={item.el}></div>` | **yes** | **yes** |
| `<svelte:element this={'div'} bind:this={item.el}></svelte:element>` | **yes** | **yes** |

corpus の当該 fixture（`ownership-invalid-mutation-binding-ignore/Item.svelte`）は official 出力と末尾改行を除いてバイト一致です。rsvelte_core の全 694 テストバイナリ green、clippy 0 警告。

**残り2セル**は #4403 に切り出しました。コンポーネントの過剰抑制（本 bump 以前からの既存挙動）と `<svelte:window|document|body>` の抑制漏れ（`SvelteElement` に `Vec<String>` を足すと `size_of::<TemplateNode>() <= 128` を割るため、コンパクトな表現が要る）です。

Issue #4385, opened 2026-09-06
lsp: six of the nine start-tag opener ports still apply as one overwrite, so every expression inside them is unmapped


baseballyama:
`<svelte:element>`'s start tag was applied to the `MagicString` as **one** `str.overwrite`, so every expression inside it collapsed to a single mapping and hover answered about the wrong text (#4102). That is one host of a class: **nine** call sites apply a start tag with a single overwrite, and converting one of them leaves six still broken.

## Measured: one cell per port

One `.svelte` file per cell — a construct one compiler rejects must not change another cell's output — each holding `const probe = 1;` in `<script lang="ts">` and one expression attribute inside the start tag under test. The measurement is `textDocument/hover` on `probe`, both servers over stdio, `TSGO_BIN` pinned.

| # | port | site | official | rsvelte |
|---|---|---|---|---|
| 1 | `<div title={probe}>` | `element.rs` (segmented) | `const probe: 1` @6:12-17 | **same** ✅ |
| 2 | `<svelte:element this={tag} title={probe}>` | `dynamic_element.rs` | `const probe: 1` @6:34-39 | **same** ✅ (this is #4102's fix) |
| 3 | `<Comp a={probe} />` | `inline_component.rs::handle_component` | `const probe: 1` @6:9-14 | **same** ✅ |
| 4 | `<svelte:component this={Comp} a={probe} />` | `inline_component.rs:737` | `const probe: 1` @6:33-38 | **`null`** ❌ |
| 5 | `<svelte:self a={probe} />` | `inline_component.rs:1091,1096` | `const probe: 1` @6:16-21 | **`null`** ❌ |
| 6 | `<svelte:body onclick={() => probe} />` | `special_element.rs:331` | `const probe: 1` @6:28-33 | **`null`** ❌ |
| 7 | `<slot title={probe} />` | `slot_element.rs:122` | `const probe: 1` @6:13-18 | **`null`** ❌ |
| 8 | `<Comp><div slot="s" title={probe}></div></Comp>` | `component_slots.rs:571,675,679` | `const probe: 1` @6:27-32 | **`null`** ❌ |
| 9 | `<svelte:head><title data-x={probe}>t</title></svelte:head>` | `element.rs:385` (`handle_title_element`) | `const probe: 1` @6:28-33 | **`null`** ❌ |

[...]

Issue #4103, opened 2026-09-01
lsp: bind:this={el} is not projected — hover answers with the previous attribute, and 6133 fires on el


baseballyama:
`bind:this={el}` on a regular element is not projected as a reference to `el`. Two observable consequences: `textDocument/hover` inside the binding answers with the **previous attribute's** symbol at the **previous attribute's** range, and TypeScript reports `6133 'el' is declared but its value is never read` where the official server reports nothing.

## Repro

`src/App.svelte` (workspace: a plain `tsconfig.json`, no `node_modules`):

```svelte
<script lang="ts">
	const tag = 'div';
	let el: HTMLElement | undefined = undefined;
</script>

<div title={tag} bind:this={el}>a</div>
```

`initialize` (`initializationOptions` as `scripts/compat-lsp/verify.mjs` builds them, `TSGO_BIN` at the pinned `@typescript/native-preview`) + `didOpen` + `textDocument/hover` + `textDocument/diagnostic`.

## Actual output, both sides

**hover on `el` inside `bind:this={el}` (5:29)**

```
official {"range":{"start":{"line":5,"character":28},"end":{"line":5,"character":30}},
          "contents":"```typescript\nlet el: HTMLElement | undefined\n```"}

rsvelte  {"contents":"```typescript\nconst tag: \"div\"\n```",
          "range":{"start":{"line":5,"character":12},"end":{"line":5,"character":15}}}
```

rsvelte returns the hover for `tag` — the value of the *preceding* `title={tag}` attribute — and its range, 16 characters to the left of where the cursor is.

**Control, same line:** hover on `tag` in `title={tag}` (5:13) is identical on both servers (`const tag: "div"` @ 5:12–15). So the element and the attribute list are projected; it is the `bind:this` value that is not.

**`textDocument/diagnostic`**

| | items |
|---|---|
| official | 0 |
| rsvelte | 1 — `2:5 [6133] 'el' is declared but its value is never read.` |

The `6133` is the same fact from the other side: nothing in the generated `.svelte.tsx` reads `el`, so hovering inside the binding falls back to the nearest mapped token to its left, and TypeScript sees the declaration as unused. This makes it a projection defect rather than a hover-formatting one — a user with `noUnusedLocals` gets a false error on every `bind:this`.

## Scale

[...]

PR #4422, opened 2026-09-07
fix(svelte2tsx): apply a `<svelte:element>` opener as segments, not one overwrite (#4102)

### LINKED ISSUE #4102: lsp: every expression inside a <svelte:element> start tag has no hover
Every expression inside a `<svelte:element …>` start tag gets `null` from `rsvelte-language-server`'s `textDocument/hover`. The identical attributes on a plain `<div>` — single-line **or** multi-line — answer correctly, so it is the element, not the attribute syntax and not the line layout.

## Repro

`src/App.svelte` (workspace: a plain `tsconfig.json`, no `node_modules`):

```svelte
<script lang="ts">
	const tag = 'div';
	let hovered = false;
</script>

<div title={tag} onmouseenter={() => { hovered = true; }}>a</div>

<div
	title={tag}
	onmouseenter={() => { hovered = true; }}
>b</div>

<svelte:element this={tag} title={tag}>c</svelte:element>
```

`initialize` (`initializationOptions` as `scripts/compat-lsp/verify.mjs` builds them, `TSGO_BIN` at the pinned `@typescript/native-preview`) + `didOpen` + `textDocument/hover`.

## Actual output, both sides

| cell | official | rsvelte |
|---|---|---|
| A single-line `<div>`: `tag` in `title={tag}` (5:13) | `const tag: "div"` @ 5:12–15 | `const tag: "div"` @ 5:12–15 ✅ |
| A single-line `<div>`: `hovered` in the handler (5:40) | `let hovered: boolean` @ 5:39–46 | same ✅ |
| B multi-line `<div>`: `tag` in `title={tag}` (8:9) | `const tag: "div"` @ 8:8–11 | same ✅ |
| B multi-line `<div>`: `hovered` in the handler (9:24) | `let hovered: boolean` @ 9:23–30 | same ✅ |
| **C `<svelte:element>`: `tag` in `this={tag}` (12:22)** | `const tag: "div"` @ 12:22–25 | **`null`** |
| **C `<svelte:element>`: `tag` in `title={tag}` (12:34)** | `const tag: "div"` @ 12:34–37 | **`null`** |

Rows A and B are the controls and they hold. The multi-line control matters: the first reduction attempt blamed the attributes being on their own lines, and B refutes that.

The same file also yields **3 diagnostics from rsvelte against 0 from official**, so the projection difference is not confined to hover.

## Scale

`submodules/melt-ui`, full run (43 files / 12,043 identifiers / 72,258 requests). Positions that fall inside a `<svelte:element …>` start tag, counted by byte offset with brace-aware tag-end detection (a naive `indexOf('>')` stops at an arrow function's `=>` and undercounts):

| | |
|---|---|
| identifier positions inside a `<svelte:element>` start tag | 30 |
| of those, `textDocument/hover` diverges | 16 |
| of those, rsvelte returns `null` | **13** |

So it is **13 of the 30 positions the construct offers in this corpus**, not 13 of 12,043 — melt-ui simply contains one such element (`docs/src/components/motion.svelte:171-195`). Read the rate, not the count: the construct is under-sampled here and common in real components.

`textDocument/definition` and `textDocument/completion` at these positions are **not** classified yet.

baseballyama:
Closes #4102.

## Mechanism

`handle_svelte_dynamic_element` assembled the whole opening tag as a string and applied it with a single `str.overwrite(el.start, opening_tag_end, …)`. `magic-string` emits a segment per character for an **unedited chunk** and exactly one segment for an `addEdit`, so the `this={…}` expression and every attribute value reached the `.svelte.tsx` shadow as generated text sharing the element's start mapping. A `textDocument/hover` inside the start tag then resolved through the nearest mapping to its left and answered `null`.

Upstream never does this: `htmlxtojsx_v2`'s `transform` takes a `TransformationArray` whose entries are strings **or `[start, end]` ranges**, and it moves each range so the source chunk arrives unedited.

## Fix

The opener goes through `build_attribute_segments` + `bake_out_of_order_src` + `emit_segmented_overwrite` — the path `element.rs` already used for plain elements. `ordered_dynamic_suffix` now takes and returns `Vec<Seg>`, so the `class:` / `style:` suffix (already segmented) and the `bind:` suffix keep their own mappings instead of being flattened on the way in. `binding.rs` gains `build_bind_directive_suffix_segs`; the string form stays as a thin wrapper.

`this="div"` deliberately stays generated — the parser stores the bare text with no range, and upstream emits `createElement("div", …)` from that text too.

## Ratchet — both directions, because they are the same size

| | stale (removed) | new (added) |
|---|---|---|
| `differential` `extra-rsvelte` | `count=2` | `count=1` |
| `differential` `missing-rsvelte` | `count=2` | `count=1` |
| `expected` `extra-rsvelte` | `count=4` | `count=3` |
| `expected` `missing-rsvelte` | `count=2` | `count=1` |

×2 for `phase=edit` — **8 removed, 8 added**, all on `upstream-features/svelte-element/input.svelte`, entry total unchanged at **23734**. `count=` is part of the key, so a divergence that shrinks without reaching zero re-keys rather than retiring. Under a net figure alone that is indistinguishable from nothing having happened, which is why both directions are printed. `lsp-mechanisms.json` moves the same 8/8, so the bijection its checker requires holds.

## Re-measured on this tree, not carried over

The eight hashes were computed before two rebases. `git log` says 14 commits touched this fixture's inputs since the branch's old merge base — including #4401 (Svelte 5.57.0) and #4388 (the same `bind:` suffix path) — so a blob comparison could not retire the question.

`pnpm run lsp:verify:fixtures`, release arm, against this branch's own baseline: **all 802 `textDocument/diagnostic` keys in the measured scope reproduce exactly — 0 new, 0 stale.**

## Two arms, one control

The same gate on `origin/main` at `b92218895`, with this change absent, reports an identical churn of **208 added / 208 removed**, every one `textDocument/inlayHint`, over 89 fixtures, with all 208 removed stems having an added twin differing only in the payload — **0 unpaired on either side**. None of the 74 of those fixtures I could locate contains a `<svelte:element>` at all.

That churn is this checkout, not this change. `main` was green on the Language server job at `939672df8`, which already carries the Svelte 5.57.0 bump, and the pinned tsgo here is `7.0.0-dev.20260703.1` exactly as the submodule specifies. **The cause is not identified and nothing here depends on identifying it** — flagged so the next person to run this gate locally does not spend hours attributing it to their own change.

The two arms are distinct binaries rather than one measured twice: their `current` sets differ on exactly 8 keys, and those 8 are the `svelte-element` diagnostic keys at `count=1/3` against `count=2/4`. Note that `divergentFields` reads **2104 on both arms** — a `[count=N]` key is one field whatever `N` is — so the aggregate cannot see this fix, and reading it would have scored the change inert.

## Ablated, not merely run

| | result |
|---|---|
| with the fix | `ok. 11 passed; 0 failed` |
| two source files reverted to `origin/main` | `FAILED. 2 passed; 9 failed` |

The two that still pass are the two named **in advance** as controls rather than discriminators: `a_string_literal_this_stays_unmapped`, because `this="div"` must stay generated either way, and `a_plain_element_attribute_still_carries_its_map_segments`, which exercises the untouched plain-element path and is a regression guard on the shared `build_attribute_segments`. Restored afterwards to blobs identical to `HEAD`.

## Measured before the rebases, and still describing this code

All three files come through both rebases byte-identical — `dynamic_element.rs` `836672a9b`, `binding.rs` `7cd81d60e`, the test `93deb76cb`, pre and post — so these hold of the code as merged:

- The generated TSX is unchanged: 18 `<svelte:element>` shapes × 2 `is_ts` (named slot, `let:`, self-closing, multi-line, `this` after another attribute, `this={"div"}`, spread, each directive kind), digests equal on both arms, 0 errors, and all 36 digests distinct so the sweep discriminates shapes.
- The issue's own table is **6/6 EQ** end to end, both `<svelte:element>` cells matching official's range exactly (`12:22-25`, `12:34-37`).
- The issue additionally claims 3 diagnostics from rsvelte against 0 from official on that file. **That half does not reproduce**: both servers emit the same 3 (`a11y_no_static_element_interactions` twice and 6133), identical item for item. This is the second issue in the LSP cluster whose diagnostic half is wrong in the same direction, #4103 being the first.

## Local checks

`cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --all-features -- -D warnings` exit 0 with 0 warnings and a positive control confirming it reached `rsvelte_projection`; the new test file 11/11 read by name.

## Not fixed here

This is 1 of the 9 start-tag opener ports (#4385). Six others still apply as one overwrite; converting this one leaves those unaffected, and the two ports that were already segmented (`element.rs`, `handle_component`) are the live positive controls that keep that grid meaningful.


baseballyama:
Not my PR — posting one measurement because it is invisible from the PR page.

**This PR's green is a verdict about a tree 17 merges old.** Its newest check per name started `2026-09-07T21:09:06Z`; `main`'s most recent merge is `2026-09-08T07:36:23Z` (`bf815241a`). The branch has not been pushed since, so nothing recreated the checks and the rollup still reads `total=62 done=62 bad=[]`.

```
head            44fc1a07c
merge base      b92218895
origin/main     bf815241a   (17 commits ahead of the base)
newest verdict  2026-09-07T21:09:06Z
last main merge 2026-09-08T07:36:23Z
```

Two of the 17 are in this PR's own neighbourhood rather than merely nearby: `2c7d0f345 fix(svelte2tsx): a generic arrow's type-parameter list is copied verbatim (#4431)`, which is the same start-tag/segment machinery, and `9e2a02c9d fix(parse): keep TypeScript type fields on parse() output (#4430)`, which changes what the projection reads. Whether either interacts is **unmeasured** — I am reporting the staleness, not predicting a conflict.

Worth stating which direction this is dangerous in. A branch that goes red when `main` moves announces itself; one that stays green does not, and the green here is not evidence about today's tree either way. A push (rebase or empty) recreates every check, and the two-sided ratchets are exactly the gates that can flip from a `main` move without this branch changing a byte.

PR #4439, opened 2026-09-08
fix(parse): a leftover bracket the expression wrapper closes is acorn's, not the wrapper's

### LINKED ISSUE #4387: A leftover token the `(…)` expression wrapper can absorb reports the wrapper's error, not acorn's
## Summary

#3350 established the rule: upstream parses one maximal expression and reports whatever is left over as `expected_token` at the leftover token; everything else is `js_parse_error` at `err.pos`. It was closed, and the rule holds for every leftover token **except a closing bracket**, because that is the one token the `(…)` this parser wraps the body in can absorb.

Two divergences, both on `main` (`1512bcdcf`), both `client` and `server`, measured against `submodules/svelte/.../compiler/index.js`.

## 1. `{a)}` — different code *and* different position

```svelte
<script>let a;</script>
<p>{a)}</p>
```

| | code | message | position |
|---|---|---|---|
| official | `expected_token` | `Expected token }` | 1:35 |
| rsvelte | `js_parse_error` | `Unexpected token` | 1:36 |

Upstream stops after `a`, sees `)` where it wants `}`, and reports `expected_token` at the `)`. rsvelte wraps the body as `(a)\n)`, which parses `(a)` and then trips on the *second* `)` — a token that does not exist in the source — so both the classification and the column come from the wrapper.

## 2. `{(a}` — same code, different message

| | code | message | position |
|---|---|---|---|
| official | `js_parse_error` | `Unexpected token` | 1:36 |
| rsvelte | `js_parse_error` | ``Expected `)` but found `EOF` `` | 1:36 |

The position agrees; the text is OXC's rather than acorn's. Same wrapper: the unbalanced `(` in the body consumes the wrapper's own `)`, so OXC reports EOF where acorn reports the delimiter.

## Live controls — #3350's rule is right everywhere else

Twelve expression bodies in the mustache slot, both compilers, `client`:

| body | verdict |
|---|---|
| `a b` | SAME (`expected_token` 1:36 both) |
| `a;` | SAME (`expected_token` 1:35 both) |
| `a]` | SAME (`expected_token` 1:35 both) |
| `a +`, `)`, `a,`, `a=>`, `a.`, `a??` | SAME (`js_parse_error`, same position) |
| `a}` | SAME (both compile) |
| **`a)`** | **DIFF — code and position** |
| **`(a`** | **DIFF — message** |

`a]` is the discriminating control: a leftover **square** bracket classifies correctly, because the wrapper is round. So the axis is not "a leftover bracket" but "a leftover bracket the wrapper itself can consume".

## Reach

Not measured against the collected corpus, and expected to be 0 — published code compiles, so a stray `)` inside `{…}` does not occur. This is generated-input territory: AGENTS.md records the error `code` field as saturated at 0 divergences over 2,843 both-reject corpus pairs, which is exactly why a case like this can only arrive from a constructed input.

## Related

- #3350 (closed) — the rule and its five original rows.
- #4084 / #4083 — the same `(…)` wrapper producing diagnostics that are the wrapper's rather than the body's, in the accept direction rather than the message direction.

baseballyama:
Closes #4387.

#3350's rule — acorn parses one maximal expression and reports whatever is left
over as `expected_token` at that token, everything else as `js_parse_error` at
`err.pos` — held for every leftover token **except a closing bracket the probe's
own wrapper can consume**.

`trailing_token_offset` wraps the mustache body in `(…)`, so `{a)}` read as the
complete `(a)` followed by the probe's own `)`: the leftover moved past the end
of the content and the classification fell through to `js_parse_error`, at a
column that exists only in the wrapper. `{a]}` was already right, and that is the
discriminating control — the axis is not "a leftover bracket" but "a leftover
bracket **this wrapper closes**". The probe now runs with both bracket pairs and
takes the earlier leftover, so each covers the other's blind spot.

The second half is the message. An expression that runs out of input has no
offending token for acorn to name, so `parseExpressionAt` reports `Unexpected
token`; OXC names the delimiter it wanted (`` Expected `)` but found `EOF` ``).
The position already agreed.

## Measured

A 49-body probe in the mustache slot, both compilers, arm-vs-arm against the
oracle at `VERSION === '5.57.0'`:

```
moved = 10    DIFF -> EQ = 10    EQ -> DIFF = 0
```

Every moved cell:

| body | before | after / official |
|---|---|---|
| `a)` | `js_parse_error` `Unexpected token` 2:6 | `expected_token` `Expected token }` 2:5 |
| `a))`, `a)]` | `js_parse_error` `Expected a semicolon…` 2:6 | `expected_token` 2:5 |
| `(a`, `((a`, `f(a`, `f(a,`, `new Map(`, `a.b(` | `` Expected `)` but found `EOF` `` | `Unexpected token` |
| `function(` | `` Expected `{` but found `EOF` `` | `Unexpected token` |

Both of the issue's rows are among them, and the fix reaches eight neighbours the
issue did not report.

## Tests

`crates/rsvelte_core/tests/wrapper_absorbs_leftover_bracket_4387.rs`, 4 tests, 21
cells, every expectation generated by running the source through the pinned
official compiler rather than written by hand.

* `every_cell_matches_the_official_compiler` — 19 error cells plus one that must
  compile.
* `the_grid_holds_all_three_verdicts` — a build answering `js_parse_error` for
  everything, `expected_token` for everything, or rejecting everything must fail.
* `both_bracket_shapes_are_present_in_the_leftover_rows` — `a]` **and** `a)`, so
  a wrapper blind to one of them cannot pass a grid holding only the other.
* `the_residue_is_still_the_residue` — see below.

## The residue is measured, not assumed

Of the 49 probed bodies, 15 still differ from official and this change moves none
of them. Two are the closest neighbours of the fix and are pinned in the same
file rather than left implicit, because they say where the fix stops:

```
[a)      rsvelte `Expected `,` or `]` but found `)`` 2:6   official `Unexpected token` 2:6
{a: 1)   rsvelte `Unexpected token` 2:13                   official `Unexpected token` 2:9
```

These are a leftover `)` the wrapper **cannot** reach, because an opener in the
body claims it first — a different mechanism, whose message comes from inside an
unfinished array or object. The pin is two-sided, so a later fix has to come
through this file rather than silently deleting the row. Widening the message
map to cover them would be exactly the enumeration-from-observed-shapes hazard
this repository records, so they stay listed.

## Corpus sweep

`MOVED 0 / 135,560` (33,890 `.svelte` × 4 targets), swap positive control `1`,
both arms' own `Compiling rsvelte_core (<path>)` lines read, tree restored
byte-identical to the fix arm after the swap.

That zero is not a dead instrument twice over. The harness folds the error text
into a rejected unit's payload on purpose (`zz_sweep.rs:105-113`, "a change in
WHICH error is reported shows up, rather than folding every rejection onto one
hash"), so the reject population — the only one this change can reach — is
**inside** the 135,560, not outside it. And the two arms demonstrably differ: the
49-cell probe above separates them on 10 inputs.

The reach is 0 because published code compiles: a stray `)` inside `{…}` does not
occur in the collected corpus. `AGENTS.md` records the error `code` field as
saturated at 0 divergences over 2,843 both-reject corpus pairs, which is exactly
why this class can only arrive from a constructed input.


baseballyama:
`Test (ubuntu-latest, 1)` was real and it was **an existing expectation pinned to rsvelte's own answer rather than the oracle's**, which this branch's fix corrects.

```
empty_expression_message_3319::a_body_with_code_in_it_keeps_its_own_message
  assertion `left == right` failed: for "{f(}"
   left: "Unexpected token"
  right: "Expected \`)\` but found \`EOF\`"
```

`svelte.compile` at `VERSION === '5.57.0'` answers `js_parse_error` / `Unexpected token` at 3 for `{f(}`. So `left` is the oracle and `right` was the pin — and that file's own header says *"Code, message and both endpoints below were measured against `svelte.compile`"*, which was true of its other tests and not of this one. All four cells are regenerated from the oracle and the assert is widened from the message alone to `(code, message, start, end)`; a right message at a wrong position used to pass.

**Two lexical run-outs are pinned as residue, not fixed, and the arms say why.** I added `` `a `` and `'a` as negative controls for the ``ends_with("but found \`EOF\`")`` rewrite, they went red, and the reading that presents itself is *the fix over-reaches*. Two arms, built by editing out only that one line:

| body | ablated arm | this branch |
|---|---|---|
| `` `a `` | `js_parse_error` / `Unexpected token` / 33 | **same** |
| `'a` | `js_parse_error` / `Unexpected token` / 33 | **same** |
| `(a` | `js_parse_error` / ``Expected \`)\` but found \`EOF\``` / 30 | `js_parse_error` / `Unexpected token` / 30 |
| `a b` | `expected_token` / `Expected token }` / 30 | **same** |

Only `(a` moves. The two lexical rows are byte-identical across the arms, so they predate the branch — official says `Unterminated template` at 2:5 and `Unterminated string constant` at 2:4 while rsvelte says `Unexpected token` at 2:9 for both, wrong message *and* wrong position, and OXC's own wording never reaches `check_js_parse_error_with_pos` at all, so there is nothing there for this remap to rewrite. They are now two-sided `RESIDUE` pins with the official answer in the comment.

The generalizable half: **a red on a cell you have just added is indistinguishable from a red the fix caused**, and re-reading the red cannot separate them — only an independent measurement of the other candidate author can. Here that is the ablated arm, and it is cheapest to build the moment the fix is written, before there is a red to explain.

Locally green at `58e37ae28`: `wrapper_absorbs_leftover_bracket_4387` 5 passed, `empty_expression_message_3319` 4 passed, plus `tag_expression_parse_errors_3202` 9 and `js_parse_error_position_3700` 4; `compiler_errors`, `parser_fixtures`, `early_errors_3243` and `sourcemaps_gate` (`808/808` official segments, 0 missing, 0 wrong) all pass after initializing `submodules/svelte` and generating fixtures.

PR #4443, opened 2026-09-08
fix(parse): a destructured parameter's default keeps the span the source wrote

### LINKED ISSUE #4432: parse: the expression path's `offset - 1` base underflows at offset 0
### What

`parse_expression_with_typescript` spells an absolute position as `offset + span - 1`, so the
base handed to the TS conversion subtree is `offset - 1`. That value is **-1 when the entry
point is called at offset 0**, and `usize` has no spelling for it — the subtraction underflows.
Debug panics; release wraps and emits a span nobody reads back.

### Measured on `main` (8916f851), debug, `parse_expression_with_typescript(arena, src, 0, .., true)`

| cell | verdict |
|---|---|
| `work()` / `new C(1)` / `f(g(1))` / `[1, 2].map(f)` | ok |
| `() => 1` / `a + b` / `o.m(1)` / `({ x: 1 })` | ok |
| `function q(a) { return a; }` | ok |
| `f<number>(1)` / `new C<number>(1)` | ok *(main emits no type arguments)* |
| `class K<T> extends M<number> { p: number = 1; }` | ok *(same reason)* |
| **`(v) => v`** | **PANIC** |
| **`(v: number) => v`** | **PANIC** |
| **`o as string`** | **PANIC** |

So the class is not TypeScript-specific: an **arrow with any parameter list** reaches it, via
`convert_formal_parameter(arena, param, offset - 1, ..)`.

### Reach

The two production callers (`parse_expression`, `parse_expression_with_end`) receive an
`offset` derived from a position inside a component, which is never 0 — there is always a `{`
or a `<script>` ahead of it. Both are `pub` within the crate, so this is a property of today's
callers rather than of the signature.

### Why it is not a one-line fix

The base genuinely is `-1`; laziness only defers it. Making it signed means changing
`adjusted_offset: usize` on 12 signatures **and** the ~52 arithmetic sites under them — and the
boundary is ragged, because the same subtree also calls converters that take the *unadjusted*
`offset` (`convert_expression`, `convert_ts_type`, `convert_formal_parameter`), so those bridge
sites have to be decided one at a time. Attempted as part of #4335 it reached 81 compiler errors
and was reverted as out of scope rather than half-applied.

### Note

#4335 narrows this slightly in one direction and widens it in another: it makes the three
`opt_type_*` helpers take their base lazily (so a node carrying **no** TS construct no longer
evaluates `offset - 1` at all — that regression broke
`function_body_directives_remain_expression_statements`), while newly routing type-argument,
type-parameter and type-annotation nodes through the same base, so the three `ok` rows above
marked *"main emits no type arguments"* become panics at offset 0.

baseballyama:
## What

A template expression is parsed inside `wrap_for_parse("(", content, ")")`, so a node at
relative `p` sits at `base + p - 1`. **Two conventions for carrying that one byte live in
`1_parse/read/expression.rs`**, and they are never named:

- the twelve pattern converters take a base with the wrapper **already subtracted**
  (`convert_formal_parameter(arena, param, offset - 1, …)`), then add spans to it;
- `convert_expression` takes the **unsubtracted** base and removes the wrapper from the span
  itself (`offset + id.span.start as usize - 1`).

They meet at four places. **One of the four converted between them** —
`convert_formal_parameter_inner` passes `adjusted_offset + 1`, with no comment saying why —
and the other three did not. So every destructured parameter default was reported one byte
early. A second instance of the same class sits in `create_function_expression`, which is
written entirely in the paren convention (`id` and the rest element both spell `- 1`) and
handed its parameters the raw `offset`, one byte late.

Neither is visible in generated code; both are in the public `parse()` AST, which
`svelte2tsx`, `rsvelte-lint` and the language server all read.

## Measured against the pinned oracle

`submodules/svelte`'s own `parse()`, 23 cells, spans asserted as the source text they slice:

| | before | after |
|---|---|---|
| destructured parameter defaults (16 cells) | **14 wrong** | 0 wrong |
| function/arrow parameters (7 cells) | **4 wrong** | 0 wrong |

Before, the defaults sliced to `" "` and the parameters to `")"` — one byte off in opposite
directions, which is what named the two conventions rather than one bug:

```
                    official                    rsvelte (before)
obj-default         62..63="1"                  61..62=" "
default-is-arrow    62..69="() => 1"            61..68=" () => "
fnexpr-param        38..39="a"                  39..40=")"
fnexpr-two          38..39="a",41..42="b"       39..40=",",42..43=")"
```

Two cells were already correct and are kept as controls, because a fix that applies one
convention everywhere passes a table that contains only the broken shapes: `plain-default`
(`(a = 1) => a`) is the site that already compensated, and `each-pattern` reaches a different
entry point entirely.

## What changed

`AdjustedOffset { base, prefix }` replaces the pre-subtracted `usize` on the twelve
converters. It hands out only sums, so the two conventions become two named accessors
(`for_paren_converter`, `for_plain_converter`) instead of a `+ 1` at one call site and nothing
at three. The three missing conversions are then a type error rather than an off-by-one, and
`create_function_expression` gets the `- 1` its neighbours already had.

## What this does NOT close

#4432 is the same file's other symptom — `offset - 1` underflows when the entry point is
called at offset 0 — and it is **narrowed, not closed**. The twelve converters no longer
evaluate a negative base, so `(v) => v` stops panicking; `(v: number) => v` and `o as string`
still reach `convert_ts_type`, which takes a `usize` base and adds to it in checked
arithmetic. Giving that converter the same type cascades: measured, its signature alone takes
the build from 4 errors to 44, and its callers are themselves `offset: usize` functions. That
is a separate change and #4432 stays open for it. No production caller passes 0 — both
`parse_expression` and `parse_expression_with_end` derive their offset from a position inside
a component, which always has a `{` or a `<script>` ahead of it.

## Verification

- 23 cells against the pinned oracle, above; expectations generated by running that oracle,
  not inferred from rsvelte's output.
- `crates/rsvelte_core/tests/param_pattern_spans_4432.rs`, with a liveness test asserting both
  tables carry both hosts — a table of only-arrow cells cannot separate the two conventions.
- 16 suites, 2041 tests, 0 failed — `--lib` (1969) plus every parser, span, snippet and
  `compiler_errors` target. `parser_fixtures` printed its own denominators (modern 28/28,
  legacy 83/84, the one incompatible sample being the pre-existing documented case).
- **Two-arm corpus sweep, `MOVED=0`.** 33,891 sources x 4 targets over
  `compatibility/sources`: `rows=135560 distinct=135560 unparsed=0 live=130585 dead=4975`,
  0 hashes differing. That is the wanted answer rather than a null result — these spans are
  read by `parse()` consumers (svelte2tsx, the linter, the language server) and not by
  codegen, so a compiled-output sweep is the control that says the change is confined.
  Arms identified by provenance rather than by label: the fix commit is an ancestor of the
  fix arm and **not** of the base arm, the two `corpus_hash` binaries hash differently, and
  `git diff` between the two trees is exactly the two files above.

### DIFF
--- a/.changeset/param-pattern-span-conventions.md
+++ b/.changeset/param-pattern-span-conventions.md
@@ -1,0 +1,11 @@
+---
+'@rsvelte/compiler': patch
+---
+
+`parse()` reports a destructured parameter's default value and a `function`
+expression's parameters at the spans the source wrote. A template expression is
+parsed inside a `(`-wrapper, and the two ways this file carries that one-byte
+shift — pre-subtracted into the converter's base, or subtracted from the span
+by `convert_expression` — meet at four places, of which one converted between
+them. Every destructured default (`{(({ a = 1 }) => a)({})}`) was reported one
+byte early and every `function` expression parameter one byte late.
--- a/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
+++ b/crates/rsvelte_core/src/compiler/phases/1_parse/read/expression.rs
@@ -2728,8 +2728,12 @@ pub fn parse_typescript_params<'a>(
             if let Some(rest) = &arrow.params.rest {
                 let rest_start = (offset - 1) + rest.span.start as usize;
                 let rest_end = (offset - 1) + rest.span.end as usize;
-                let argument =
-                    convert_binding_pattern(arena, &rest.rest.argument, offset - 1, line_offsets);
+                let argument = convert_binding_pattern(
+                    arena,
+                    &rest.rest.argument,
+                    AdjustedOffset::wrapped(offset, 1),
+                    line_offsets,
+                );
                 p.push(Expression::from_node(JsNode::RestElement {
                     start: rest_start as u32,
                     end: rest_end as u32,
@@ -5853,7 +5857,12 @@ fn create_function_expression<'a>(
         // The whole conversion is in the paren-wrapper's coordinates, as the
         // `id` and rest spans above and below already are.
         .map(|param| {
-            convert_binding_pattern(arena, &param.pattern, offset.wrapping_sub(1), line_offsets)
+            convert_binding_pattern(
+                arena,
+                &param.pattern,
+                AdjustedOffset::wrapped(offset, 1),
+                line_offsets,
+            )
         })
         .collect();
     // Handle rest parameter (`...args`).
@@ -5863,7 +5872,7 @@ fn create_function_expression<'a>(
         let argument = convert_binding_pattern(
             arena,
             &rest.rest.argument,
-            offset.wrapping_sub(1),
+            AdjustedOffset::wrapped(offset, 1),
             line_offsets,
         );
         params.push(JsNode::RestElement {
@@ -7563,15 +7572,24 @@ fn convert_binding_pattern_for_decl_as_node(
                 expr_to_node(create_identifier(&id.name, start, end, line_offsets))
             }
         }
-        oxc_ast::ast::BindingPattern::ObjectPattern(obj_pat) => {
-            convert_object_pattern(arena, obj_pat, offset - 1, line_offsets)
-        }
-        oxc_ast::ast::BindingPattern::ArrayPattern(arr_pat) => {
-            convert_array_pattern(arena, arr_pat, offset - 1, line_offsets)
-        }
-        oxc_ast::ast::BindingPattern::AssignmentPattern(assign_pat) => {
-            convert_assignment_pattern(arena, assign_pat, offset - 1, line_offsets)
-        }
+        oxc_ast::ast::BindingPattern::ObjectPattern(obj_pat) => convert_object_pattern(
+            arena,
+            obj_pat,
+            AdjustedOffset::wrapped(offset, 1),
+            line_offsets,
+        ),
+        oxc_ast::ast::BindingPattern::ArrayPattern(arr_pat) => convert_array_pattern(
+            arena,
+            arr_pat,
+            AdjustedOffset::wrapped(offset, 1),
+            line_offsets,
+        ),
+        oxc_ast::ast::BindingPattern::AssignmentPattern(assign_pat) => convert_assignment_pattern(
+            arena,
+            assign_pat,
+            AdjustedOffset::wrapped(offset, 1),
+            line_offsets,
+        ),
     }
 }
 
@@ -9982,7 +10000,7 @@ fn convert_statement_for_program(
                     arena.alloc_js_node(convert_binding_pattern(
                         arena,
                         &param.pattern,
-                        offset,
+                        AdjustedOffset::plain(offset),
                         line_offsets,
                     ))
                 });
@@ -11450,7 +11468,8 @@ fn convert_variable_declarator_for_program(
     // carrying the annotation as an opaque boundary blob; an annotated
     // destructuring pattern (`let { a }: T = …`) keeps the Value (Raw) form
     // since the annotation hangs off a pattern node.
-    let id_pattern = convert_binding_pattern(arena, &decl.id, offset, line_offsets);
+    let id_pattern =
+        convert_binding_pattern(arena, &decl.id, AdjustedOffset::plain(offset), line_offsets);
     let id_node = if let Some(type_annotation) = &decl.type_annotation {
         let ts_start = type_annotation.span.start as usize + offset;
         let ts_end = type_annotation.span.end as usize + offset;
@@ -11708,7 +11727,12 @@ fn convert_expression_for_program<'a>(
                     oxc_ast::ast::ObjectPropertyKind::ObjectProperty(p) => {
                         let prop_start = offset + p.span.start as usize;
                         let prop_end = offset + p.span.end as usize;
-                        let key = convert_property_key(arena, &p.key, offset, line_offsets);
+                        let key = convert_property_key(
+                            arena,
+                            &p.key,
+                            AdjustedOffset::plain(offset),
+                            line_offsets,
+                        );
                         let value =
                             convert_expression_for_program(arena, &p.value, offset, line_offsets);
                         let mut value_node = expr_to_node(value);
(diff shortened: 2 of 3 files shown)


baseballyama:
Rebased onto `bf815241a`. One conflict and one semantic conflict, both from #4430 landing in the same file:

- #4430 moved the `TSDeclareMethod` `returnType` block out of the `is_abstract` arm and into `convert_function_expression_for_program` itself. Resolved to main’s side — the block this branch had edited no longer exists, and the converter now sets the field.
- #4430 also added `opt_type_annotation`, whose doc comment states this defect in words: *“the caller supplies the adjusted offset because the two parse contexts disagree: an expression is parsed from a slice wrapped in parens, so its spans need `offset - 1`, while a program’s need `offset`.”* Its seven callers each spelled the convention by hand, four of them as `|| offset - 1` — the underflowing spelling this PR removes. Its closure now returns `AdjustedOffset` and the callers say which convention they are in.

`opt_type_args` / `opt_type_params` keep the pre-subtracted `usize`: they feed `convert_ts_type`, whose cascade this PR leaves out of scope (measured earlier: changing that one signature takes the build from 4 errors to 44).

Re-verified on the rebased tree — `param_pattern_spans_4432`, `arrow_param_spans`, `parser_fixtures`, `import_export_parser_shapes`, `early_errors_3243`, `param_pattern_dependencies`: 46 tests, 0 failed.

baseballyama:
Registering a prediction before the gate reports it, so the red is not read as a regression — and so it is falsifiable rather than explained afterwards.

**This PR's own `Compiler parity` may fail on the `parse-ast` two-sided ratchet, and `main` will not.** That gate runs `parse-ast-verify.mjs`, whose ratchet fails both directions: a new divergence *and* a listed key that no longer diverges. This PR re-bases the spans of exactly the nodes some listed keys name. On `bf815241a` the ratchet holds 180 keys / 102 distinct field keys, of which 18 name a node this change can move:

```
{legacy,modern}::AssignmentPattern#span      <- the default's own node
{legacy,modern}::FunctionExpression#span     <- the `create_function_expression` half
{legacy,modern}::RestElement#span
{legacy,modern}::ObjectPattern#span
{legacy,modern}::ArrayPattern#span
{legacy,modern}::Property#span
{legacy,modern}::Identifier#span
{legacy,modern}::ArrowFunctionExpression#span
{legacy,modern}::AssignmentPattern.trailingComments#missing
```

If any of those stop diverging the job prints `listed key no longer diverges` and I re-baseline in the same PR, which is what the two-sided ratchet exists to force.

The prediction is not "18 retire" — it is the weaker, checkable one: **the retirements, if any, are a subset of those 18, and no key outside that set moves.** A retirement outside the list would mean this change reaches nodes it has no path to, and I would want to know that before merging rather than after. The list is derived from the ratchet's own keys, not from the files I edited, so it is not a restatement of my own diff.

Two things it does not cover, stated rather than left blank. `#span` folds `start`, `end` and `loc` into one key, so a key that stays listed can still have half its carriers repaired — **the ratchet cannot report a partial retirement**, and I have not measured the carrier counts. And I have not run the gate: this is a set intersection over key names, so it names candidates and not outcomes.

baseballyama:
The two reds now on this PR are **not** the expected one I registered above, and I want that said before the two get conflated — a too-wide "this red is expected" is how a real failure gets waved through.

`Test unit` and `MSRV, package, and docs` both fail on **one** test, and it is mine:

```
the_expression_entry_point_accepts_offset_zero ... FAILED
panicked at .../1_parse/read/expression.rs:13369:25: attempt to add with overflow
Test unit: 3531 passed, 1 failed        MSRV: 1969 passed, 1 failed
```

`create_function_expression` passes `offset.wrapping_sub(1)` into `convert_binding_pattern`; at offset 0 that is `usize::MAX` and `offset + span.start` overflows in a debug build. So the fix for a span convention reintroduced the underflow this PR is named after — by patching a **call site** with a `wrapping_sub` instead of extending the type, which is the anti-pattern the change argues against, committed inside the change that argues it. The cell that caught it is the one written for offset 0 in the first place.

The repair extends `AdjustedOffset` through the five-function binding-pattern family (`convert_binding_pattern`, `convert_object_pattern`, `convert_array_pattern`, `convert_assignment_pattern`, `convert_binding_property`), so a pre-subtracted base is not spellable there either. That also closes a **pre-existing** instance: `expression.rs:2732` spells `offset - 1` into that same family on `main` today, so a rest element in that host underflows at offset 0 before this PR touches anything.

To be explicit about scope, since I registered a prediction two comments up: that one is about `Compiler parity` and the `parse-ast` two-sided ratchet, and it has not run yet. This red is a different failure with a different cause, `main` is not red, and neither comment excuses the other.

baseballyama:
Pushed the repair. `--lib` goes from `1969 passed; 1 failed` to `1970 passed; 0 failed`, and the six parse targets are 46 tests / 0 failed with the `Running tests/…` lines present, so the denominator is six and not a filter that matched nothing.

Three things the repair found that the regression did not.

**Four pre-existing `offset - 1` spellings into this family are on `main` today** — `expression.rs:2732`, and three in `convert_binding_pattern_for_decl_as_node` (7567/7570/7573), a host I had not looked at. Extending the type is what enumerated them; a `wrapping_sub` at my one call site would have left all four live and invisible. **What I am not claiming**: whether any input reaches those four at offset 0 is *unmeasured*. My cells go through `create_function_expression`, not those hosts. The spelling is gone; a latent panic is not demonstrated.

**Adding the pattern cells failed again, in `convert_property_key`, because my own bridge spelled `offset.for_plain_converter()`** — which pre-subtracts. So the fix for the underflow reintroduced the underflow one function over, an hour after the first time. The cell list growing is the only reason it did not ship twice.

**The next cell (`{ a = 1 }`) failed one level deeper and is out of scope, measured rather than judged.** A parameter default resolves through `convert_expression_for_program`, whose base in a paren-wrapped host genuinely is `-1`. Giving that function the type produces **59** further type errors — against the 4 → 44 this PR already records for `convert_ts_type` — so those two cells are *removed* with the 59 written beside them. A cell that cannot go green is worse than no cell: it reads as coverage and can only ever be deleted or ignored.

Every sum is unchanged for a non-zero base — `wrapped(o, 1) + s` is `o + s - 1`, `plain(o) + s` is `o + s` — so this commit moves no span; it moves what is spellable.

Issue #4397, opened 2026-09-06
A one-line erased TS annotation's two comment copies land on two lines where upstream keeps one


baseballyama:
### What

`@sveltejs/acorn-typescript@1.0.10` emits a comment in a speculative type head twice, and rsvelte reproduces that (see #4373 for the model and #2990 for why a divergence whose cause is upstream is still ours). For a **single-line** annotation the two copies are emitted — the count is right — and they land on two lines where upstream keeps them on one.

```svelte
<script lang="ts">
	let a: { /* c */ b: number } = { b: 1 };
</script>
<i>{a.b}</i>
```

```js
// official, client                      // rsvelte, client
let a = /* c */ /* c */ { b: 1 };        let a = /* c */
                                         /* c */
                                         { b: 1 };
```

The multi-line sibling is byte-equal, which is what localises this:

| cell | target | o | r | |
|---|---|---:|---:|---|
| multi-line annotation, initialized | client | 2 | 2 | **EQ** |
| one-line annotation, initialized | client | 2 | 2 | **DIFF** |
| one-line annotation, initialized | server | 2 | 1 | DIFF — #4373, a different mechanism |
| one-line `type` alias declaration | client | 2 | 2 | **EQ** |
| one-line `interface` declaration | client | 1 | 1 | **EQ** |

So it is not "one-line inputs" and not "annotations": it is a one-line **annotation**, where the removal point sits mid-line and the printer breaks after re-emission. The two alias/interface rows are the controls that say so — both are one-line, both are byte-equal.

### Why a count key cannot see it

`o=2 r=2`. Any grid keyed on how many copies of the comment reach the output scores this cell EQUAL. It was found only because the guard compares the whole `js.code` against text generated from the oracle. Recording that here because the obvious cheap instrument for #4373's family is a count, and this is the cell it is blind to.

### What was tried and does not reach it

Changing the separator the re-emission writes (`RSVELTE_ERASED_SEP=space`) does not move it: the break is inserted downstream by the printer, not by the separator. So the fix is printer-side, not in `2_analyze/types.rs`.

### Reach

[...]

Issue #4396, opened 2026-09-06
A comment in an erased TS annotation on an UNINITIALIZED declarator is dropped; upstream floats it to the next node


baseballyama:
### What

When TypeScript erasure removes a type **annotation** on a declarator that has **no initializer**, upstream keeps the comment and rsvelte drops it.

```svelte
<script lang="ts">
	let a: {
		/* c */
		b: number;
	} | null;
</script>
<i>{a}</i>
```

```js
// official, client        // rsvelte, client
let a;                     let a;
var /* c */
/* c */
i = root();                var i = root();
```

Upstream's declaration ends at the identifier, so the comment is not flushed with the declaration — it floats to whatever the printer locates next, which is the `var i = root()` on the client and the template's own first statement on the server. rsvelte emits nothing.

### Why it is not simply "re-emit it like the initialized case"

#4395 re-emits an erased annotation's comment at the removal point, and this is the one shape it deliberately leaves alone. Re-emitting here puts the comment **after the `;`**, and the client's legacy state lowering stops scanning at that point: measured over the corpus, `let stats = $.mutable_source(…)` came out as `let stats;` — text that parses, runs and is no longer reactive. So matching upstream's *count* here is a regression; matching its *placement* needs the comment attached to the following node instead, which is a different mechanism from the one #4395 implements.

That is the whole reason this is a separate issue rather than a residue line in that PR: the fix is not "widen the gate", it is "attach forward".

### Grid

Oracle `submodules/svelte/…/src/compiler/index.js` (`VERSION 5.56.10`), `dev: false`. `o` / `r` are counts of the comment in `js.code`; the verdict is **full `js.code` byte equality**, because a count key cannot separate this from the sibling shape below.

| cell | target | o | r | |
|---|---|---:|---:|---|
| annotation, no initializer | client | 2 | 0 | **DIFF** |
| annotation, no initializer | server | 2 | 0 | **DIFF** |
| annotation, with initializer (control) | client | 2 | 2 | EQ |
| annotation, with initializer (control) | server | 2 | 1 | DIFF — #4373, a different mechanism |
| value position, nothing erased (control) | client | 1 | 1 | EQ |

The second control is the one that stops this being read as "rsvelte drops erased comments": with an initializer the client is byte-equal, so the axis is the initializer's presence and not erasure.

[...]

Issue #4373, opened 2026-09-06
The acorn-typescript comment duplication is modelled per comment on the client and not at all on the server


baseballyama:
### What

`@sveltejs/acorn-typescript@1.0.10`'s `tsLookAhead` does not set `isLookahead`, so a comment consumed while speculatively parsing an object type fires `onComment` twice — once during the lookahead and again after the rewind. rsvelte already models this (`2_analyze/types.rs:64`, `repeated_comment_outputs`), and the model is wrong in two independent ways:

1. **The client repeats each comment in place; upstream repeats the whole speculated run.** For one comment the two rules coincide, which is why nothing has seen it.
2. **The server does not repeat at all.** The only consumer of `repeated_comment_outputs` is `3_transform/client/mod.rs:749`; nothing in the server pipeline reads it.

Per the #2990 rule — byte equality is the goal, so a divergence whose *cause* is upstream still has to be reproduced — both are ours to close.

### Repro

```svelte
<script lang="ts">
type P = {
  // c
  // d
  a: number;
};
let { a }: P = $props();
</script>
<i>{a}</i>
```

```js
// official, client            // rsvelte, client
var // c                       var // c
// d                           // c
// c                           // d
// d                           // d
i = root();                    i = root();

// official, server            // rsvelte, server
// c                           // c
// d                           // d
// c                           let { a } = $$props;
// d
let { a } = $$props;
```

### Grid

Comment sequence in the generated output, `dev: false`, oracle `submodules/svelte/.../compiler/index.js` `VERSION 5.56.10`, arm `origin/main` `8057104f8`.

| cell | client official | client rsvelte | server official | server rsvelte |
|---|---|---|---|---|
| 1 line comment in the head | `c c` | `c c` EQ | `c c` | `c` **DIFF** |
| 1 block comment in the head | `/*c*/ /*c*/` | same EQ | `/*c*/ /*c*/` | `/*c*/` **DIFF** |
| 2 line comments in the head | `c d c d` | `c c d d` **DIFF** | `c d c d` | `c d` **DIFF** |
| 3 line comments in the head | `c d e c d e` | `c c d d e e` **DIFF** | `c d e c d e` | `c d e` **DIFF** |
| 1 in the head + 1 after member 1 | `c c d` | `c c d` EQ | `c c d` | `c d` **DIFF** |
| `interface` body, 1 comment | `c` | `c` EQ | `c` | `c` EQ |

[...]

PR #4425, opened 2026-09-07
fix(compiler): an erased annotation's comment goes ahead of the initializer

baseballyama:
Fixes #4424. **`main` has been red since #4395 landed, and the release gate reads `main` — this is what stopped #4345 from publishing.**

## The defect

#4395 re-emits a comment left inside an erased declarator annotation **at the removal point** — the annotation's own start, immediately after the identifier:

```js
  let tabs// c
 = [];
```

Upstream's printer flushes it ahead of the **initializer**:

```js
	let tabs = $.mutable_source(
		// ) c
		[]
	);
```

The position is not cosmetic. `transform_legacy_state_declarations` searches for the literal `"<keyword> <var> ="`; with the comment between the identifier and the `=` the needle misses, the `$.mutable_source()` wrapping is dropped, and the declaration keeps its raw value while every read and write around it still goes through `$.get(tabs)` / `$.set(tabs, …)`. That component throws at first render — it does not merely print differently.

## Fix

Hold the comment and flush it at the initializer's start. A declarator with **no** initializer keeps #4395's answer: upstream attaches that comment to the next located node, so re-emitting it there would put it after the `;`.

## Why the PR that introduced it was green

The mutate gate samples **2,969** seeds on a PR and runs all **34,902** on `main`. The denominator, not the shape, is what hid it. The four findings are byte-identical in every failing `main` run since `f7d6bae39`.

## Verification

Through the **NAPI binding** (what the gate uses), with the gate's own `codeIdentity` normalization, on both failing seeds × 4 targets:

| arm | mismatches |
|---|---|
| before | **4** |
| after | **0** |

The 4 match the CI report exactly, and the before-arm is the negative control (built by stashing the fix and rebuilding the binding, not by inference).

All eight `COMMENT_KINDS` reproduce and are fixed — `block`, `block-with-brace`, `block-with-paren`, `line`, `line-with-brace`, `line-with-paren`, `line-with-semi`, `svelte-ignore`. The gate picks one kind per seed, so the corpus could only ever show two; the defect is the comment's presence, not a `)` or `}` inside it.

`crates/rsvelte_core` is 697/697 test binaries green, clippy clean. `erased_annotation_comment.rs` (#4395's own suite) still passes, and the new suite pins the comment is **moved, not dropped** — dropping it would satisfy every other assertion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Uyd6xc1EN5Jt4yNWeiWtuy

PR #4447, opened 2026-09-08
fix(client): a comment in a removed `$props()` pattern flushes where upstream flushes it

### LINKED ISSUE #4398: An inline TS annotation on a destructured `$props()` loses its comment on the client — a three-term conjunction
### What

A comment inside an **inline** type annotation on a **destructuring** pattern initialized by **`$props()`** is dropped entirely on the client. Each of those three alone is fine; only all three together lose it.

```svelte
<script lang="ts">
	let { a }: { /* c */ a?: number } = $props();
</script>
<i>{a}</i>
```

```js
var /* c */
/* c */
i = root();                var i = root();
```

### The axis is a conjunction, and each single-term cell is the control that says so

Oracle `submodules/svelte/…/src/compiler/index.js` (`VERSION 5.56.10`), `dev: false`, arm = #4395's tree. `o` / `r` count the comment in `js.code`; the verdict is full `js.code` byte equality.

| cell | destructure | inline annotation | `$props()` | target | o | r | |
|---|---|---|---|---|---:|---:|---|
| `let { a }: { /* c */ … } = { a: 1 }` | ✓ | ✓ | | client | 2 | 2 | DIFF (line placement, #4397) |
| `let a: { /* c */ … } = $props()` | | ✓ | ✓ | client | 2 | 2 | DIFF (line placement, #4397) |
| `type P = { /* c */ … }; let { a }: P = $props()` | ✓ | | ✓ | client | 2 | 2 | **EQ** |
| `let { a }: { /* c */ … } = $props()` | ✓ | ✓ | ✓ | client | 2 | **0** | **DIFF** |

Three of the four reach the re-emission and two of them are already byte-equal or differ only by #4397's line placement. Dropping any one term takes the count from 0 back to 2, so this is not "the destructure path is broken" and not "runes are broken" — it is the cell where all three meet.

### Where it goes

The runes destructure is lowered through the projection path and never reaches the client text fallback that `repeated_comment_outputs` feeds (`3_transform/client/mod.rs`). That is the same one-consumer fact #4373 records for the server; here it costs the comment entirely rather than one copy of it. **Not verified by instrumentation** — it is read off the call graph, and the cheap check is a `#[track_caller]` counter on the fallback showing 0 for this input and non-zero for the alias cell one row up.

### Reach

Source-shape screen over `compatibility/sources` — 33,890 `.svelte` files, controls `lang="ts"` 22,274 and an impossible token 0 — selects **7** files:

```
appwrite-console/src/lib/components/filters/content.svelte
appwrite-console/src/lib/components/multiSelectTable.svelte
cpi-ide/src/lib/components/editor/monaco/SecondaryMonacoEditor.svelte
svelte-commerce/src/lib/theme/default/DefaultHomepage.svelte
svelte-maplibre-gl/extensions/pmtiles/src/lib/PMTilesProtocol.svelte
svelte-maplibre-gl/extensions/terradraw/src/lib/TerraDraw.svelte
threlte/packages/studio/src/lib/components/Portal.svelte
```

All 7 are DIFF against the oracle on **both** client and server on the #4395 arm, and every one of the 14 differences is **comment-only** under acorn's own `onComment` ranges (0 CODE-DIFF, 0 unparseable; positive control an injected code line → CODE-DIFF, negative control a side against itself → comment-only). `ast_equiv_batch` rescues comment-only pairs under `CommentPolicy::Ignore`, which is why the corpus gate is green on all 7 and will stay green through the fix — **the guard has to be a unit test**.

One caution on those verdicts: the screen selects by source text, so a file it selects can be DIFF for a mechanism other than this one. What the 7 establish is that real published components carry the shape, not that this mechanism is the sole cause of each file's divergence.

### LINKED ISSUE #4453: A comment inside a `$props()` destructuring pattern is dropped, or overtakes an earlier comment, when the declaration is removed
### What

A comment written **inside a `$props()` destructuring pattern** is lost when the client lowering removes the declaration. Which way it is lost depends on whether another comment sits ahead of it in the same script:

```svelte
<script>
	let { a /* c */ } = $props();
</script>
<i>{a}</i>
```

```js
// official, client          // rsvelte, client
var /* c */                  var i = root();
i = root();
```

With a sibling comment ahead of it the comment survives and **overtakes** the sibling, so the two come out in reverse source order:

```svelte
<script>
	// lead
	let {
		a,
		// c
		b
	} = $props();
</script>
```

| | official | rsvelte |
|---|---|---|
| comment sequence in `js.code` | `// lead` , `// c` | `// c` , `// lead` |

### The two symptoms are one mechanism, and a grid separates them

Oracle `submodules/svelte/…/src/compiler/index.js` (`VERSION 5.57.0`), `generate: client`, `dev: false`, arm `cc-clean-main.node` @ `b003aae56` (`sha256 d5cb72ce…`), component function name normalized on both sides. Verdict is full `js.code` byte equality; the bracket shows the order the comments appear in.

| cell | official | rsvelte | |
|---|---|---|---|
| block, 1 prop, no other comment | `[/* c */]` | `[(none)]` | DIFF — **dropped** |
| line, 1 prop, no other comment | `[// c]` | `[(none)]` | DIFF — **dropped** |
| line, 2 props, leading script comment | `[// lead, // c]` | `[// c, // lead]` | DIFF — **reversed** |
| line, 2 props, leading + `lang="ts"` alias | `[// lead, // c]` | `[// c, // lead]` | DIFF — **reversed** |
| **control** no comment anywhere | `[(none)]` | `[(none)]` | **EQ** |

The control reaches the same branch (the declaration is still removed) and asserts nothing is invented, so it is not EQ by never entering the code under repair.

### Where it goes

`3_transform/client/ast_state_transform.rs` replaces the declaration's **source range** rather than reprinting it, so when the helper returns nothing the range — and every comment inside it — is deleted with the declaration. Upstream's printer instead flushes a removed node's comments before the next node it reaches. Where the script holds no other comment there is nothing to flush against and the comment is gone.

The **reversal** half of that sentence is weaker evidence than the drop half and is marked as such: that a later flush emits the pattern's comment at the sibling's position is **read off the call graph, not instrumented**. What is measured is the output — official `[// lead, // c]`, rsvelte `[// c, // lead]`, on two cells and on two corpus files. The cheap check is a `#[track_caller]` counter on the flush showing it fires for the pattern's comment on the reversal cells and not on the drop cells.

### This is NOT #4398

#4398's axis is a comment inside an **erased inline type annotation**, and its own four cells are unmoved by any change to the pattern path — measured on both arms:

| #4398's cell | main | fix |
|---|---|---|
| `let { a }: { /* c */ a?: number } = $props()` | DIFF o=2 r=0 | DIFF o=2 r=0 |
| `let { a }: { /* c */ … } = { a: 1 }` | DIFF o=2 r=2 | DIFF o=2 r=2 |
| `let a: { /* c */ … } = $props()` | DIFF o=2 r=1 | DIFF o=2 r=1 |
| `type P = { /* c */ … }; let { a }: P = $props()` | EQ o=2 r=2 | EQ o=2 r=2 |

The two also have disjoint carriers: #4398's reach screen selects 7 files and **neither** of this defect's 2 carriers is among them.

### Reach

Two-arm hash sweep over the collected corpus — 33,890 `.svelte` ids × 4 targets = **135,560 pairs**, `135,560 rows / 135,560 distinct keys` on both arms, live 130,586 / dead 4,974 identical on both:

```
svelte-maplibre-gl/svelte-maplibre-gl/src/lib/layers/CustomLayer.svelte   client, client-dev
svelte-maplibre-gl/svelte-maplibre-gl/src/lib/layers/RawLayer.svelte      client, client-dev
```

Both carry `// Events` and a URL comment between properties of a `}: Props = $props()` pattern, and both are the **reversal** symptom.

### Why the corpus gate is green on it

Those 4 pairs are byte-different on `main` and scored a **pass**: `verify.mjs` hands the differing pairs to `ast_equiv_batch` with **empty argv**, which is `CommentPolicy::Ignore`, so a divergence living only in comments is rescued. The guard therefore has to be a unit test — the same conclusion #4398 and #4452 reach.

baseballyama:
Closes #4453.

**This PR does not close #4398, and it was filed against it by mistake.** #4398's axis is a comment inside an **erased inline type annotation**; this one is a comment inside the **destructuring pattern**. Measured on both arms, all four of #4398's own cells are unmoved (table below), and the two defects have disjoint corpus carriers, so #4398 stays open. #4453 is the defect this actually fixes.

### What changes

`3_transform/client/ast_state_transform.rs` replaces the declaration's **source range** rather than reprinting it, so when `transform_props_destructuring` returns nothing the range — and every comment inside it — is deleted with the declaration. Upstream's printer instead flushes a removed node's comments before the next node it reaches. This carries the pattern's own comments into the replacement, which puts them at that same position.

Only the **pattern's** comments. A comment in an erased type annotation is deliberately not carried: which of those upstream keeps is position-dependent, and carrying them emits text official does not (measured on two shadcn-svelte components). A comment outside the pattern is already emitted by the caller's own re-emission, which runs ahead of this text, so carrying a sibling here would put the two out of source order.

### The two symptoms are one mechanism

Oracle `submodules/svelte/…/src/compiler/index.js` (`VERSION 5.57.0`), `generate: client`, `dev: false`; component function name normalized on both sides; the bracket shows the order the comments appear in `js.code`.

| cell | official | main | fix | |
|---|---|---|---|---|
| block, 1 prop, no other comment | `[/* c */]` | `[(none)]` | `[/* c */]` | DIFF → **EQ** |
| line, 1 prop, no other comment | `[// c]` | `[(none)]` | `[// c]` | DIFF → **EQ** |
| line, 2 props, leading script comment | `[// lead, // c]` | `[// c, // lead]` | `[// lead, // c]` | DIFF → **EQ** |
| line, 2 props, leading + `lang="ts"` alias | `[// lead, // c]` | `[// c, // lead]` | `[// lead, // c]` | DIFF → **EQ** |
| **control** no comment anywhere | `[(none)]` | `[(none)]` | `[(none)]` | EQ, unmoved |

The control reaches the same branch — the declaration is still removed — and asserts nothing is invented, so it is not EQ by never entering the code under repair.

### #4398 is unmoved

| #4398's own cell | main | fix |
|---|---|---|
| `let { a }: { /* c */ a?: number } = $props()` | DIFF o=2 r=0 | DIFF o=2 r=0 |
| `let { a }: { /* c */ … } = { a: 1 }` | DIFF o=2 r=2 | DIFF o=2 r=2 |
| `let a: { /* c */ … } = $props()` | DIFF o=2 r=1 | DIFF o=2 r=1 |
| `type P = { /* c */ … }; let { a }: P = $props()` | EQ o=2 r=2 | EQ o=2 r=2 |

### Corpus sweep

Two arms, one process each, both identified by a discriminating probe on their output rather than by a path:

```
main  cc-clean-main.node  b003aae56  sha256 d5cb72ce2ec8a34e  carries_4398_fix=false  invents_a_comment=false
fix   cc-clean-fix.node   f2ea19784  sha256 22975dbf76a5ae62  carries_4398_fix=true   invents_a_comment=false
```

33,890 `.svelte` ids × 4 targets = 135,560 pairs per arm; `135,560 rows / 135,560 distinct keys` on both; live 130,586 / dead 4,974 identical on both.

```
MOVED = 4 of 135560 pairs        client: 2   client-dev: 2   distinct ids: 2
  svelte-maplibre-gl/…/layers/CustomLayer.svelte
  svelte-maplibre-gl/…/layers/RawLayer.svelte

direction over the moved set (main -> fix):   match -> match: 8
  match -> MISMATCH: 0
  MISMATCH -> match: 0
```

`match -> match` is the right verdict and understates the change: on `main` those 4 pairs are **byte-different and rescued** by `ast_equiv_batch`; with the fix they are **byte-identical**. Measured directly, one arm per process:

```
main  client / client-dev   byte-equal=false   CustomLayer.svelte, RawLayer.svelte
fix   client / client-dev   byte-equal=true    CustomLayer.svelte, RawLayer.svelte
```

Neither file is among the 7 that #4398's reach screen selects.

**What that sweep does and does not read.** It hashes `js.code`. This change moves *comments in generated output*, and a comment shifts every position after it, so **`js.map` is precisely the field it could also move** and no `js.code` hash carries one. A second sweep over `js.map` + `css.code` on the same two arms is running; until it lands the claim is `js.code` **0 regressions over 135,560 pairs**, `js.map` **UNMEASURED**. The field was confirmed live first (`mappings length 2564` on one of the two moved carriers), so a `MOVED 0` from it will be a measurement rather than a sweep over a `null`.

### Both reach margins

A backward scan over all 33,890 ids (seven controls, three positive and four negative, all passing) finds **110 files carrying the source shape**. Compiling those on both arms, `client` + `client-dev`, byte equality against official — 13 throw on both sides, 97 compared:

| main → fix | files |
|---|---:|
| EQ → EQ | 64 |
| DIFF → **EQ** | **2** |
| DIFF → DIFF | 31 |

The fix closes 2 of the 33 diverging carriers. The 31 are recorded as residue and **not** claimed to be this mechanism — a file carrying the shape can diverge for any of #4373 / #4396 / #4397 / #4398 or for something unrelated. Carrying the shape is not sufficient to be a carrier of this defect, which is the same caution #4398's reach section states about its own 7.

### Why the guard is a unit test

`verify.mjs` hands the differing pairs to `ast_equiv_batch` with **empty argv**, which is `CommentPolicy::Ignore`, so a divergence living only in comments is scored a pass for every entry and every target. Re-running the gate's own normalization over `pattern/issues` on each arm confirms it: the same 11 files are byte-different-but-rescued on both, and the same single file would FAIL on both (`dollar-function-parameter`, which is exactly the existing ratchet). A pattern-corpus repro would land green and stay green through a regression.


baseballyama:
Labelled `skip-changeset`. This PR ships no package change — it is a test-only pin, and the check itself exits 0 for a `test(` title.

The failing `Changeset` run is stale rather than wrong: its log records `PR_TITLE: fix(client): …`, the title this PR carried before it was reduced to a pin, and `changeset.yml` sources `PR_TITLE` from `github.event.pull_request.title`, so a `gh run rerun` would replay the same stale payload. The workflow does not list `edited` among its trigger types, so the retitle started no new run either — `labeled` is the one trigger here that both re-runs it and records why.

baseballyama:
**Holding this as a draft: the first test fails on CI, and the reason invalidates the conclusion this PR was reduced to.**

`Test (ubuntu-latest, 3)` on the sibling docs PR — whose merge ref happens to contain this file — reports:

```
props_destructure_comment_4398 a_comment_in_a_removed_props_pattern_lands_where_upstream_puts_it FAILED
  "<script>\n\tlet { a /* c */ } = $props();\n</script>\n<i>{a}</i>\n"
  expected to contain "\n\tvar /* c */\n\ti = root();\n"
  got: … export default function R($$anchor, $$props) { var i = root(); …
```

The comment is absent. So on `origin/main` (`b003aae56`) the shape #4398 reports is **not fixed**, and the claim in this file's header — that #4425 moves every cell and a compiler change written for #4398 is therefore dead — does not survive.

### What went wrong, because it is the interesting part

The claim rested on a two-arm comparison, and **the arm I labelled `b003aae56` is not `b003aae56`**. Measured just now, same input, same option object, one addon per process:

```
/tmp/cc-arm-base.node   (bf815241a)   MISS  nothing follows / line comment / two comments
/tmp/cc-arm-base2.node  ("b003aae56") HAS   all three
CI, Rust `compile()` on main + this file only:  MISS
```

CI is a checkout of `main` plus one test file, made by GitHub rather than by me, and it disagrees with my arm on the same input. The arm was built from a worktree whose working copy still carried the uncommitted compiler change, so what I read as "#4425 already does this" was my own reverted code answering.

`filename` is not the variable — I checked both option shapes and the answer does not move.

The probe I used for arm identity was `let { /* c */ a } = $props()`, chosen because #4425 is about a comment ahead of an initializer. **My own change satisfies that cell too**, so the probe separates the arm from `bf815241a` and does not separate "#4425 is in" from "my fix is in" — which is the one distinction the whole conclusion turned on. A probe answers only the hypothesis it is handed, and I handed it the wrong one.

Two consequences to be honest about rather than tidy away:

- The `0 of 337 real-world carriers / 0 of 3,000 non-carriers` sweep was run between two arms at least one of which is not what its name says. A `MOVED 0` from arms that may not differ measures nothing, and I have no reason yet to prefer either reading.
- Anything I said elsewhere on the strength of "#4425 fixed #4398" — including on #4452 and in messages to other sessions — needs the same correction, and I am posting it there too.

### Next

Build an arm from a pristine `origin/main` checkout, re-run the same rows and the same product, and then decide whether the reverted compiler change is the fix. The test file itself is unchanged and its expected strings came from the oracle, so if the change is restored the rows are already the pin; if it is not, the rows are the reproduction of a live defect and belong to #4398 rather than to a merged PR.

baseballyama:
**Restored the compiler change, and ran the ablation properly this time — on the current base.**

`f2ea19784`, three files: the transform, the test, and a changeset (`skip-changeset` removed, since this ships a fix again).

```
fix in       4 passed; 0 failed
fix ablated  3 passed; 1 failed   <- a_comment_in_a_removed_props_pattern_lands_where_upstream_puts_it
restored     git diff empty, 4 passed; 0 failed
```

The ablation reverts only `ast_state_transform.rs` to `origin/main` and leaves the test file alone, so the red is attributable to the change and not to the harness. `Compiling rsvelte_core (…/wt3/…)` in each run names the tree that was read. This is the control that was run against the wrong tree the first time, and it settles what the arm comparison could not: on `b003aae56`, without this change, the comment is gone.

### What is corrected

The claim that #4425 had already fixed this is withdrawn. Its evidence was a two-arm comparison in which the arm labelled `b003aae56` had been built from a worktree still holding this change, so it was this code answering. `filename` is not the variable; I checked both option shapes. CI on a tree of `main` plus the test file alone reports the comment absent, which is what named the mislabelling.

The probe I used for arm identity was `let { /* c */ a } = $props()`, chosen because #4425 is about a comment ahead of an initializer — and **this change satisfies that cell too**. It separated the arm from `bf815241a` and did not separate the two candidate causes, which was the only distinction the conclusion rested on.

### What is not restored, deliberately

The corpus figures the commit message used to carry — 2 `MISMATCH -> match` over 337 real-world carriers, 0 the other way, 0 of 3,000 non-carriers — were measured with `bf815241a` as the base, before the rebase. They are removed from the message rather than corrected, because at least one arm of the later re-measurement was mislabelled and I will not restate either number as current. The message says so explicitly.

Re-measuring needs two arms built from pristine checkouts (`origin/main` and this head). I have not built them: a peer's `rustc` is at 100% on the shared box and that is not mine to interrupt. Until then, the unit test and its ablation are the evidence, and the interaction worth watching is with #4425 specifically — that commit moves erased-annotation comments, and this change deliberately declines to carry them, so the two touch adjacent decisions.

Staying a draft until the sweep is re-measured on the current base.

baseballyama:
### The `js.map` sweep landed, and it found something the `js.code` sweep structurally could not. **Do not merge this yet.**

I have put the PR back to draft.

**The map moves on 4 files, not 2.** Same two arms, same 33,890 ids × 4 targets, `135,560 rows / 135,560 surviving keys` on both (that guard is now the writer's row count against surviving keys — the previous spelling compared a dict to itself and was vacuous):

```
MOVED = 8 of 135560 pairs        client 4, client-dev 4, distinct ids 4
  svelte-maplibre-gl/…/CustomLayer.svelte        (js.code moved too)
  svelte-maplibre-gl/…/RawLayer.svelte           (js.code moved too)
  svelte-pivottable/src/lib/UI/Draggable.svelte        js.code IDENTICAL on both arms
  svelte-virtuallists/src/comp/PrerenderedArea.svelte  js.code IDENTICAL on both arms
```

So two files' maps move while their generated code does not — reach the `js.code` hash cannot see at any corpus size.

**Direction, on the two files where the comparison is meaningful.** Generated positions can only be compared when the generated text is the same, so only `Draggable` and `PrerenderedArea` qualify (`codeEQ=true` on *both* arms). Partitioning official's generated positions by whether rsvelte maps them to the same source position:

| file | arm | correct | wrong src | missing | spurious |
|---|---|---:|---:|---:|---:|
| PrerenderedArea | main | 175 | 2 | 5 | 16 |
| PrerenderedArea | **fix** | **179** | 1 | **2** | 16 |
| Draggable | main | 28 | 138 | 7 | 50 |
| Draggable | **fix** | 28 | **13** | **132** | 9 |

`PrerenderedArea` improves. `Draggable` **gains nothing correct and loses 125 positions to unmapped** — it trades wrong mappings for no mapping, taking rsvelte from 216 mapped generated positions to 50 against official's 173.

Neither direction is the whole story, and with n=2 comparable files there is no majority to read. That is why this is a hold rather than a note.

**The minimization reproduces the wrong quantity, which is worth recording.** A constructed cell (destructure with a line comment, first statement, long tail) reproduces the *segment-count drop* — 136 → 115, with the no-comment control unchanged at 138 — but its correctness partition goes the **other way**: correct 95 → 100, wrong 15 → 2, missing 3 → 11. So the minimized repro says "better" where the real file says "unchanged and much less mapped". I keyed the minimization on segment count, which is not the quantity that decides anything; the cell is real and it is not this defect.

**Context, measured on `main` and filed separately as #4454.** rsvelte's client map is already far from official's on these files independently of this change — 8.2% of segments in a 288-file corpus sample point past the end of the source line they name (official: 0 of 48,346), and `Draggable` is 28-of-173 correct on `main` too. So this change is moving a map that is already badly wrong; that is context for a decision, not a licence.

**What is still unmeasured.** `server` and `server-dev` map behaviour: the sweep covers all four targets and only `client` / `client-dev` moved, but I have not partitioned the server maps for correctness. And *why* `Draggable` collapses is read off nothing — I have no instrumented cause, only the shape (the destructure is the script's first statement and the carried text is inserted at its start).

baseballyama:
## Retracting the hold: the numbers were right and my reading of them was wrong

I put this back to draft over a "map regression" on `Draggable.svelte`. Re-measured
against the predicate `sourcemaps_gate.rs` actually budgets, **nothing regresses**, and
the quantity I called a regression is one nobody scores.

The earlier partition was run on a *constructed* cell. This one is the four real corpus
ids the `js.map` sweep moved, compiled from the pinned submodule bytes (`wt3` pin ==
`main` pin == the main checkout's HEAD for all three repos), one arm per process.

### The two budgets

`reproduced` = official generated positions rsvelte maps to the same source position.
`oob` = rsvelte segments pointing past the end of their source line (#4454's predicate;
detector positive-controlled by pushing one segment 100000 lines past EOF — it counts it).

| file | target | code==official | reproduced main→fix | oob main→fix |
|---|---|---|---|---|
| CustomLayer | client | **n → Y** | 26/299 → 26/299 | 360/476 → 362/478 |
| CustomLayer | client-dev | **n → Y** | 26 → 26 | 360 → 362 |
| RawLayer | client | **n → Y** | 28/799 → **30** | 1030/1168 → 1032/1170 |
| RawLayer | client-dev | **n → Y** | 28 → **30** | 1102/1240 → 1104/1242 |
| Draggable | client | Y → Y | 28/173 → 28/173 | **5/232 → 0/52** |
| Draggable | client-dev | Y → Y | 28 → 28 | **4/248 → 0/52** |
| PrerenderedArea | client | Y → Y | 175/182 → **179** | 0/208 → 0/212 |
| PrerenderedArea | client-dev | Y → Y | 163/182 → **167** | 0/195 → 0/199 |

official is `0/…` out-of-range on every row, both arms — the negative control for the detector.

**Only four of those rows are comparable.** A map partition keyed on generated position
means nothing when the two sides' generated text differs, and on `main` CustomLayer and
RawLayer differ from official; the fix makes them byte-identical, which is what the
`js.code` sweep found. Over the four comparable rows (Draggable, PrerenderedArea):
**reproduced 394 → 402, out-of-range 9 → 0.** Neither moves the wrong way.

### What I mistook for a regression

On Draggable the transfer is `wrong 138 → 13` and `missing 7 → 132`. I read the second
half alone as "125 positions lost". Both states are failures of the same official
position — a mapping to the wrong source position, or none — and I have no argument that
either is better than the other, so **that transfer is unbudgeted and I am claiming no
direction for it.** `reproduced` is flat at 28 across it; `oob` goes 5 → 0.

### Why Draggable's segment count collapses (232 → 52), measured

This is not the fix's mechanism. Three cells on Draggable, both arms:

| cell | main | fix |
|---|---|---|
| A `// query` in the `$props()` pattern (as-is) | 232 | **52** |
| B comment deleted (control) | 234 | 234 |
| C `// keep` in an ordinary function body (control) | **56** | **56** |

Cell C carries no `$props()` comment and is untouched by this PR, and it already collapses
to 56 on `main`. **A comment reaching rsvelte's generated instance script coarsens the
client map from ~234 segments to ~56, on `main`, independent of this change.** Cell A on
`main` is the outlier — 232 segments *with* a generated comment — and the fix moves that one
file into the behaviour every other comment-bearing component already has. Both arms emit
`// query` at generated line 6 and the generated code is byte-identical, so this is a
difference in position bookkeeping for identical output.

Reach: the whole-corpus `js.map` sweep moved **4 ids of 33,890**, and Draggable is the only
one that loses segments.

### Still unmeasured, and not inferred

- `server` / `server-dev` correctness partitions. The sweep covers all four targets and
  only `client` / `client-dev` moved, but I have not partitioned the server maps.
- *Why* `main` keeps a fine-grained map for cell A when every other comment-bearing program
  gets the coarse one. That outlier is `main`'s behaviour, not this PR's, and I have no
  instrumented cause for it — only the three cells above.

Taking this out of draft.

baseballyama:
## Correcting the magnitude in the comment above

I wrote "a comment reaching rsvelte's generated instance script coarsens the client map from
~234 segments to ~56, on `main`, independent of this change". The **direction** is right and
the **magnitude was read off one file** — Draggable's cell C — which is the failure mode this
repo already records for reading a sign off a family's cells.

Fingerprinted properly: same file with and without one injected `// probe`, 400 components
from the svelte test tree + pattern-corpus, both compilers on an identical file set.

| arm | segments after/before, median | p10 | worse | same | better |
|---|---|---|---|---|---|
| official | 1.000 | 1.000 | 3 | 397 | 0 |
| rsvelte | **0.833** | 0.519 | 274 | 97 | 29 |

So the general cost is ~17% of segments, not 76%. Draggable's 232 -> 52 is **0.22x**, below
the p10 of 0.519 — it is in the tail of the class, not a typical member of it.

An intermediate two-group comparison (comment-bearing files vs comment-free files, 4,955
files, ratio against official's own segment count to control for size) reported medians of
1.06 vs 1.21 and would have been written up as "a small effect". That comparison is between
*different files*; the paired one holds the file fixed and is the one that found it.

**None of this changes the verdict on this PR.** Over the four comparable rows, `reproduced`
goes 394 -> 402 and out-of-range 9 -> 0; the reach of any map movement is 4 ids of 33,890;
Draggable's `reproduced` is flat. What changes is the description of the class this PR moves
one file into, and that class is now filed on its own as #4455 with the oracle control and
the cost against the gate's own predicate (a 2.4pp mean drop in reproduced positions, 84
files worse and 58 better of 395 — smaller and less one-directional than the segment count).

baseballyama:
## What else this moves, and what it does not

Eight cells, both arms, one process per arm, keyed on a **sha256 of the whole generated
output** rather than on an `eq` boolean — an earlier version of this table keyed on a base64
prefix that is `import '…` for every cell, which cannot distinguish "unmoved" from "moved and
still unequal", so its `unmoved` rows were unsupported.

| cell | client | server |
|---|---|---|
| #4448 — comment between `=` and `$props()`, non-destructured | unmoved | unmoved |
| #4396 — comment in an erased annotation, uninitialized declarator | unmoved | unmoved |
| #4453 carrier (positive control, must move) | **n → Y** | see below |
| same shape with no comment (negative control, must not move) | unmoved | unmoved |

**Exactly one of the eight rows differs between the arms**, and it is the positive control.
#4448 and #4396 are byte-identical across the arms on both targets, so neither is touched by
this PR and neither should be closed by it.

### The server row is a candidate, not a finding

On raw bytes the #4453 carrier's server output is unequal to official on both arms, which
looked like an unrecorded server-side half of the same defect. It is not — the difference is
one leading space on the line after the trailing `//` comment:

```
official:  let { children, handle, // query
           	 close } = $$props;
rsvelte:   let { children, handle, // query
           	close } = $$props;
```

Run through the gate's own normalization (`compatibility/.oxfmtrc.json` + blank-line strip):
**EQUAL**. So the raw comparison was stricter than the gate and its non-zero is a candidate
list, exactly as a strict reconstruction's non-zero always is. There is no server defect for
this shape and #4453 stays a client-only issue.

PR #4450, opened 2026-09-08
fix(svelte2tsx): anchor an attribute key's opening quote on the name it opens (#4104)

### LINKED ISSUE #4104: lsp: every definition range's start snaps to the previous mapped token, crossing lines in a multi-line start tag
Every range `rsvelte-language-server` returns from `textDocument/definition` has its **start** snapped back to the start of the previous mapped token instead of to the symbol. On a single-line start tag it is a few characters early; when the tag spans lines it lands on the **previous line**. The end is correct, so the range is not merely offset — it is widened leftwards to swallow whatever precedes the symbol.

## Repro

`src/App.svelte` (workspace: a plain `tsconfig.json`; `initialize` must declare `textDocument.definition.linkSupport: true`, as `scripts/compat-lsp/verify.mjs` does — without it both servers return plain `Location` and the defect does not appear):

```svelte
<script lang="ts">
	const open = true;
</script>

<div class="a" data-open={open}>one-line</div>

<div
	class="a"
	data-open={open}
>multi-line</div>
```

## Actual output, both sides

**`data-open` attribute name, single-line tag (4:17)**

```
official  targetRange/targetSelectionRange/originSelectionRange all {"start":{"line":4,"character":15},"end":{"line":4,"character":24}}
rsvelte   all three                                                {"start":{"line":4,"character":13},"end":{"line":4,"character":24}}
```

The end (`4:24`) agrees; the start is 2 characters early — it sits on the `"` that closes the preceding `class="a"`.

**`data-open` attribute name, multi-line tag (8:3)**

```
official  {"start":{"line":8,"character":1},"end":{"line":8,"character":10}}
rsvelte   {"start":{"line":7,"character":9},"end":{"line":8,"character":10}}
```

The start has crossed onto **line 7**, at the end of `class="a"`.

## Cause

`TsgoOverlay::map_generated_position` (`crates/rsvelte_language_server/src/tsgo_overlay.rs`) falls back, when the `exact_map` segment lookup misses, to

```rust
let previous = index.checked_sub(1).and_then(|i| entry.tokens.get(i))?;
…
let (line, column) = previous.source?;
```

— the **start** of the nearest mapping token at or before the position, discarding the offset inside that token. `map_generated_range` maps `start` and `end` independently through it, so a range whose end happens to land on a token boundary keeps its end and loses its start.

The `#[ignore]`d `projection_roundtrip_identity_measure` in the same file measures the same fallback from the request side (melt-ui: 1,239 of 12,043 requested positions do not round-trip). It **cannot** see this defect, because the offsets in a *response* range are not the offset the request was made at — which is why the round-trip measure reported these positions as `identity`.

## Scale

`submodules/melt-ui`, full run (43 files / 12,043 identifiers / 72,258 requests). Of the 1,771 open-phase `textDocument/definition` divergences, 545 are "same target files, same count, different ranges", and they split:

| shape | count |
|---|---|
| `originSelectionRange` start collapsed | 269 |
| `targetRange` + `targetSelectionRange` collapsed to **zero width** | 106 |
| all three collapsed | 68 |
| all three zero-width | 13 |
| **range collapse, total** | **456** (83.7% of the 545) |
| `targetRange` differs by a `.d.ts` line offset — a different TypeScript lib copy, unrelated | 89 |

The edit-phase set is byte-identical, so each appears twice in `compatibility/lsp-known-failures.json`.

## A second, different defect visible in the same repro

Hovering *definition* on the `class` attribute name: official jumps to `svelte/elements.d.ts:755` (where `class` is declared), while rsvelte returns a self-reference into `App.svelte` spanning `<div class`. That is a target-**resolution** difference, not a range one, and it belongs to the "different target files" bucket (404 of 1,771). Filed here only so the repro is not mistaken for a single defect.

baseballyama:
Closes #4104.

## The issue's mechanism is right; its implied repair is not

`map_generated_position`'s lower-bound fallback really is what produces the reported `(4,13)` — I claimed otherwise on the issue from four sampled lookups and have retracted that there. Columns 78 and 84 interpolate correctly because they *have* segments; column 74 does not, which is a question a sample landing on mapped columns cannot answer.

Printing the whole segment list rather than four lookups, on the issue's own repro, with the generated line byte-identical in both compilers:

```
GEN LINE 7:  { svelteHTML.createElement("div", {   "class":`a`,...__sveltets_2_empty({"data-open":open}),});  }
                                                                                     ^ col 74 = the opening quote

  rsvelte(main)              upstream
  gen 49 -> src (4,13)       gen 49 -> src (4,13)
       (no segment 50..74)        (no segment 50..73)
                             gen 74 -> src (4,15)
  gen 75 -> src (4,15)            (no segment at 75)
  gen 76 -> src (4,16)       gen 76 -> src (4,16)
  ...                        ...
```

One column. Because rsvelte has no segment at 74, the fallback fires, walks back to the previous token at generated column 49, and returns its source start — `(4,13)`, the `"` closing `class="a"`. Interpolating inside that token cannot give `15`: it sits 25 generated columns away at source 13, so interpolation gives 38. The repair belongs in the emitter.

## The fix

`push_attribute_name` emits `"` + the name as a preserved chunk + `"`, and `emit_segmented_overwrite` flushed the preceding literal as one `overwrite(cursor, *s, pending)`. `MagicString` gives an edited chunk a single segment at its own start, so the inserted quote mapped to `cursor`.

The delimiter is now written over the name's own first character. That is what the reference does — and `push_attribute_name`'s doc comment has said so all along: *"upstream `Attribute.ts` overwrites only the name's first character (`contentOnly`)"*. `Seg::LitOpen` marks a delimiter that opens the chunk after it, so the fold is opt-in at the producer.

## The rule is the delimiter, not the character

A first attempt folded *every* literal preceding a chunk. Measured against upstream over 665 language-tools components: **386 fixes against 358 regressions**, with a per-file signature of `fix 26 / reg 26` — it fires where upstream folds and where it does not (upstream folds the key's `"`; it does not fold the value's backtick).

Deriving the rule from upstream's own maps rather than from three cells: of 425 folds upstream performs on that population, **386 are `"` and 39 are `)`** (a different construct, left alone). Keying on the character would still over-apply, because there are **201** places where a quote precedes a chunk because the *user* wrote it — a string literal in their script — and upstream folds none of them. Hence the producer-side opt-in.

## Measurement

Same 665 components, comparing the lower-bound answer at **every generated column of every generated line whose text is identical on both sides** (8,304 lines compared; 2,853 skipped because rsvelte emits `exports: {}, bindings: ""` in the render return and the pinned `submodules/language-tools` build does not — a version skew in the oracle artifact, unrelated to this):

| | |
|---|---:|
| cells compared | 238,443 |
| changed by the patch | 386 |
| **FIXED** (main ≠ upstream, patched = upstream) | **386** |
| **REGRESSED** (main = upstream, patched ≠ upstream) | **0** |

`changed == fixed` means no cell moved to a third value. Generated text is unchanged in all 763 projected files, and the two arms' dumps hash differently, so the zero is not two copies of one binary.

## Tests

Five new cells; two are discriminating. Ablated, `the_opening_quote_of_an_attribute_key_anchors_on_the_name` and `the_first_attribute_in_a_tag_folds_too` fail and the other three pass — as named in advance. `the_name_itself_still_maps_per_character` is a regression guard, not a discriminator: it holds under either anchoring. The two controls are a quote the user wrote (must not fold) and a rewritten attribute name (no chunk to open). Restoring after the ablation left the tree byte-identical.

`cargo test -p rsvelte_projection`: 734 passed, 0 failed, including the 253 svelte2tsx fixtures — which is why neither they nor the map gate (structural well-formedness, not equality against upstream) can see this change.

## Residue

The 39 `)` folds upstream performs are a different construct and are untouched. This addresses the `originSelectionRange` start-collapse class the issue reports; whether it also moves the `targetRange` zero-width class is **unmeasured** here — that population lives in the scheduled `lsp-corpus` run, which a crates-only PR does not admit.

## The LSP ratchet will be red until this PR re-baselines it

`compatibility/lsp-known-failures.json` is not touched yet, so the `Language server` job
is expected to fail on this push. Measured locally as an ABBA control against a clean arm
built at this branch's own merge base (both binaries hashed distinct, `transportTimeouts`
0 on every arm), the residue attributable to the change is:

| direction | n |
|---|---|
| re-keyed (`count=` / `hash=` payload moved, stem unchanged) | 104 |
| genuinely retired | 14 |
| genuinely new | 4 |

The two directions are printed separately on purpose — an over-collection and an
under-collection of the same size are the same total.

The clean arm reports the identical divergence set as an arm built nine commits earlier,
so nothing between them (including #4431) moves this gate; that half was pre-registered
from a 0-carrier count and is now measured. The 4 new keys are one pre-existing,
position-independent divergence on `checkjs-nostrict/component-props-ts` becoming
reachable because the fix pairs the `/items` that were previously unpairable — both
messages recover from their digests to range 8:57-8:68, code 2353, differing only
`string` vs `string | undefined` under `exactOptionalPropertyTypes`. GATES.md 39b.

## The re-baseline

`verify.mjs:59` refuses a direct `--update-baseline`, and `merge-current.mjs` accepts only the
complete 17-artifact union. Both guards exist for the same reason: rewriting the whole ratchet
from a partial measurement deletes every baseline id the run did not measure. That is a property
of the rewrite, not of the file, so the re-baseline here is a targeted edit rather than a
dispatched Corpus Compat run.

What makes it safe is a partition the tooling already asserts. `mergeCurrentArtifacts` rejects a
fixture artifact carrying a key outside `^(differential|expected):(fixtures|upstream-features|upstream-testfiles)/`
and a corpus artifact carrying one outside `aggregate:corpus/`, so the two scopes partition the
ratchet by prefix. The measuring run covered all three fixture suites completely (423 measured ids,
`projectRevision` e78437080 — the artifact names its own tree), and the edit touches only that
prefix. Two postconditions are asserted rather than described, because they fail silently and in
opposite directions: no out-of-scope key was **added**, and the corpus-scope key set is
byte-identical **before and after** — the second is the one that catches a drop.

Measured on that run: 344 stale, 328 new, `24136 -> 24120`. Stripping `[...]` brackets to compare
divergence identities rather than key strings gives **324 re-keyed, 20 retired, 4 new**; the raw
counts alone cannot separate a moved count from a moved defect.

Two controls on the measurement itself. The oracle calibration reproduces CI's exactly —
`official server projects 365/423 … svelte 5.57.0 (58 fail, 13.7%)`, the same line the failing
run printed — which is a positive control on the oracle half that a hash comparison cannot vouch
for. And no request hit the transport deadline, which `verify.mjs` guards precisely because a
timeout changes a run's keys and the next run would then disagree.

**What no local run can settle** is the SUPERSET direction: entries this machine produces that CI
cannot reproduce. Nothing above covers that, and it should not be read as covered. The two-sided
ratchet is what checks it — an over-shrink comes back as stale and an under-shrink as NEW — so
this push is the verification, not a claim to have skipped one.

One note for the reader of the ratchet diff: `Language server` has never passed on this branch.
Its pre-rebase run (34210544210) failed the same way, and its log lists only the first 30 of each
list (`added.slice(0, SHOW)`, `SHOW = 30`) with nothing saying so — so the log's visible entries
are a display sample, not the population.

---

## Correction: the local measurement above was made by an unidentifiable arm

The numbers in the section above came from a local run whose arm I cannot identify, and the first
push of the re-baseline carried 208 of its values. Both are corrected here; the fix itself is
unchanged.

`verify.mjs:171` resolves the rsvelte arm as `path.join(ROOT, "target/debug/rsvelte-language-server")`
— a path — and the artifact records `projectRevision`, `sourceRevisions`, `universeHash` and
`configurationId`, every one of which is a property of `ROOT` rather than of the binary. The binary
that produced the local artifact is gone; `ROOT/target` is empty. So the artifact self-reports
`projectRevision: e784370806` — this PR's fix commit, **correct** — while nothing in it, or beside
it, says what computed the answers.

I attributed the 208 keys to a specific sibling branch's binary and had to retract it. The
attribution was consistent with the only binary still on disk and contradicted by a row of my own
table: the local artifact contains this PR's 116 keys, which a binary lacking this PR's fix cannot
produce, and that worktree's tree is clean and matches `origin/main` on this PR's files. Two
hypotheses remain open and this data does not separate them — a deleted binary carrying this PR's
fix plus another in-flight change, or instability in the inlayHint payloads themselves. The second
is weakly disfavoured, since CI and `origin/main` agree exactly on all 208 across two separate runs,
but it is not excluded. Filed as #4479.

What *is* measured is the decomposition, and it is exact rather than a bound:

```
main -> local artifact   re-keyed 324  new 4  retired 20
                          inlayHint 208, diagnostic 84, documentSymbol 30, completion 2
main -> CI at merge ref  re-keyed 116  new 4  retired 20
                          diagnostic 84, documentSymbol 30, completion 2
CI   -> local artifact   re-keyed 208  new 0  retired 0   inlayHint ONLY
```

`324 - 208 = 116`, with `new`/`retired` identical in both, so the contamination is separable to the
key: these 208 and no others. The independent check is that a key this PR does not move must agree
with `origin/main`, and CI's value does for all 208 while the first push's matched neither.

### What this PR actually moves, measured at the merge ref

From the failing run's own `lsp-current-fixtures` artifact, whose `projectRevision` is the merge ref
CI measured — fixture scope **2104 → 2088**:

| | |
|---|---:|
| re-keyed (same divergence, moved payload) | 116 |
| **genuinely retired** | **20** |
| genuinely new | 4 |

The 20 retired are `textDocument/diagnostic` on `element-attributes`, `generics-runes.v5`,
`ignore-generated-code` and `svelte-element`, across both phases and both arms. The 4 new are one
mechanism on `checkjs-nostrict/component-props-ts` over 2 phases × 2 arms, and are the expected shape
of `GATES.md` 39b: repairing a whole-response divergence makes its child fields comparable for the
first time, so a correct fix can add keys.

### Retracted

The first push's commit message said these payloads "move with the toolchain rather than with the
projection". **Falsified**: `@typescript/native-preview` is pinned exactly
(`~7.0.0-dev.20260703.1`, lockfile-resolved, identical locally and in CI). It was a mechanism stated
as measured, sitting beside four things that were measured, which is how it read as one.

The oracle-calibration control quoted above still passed and still could not have caught this: it
compares the **official** server's projection counts, and the half that moved was rsvelte's. A control
on the oracle certifies the oracle. The check that catches this class costs one command —
`git merge-base --is-ancestor <the fix> <the arm's HEAD>`.


baseballyama:
Status on the three red rows, so the next reader does not re-derive it.

**One root cause.** `Tests` is a rollup — its log is `Test job(s) FAILED:
language-server` and nothing else — and `LSP fixture parity current` and
`Language server` both fail at *Run LSP differential fixture and upstream
gates*. So there is one failing gate, not three.

**It is not a regression on `main`.** The most recent completed `CI` run on
`main` (`3e89b6033`, 10:21Z) is green on the same job, and this PR's run is at
09:41Z on the merge ref — so the movement is this branch's.

**The movement is 108 NEW / 118 stale, and on the visible sample it is mostly a
re-key rather than a behaviour change.** `verify.mjs` truncates each listing at
`--show 30`, so this is 30 of 108 and 30 of 118, not the whole thing. Stripping
the `[count=…,hash=…]` bracket and intersecting the two samples:

| | |
|---|---|
| same key modulo payload bracket (a re-key) | **26** |
| only in NEW | 4 |
| only in stale | 4 |

which is what an attribute-key mapping change should look like: the divergence
survives, its payload moves, the key changes with it. `element-attributes`'
four entries are inside NEW's alphabetical window and absent from it, so those
are genuine retirements.

**Why it cannot be re-baselined from this run.** `mergeCurrentArtifacts`
(`scripts/compat-lsp/artifacts.mjs:102-147`) requires exactly
`CORPUS_SHARDS + 1` artifacts and throws otherwise, and `--update-baseline`
replaces the whole ratchet with `merged.current` — so updating from the
fixture/upstream artifacts alone would delete every `aggregate:corpus/` entry.
Every listed key here is `differential:fixtures/` or
`differential:upstream-features/`, but the update still needs the full
17-artifact union from a `workflow_dispatch` `Corpus Compat` run on this branch.

**Blocked on capacity, not on the code.** Census a moment ago: 19 jobs
in progress against the 20-job account ceiling with **311 queued** (summed over
`runs/<id>/jobs` for every run at run-level `in_progress` *or* `queued` — the
run-level filter alone reports neither number). The dispatch runs when that
drains.

baseballyama:
Notes before anyone diagnoses this PR's red, because two things about it are misleading.

## The verdicts are stale

The three failing checks started at **09:41–10:18Z on 2026-09-08**. `origin/main` has taken
five merges since, the most recent `90a45850d` at **17:23Z** — including `392273f79`
(`fix(svelte2tsx): reproduce upstream's case-sensitive script and style scan`, #4456), which is
itself an svelte2tsx change. So these runs describe a tree six hours and five merges old, and
nothing here should be read as a current measurement. The rebase will recreate every check.

## The failure is not what our notes said

I had this recorded as "118 stale ratchet entries on the fixtures suite". Reading the log, it is
**`[lsp-verify] 108 NEW divergence(s)`** — a different number *and the opposite direction*.
Stale would mean listed entries now pass; NEW means entries the ratchet does not list. Correcting
the record rather than the count alone, because the two want opposite responses.

## What the 108 look like — measured on 27% of them

The verifier prints **30** of the 108 and then stops (a display cap), so the split below is over
30, not 108. Against the ratchet at `e1aa3aa9f`:

| | n |
|---|---|
| already listed verbatim (control — must be 0) | **0** |
| base key already listed, only the `[count=,hash=]` detail moved | **28** |
| base key not listed at all | **2** |

The 2 are one divergence counted in both phases (`opened` and `phase=edit`) on
`upstream-features/checkjs-nostrict/component-props-ts`. The 28 are `textDocument/diagnostic`
(26) and `textDocument/completion` (2) on fixtures that already carry diagnostic keys —
`upstream-features/$$props-usage`, `component-invalid` and `custom-types` hold 8 each today, and
`upstream-features` holds 472 diagnostic keys in total.

So on the sampled quarter this reads as **`[count=,hash=]` churn on already-listed entries**, not
as a wave of new defects — a fixture key carries its payload's digest, so any change to the
diagnostic set produces a *different key* and reads as NEW rather than as a modified entry. The
prominent payload is TypeScript's deprecation diagnostic on svelte2tsx's generated constructors
(`The signature '…' of '$$_skroW0C' is deprecated.`), which is consistent with a mapping change
moving those diagnostics into or out of a region that maps back to the document.

**The other 78 are UNMEASURED.** Two earlier attempts to recover them by grepping the whole log
were contaminated — the first reported 28 keys as "already listed", which is impossible for a NEW
key and is what said the instrument was wrong rather than the world. I am not refining an
extractor against a superseded tree; re-measure after the rebase.

## Rebase

Merge-tree against `e1aa3aa9f` gives **one** conflict hunk, and it is the `use` line in
`attributes/mod.rs` (this branch needs the wider brace list for `segs_push_lit_open` and
`segs_to_string`).

The risk worth checking was not the conflict: this branch adds a `Seg::LitOpen` **enum variant**,
and a match site added on the other side would be a type error rather than a textual conflict —
invisible to merge-tree. `Seg::Lit(` goes from 33 occurrences across 6 files at the merge base to
**71 across 10** at `e1aa3aa9f`. Measured, the risk is zero: the four newly-segmenting files
(`component_slots`, `dynamic_element`, `slot_element`, `special_element`) hold only construction
sites, and every real match arm is in `attribute.rs`, `attributes/mod.rs`, `segs.rs` and
`opener_spacing.rs` — all four already patched here, including `append_segments`, whose match is
exhaustive.

baseballyama:
Status note so the red is not read as a measurement of the current tree.

Every verdict on this PR was taken between **09:41Z and 10:18Z**. The last merge into `main` was **20:33:52Z**, several merges later, and GitHub now reports the PR `CONFLICTING`. A `pull_request` run checks out `refs/pull/N/merge`, so these checks describe a merge with a `main` that no longer exists — the branch has not moved (`73de20db8`), and that is exactly the case where a stale verdict is *not* self-announcing.

So the three failures are not currently attributable:

| check | newest run | failing step |
|---|---|---|
| `Tests` | 10:18:33Z | `Verify all test jobs passed` — a rollup, so it reports its shards' state rather than a test |
| `Language server` | 09:41:09Z | past `Install dependencies` |
| `LSP fixture parity current` | 10:07:19Z | past `Install and build` |

I checked the step lists rather than the conclusions, because a cancelled shard makes the `Tests` rollup red with no test having run. These do not look like cancellations — the shown steps all succeeded and the failures are later — but at a ten-hour-old merge base that distinction cannot be settled here either.

Not rebasing yet, deliberately: this PR touches the LSP ratchet, which re-admits the ~950-job-minute `lsp-corpus` job on a pull request, and the account is at 20 jobs in progress against a ceiling of 20 with 57 queued. Rebasing now would put 16 shards into that queue to measure a tree I already know I have to re-measure. It is queued behind the branches that do not open the hatch.

When it is rebased, the number to re-derive rather than carry over is the `108 NEW divergence(s)` recorded earlier — that was measured against a different merge base, and a NEW-key count is a property of the tree it was measured on.

baseballyama:
Shard 2/16 is red, and it is the **stale** direction of the two-sided ratchet — listed entries that now pass — not a new failure. This is the prediction I pre-registered when I re-baselined the fixture scope only: if corpus-scope movement existed, the corpus half is what would say so. It did.

```
[lsp-verify] 250176/250176 compared, 66034 divergent requests, 0 skipped
[lsp-verify] 22 stale ratchet entry/entries        <- all aggregate:corpus/…|textDocument/definition
```

Every other step of the job succeeded, the artifact uploaded, and the compare step ran 52 minutes — so this is a real verdict, not a cancellation or a lost runner.

**Attribution, from the sidecar rather than from a mechanism story.** Three commits sit between the last green scheduled run (`629cd1cff`, all 16 shards + full-population green) and this head: my `e78437080`, plus main's `c1571c6c0` (LSP document store, symlinked URIs) and `f3193eae1` (client codegen via `esrap/printer.rs`). All 22 stale entries carry **exactly one** mechanism label:

```
projection-origin-range   22/22   (no other label on any of them)
```

That is the source↔generated range mapping — which `e78437080` changes (`svelte2tsx/magic_string.rs`, `template/segs.rs`, `template/attributes/`) and which neither of the other two touches. So this PR fixes them, and `main` is not stale on this axis.

**Pre-registration, before the other 15 shards report.** The ratchet holds **230** solo-labelled `projection-origin-range` entries, all corpus-scope, all `textDocument/definition`. Shard 2's 22 are a strict subset. So **208 candidates remain** across the other 15 shards, split bits-ui 76 / flowbite-svelte 76 / shadcn-svelte 56.

- If the 16 shards total **230**, this fix retires the entire solo `projection-origin-range` population and the re-baseline is that whole set.
- If the total is **< 230**, the fix is file-dependent, and the entries that did *not* move are the interesting ones — I will name them rather than just re-baselining the difference.

Both readings come from the same artifact (each shard's own stale list), which is the half a pre-registration usually gets wrong.

An arithmetic coincidence worth naming so nobody merges the two: **208 also appears in my #4464/#4479 work**, as the count of platform-churning inlayHint keys. Different scope, different method, disjoint sets. Equal cardinality is not identity.

**Plan:** wait for all 16 shards, take the union of their stale lists, and re-baseline the corpus scope in this PR — removing the entries from `lsp-known-failures.json` **and** their records from `lsp-mechanisms.json`, since that sidecar is two-sided and a record naming an unlisted id fails on its own. A partial re-baseline from one shard would simply go red again on the next run, so I am not touching the ratchet until all 16 have reported.

baseballyama:
**My pre-registration above is falsified, and by the second shard.** Reporting it before the rest land, since the whole value of registering it was that it could fail.

I predicted the stale set would be a subset of the **230** solo-labelled `projection-origin-range` entries. Shard 4 broke that:

```
union after 2/16 shards        38
  solo-labelled                36
  MULTI-labelled                2   <- ['projection-origin-range', 'projection-target-position-workspace']
union ⊆ solo                False
```

The premise was wrong, not the arithmetic. **A multi-label entry goes stale when *all* of its divergences resolve**, not only when it carries one label — so a fix that repairs two mechanisms on the same request retires the entry, and I had silently assumed one entry carries one defect. The label set is a conjunction for staleness, exactly as `lsp-mechanisms-check` treats it for attribution, and I have been reading that file all day without applying it here.

**The corrected bound.** `projection-origin-range` appears on **4990** ratchet entries, not 230; 230 is only the subset where it is the *sole* label. So the upper bound on this fix's corpus reach is 4990, and the 230 was an artifact of my own filter:

| | entries |
|---|---:|
| solo `projection-origin-range` remaining | 194 |
| multi-label `projection-origin-range` remaining | 4758 |
| co-occurring labels (top) | `rsvelte-empty` 4224, `projection-target-position-workspace` 4054, `official-empty` 3760, `projection-target-position-declaration` 2724 |

I am **not** replacing one prediction with another. 4990 is a bound, and the run so far (38 after 2 shards) is consistent with anything from ~300 to a few thousand. The honest statement is that I no longer have a prediction, and the reason I no longer have one is worth more than the one I had: **the quantity I could compute (solo labels) is not the quantity that governs (all labels resolving together).**

What does not change: every entry in the union is present in both the ratchet and the sidecar (38/38 each), all are `aggregate:corpus/…|textDocument/definition`, and all carry `projection-origin-range` — so the attribution to `e78437080` stands, and `main` is still clean on this axis. The plan is unchanged: wait for all 16, take the union, remove it from `lsp-known-failures.json` **and** `lsp-mechanisms.json`.

One correction to my own numbers in the same breath: the script that produced this first printed "remaining candidates 192" as `230 − 38`, which is wrong the moment the subset check fails — 2 of the 38 are outside the 230, so 194 solo remain. A derived count computed from a premise the same output has just falsified is worth re-deriving rather than carrying forward.

baseballyama:
Correcting the method I stated above: I said I would take the union of the shards' **printed** stale lists. That source is capped and I have switched off it.

`verify.mjs` prints `removed.slice(0, SHOW)` with `SHOW = 30` (`:61`, `:1118`) and no "… and N more" line, and `--show` appears 0 times in `.github/workflows/` (positive control: `compat-lsp` appears on 8 lines there). So above 30 the log hands you a short list that looks complete, the re-baseline removes a subset, and the next 16-shard run goes red on exactly the entries the cap hid — which reads as a fresh regression rather than as my incomplete edit. Filed as #4484.

It did not bite here, and that is the population rather than the design. Per-shard stale counts on the 8 shards reported so far: `8, 10, 16, 16, 18, 20, 22, 24`. All under 30, with shard 0 the closest.

**New source.** `--write-current` uploads `lsp-current-corpus-<shard>.json` with `if: always()`, so the artifact survives the failing verify. It holds `current` but not `removed` — `report.ratchet = { current, added, removed }` (`:1093`) goes to the report file, which is not uploaded — so `removed` is recomputed with the gate's own exported selector, leaving only `verify.mjs:1091`'s one-line filter reconstructed:

```js
const selected = selectKnownForScope(known, a.suites, a.repos, a.shard);
const removed  = selected.filter((e) => !new Set(a.current).has(e));
```

**Validated against the four shards this run had already ruled on** (2, 4, 11, 14, where the printed lists are complete because each declared count is under the cap): the recomputed set equals the gate's own printed list exactly, 4/4. Negative control: `shard2 ∩ shard4 = 0`, so the reconstruction does not collapse distinct shards into one answer.

**State at 8/16 shards.** Union 134, and 134 rows against 134 distinct keys, so no key collapse. `new = 0` on every shard — there is no NEW divergence anywhere in this run, in either direction of the two-sided ratchet. All 8 artifacts carry one arm identity (`projectRevision eb3729d71`, `configurationId lsp-diff-v12`, one `universeHash`), so the union is not being assembled across two measurements.

Composition of the 134: every key is `textDocument/definition`; every key carries `projection-origin-range` (6 of them also carry `projection-target-position-workspace`); 67 opened / 67 edit, i.e. 67 carriers × 2 phases exactly; 0 keys absent from the ratchet and 0 absent from the sidecar. By repo: bits-ui, flowbite-svelte, shadcn-svelte — melt-ui contributes nothing so far.

Still waiting for all 16 before touching the ratchet, for the reason I gave: a partial re-baseline simply goes red again on the next run.

baseballyama:
Rebased onto `main` and re-baselined from the complete 16-shard run.

**The union is 240**, from all 16 shards — every shard reported, all 16 artifacts present, and per-shard stale counts `4, 8, 8, 10, 10, 14, 14, 16, 16, 18, 18, 18, 20, 20, 22, 24`. Composition: every key is `textDocument/definition`; 120 carriers × 2 request phases exactly; every key carries `projection-origin-range`, 10 of them also `projection-target-position-workspace`. By repo: bits-ui 92, flowbite-svelte 86, shadcn-svelte 62; melt-ui contributes none.

**`new = 0` on every one of the 16 shards**, so nothing moves in the adding direction of the two-sided ratchet. All 16 artifacts agree on `projectRevision`, `configurationId` and `universeHash`, so this is one measurement rather than an assembly of several.

## The rebase, and why the conflict could not be resolved by picking a side

`main` had moved 4 commits, and #4476 re-baselined this same ratchet — so the branch was carrying an arm and a ratchet from two different trees. Measured before acting: the two re-baselines are **disjoint**. All 240 of these entries were still listed in `main`'s ratchet, and `main`'s own churn (132 removed / 138 added) is entirely `diagnostic`/`documentSymbol`/`hover`/`completion`. Corpus scope was unchanged at 22032 with identical per-repo counts, so #4476's movement is entirely `differential:`/`expected:` scope.

Both JSONs auto-merged (all four add/remove intersections between the two edits are 0, so the merge is clean rather than lucky). `KNOWN-FAILURES.md` conflicted on three gated numbers, and **neither side was correct for the merged tree** — `main` says 24126, this branch said 24120, the merge is **24110**. All three were derived from the merged JSON, then each was confirmed live by injecting a wrong value and requiring the checker to go red; all three do, and the restore is byte-clean. Passing was not evidence they were being read.

Rebase verification: `main` is an ancestor; all 7 code and changeset files byte-identical to the pre-rebase state; merged ratchet **set-equal** to the count predicted beforehand (24136 −136 +120 −12 +2 = 24110), not merely count-equal.

## Final state

```
ratchet   24110 -> 23870     sidecar 24110 -> 23870 (same order)
key kind  22032 + 1744 + 334 -> 21792 + 1744 + 334
phase     12060 + 12050      -> 11940 + 11930
repos     3662 + 7672 + 258 + 10440 -> 3570 + 7586 + 258 + 10378
```

`known-failures-md-check` and `lsp-mechanisms-check` both pass; `attribution-check`'s 3 problems are pre-existing and identical on the untouched tree. The `mechanisms` label table is left alone — the checker has no "every declared label is used" rule, and both labels here survive with 4848 and 5784 carriers remaining anyway.

The stale list came from the artifacts rather than the shard logs, for the reason in the previous comment (`SHOW = 30`, #4484). The largest shard here is 24, so the cap did not bite — that is the population, not the design.

PR #4461, opened 2026-09-08
fix(svelte2tsx): apply the six remaining start-tag openers as segments (#4385)

### LINKED ISSUE #4385: lsp: six of the nine start-tag opener ports still apply as one overwrite, so every expression inside them is unmapped

baseballyama:
Fixes #4385.

`<svelte:element>` (#4102) was one host of a class. Nine call sites build a start
tag with `format!` and apply it with a single
`str.overwrite(el.start, opening_tag_end, &opener)`. `magic-string` emits exactly
one segment for an `addEdit` and a segment per character for an unedited chunk, so
every expression inside such a tag shares the element's start mapping, and a
request inside the tag resolves through the nearest mapping to its left. Upstream
never does this: `htmlxtojsx_v2`'s `transform` takes a `TransformationArray` whose
entries are strings **or `[start, end]` ranges**, and it `move`s each range so the
source chunk reaches the shadow unedited.

Six ports move to the path `element.rs` has used for plain elements since the
structured bake landed — `build_attribute_segments` + `bake_out_of_order_src` +
`emit_segmented_overwrite`:

| file | ports |
|---|---|
| `inline_component.rs` | `<svelte:component>`, `<svelte:self>` |
| `special_element.rs` | `<svelte:body\|window\|document\|head>`, boundary snippet props |
| `slot_element.rs` | `<slot>` |
| `component_slots.rs` | a named-slot element, `<svelte:fragment>` |
| `element.rs` | `<title>` inside `<svelte:head>` |

The string attribute-building path then has no callers and is deleted — 15
functions, 514 lines, including `build_attributes_string`, which was literally
`segs_to_string(build_attribute_segments(…))`. `trailing_attr_comment_text`'s unit
test is re-pointed at `trailing_attr_comment_segs` + `segs_to_string` rather than
dropped, so the assertion survives and now exercises the segment path.

## Two oracles, and only one of them reads upstream

`tests/svelte2tsx_start_tag_opener_segments_4385.rs` is one file per cell (a
construct one compiler rejects must not change another cell's output), each with
`const probe = 1` and one expression attribute in the start tag. It asserts a
property of rsvelte's own map — some generated segment is anchored on exactly the
source position of the second `probe` — because a lower-bound lookup answers
"something maps near here", which is the thing that is wrong.

    before   3 passed; 6 failed
    after    9 passed; 0 failed

The same nine cells were then driven through `textDocument/hover` against the
pinned official language server over stdio, with the `initializationOptions`
`verify.mjs` sends. Re-measured on 2026-09-09 at two rebuilt arms:

| # | cell | | base | head |
|---|---|---|---|---|
| 1 | `<div>` | control | EQ | EQ |
| 2 | `<svelte:element>` | control | EQ | EQ |
| 3 | `<Comp>` | control | EQ | EQ |
| 4 | `<svelte:component>` | | DIFF (`null`) | EQ |
| 5 | `<svelte:self>` | | DIFF (`null`) | EQ |
| 6 | `<svelte:body>` | | DIFF (`null`) | EQ |
| 7 | `<slot>` | | DIFF (`null`) | EQ |
| 8 | component named slot | | DIFF (`null`) | EQ |
| 9 | `<title>` in `<svelte:head>` | | DIFF (`null`) | EQ |
| | | | **3/9** | **9/9** |

    sha256  base fc1a0b917b6f259b   head e6ac92eee4764865   distinct

Same partition as the structural grid, from an instrument that shares almost
nothing with it: a live server answering a real request, judged by upstream rather
than by a rule written here. Cells 1-3 are pre-registered positive controls — ports
that already build their opener as `Vec<Seg>` — so a grid on which every cell was
red would say nothing about the mechanism, and they are what pins arm identity. The
six DIFFs are `null` on the base arm, not a wrong range: no answer, which is what a
request resolving through the element's start mapping produces. The probe spells a
timeout as `NO RESPONSE (…) — not a null result` so a dead server cannot be read as
an answer of `null`.

## Generated text: measured, not assumed

Two arms, each identified by running the grid from its own build, sweeping every
`.svelte` under `submodules/` in both `is_ts` modes:

    rows 69,766      distinct keys 69,766      only-base 0   only-head 0
    live 69,434      both-reject 332           rejection message differs 0
    GENERATED TEXT differs        0
    SOURCE MAP differs        2,498     (75 corpus sources)

The map column is the sweep's own positive control on the same rows: `0 text / 0
map` and a dead instrument are the same output. Units are keyed `(path, is_ts)` and
`rows == distinct keys` is asserted, because a key collapse reports a small number
rather than a zero. Population named rather than counted: 119 submodule
directories, 112 contribute a `.svelte`; six of the seven that do not are checked
out and genuinely hold none, and `vize` is uninitialised, so it is **unmeasured**
rather than empty.

## LSP ratchet: 14 keys re-keyed here, and 208 that are not this change

`pnpm run lsp:verify:fixtures` moves 14 non-inlayHint keys — the same diagnostics
and document symbols at different positions, `count=` unchanged in all 14 — and
they are re-keyed in **both** `lsp-known-failures.json` and `lsp-mechanisms.json`
in this commit. Attribution rather than coincidence: all four carrier files hold a
shape this commit converts (`$$slots` and `rename-prop-with-slot-events` a
`<slot>`, `component-invalid` six `<svelte:component>`, `element-events` a
`<svelte:window>` and a `<svelte:body>`), and all four carry zero `svelte:element`
against a positive control of 6 in the `svelte-element` fixture, so #4102
underneath cannot reach them.

A separate 208 `textDocument/inlayHint` pairs move on this checkout, on
`origin/main` too. Re-measured on 2026-09-09 at a rebuilt arm, the gate reports
**208 added / 208 stale, non-inlayHint 0/0**, and the 208 are element-for-element
identical to an `origin/main` run — so nothing further is re-baselined here. That
churn is recorded, unexplained, in #4102's commit message; CI does not reproduce
it, which is why `Language server` is green on `main`.

`cargo test --release -p rsvelte_projection` is 743 passed / 0 failed over 99 test
binaries plus the lib, including `svelte2tsx_fixtures`. Workspace clippy with
`-D warnings` is clean over 17 crates.


baseballyama:
### The six red `LSP real-world parity` shards are not this PR

Run [34249324706](https://github.com/baseballyama/rsvelte/actions/runs/34249324706), shards 2/4/9/13/14/15. Step states say the comparison step itself failed (not a cancellation, not a lost runner); logs are not retrievable until the run finishes, so this was measured from the shards' own uploaded artifacts against this PR's ratchet blob.

**Two-sided compare over the six completed shards** (ratchet read at the merge ref `ceca269c`, 23,734 entries):

```
NEW   156      STALE 0
NEW by method  {completion: 52, definition: 52, hover: 52}
```

156 = **26 files x 3 methods x 2 phases**, uniform — the shape of files entering the corpus, not of a behaviour change. All 26 are under `corpus/shadcn-svelte/`, and **none of them carries a single pre-existing ratchet key**.

**Cause.** `#4407` bumped `submodules/shadcn-svelte` `50e5c391` -> `3313e3ee` and merged 2026-09-08 14:44Z. `lsp-corpus` does not run on a pull request unless the `lsp-ratchet` escape hatch fires (`compatibility/lsp-known-failures*.json` or `scripts/compat-lsp/**`), and a submodule bump fires neither — so the bump landed with the LSP corpus never measured. This PR touches the ratchet, so it is the first run to see the new files.

**Control.** The last green nightly ([34162791979](https://github.com/baseballyama/rsvelte/actions/runs/34162791979), 2026-09-07T21:20Z, all 16 LSP shards `success`) predates the bump. Its 16 artifacts give the pre-bump universe: 3,637 measured ids, of which **1,681 are shadcn files** (live positive control). Of the 26 files carrying the new keys, **0 are present** in that universe and 26 are absent. They are new files, so every divergence they carry is new.

**What this PR will do about it.** The re-baseline needs a complete 17-artifact set at one revision, and this run is one — 16 corpus shards plus `lsp-current-fixtures`, all at merge ref `ceca269c`. So the baseline is taken here, from this run, with `merge-current.mjs --update-baseline`, and both directions reported separately.

It is deliberately **not** split into a separate main-based PR: this PR rewrites the svelte2tsx start-tag opener map across every node kind, so it can move `hover`/`definition`/`completion` on an arbitrary component. Baselining a merge-ref measurement onto `main` could therefore enrol an entry that passes on `main` — a stale entry, and the two-sided ratchet's other direction. Taken here the ratchet describes exactly the tree CI measures and exactly the tree `main` becomes on merge.

baseballyama:
Correction to the comment above: **67 new files, not 26** — 26 was the count over the six shards that had finished when I wrote it, generalised to the corpus without saying so.

Derived independently of the gate, in the submodule:

```
git ls-tree -r --name-only 50e5c391 | grep -c '\.svelte$'   ->  1681   (old pin)
git ls-tree -r --name-only 3313e3ee | grep -c '\.svelte$'   ->  1748   (new pin)
```

1,681 agrees exactly with the pre-bump nightly's measured shadcn universe and with `scripts/compat-lsp/corpus-population.json`, which is the control that the two counts are of the same set. So the bump adds 67 files and the new-key ceiling is 402, not 156.

Running total at 9 of 16 shards: `NEW 216, STALE 0`, 36 distinct files, all `corpus/shadcn-svelte/`, none present in the pre-bump universe, and **none carrying a `[count=,hash=]` detail** — corpus aggregate keys have no digest, so a hand-add of exactly these keys bakes in nothing arm-specific.

One further thing the control turned up, which changes how the baseline is taken here. `artifacts.mjs:203-206` asserts `corpus-population.json` with `!==` on `files`/`identifiers`/`requests` per repo — it is an exact declaration, not a floor despite the parameter name. So `merge-current.mjs --update-baseline` **throws** on a grown corpus until that file is updated; and a blanket re-baseline would also rewrite the churny `[count=,hash=]` fixture keys with this run's values (measured: the pre-bump nightly's own artifacts differ from `main`'s ratchet by 8 new / 8 stale, every one `upstream-features/svelte-element/input.svelte|textDocument/diagnostic`, differing only in `count=`/`hash=`).

So this PR will hand-add the corpus growth keys and update `corpus-population.json` from the summed artifact population, and leave every digest-bearing key untouched.

baseballyama:
The 48 `lsp-verify` divergences on this PR are **not caused by this change** — they came in with the shadcn-svelte submodule bump I merged in #4407, and this PR is simply the first one to run the LSP corpus gate on top of it.

Measured:

- All 48 keys are `completion` / `definition` / `hover`, each with and without `phase=edit`, over exactly 8 files under `shadcn-svelte/docs/src/lib/registry/examples/` (`attachment-image`, `bubble-demo`, `bubble-reactions`, `bubble-variants`, `marker-icon`, …) — 8 x 3 x 2 = 48.
- Those files are **new in the bump**: `git cat-file -e` finds every one of them at the new pointer `3313e3eeb9b7` and **none** of them at the previous pointer `50e5c3916888`.
- `main` is green because `lsp-corpus` does not run on `main` pushes (schedule/dispatch only, or a PR touching `scripts/compat-lsp/**` or its ratchet), so main is already carrying these — this PR just surfaced them.

I introduced them, so the ratchet side is mine to handle. Nothing here is yours to fix.

baseballyama:
### Baseline applied — rebased onto `main` and force-pushed, head `e1aa3aa9f`

The run completed at 16/16. Final delta against the ratchet blob at its merge ref, with every guard reported:

```
shards 16   measuredIds 3704   (= 3637 + 67, the pre-bump universe plus the bump)
NEW 402     STALE 0            (= 67 files x 3 methods x 2 phases, the exact ceiling)
NEW distinct files 67, all corpus/shadcn-svelte/
  of those present in the pre-bump universe: 0
NEW carrying a [count=,hash=] detail: 0
population  bits-ui 617/135920/407760      unchanged
            flowbite-svelte 1296/211845/635535   unchanged
            melt-ui 43/12043/36129         unchanged
            shadcn-svelte 1681->1748 files, 278269->285961 ids, 834807->857883 requests
```

402 is not a number I read off the run and accepted — it is the ceiling predicted **before** the run finished, from `git ls-tree -r --name-only <pin> | grep -c '\.svelte$'` in the submodule (1,681 -> 1,748) times the six `(method, phase)` keys a diverging file carries. It came back equal, and the three untouched repositories came back identical, which is what says the summation is measuring the bump and nothing else.

The applier refuses unless all 16 shards are present, `STALE` is 0, every added key is a digest-free corpus aggregate carrying a mechanism set, and no added key sits on a file that existed before the bump. Controls, both directions: the completeness guard fires on a 10/16 set; the pre-existing-file guard fires on a synthetic complete set built from the pre-bump nightly's own artifacts (12 keys, 2 files, named); the branch guard fires with the real branch name; and the population summation reproduces the committed pre-bump `corpus-population.json` exactly on all twelve numbers.

**Files.** `lsp-known-failures.json` 23,734 -> 24,136. `lsp-mechanisms.json` is a pure insertion — 402 added, 0 removed, 0 values changed, shared keys in the same relative order (the first write re-sorted by `localeCompare`, which is what `merge-current.mjs` uses but not what the committed file was in; restored to the committed order so the diff is an insertion rather than a 7,000-line reshuffle). `corpus-population.json` is one line, prettier-clean. `KNOWN-FAILURES.md`'s four gated numbers are recomputed from the JSON, including the per-repository partition, whose own prose says it "is the count that moves when a corpus submodule is bumped".

**`GATES.md` blind spot 27o** gets the growth mechanism. Its existing table re-derives exactly today — 21,630 aggregate keys, 3,632 distinct files, 3,551x6 + 81x4, per-repo 616/1293/43/1680, and all five "clean" files really are 0 bytes — so the argument is sound and only its domain was unstated: "that population" is a set of submodule pins, and moving one is the single way the corpus half can report a NEW. The table is marked as a snapshot rather than restated, because this commit is what makes it move. The population-floor row's `eight stable-hash shard union` is corrected to sixteen (`CORPUS_SHARDS = 16`, `artifacts.mjs:11`).

**Rebased first, not after.** `origin/main` moved one commit (#4462, layercake) while the run was going. Compared by blob rather than by revision, none of the gate's inputs moved — the four corpus submodules, `language-tools`, `scripts/compat-lsp/**` and the LSP ratchet are identical between the merge ref's own main parent and today's `main` — so the measurement still describes the tree CI will re-measure.

`known-failures-md-check`, `lsp-mechanisms-check` (24,136 / 24,136), `check:attribution --gate-known`, `attribution-progress-check` and `check-core-consumer-changesets` all pass.

Issue #4452, opened 2026-09-08
Eleven pattern-corpus repro files diverge from official and the output gate passes all of them


baseballyama:
Eleven checked-in `pattern-corpus` repro files produce client/server output that is byte-different from the official compiler, and the corpus output gate scores every one of them a pass. Ten of the eleven were landed *as the repro for a comment issue*, so the convention that a fix ships with a corpus repro does not hold for this defect class — the repro file is present, the divergence is live, and nothing can fail on it.

### Measurement

Tree `origin/main` = `b003aae5656e2609ceac490f6a9b3661bf8ff78d`. The rsvelte arm is a `--release` NAPI build of that commit (`sha256` recorded, `/tmp/cc-arm-base2.node`); the oracle is the recorded submodule pin `7bc0a70fe` (`VERSION 5.57.0`) reached through `submodules/svelte/packages/svelte/src/compiler/index.js`, the source path the gates use.

**Note for anyone reproducing from the primary checkout**: its `submodules/svelte` working tree sits at `56a036f4c` / `VERSION 5.56.10` while the gitlink at `origin/main` is `7bc0a70fe` / `5.57.0`. Probing the oracle from there measures a different compiler than the tree declares.

Population: every `.svelte` file under `compatibility/pattern-corpus/issues/` — 586 files × 4 targets = 2,344 pairs.

The comparison is a reconstruction of `scripts/compat-corpus/verify.mjs`, stage for stage: compile with the gate's own options (`{generate, dev, filename: id, css: 'external'}`), `flattenTemplateHoles`, `oxfmtTree` with `compatibility/.oxfmtrc.json`, `stripBlankLines`, byte compare, then `target/release/ast_equiv_batch` invoked with **empty argv** — which `verify.mjs:663` marks as load-bearing, because no `--comments` means `CommentPolicy::Ignore`.

```
error axis          both-reject 412   only-official 0   only-rsvelte 0   differing code 0
post-normalization  byte-different 32 of 2344 pairs
gate would FAIL     1 file, 2 targets
gate scores a PASS  11 files, 30 byte-different pairs
```

[...]

PR #4469, opened 2026-09-08
fix(client): a method's parameter list runs its comment window to the body (#4468)

### LINKED ISSUE #4468: A comment between a method's `)` and its body is dropped; a `function`'s is kept, and the two go through different ports of one upstream expression
A comment between a **method's** `)` and its body is dropped from rsvelte's output on all four targets. The same comment on a `function` declaration or an arrow is kept, by rsvelte and by official alike.

```
<script>
  class C { m() /*W*/ { return 1; } }
  export const a = new C();
</script>

official  m(/*W*/) {
rsvelte   (the comment appears nowhere in the output)
```

### Mechanism

esrap ends a parameter list's comment window with `sequence(context, node.params, <until>, false)`. Upstream spells `<until>` **twice**, split by whether the node has a body:

| upstream | expression | reaches |
|---|---|---|
| `esrap/src/languages/ts/index.js:744` | `(node.returnType ?? node.body).loc?.start ?? null` | `FunctionDeclaration` / `FunctionExpression` — which in the acorn model is also every method, getter, setter and object-literal method |
| `:988` | *the same expression* | `ArrowFunctionExpression` |
| `:1771` | `node.returnType?.loc?.start ?? node.loc?.end ?? null` | bodyless TS signatures |

rsvelte spells it **three** times, and the split does not line up with upstream's:

| # | rsvelte site | expression | reaches |
|---|---|---|---|
| 1 | `crates/rsvelte_esrap/src/printer.rs:2493-2501` | `return_type.start` ?? `body.start` ?? `node.span.end` | `function` declarations and expressions |
| 2 | `printer.rs:3173-3175`, `formal_parameters` | **`params.span().end`** — the `)` | 8 call sites: `:2805` methods/getters/setters, `:4407` object-literal methods, `:5019/:5032/:5271/:5282/:5294` TS signatures |
| 3 | `printer.rs:3294-3308` | a four-arm `match` that can yield `None` | arrows |

Site 2 is the defect. Upstream routes a method through `:744`, whose window ends at the **body**, so a comment sitting between `)` and `{` is inside it and gets flushed. rsvelte ends the window at `)`, the comment is outside it, and nothing downstream claims it.

### Measurement

Arm: `--release` NAPI build of `90a45850d` (`/tmp/cc-base2.node`). Oracle: `submodules/svelte/packages/svelte/src/compiler/index.js`, the source path the gates use.

**Axis 1 = the host (which `until` port prints the parameter list). Axis 2 = the comment's slot.** The `before )` slot is the negative control: every port must keep it, so a grid that is DIFF everywhere would indict the instrument rather than the port.

```
host                   slot                  rsvelte   official  verdict
function decl   :2497  before )  [neg ctl]   true      true      EQ
function decl   :2497  after  )  [subject]   true      true      EQ
arrow           :3308  before )  [neg ctl]   true      true      EQ
arrow           :3308  after  )  [subject]   true      true      EQ
class method    :2805  before )  [neg ctl]   true      true      EQ
class method    :2805  after  )  [subject]   false     true      DIFF
object method   :4407  before )  [neg ctl]   true      true      EQ
object method   :4407  after  )  [subject]   false     true      DIFF
class getter    :2805  before )  [neg ctl]   true      true      EQ
class getter    :2805  after  )  [subject]   false     true      DIFF
class setter    :2805  before )  [neg ctl]   true      true      EQ
class setter    :2805  after  )  [subject]   false     true      DIFF

EQ 8  DIFF 4  (cells 12)
```

The control fires as it must on all six hosts, so the divergence is the slot and not the host's ability to carry a comment at all.

Crossed against the four targets (`Y` = the comment is somewhere in the output, `.` = it is nowhere):

```
host                  client     client-dev server     server-dev   (rsvelte/official)
function decl  :2497  Y/Y        Y/Y        Y/Y        Y/Y
arrow          :3308  Y/Y        Y/Y        Y/Y        Y/Y
class method   :2805  ./Y        ./Y        ./Y        ./Y
object method  :4407  ./Y        ./Y        ./Y        ./Y
class getter   :2805  ./Y        ./Y        ./Y        ./Y
class setter   :2805  ./Y        ./Y        ./Y        ./Y

cells 24  DIFF 16
```

The predicate is whole-output containment, so `.` means dropped rather than relocated.

### Reach

**0 carriers, against a denominator that is the caller population rather than a source-text
guess.** Every `.svelte` file under `submodules/svelte/packages/svelte/tests` and
`compatibility/pattern-corpus` was parsed with the official compiler and walked for the two
things that route to port 2 — a `MethodDefinition` and an object `Property` that is a method,
getter or setter — and then, for each one found, the source between the last parameter's `end`
and the body's `start` was read for a comment marker:

```
.svelte files              5,920
parsed by official         5,737   (183 rejected — the deliberately-invalid fixtures)
reach port 2                 214   <- the live control: the port is exercised in this population
carry the disputed slot        0
```

Both predicates carry two-sided controls: the reach predicate answers `true` for a class method
and an object method and `false` for a plain `function` and an arrow; the slot predicate answers
`1` for `m() /*W*/ {}` and `0` for `m() {}`.

The 214 is what makes the 0 readable — a scan that found no methods at all would produce the same
zero. One qualification stands: the collected corpus (~33.9k components) is not materialised on
this machine, so 5,737 is not the denominator this project normally quotes.

### Explicitly unmeasured

- **The five TS-signature call sites** (`:5019`, `:5032`, `:5271`, `:5282`, `:5294`). Upstream's counterpart for those is `:1771`, a *different* expression from `:744`, so `params.span().end` may or may not be wrong there — and their output is type-only, which the client erases. Neither direction is claimed.
- **Whether site 3's `None` arm diverges.** Upstream's `:988` yields `null` only when a `loc` is missing; rsvelte's four-arm `match` has a `Some(_) => None` arm reached on a different condition. The arrow row above is EQ, so this input does not separate them; nothing else was tried.
- **Whether fixing site 2 moves anything else.** `formal_parameters` has 8 callers and the fix changes the window for all of them.

### Where this sits

`compatibility/GATES.md#two-ports-inventory` is the artifact that owns "one upstream decision, N rsvelte implementations". This is a row for it, and the row is what makes the *other* two sites findable — a maintainer reading site 1, which is correct, has no way to learn that the same decision is answered twice more.

Not a duplicate of #4452, which reports that eleven checked-in repro files diverge with the gate passing them; this shape has no repro file and no carrier.

baseballyama:
Closes #4468.

A comment written between a method's `)` and its body was dropped from the output entirely, on all four targets. The same comment on a `function` declaration or an arrow is kept, by rsvelte and by official alike.

```
class C { m() /*W*/ { return 1; } }

official  m(/*W*/) {
rsvelte   (the comment appears nowhere in the output)
```

## The rule

esrap ends a parameter list's comment window with `sequence(context, node.params, <until>, false)`, and spells `<until>` twice — split by whether the node has a body:

| upstream | expression | reaches |
|---|---|---|
| `esrap/src/languages/ts/index.js:744` | `(node.returnType ?? node.body).loc?.start ?? null` | `FunctionDeclaration` / `FunctionExpression` — which in the acorn model is *also* every method, getter, setter and object-literal method |
| `:988` | the same expression, character for character | `ArrowFunctionExpression` |
| `:1771` | `node.returnType?.loc?.start ?? node.loc?.end ?? null` | bodyless TS signatures, which have no `node.body` to ask |

rsvelte spelled it three times, and the split did not line up: `formal_parameters` ended the window at `params.span().end` — the `)` — for eight callers, four of which upstream routes through `:744`. A comment after the `)` therefore sat outside every window and nothing downstream claimed it.

The three bodied sites now share `function_params_until`. `formal_parameters` keeps the paren-ending default and is, after this change, called from exactly the five bodyless TS signature sites — which is what its doc comment now claims, and `grep -n 'formal_parameters('` can check.

## Measurement

Two `--release` NAPI arms, both built from this tree at `2de8dfce1`: base `0b22513a83fc84ec`, head `65bfced0b78140d5`, **distinct** by `sha256`. The base arm is built at this branch's own merge base rather than reused from an older binary — reusing one would have differed by #4467 as well as by the change under test.

Arm probe, two-sided:

```
base   method (SUBJECT, must be false): false    function (CONTROL, both true): true
head   method (SUBJECT, must be false): true     function (CONTROL, both true): true
```

Host × comment slot against the oracle, before and after. `before )` is the negative control every port must keep:

```
host                   slot                  before fix   after fix
class method    :2805  before )  [neg ctl]   EQ           EQ
class method    :2805  after  )  [subject]   DIFF         EQ
object method   :4407  before )  [neg ctl]   EQ           EQ
object method   :4407  after  )  [subject]   DIFF         EQ
class getter    :2805  before )  [neg ctl]   EQ           EQ
class getter    :2805  after  )  [subject]   DIFF         EQ
class setter    :2805  before )  [neg ctl]   EQ           EQ
class setter    :2805  after  )  [subject]   DIFF         EQ
function decl   :2497  both slots [ctl]      EQ           EQ
arrow           :3308  both slots [ctl]      EQ           EQ
```

## Blast radius

```
arm A 0b22513a83fc84ec   arm B 65bfced0b78140d5   (distinct)
files 6,155   pairs compared 24,620   rows 24,620, distinct ids 6,155   (no key collapse)
live 21,174 / dead 3,446 on both arms

MOVED = 0 of 24,620
```

**That zero has a manufactured pin.** Re-running the same sweep with two injected files:

```
files 6,155 -> 6,157                            <- both entered the population
MOVED = 4
   manufactured-carrier.svelte -> client, client-dev, server, server-dev
   manufactured-noncarrier.svelte -> (0 targets)
   any real corpus file -> 0
```

The carrier is `m() /*W*/ {}` and the non-carrier is `m(/*W*/) {}`, one position to the left. So the sweep is demonstrably able to report movement from *this* change, and rejects the near-miss; the zero is a property of the population.

## Why a local sweep rather than the merge queue

The corpus output gate cannot see this class. `ast_equiv_batch` is invoked with empty argv, which is `CommentPolicy::Ignore`, so a comment-only divergence normalizes to `match` — that is #4452's whole subject. A comment-placement regression from this change would be invisible to CI, which is why the evidence is two local arms rather than a green corpus job.

For the same reason there is no `pattern-corpus` repro: the file would be checked in, the divergence live, and the gate would pass it. The regression suite is `crates/rsvelte_core/tests/method_param_comment_window_4468.rs` instead.

## Tests

Six, one per host so that a single "all four are kept" assertion cannot be satisfied by three. Expected strings are the official compiler's own output, generated rather than inferred. Ablated — the fix reverted, the suite re-run, the fix reapplied and the restore confirmed byte-identical by patch `sha256`:

```
with the fix       6 passed
fix reverted       4 failed, 2 passed
                   FAILED  a_class_method / a_class_getter / a_class_setter / an_object_literal_method
                   ok      a_function_declaration_still_keeps_it        <- already agreed
                   ok      a_comment_inside_the_parens_is_unmoved       <- the near-miss
```

## Reach

0 carriers, against the caller population rather than a source-text guess. Every `.svelte` under `submodules/svelte/packages/svelte/tests` and `compatibility/pattern-corpus` parsed with the official compiler and walked for the two node shapes that route through the changed site:

```
parsed by official   5,737   (183 rejected — the deliberately-invalid fixtures)
reach the site         214   <- the live control
carry the slot           0
```

The 214 is what makes the 0 readable; without it a population holding no methods produces the identical zero. The collected corpus (~33.9k components) is not materialised on this machine, so 5,737 is not the denominator this project normally quotes.

## The GATES row

`compatibility/GATES.md` row 34 of the two-ports inventory. It records the closed pair *and* the one that remains: `formal_parameters` still answers upstream's `:1771` with the `)`, and nothing reaches it — a TS signature's type is erased from generated output by both compilers, while the same marker written as a value survives on both, so the probe is live rather than blind. It also records both instrument corrections made while measuring the reach, neither of which moved the number.

### DIFF
--- a/.changeset/method-param-comment-window.md
+++ b/.changeset/method-param-comment-window.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Keep a comment written between a method's `)` and its body. esrap runs a parameter list's comment window until the body starts, and in the acorn AST a method is a `FunctionExpression`, so upstream reaches methods, getters, setters and object-literal methods through that same rule. rsvelte routed them through a helper that ended the window at the `)` instead, which put such a comment outside every window and dropped it from the output on all four targets. The three bodied sites now share one expression; the five bodyless TS signature callers keep the paren-ending default, which is the only place it is correct.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2643,7 +2643,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_esrap"
-version = "0.10.39"
+version = "0.10.40"
 dependencies = [
  "compact_str",
  "memchr",
--- a/compatibility/GATES.md
+++ b/compatibility/GATES.md
@@ -8090,6 +8090,101 @@ same holds for `void 0`. Which inputs *do* reach it is **unmeasured**, so the ro
 port that disagrees with upstream and is currently unreachable is one refactor away from being
 reachable, and nothing here would notice.
 
+#### 34. Where does a parameter list's comment window end? — [D], one port pair closed, one left unreachable
+
+**Upstream** ends it with `sequence(context, node.params, <until>, false)` and spells `<until>`
+**twice**, split by whether the node has a body
+(`esrap/src/languages/ts/index.js`, pinned at `esrap@2.2.12`):
+
+| upstream | expression | reaches |
+|---|---|---|
+| `:744` | `(node.returnType ?? node.body).loc?.start ?? null` | `FunctionDeclaration` / `FunctionExpression` — which in the acorn model is *also* every method, getter, setter and object-literal method |
+| `:988` | the same expression, character for character | `ArrowFunctionExpression` |
+| `:1771` | `node.returnType?.loc?.start ?? node.loc?.end ?? null` | bodyless TS signatures, which have no `node.body` to ask |
+
+**Ports, before #4468.** rsvelte spelled it three times and the split did not line up with
+upstream's: `printer.rs:2493-2501` (function declarations), `formal_parameters` (**`params.span().end`**,
+the `)`, at eight callers), and the arrow's four-arm `match`. Ports 1 and 3 answered upstream's
+`:744`/`:988`. Port 2 served callers upstream splits across `:744` — methods, which *are*
+`FunctionExpression`s — **and** `:1771`, so it was one answer where upstream has two, and for the
+`:744` half it was the wrong one.
+
+**Named input, and it separated two rsvelte ports rather than only rsvelte from upstream.** A
+comment between `)` and the body is inside port 1's window and outside port 2's:
+
+```
+function f() /*W*/ { return 1; }          port 1 -> kept
+class C { m() /*W*/ { return 1; } }       port 2 -> dropped from the output entirely
+```
+
+Measured host × comment slot, with `before )` as the negative control every port must keep:
+4 DIFF of 12 cells, the control firing on all six hosts, and 16 DIFF of 24 across the four
+targets — the predicate being whole-output containment, so the comment was dropped rather than
+relocated.
+
+**Closed at degree 1 for the bodied half.** `function_params_until` is now the single expression
+for every bodied function, called from the function, method and object-method sites, so the
+question has one answer wherever upstream's `:744` answers it. What remains is a genuine second
+port: `formal_parameters` keeps `params.span().end` and is now called from exactly the five
+bodyless TS signature sites — which a `grep` confirms, and which is what its doc comment claims,
+so the claim is checkable rather than asserted.
+
+**The remaining pair is [S], and its divergence is unreachable rather than absent.** Upstream's
+`:1771` is `returnType?.loc?.start ?? node.loc?.end`, and a `TSFunctionType` always has a return
+type, so upstream stops at the return type where rsvelte stops at the `)`. Nothing reaches it: a
+named type threaded through `TSFunctionType`, `TSConstructorType`, `TSMethodSignature` and
+`TSDeclareFunction` is absent from `js.code` on `client` and `server` **on both compilers**, while
+the same marker written as a value survives on both — so the probe is live rather than blind.
+`rsvelte_esrap`'s only consumers besides `rsvelte_core` are `rsvelte_devtools` and `rsvelte_bench`,
+neither of which is shipped output. What that does *not* establish is that the five sites are never
+called; a `#[track_caller]` counter would, and was not built.
+
+**Port 3's `Some(_) => None` arm is a deliberate divergence, recorded rather than dropped.**
+`arrow_function` passes `owned_comment_until = None` and takes the `None => Some(body_start)` arm,
+which is upstream's expression; over 8 arrow body shapes × 4 targets that is 32 cells, 0 DIFF, with
+the method grid as the control that the same instrument does report DIFF. None of the 32 reaches
+the other arm: it needs `owned_comment_until` to be `Some`, supplied only by `printer.rs:4495` and
+`:4678`, both synthesized arrows owning a *call argument's* comments. The comment above it gives
+the reason — a generated arrow's empty parameter list would otherwise consume a located body's
+leading comments before the `=>`. Unmeasured against upstream, and kept here because a comment
+asserting a considered difference is exactly where this class hides.
+
+**Reach was 0 before the fix, against the caller population.** Every `.svelte` file under
+`submodules/svelte/packages/svelte/tests` and `compatibility/pattern-corpus` was parsed with the
+official compiler and walked for the two node shapes that route to port 2, and for each one found
+the source between the parameter list's own `)` and the body's `start` was read for a comment
+marker:
+
+```
+.svelte files              5,920
+parsed by official         5,737   (183 rejected — the deliberately-invalid fixtures)
+reach port 2                 214   <- the live control: the port is exercised here
+carry the disputed slot        0
+```
+
+The 214 is what makes the 0 readable — without it, a population holding no methods at all produces
+the identical zero and nothing in the output says so.
+
+**Two instrument corrections, neither of which moved the number.** The first version counted
+*source shapes* with a loose `\)\s*comment\s*\{` regex, which is a scan of sources where the
+population that decides port 2 is *callers*; it returned one file and printing it disqualified it
+(an object literal's `)`), so "0 carriers" and "0 carriers among the files that reach this port"
+are different claims and only the second survives being quoted. The second is sharper: the AST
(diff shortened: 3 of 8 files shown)


baseballyama:
`Changeset` is failing on the version-drift guard, not on a missing changeset (the changeset itself is found, and `check-core-consumer-changesets` passes):

```
rsvelte_esrap source changed but its crate version did not increase
(base 0.10.39, current 0.10.39). Bump crates/rsvelte_esrap/Cargo.toml;
the exact rsvelte_core dependency and compiler release-set changeset
must advance with it.
```

`crates/rsvelte_esrap/src/printer.rs` is in the diff, so the crate needs its version bumped along with the exact `rsvelte_core` dependency pin.

Say the word if you'd rather I push it — otherwise it's yours and I'll stay out of the branch.

baseballyama:
Pushed the bump as `963001dac` — three one-line changes, the same shape as the last esrap bump (#4368):

- `crates/rsvelte_esrap/Cargo.toml` `0.10.39` -> `0.10.40`
- `crates/rsvelte_core/Cargo.toml` exact pin `= "=0.10.40"`
- `Cargo.lock`

Verified locally with the gate that was red: `check-esrap-version-bump.mjs` now exits 0 (`rsvelte_esrap source changed and version advanced: 0.10.39 -> 0.10.40. ✓`), and `check-core-consumer-changesets.mjs` exits 0. Your changeset already named `@rsvelte/compiler`, so nothing there needed touching.

baseballyama:
The `Changeset` failure was the `rsvelte_esrap` version gate — a published crate whose source moved must advance, and `rsvelte_core` pins it exactly.

`963001dac` had already landed that while I was preparing the same fix. **I did not force-push over it.** Comparing the two commits at tree level rather than by inspection, they differ in exactly one file:

```
git diff --stat 963001dac <mine>
 crates/rsvelte_lint_types/Cargo.lock | 2 +-
```

So I dropped mine, reset onto `963001dac`, and added only the missing half as `04d745c45`.

That half is not optional. `crates/rsvelte_lint_types` is its own Cargo workspace, so the root build never re-resolves it, and it reaches `rsvelte_esrap` through `rsvelte_core` — on `963001dac` the checker fails:

```
crates/rsvelte_lint_types/Cargo.lock is stale: it pins in-repo crate versions that no longer exist.
  rsvelte_esrap: lock pins 0.10.39, crates/ declares 0.10.40
```

`check:lint-types-lock` runs on every PR, so it would have gone red here rather than waiting for a later PR that happens to touch the lint crates — which is the failure mode AGENTS.md describes for that lock, and the reason the gate was moved to every PR in the first place. Re-resolved with `--fix`, not hand-edited; `check-lint-types-lock` now reports up to date.

It needed `submodules/corsa-bind`, which I initialised in my own worktree only (public, no credentials, and a linked worktree keeps its submodule git dir under `.git/worktrees/<name>/modules/`). The gitlink is unchanged.

Issue #4472, opened 2026-09-08
client: a comment containing `$props(): { … }` makes the compiler emit unparseable JavaScript


baseballyama:
## Summary

A comment whose text contains `$props(): { … }` makes the client emit **JavaScript no parser accepts**, when the same source also has a multi-line TS type annotation on the `$props()` pattern with a comment inside it. Official compiles the same input to output acorn parses.

This is the `opaque-keyword` family (#2986 / #2987 / #2988): a raw text scan keyed on a source-level spelling, defeated by that spelling appearing inside a comment. Here the scan's key is not `$props()` alone — see the controls.

## Minimal reproducer

```svelte
<script lang="ts">
	let { a }: { // x
		a: string;
	} = $props(); // $props(): { a: b; }
</script>
<p>x</p>
```

`rsvelte` → **UNPARSEABLE** (`acorn`: `Unexpected token`).  `official` → PARSES.

## It is a conjunction of three conditions

Each row drops exactly one condition; all three are required.

| input | parseable | vs official |
|---|---|---|
| annotated multi-line type + comment in annotation + trailing `$props(): { … }` | **UNPARSEABLE** | DIFF |
| …no comment in the annotation | PARSES | EQ |
| …single-line type annotation instead | PARSES | EQ |
| …no type annotation at all | PARSES | EQ |
| …identifier binding (`let p = $props()`) | PARSES | EQ |
| …plain `<script>` (no `lang="ts"`) | PARSES | EQ |
| `generate: 'server'` (same source) | PARSES | EQ |

Client only.

## The scan's key is `$props()` + `:` + a brace block, not `$props()`

Varying **only** the trailing comment's text, holding everything else at the reproducer:

| trailing comment text | parseable |
|---|---|
| `$props(): { a: b; }` | **UNPARSEABLE** |
| `$props(): { a: b }` | **UNPARSEABLE** |
| `$props():{a:b;}` | **UNPARSEABLE** |
| `$props() : { a: b; }` | **UNPARSEABLE** |
| `$props(): { }` (empty braces) | **UNPARSEABLE** |
| `foo(): { a: b; }` | PARSES |
| `$props()` | PARSES |
| `$props(): x` (no braces) | PARSES |
| `{ data: any[]; }` | PARSES |
| a plain word | PARSES |

[...]

Issue #4244, opened 2026-09-03
A comment inside an erased TypeScript type annotation is dropped entirely


baseballyama:
A comment written inside an erased TypeScript **type annotation** is dropped entirely — rsvelte emits zero copies where official emits two.

This is the **opposite sign** of #4239 (which is `+1`, inside an erased `interface` body), it lives in a **different host**, and **fixing #4239 will not remove it**.

## Repro

```svelte
<script lang="ts">
let a: { /* c */ b: number } = { b: 1 };
</script>
<i>{a.b}</i>
```

`generate:'client'`, oracle `submodules/svelte/packages/svelte/src/compiler/index.js`.

## Actual output

```js
// official
let a = /* c */ /* c */ { b: 1 };

// rsvelte
let a = { b: 1 };
```

The line-comment spelling behaves the same way, and official's output there parses (checked with acorn):

```js
// official
let a = // c
// c
{ b: 1 };

// rsvelte
let a = { b: 1 };
```

Official's **two** copies are the `@sveltejs/acorn-typescript@1.0.10` `tsLookAhead` duplication surviving into generated output: a comment that is the first token after the opening `(` or `{` of a *type* position is scanned once by the speculative parse and once by the real one. That upstream behaviour is fixed in 1.0.13 (measured: 0/17 cells duplicate there, against 8/17 at 1.0.10), but svelte 5.56.10 bundles 1.0.10, so it is what the oracle emits today.

**That is not what this issue is about.** Declining to reproduce the duplication would give **1**. rsvelte gives **0**, so one comment is lost independently of the duplication question.

## Grid

`/* c */` (or `// c`) occurrences in the generated output, `dev:false`:

| host | comment kind | used | official | rsvelte | verdict |
|---|---|---|---|---|---|
| erased `interface` body | line | yes | 1 | 2 | DIFF (+) — this is #4239 |
| erased `interface` body | block | yes | 1 | 2 | DIFF (+) |
| erased `interface` body | line | no | 1 | 2 | DIFF (+) |
| erased `interface` body | block | no | 1 | 2 | DIFF (+) |
| **erased type annotation** | **line** | **yes** | **2** | **0** | **DIFF (−)** |
| **erased type annotation** | **block** | **yes** | **2** | **0** | **DIFF (−)** |
| **erased type annotation** | **block** | **no** | **2** | **0** | **DIFF (−)** |
| value position (not erased) | block | yes | 1 | 1 | EQ |
[...]

PR #4478, opened 2026-09-09
fix(compiler): repeat a speculated comment run whole, and once for both targets

### LINKED ISSUE #4373: The acorn-typescript comment duplication is modelled per comment on the client and not at all on the server

baseballyama:
Closes #4373.

## What

`@sveltejs/acorn-typescript`'s `tsLookAhead` does not set `isLookahead`, so a comment consumed while speculatively parsing an object type fires `onComment` during the lookahead and again after the rewind. Upstream prints it twice; byte equality is the goal, so rsvelte reproduces it. Two things were wrong with how:

1. **The unit is the run, not the comment.** The rewind replays everything the speculation consumed, so two comments in a `TSTypeLiteral` head print `c d c d`; rsvelte repeated each comment in place and printed `c c d d`. One comment is where the two rules agree, which is the whole reason a grid built on a single comment could not see it.
2. **The server did not repeat at all.** `repeated_comment_outputs` had exactly one consumer — a client-side text pass — so the rule lived on one of the two ports.

## How

Both halves close together because the repeat moves into the stripped script text, which is what phase 3 reads on *both* targets. Emitting the run twice there is what makes it one port rather than two: the client's text pipeline and the server's parse each see the comment list upstream's `analysis.comments` holds, and neither carries the rule. The client's text-insertion pass is deleted rather than mirrored.

**Doubling it in the server's own comment list cannot work, and that was measured rather than reasoned.** The obvious server fix is to put a duplicate into the parsed comment list carrying the same range the original does — which is exactly what upstream's list holds, since `onComment` fires twice for one range. It moved 3 of the 5 server rows and then printed `c c d d`: `print_with_comments` builds a buffer, shifts every comment onto it and lets the printer flush in **span order**, so two copies at one range cannot be separated by list position. The run has to exist twice in the text.

`collect_speculative_type_head_regions` is no longer gated on `include_projection`. That gate was sound while the marker set was projection-only and is wrong once the repeat is in the emitted text: the server reads the projection-less strip and the client reads the other, so gating it makes the two strips disagree about their own output. With the repeat in place and the gate left alone, all six client rows went green and all five server rows stayed **byte-identical to `main`** — a fix that looks complete on one target and inert on the other, with no error anywhere.

## Grid

Oracle `submodules/svelte/…/src/compiler/index.js`, the pinned `7bc0a70fe` / `VERSION 5.57.0`, `dev: false`. Every expected sequence is read out of the oracle rather than reasoned about.

| cell | client | server | before |
|---|---|---|---|
| 1 line comment in the head | `c c` | `c c` | client EQ, server `c` |
| 1 block comment in the head | `c c` | `c c` | client EQ, server `/*c*/` |
| 2 line comments in the head | `c d c d` | `c d c d` | client `c c d d`, server `c d` |
| 3 line comments in the head | `c d e c d e` | `c d e c d e` | client `c c d d e e`, server `c d e` |
| 1 in the head + 1 after member 1 | `c c d` | `c c d` | client EQ, server `c d` |
| nested literal in the first member | `c d c d` | `c d c d` | client `c c d d`, server `c d` |
| `interface` body (negative control) | `c d` | `c d` | EQ — no speculation, nothing doubles |
| `<script module>` alias (negative control) | *(none)* | *(none)* | EQ — a builder-made Program, upstream prints nothing either |

The two negative controls are there because an assertion set that only ever expects a doubling is satisfied by a compiler that doubles everything.

## The host axis, and the two rows it pins

The first grid held the annotation's **host** fixed at a `type X = {}` alias. Crossed against the other hosts, the answers split three ways, not two — so a second test pins what rsvelte answers today and names the issue that owns each row:

| host | client | server |
|---|---|---|
| inline annotation on the declarator | *(nothing)* — **#4398**, the client rebuilds a destructured `$props()` declaration and loses the comment before the repeat can apply | `c d c d` ✔ matches |
| uninitialized declarator | *(nothing)* — **#4396** | *(nothing)* — **#4396** |

The oracle prints `c d c d` for both. Fixing either issue turns that test red, which is where those rows move.

## Reach

Two-arm sweep over the collected corpus, `client` + `server`, `dev: false`. Arms: `main + #4472` (`sha256 df8c2bed…`) and this branch (`c9f34430…`); identified by a two-sided probe on their own output — a two-comment head reads `c c d d` / `c d` on the base and `c d c d` / `c d c d` on the head, and a comment-free source reads the same on both. #4472 is in the base and not the head, and it can only fire when a **comment contains `$props`**: of the 159 screened carriers, **0** do, against a live positive control of **9** in the whole 33,890-file corpus — so the confound is measured to zero rather than argued away.

```
base rows 67780  distinct keys 67780  dead 2487
head rows 67780  distinct keys 67780  dead 2487
MOVED 146  {client: 1, server: 145}   over 145 files, 0 units where either side threw
```

Scored against the oracle on the moved set only, with raw bytes — **stricter** than the gate, which normalizes with oxfmt and rescues under `ast_equiv_batch`'s `CommentPolicy::Ignore`:

| | whole `js.code` | comment sequence |
|---|---:|---:|
| `MISMATCH -> match` | 5 | 11 |
| **`match -> MISMATCH`** | **0** | **0** |
| `MISMATCH -> MISMATCH` | 141 | 135 |

A zero under a stricter comparator is a zero under the gate, so that direction needs no fidelity argument. None of the 146 moved ids is listed in any of the four output ratchets (`0 / 2 / 0 / 2` entries), so there is nothing to re-baseline.

On a source-shape screen (159 carriers with a comment in a `type X = {` head), the issue's own key — how many comments each side prints — moves on the server from **79 short / 0 extra** to **2 short / 0 extra**.

**Two numbers here correct the issue's framing.** Only **one** client unit moves in 33,890 files, because per-comment and per-run repetition give the same output for a single-comment run and a multi-comment type head is that rare; the client half is an *ordering* fix that a count key cannot see at all. And the issue read "89 of 89 unanimous short" as this mechanism's reach — the direction was right and the attribution too wide: 141 of the 146 moved units still diverge, for comment-text reasons (#4244 / #4279) and for #4396 / #4398. A direction is not an attribution.

## Gates

`cargo test -p rsvelte_core --lib` (1,971 passed) plus `ssr`, `compiler_fixtures`, `compiler_errors`, `print`, `sourcemaps`, `parser_fixtures`, `validator`, `css`, `erased_annotation_comment`, `erased_annotation_comment_initializer`, `type_literal_head_comment_repeat`, `client_script_comments_pin`, `module_comment_ownership`, `module_server_comment_ownership_2307`, `server_removed_statement_comments_2567`, `ssr_typescript_erasure` — all green. `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean.

### DIFF
--- a/.changeset/repeated-comment-runs.md
+++ b/.changeset/repeated-comment-runs.md
@@ -1,0 +1,12 @@
+---
+'@rsvelte/compiler': patch
+---
+
+repeat a speculated comment run whole, on the server as well as the client
+
+acorn-typescript's `tsLookAhead` leaves `isLookahead` unset, so a comment inside
+a `TSTypeLiteral`'s head fires `onComment` during the speculative parse and again
+after the rewind. The rewind replays the whole run, so two comments print
+`c d c d`; rsvelte repeated each comment in place (`c c d d`) and the server did
+not repeat at all. The repeat now happens once, in the stripped script both
+targets read, instead of in a client-only text pass.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
@@ -54,14 +54,6 @@ pub(crate) struct ScriptProjection {
     /// module Program drops them. Keeping the ranges separate lets Phase 3 make
     /// that distinction without trying to infer their origin from stripped JS.
     pub(crate) reemitted_comment_outputs: Vec<Range<u32>>,
-    /// The subset of `reemitted_comment_outputs` upstream prints TWICE.
-    ///
-    /// acorn-typescript's `tsLookAhead` does not set `isLookahead`, so a comment
-    /// consumed while speculatively parsing an object type fires `onComment`
-    /// once during the lookahead and again after the rewind. The region that
-    /// speculation covers is a `TSTypeLiteral`'s braces up to the end of its
-    /// first member, which is why an `interface` body never doubles.
-    pub(crate) repeated_comment_outputs: Vec<Range<u32>>,
     /// `(binding end, annotation end)` for every binding whose own type
     /// annotation was erased. Upstream's parser puts the annotation inside the
     /// binding's range, so a node ending at the first is located at the second.
@@ -672,11 +664,10 @@ fn strip_typescript_from_program_impl(
 
     let mut removals: Vec<(u32, u32)> = Vec::new();
     collect_ts_removals_from_program(program, source, &mut removals);
-    let repeat_regions = if include_projection {
-        collect_speculative_type_head_regions(program)
-    } else {
-        Vec::new()
-    };
+    // Not gated on `include_projection`: the repeat is in the emitted text now,
+    // so gating it would make the two strips disagree — and the server reads the
+    // projection-less one while the client reads the other.
+    let repeat_regions = collect_speculative_type_head_regions(program);
     let declarator_annotations = collect_declarator_annotations(program);
 
     // Text-based fallback: strip `declare global { ... }`, `declare module ... { ... }`,
@@ -761,7 +752,6 @@ fn strip_typescript_from_program_impl(
     let mut copied_chunks =
         include_projection.then(|| Vec::with_capacity(merged.len().saturating_add(1)));
     let mut reemitted_comment_outputs = include_projection.then(Vec::new);
-    let mut repeated_comment_outputs = include_projection.then(Vec::new);
     let mut pos = 0u32;
 
     // Comments an erased annotation left behind, waiting for the offset upstream
@@ -777,8 +767,6 @@ fn strip_typescript_from_program_impl(
                 &mut output,
                 copied_chunks.as_mut(),
                 reemitted_comment_outputs.as_mut(),
-                repeated_comment_outputs.as_mut(),
-                &repeat_regions,
             );
         }
         // The official compiler PARSES TypeScript and only removes the
@@ -811,49 +799,21 @@ fn strip_typescript_from_program_impl(
             if (removed.contains("/*") || removed.contains("//"))
                 && let Some(flush_at) = flush_at
             {
+                let plan = plan_reemitted_comments(*remove_start, removed, &repeat_regions);
                 if let Some(init_start) = flush_at {
-                    for (comment_offset, comment) in
-                        crate::compiler::phases::phase3_transform::server::transform_script::extract_comments_from_snippet_with_pos(removed)
-                    {
-                        let comment_start = *remove_start + comment_offset as u32;
-                        pending.push((
-                            init_start,
-                            comment_start,
-                            comment_start + comment.len() as u32,
-                        ));
-                    }
-                } else if let Some(copied_chunks) = copied_chunks.as_mut() {
-                    for (comment_offset, comment) in
-                        crate::compiler::phases::phase3_transform::server::transform_script::extract_comments_from_snippet_with_pos(removed)
-                    {
-                        let comment_start = *remove_start + comment_offset as u32;
-                        let comment_end = comment_start + comment.len() as u32;
-                        let output_start = output.len() as u32;
-                        push_source_range(
-                            source,
-                            comment_start..comment_end,
-                            &mut output,
-                            Some(copied_chunks),
-                        );
-                        let out_range = output_start..output.len() as u32;
-                        reemitted_comment_outputs
-                            .as_mut()
-                            .unwrap()
-                            .push(out_range.clone());
-                        if repeat_regions
-                            .iter()
-                            .any(|(from, to)| comment_start >= *from && comment_end <= *to)
-                        {
-                            repeated_comment_outputs.as_mut().unwrap().push(out_range);
-                        }
-                        output.push('\n');
+                    for (comment_start, comment_end) in plan {
+                        pending.push((init_start, comment_start, comment_end));
                     }
                 } else {
-                    for comment in
-                        crate::compiler::phases::phase3_transform::server::transform_script::extract_comments_from_snippet(removed)
-                    {
-                        output.push_str(&comment);
-                        output.push('\n');
+                    for (comment_start, comment_end) in plan {
+                        emit_pending_comment(
+                            source,
+                            comment_start,
+                            comment_end,
+                            &mut output,
+                            copied_chunks.as_mut(),
+                            reemitted_comment_outputs.as_mut(),
+                        );
                     }
                 }
             }
@@ -870,8 +830,6 @@ fn strip_typescript_from_program_impl(
             &mut output,
             copied_chunks.as_mut(),
             reemitted_comment_outputs.as_mut(),
-            repeated_comment_outputs.as_mut(),
-            &repeat_regions,
         );
     }
     // A flush point the copy never reached (an initializer inside a later removed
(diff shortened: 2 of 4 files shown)


baseballyama:
Corpus-side evidence added after the PR was opened, all node-only (no rebuild):

**This retires one of #4452's eleven un-watched repro files.** `compatibility/pattern-corpus/issues/typescript-declaration-comment-cursor.svelte` goes `RAW-DIFF → RAW-EQ` on `server` and `server-dev`, with `client`/`client-dev` already RAW-EQ on both arms. Over all 11 files × 4 targets: `rows 44, distinct keys 44, threw 0, keys only in either arm 0, MOVED 2` — and the 2 are those. Raw byte equality to the oracle means this is a retirement rather than a candidate; a stricter reconstruction reporting equality is equality under the gate's normalization too.

**The single moved client unit is the ordering half, and it reads cleanly.** `appwrite-console/src/lib/components/filters/content.svelte`:

```
official: "\t/* eslint  @typescript-eslint/no-explicit-any: 'off' */"
          "\t// We cast to any to not cause type errors in the input components"
rsvelte : "\t// We cast to any to not cause type errors in the input components"
          "\t/* eslint  @typescript-eslint/no-explicit-any: 'off' */"
```

Both comments present on both sides, swapped — `2 differing lines, equal lengths` → `RAW-EQ`. That is why the reach screen read `client count-diff 16 → 16` across this fix: a count key cannot see a permutation, which is the ratchet-key rule (an entry suppresses everything its key cannot tell apart) arriving inside a screen.

**Arm-pair caveat, priced.** The base arm for these measurements carries #4477's fix and lacks this one; the head arm is the reverse. A discriminating probe on each says so (`FP-4472 steered 0/5` vs `5/5`; `FP-4373 "c d"` vs `"c c d"`), so the pair is two changes apart in opposite directions. Priced per moved unit rather than by rebuilding: the pattern-corpus file contains `$props` **0 times** (live control: a sibling contains it once), so #4477's client `$props()` splice cannot reach it; and for the `appwrite-console` unit — which does contain one — the head arm **lacks** #4477's fix and still reaches RAW-EQ, a direction that absence cannot produce.

**One independent confirmation of the mechanism.** The `FP-4373` probe is `let o: { /* c */ a: number; /* d */ b: number } = …` on `server`. The oracle answers `"c c d"`, this branch answers `"c c d"`, `main` answers `"c d"` — so the repeat covers the run inside the speculated region (`{` up to the end of the **first** member) and not the whole annotation, which is what the fix implements. My own prior was `"c d c d"` and was wrong; the oracle settled it.

Issue #4479, opened 2026-09-09
lsp: `verify.mjs` resolves the rsvelte arm by path and records nothing about it, so `projectRevision` names the tree and not the binary


baseballyama:
`scripts/compat-lsp/verify.mjs` resolves the rsvelte arm by **path** and records nothing about
which binary it got, so a `lsp-current-*` artifact cannot say what produced it.

```js
// verify.mjs:171-172
const rsvelteCommand = parseCommand(process.env.RSVELTE_LSP_COMMAND, [
  path.join(ROOT, "target/debug/rsvelte-language-server"),
// :186
requireExecutable(rsvelteCommand, "rsvelte language server");
```

`requireExecutable` checks that the file exists. The artifact then records `projectRevision`,
`sourceRevisions`, `universeHash` and `configurationId` — every one of which describes the
**source tree** — and no field describes the arm. So the one label that reads like arm identity,
`projectRevision`, is a measurement of `ROOT`, while the arm is whatever binary happens to sit at
`ROOT/target/debug/`.

## The carrier, measured today

Re-baselining #4450 from a local run produced 208 `textDocument/inlayHint` keys whose values match
**neither** `origin/main` **nor** the CI run at this PR's own merge ref, on keys this PR does not
move. Decomposing the three artifacts against `origin/main` — one command each, no rebuild:

| arms | re-keyed | new | retired | by method |
|---|---:|---:|---:|---|
| main → the local run | 324 | 4 | 20 | inlayHint 208, diagnostic 84, documentSymbol 30, completion 2 |
| main → CI at the merge ref | 116 | 4 | 20 | diagnostic 84, documentSymbol 30, completion 2 |
| CI → the local run | **208** | 0 | 0 | **inlayHint only** |

`324 - 208 = 116`, and `new`/`retired` are identical in both, so the contamination is separable to
the key.

**And I cannot say what produced the local artifact.** That is the finding, not an aside. What is
excluded is narrow: the only `rsvelte-language-server` binary still on the machine sits in a
worktree whose tree is clean and matches `origin/main` on this PR's files, so it cannot have
produced the 116 — it is not the arm. The arm that was used is gone; `ROOT/target` is empty. And
nothing recorded it:

```
artifact projectRevision : e784370806   <- this PR's fix commit. CORRECT, and a property of ROOT.
artifact fields          : projectRevision, sourceRevisions, universeHash, configurationId
                           -- every one a property of the source tree
arm identity recorded    : none
```

[...]

Issue #4465, opened 2026-09-08
ci: a corpus submodule bump enrols new files into the LSP gate without running it, so main goes red and every ratchet-touching PR inherits it


baseballyama:
## What happened

`#4407` bumped `submodules/shadcn-svelte` `50e5c391` -> `3313e3ee` and merged 2026-09-08T14:44Z. Counted directly in the submodule, that pin moves shadcn from **1,681 to 1,748 `.svelte` files** — **67 new files** in the LSP differential gate's population, none of them measured before the merge. Every new file seen so far diverges on `hover`, `definition` and `completion` in both phases, i.e. six ratchet keys each.

The first run to see them was an unrelated PR ([#4461](https://github.com/baseballyama/rsvelte/pull/4461), run [34249324706](https://github.com/baseballyama/rsvelte/actions/runs/34249324706)), where six of sixteen `LSP real-world parity` shards failed with `NEW divergence`. That PR has not changed by a byte in a way that reaches those files; it inherited `main`.

## Why the gate did not run on the bump

`corpus-compat-job-filter.mjs:157-161` opens the `lsp-ratchet` escape hatch for exactly two path shapes:

```js
enabled['lsp-ratchet'] = changedFiles.some(
  (file) =>
    file.startsWith('scripts/compat-lsp/') ||
    (file.startsWith('compatibility/lsp-known-failures') &&
      file.endsWith('.json')),
```

and the job's `if:` additionally requires `schedule`, `workflow_dispatch`, or that hatch. A submodule bump touches `submodules/<repo>` and fires neither, so `lsp-corpus` does not run on the PR — and `push: main` is excluded on the same 950-job-minute arithmetic. The gate is not blind, it is **delayed to the next nightly**; the cost of the delay is that `main` is red in the meantime and every ratchet-touching PR inherits a failure it did not cause.

The four repos whose bumps carry this are declared in `scripts/compat-lsp/corpus-population.json` and reported in every artifact's `repos`: `bits-ui`, `flowbite-svelte`, `melt-ui`, `shadcn-svelte`.

## Measurement

Two-sided compare of the completed shards' artifacts against the ratchet blob at that run's own merge ref (23,734 entries). At 9 of 16 shards:

```
shards 0,2,3,4,8,9,13,14,15   measuredIds 2076
NEW 216      STALE 0
NEW by method  {completion: 72, definition: 72, hover: 72}
[...]

Issue #4458, opened 2026-09-08
lsp: a document opened under a symlinked path is not found by `self.documents.get`, so the hover quote-widening never runs


baseballyama:
When a document is opened under a path that contains a symlink, rsvelte's
`self.documents.get(source_uri)` misses, because the key the client sent and the
key the server derives are two different URIs for the same file. Official is not
sensitive to this.

The observable loss in the case measured here is
`widen_hover_range_over_string_quotes` (`tsgo_response.rs:783`), which is guarded
by that lookup (`server.rs:3316-3323`), so a hover on a module specifier comes
back with tsgo's inner-text range instead of the quote-inclusive range
TypeScript's quick info spans.

## Measured

One probe, one workspace, **two arms differing only in whether the temp
directory is realpathed** — macOS `/var` is a symlink to `/private/var`, so the
client opens `file:///var/folders/…/App.svelte` and rsvelte resolves it. Source
line 1 is `\timport { helper } from './other';`, hover at 1:28.

| arm | client's document URI | official | rsvelte |
|---|---|---|---|
| `/var/folders/…` (symlinked) | `file:///var/folders/…` | `1:24-1:33` | **`1:25-1:32`** |
| `/private/var/folders/…` (realpathed) | `file:///private/var/…` | `1:24-1:33` | `1:24-1:33` |

Both quote styles move together — `"./two"` reads `3:21-3:28` official against
`3:22-3:27` rsvelte on the symlinked arm and `3:21-3:28` on both arms on the
realpathed one — so it is 2 cells, not 1.

## Attribution

An `eprintln!` at the guard names it directly, and the same line is printed for
**every** hover request in the run, not only the diverging ones:

```
[zz] hover source_uri=Some("file:///private/var/folders/…/App.svelte")
     found=false range_before={"start":{"line":1,"character":25},"end":{…,"character":32}}
```

`found=false` on 9 of 9 requests on the symlinked arm, `found=true` on 9 of 9 on
the realpathed one. So the widening is not wrong — it is never reached.

## Not measured

`self.documents` is read in many places, and this probe only shows what the
hover guard loses. Whether other features that key on the same map degrade under
a symlinked workspace is **UNMEASURED**; the fix should be at the key, not at
this one call site.

## Why no gate sees it

`scripts/compat-lsp/` drives both servers against paths under the repository
checkout, which is not symlinked, so every unit of the LSP differential gate is
on the insensitive side of this axis.

## How it surfaced

[...]

PR #4483, opened 2026-09-09
fix(svelte2tsx): give a component's tag name its own source range

### LINKED ISSUE #4097: lsp: hover returns null on a .svelte module specifier and on a component tag name
`rsvelte-language-server` returns `null` for `textDocument/hover` at exactly two positions that name a component: the `'./X.svelte'` **module specifier** of an import, and a **component tag name** in markup. Every other position in the same file answers correctly, including a value imported *from* that same `.svelte` module, so this is not "`.svelte` modules are unresolved" — it is the two positions whose subject is the component itself.

## Repro

Workspace: a plain `tsconfig.json` and three files. No `node_modules`.

`src/other.ts`
```ts
export const helper = 1;
```

`src/Other.svelte`
```svelte
<script module lang="ts">
	export const fromSvelte = 1;
</script>

<p>other</p>
```

`src/App.svelte`
```svelte
<script lang="ts">
	import { helper } from './other';
	import Other, { fromSvelte } from './Other.svelte';
	const h = helper;
	const v = fromSvelte;
</script>

<Other />
{h}{v}
```

Driven over stdio: `initialize` (`initializationOptions` as `scripts/compat-lsp/verify.mjs` builds them, `TSGO_BIN` pointing at the pinned `@typescript/native-preview`) + `didOpen` + `textDocument/hover`.

## Actual output, both sides

| position | official | rsvelte |
|---|---|---|
| `'./Other.svelte'` specifier, `2:39` | `` ```typescript\nmodule "./Other.svelte"\n``` `` | **`null`** |
| `Other` tag in markup, `7:2` | `` ```typescript\nimport Other\n``` `` | **`null`** |
| `'./other'` specifier (a `.ts` module), `1:27` | `module "./other"` | `module "./other"` ✅ |
| `h` in a mustache, `8:1` | `const h: 1` | `const h: 1` ✅ |
| `v` in a mustache (imported from the `.svelte` module), `8:5` | `const v: 1` | `const v: 1` ✅ |
| `fromSvelte` use, `4:12` | `(alias) const fromSvelte: 1`⏎`import fromSvelte` | `(alias) const fromSvelte: 1` (separate issue: the missing `import` line is a tsgo-vs-tsc difference) |

Raw JSON for the two null cells:

```
specifier './Other.svelte' (2:39)
  official {"range":{"start":{"line":2,"character":35},"end":{"line":2,"character":51}},"contents":"```typescript\nmodule \"./Other.svelte\"\n```"}
  rsvelte  null

`Other` in markup (7:2)
  official {"range":{"start":{"line":7,"character":1},"end":{"line":7,"character":6}},"contents":"```typescript\nimport Other\n```"}
  rsvelte  null
```

## Which cells reproduce and which do not

The rows marked ✅ are the controls, and they matter: a markup mustache, a `.ts` specifier, and a value imported out of the very same `.svelte` file all answer identically on both servers. So the null is not the projection failing on this file, not markup positions in general, and not `.svelte` imports in general.

Two further discriminating measurements from the corpus run below:

- All **187** of the corpus's `module "…"` hovers where rsvelte returns null have a specifier ending in `.svelte` — **187 / 187**, no non-`.svelte` specifier is affected.
- Every one of these positions round-trips through the projection **exactly**: `map_source_position` → `map_generated_position` returns the original position (`identity`) for 187/187 of the specifier cases. The position mapping is not the cause.

## Not a backend difference

The two TypeScript backends agree that a module specifier has a hover. Measured on one plain `.ts` file, same position, no Svelte involved:

| | tsc 6.0.3 (what official loads) | tsgo (pinned `@typescript/native-preview`) |
|---|---|---|
| `'./other'` specifier | `module "./other"` | `module "./other"` |

So the `null` is produced on rsvelte's side of the tsgo proxy, not by tsgo.

## Scale

On the LSP differential gate's own corpus, `submodules/melt-ui` in full (43 files / 12,043 identifiers / 72,258 requests), over the 2,876 open-phase `textDocument/hover` divergences:

| | open-phase divergences |
|---|---|
| a `.svelte` module specifier, rsvelte null | 187 |
| a component tag name in markup, rsvelte null | 72 |

The edit-phase divergence set is byte-identical to the open-phase set, so both appear twice in `compatibility/lsp-known-failures.json`. melt-ui holds 1.9% of the gate's identifiers; the counts are measured there and are not extrapolated.

baseballyama:
Fixes #4097.

Two positions whose subject is the component itself answered `null`, and they are
two different defects that happen to share a symptom.

## (a) A component's tag name had no source range

`handle_component` interpolated the tag name into the
`__sveltets_2_ensureComponent(<name>)` header with `format!`, so the name reached
the shadow as generated text with no map segment and a hover on a `<Comp />` tag
resolved through whatever mapping sat to its left. Upstream pushes
`[nodeNameStart, nodeNameEnd]` — a source **range** — into that exact argument
position (`InlineComponent.ts:101-110`), locating the name with
`str.original.indexOf(node.name, node.start)`, and `<svelte:component this={…}>`
is the same shape with `[node.expression.start, node.expression.end]`. Both now go
through `segs_push_src`, so the name and the `this` expression reach the shadow as
unedited source chunks. The generated TSX is unchanged; only the map moves.

## (b) A `.tsx` specifier insertion was classified as generated code

The overlay rewrites `'./Other.svelte'` to `'./Other.svelte.tsx'` so tsgo can
resolve it, and registered those four inserted bytes in the same
`generated_ranges` list as the `Ωignore` regions. `is_generated_range` is an
intersection test and `map_generated_range` bails on it before mapping anything,
so a tsgo response whose range covers the specifier was discarded and the client
saw `null`.

Instrumenting `map_range` is what named it, and it falsified the reading the code
suggested: **both endpoints map, and map correctly** — the shadow range is 18
columns and its image is 14, so the insertion was already accounted for.

    in=3:35-3:53  out=None  start=Some(2:36)  end=Some(2:50)  is_gen_range=true

The two kinds are not the same thing: an `Ωignore` region is whole synthetic text
with no source, while an insertion is four synthetic bytes inside a token the user
wrote — and `map_generated_position` never consulted the predicate at all, so a
bare position already passed through an insertion while a range spanning the same
one did not. The insertions move to `specifier_insertions`, consulted for a
position as before and not for a range. A range lying wholly inside an insertion
still returns `None`, because both its endpoints fail the position check. The
three tests that pin this predicate are every one of them about `Ωignore` regions;
`git log -S` on the `extend` reaches only a squashed history commit, so nothing
recorded a reason for the union.

## Measured against the official language server

`zz-hover-4097.mjs` drives the issue's own repro against the pinned official
server over stdio with `verify.mjs`'s `initializationOptions`. Positions are
derived from the source text and printed beside the issue's claimed coordinates,
so a stale coordinate shows as a mismatch rather than as a silently different
cell — three of the issue's are stale and are flagged. Four arms, each built and
hashed here, each adding one change to the previous:

    A  fc1a0b917b6f259b   main
    B  e6ac92eee4764865   A + #4385
    C  5b9753c86dde0368   B + (a)
    D  b8eab7c48b666303   C + (b)

| cell | A | B | C | D |
|---|---|---|---|---|
| `<Other />` tag name | null | null | official's range `10:1-10:6` | same |
| `<Deep />` tag name | null | null | official's range `11:1-11:5` | same |
| `'./Other.svelte'` specifier | null | null | null | **EQ**, `module "./Other.svelte"` `2:35-2:51` |
| `'./deep/Deep.svelte'` specifier | null | null | null | **EQ**, `4:18-4:38` |
| `'./other'`, `"./two"` (`.ts`) | EQ | EQ | EQ | EQ |
| `{h}`, `{v}` mustaches | EQ | EQ | EQ | EQ |

`#4385` does not reach either position, which is what makes B a control rather
than a step. The release binary was rebuilt from the committed tree after the
probe and hashes byte-identically to D.

## What is left, and why it is not this issue

Five cells still differ and both axes are ones the issue itself separates out.
The `fromSvelte` row it calls "a tsgo-vs-tsc difference". The tag-name cells now
carry official's exact range and differ in *contents* —
`(alias) const Other: __sveltets_2_IsomorphicComponent<…>` where official says
`import Other` — and that reproduces **at the import clause**, with identical
ranges on both sides, on a position no part of this PR touches. It is a
pre-existing alias-rendering divergence, not something the tag introduces.

## Two workspace properties are measurement inputs

Both produced a wrong finding before they were controlled. macOS `/var` is a
symlink to `/private/var`, so a workspace under `os.tmpdir()` makes the server's
document lookup miss on every request — filed as #4458, and it is what produced
an earlier, retracted claim that rsvelte answers a *narrower* range than
official. And a workspace outside the checkout has no `node_modules` on any
ancestor; that one was hypothesised to explain the tag-name contents gap and was
**falsified** — re-running with the workspace under the worktree gave
byte-identical results.

## Blast radius

`crates/rsvelte_language_server`: 409 tests over 8 targets, 0 failed — main's 409 declared minus its one `#[ignore]`d, plus this branch's new test, whose name is in the run. Workspace
clippy with `-D warnings` clean. The new overlay test is two-sided and was
ablated — restoring the union reddens it at the range assertion and nowhere else.

## The LSP fixture gate moves, and what moves is `textDocument/inlayHint`

Measured as two arms in this checkout — one holding #4385 only, one holding #4385 + both
halves of this PR — so the ~208-entry baseline churn this checkout produces against
`origin/main` (and CI does not) cancels between them. Ratchet total 2104 -> 2102; the
`textDocument/inlayHint` subset 314 -> 328 keys, 143 -> 151 diverging files, 8 newly
diverging and 0 newly clean.

| `inlayHint` class | before | after |
|---|---:|---:|
| `:extra-rsvelte` (both sides lists, rsvelte has surplus elements) | 130 | **152** |
| `:missing-rsvelte` | 28 | 26 |
| value mismatch, official `null`, rsvelte a list | 92 | **110** |
| value mismatch, both lists | 58 | **36** |
| value mismatch, official `null`, rsvelte `[]` | 6 | 4 |

The `+22` extra and `-22` both-lists pair up: a file that used to differ in hint *content*
now differs by carrying *more* hints. That is this PR's doing and it is expected —
`textDocument/inlayHint` is forwarded generically (`server.rs:711`), so the response mapper
drops any hint whose position does not map, and until (a) the `__sveltets_2_ensureComponent`
argument was unmapped generated text. Giving the tag name a source range removes an
accidental suppression.

The deliberate suppression was never ported, and that is not this PR: upstream's
`InlayHintProvider.ts:66-74` filters six classes of generated hint and rsvelte ports **none**
of them. Measured on `main` with no part of this PR in the tree, a ten-line component
produces seven hints official does not (`type:`, `options:` and
`: __sveltets_2_IsomorphicComponent<…>` on each `<Comp />`, plus the render function's return
type) — filed with the probe as #4464, which also records the opposite defect the same probe
found: a request whose range **starts at 0:0** answers `[]`, which is the range a real editor
sends. The standing backlog there is 314 keys over 143 files; this PR moves it by 14, i.e.
4.5%, and porting one of six filter rules inside a hover fix would be scope creep.

One reading caution for the re-baseline: `74234e98afe7` in a key is the digest of **`null`**,
not of `[]` (`[]` is `4f53cda18c2b`). I had read those keys the other way at first, and they
are the difference between "official found no hints" and "official's arm had hints disabled".

Read CI's own delta on this PR rather than any number measured here.

**The ratchet is NOT updated in this PR as pushed**, and that is deliberate rather than an
oversight. `verify.mjs:54-59` refuses a direct `--update-baseline`, so a correct re-baseline needs
the fixture artifact plus all 16 corpus artifacts from one `workflow_dispatch` Corpus Compat run at
one revision, through `merge-current.mjs --update-baseline`. Until that run exists on this branch,
`LSP fixture parity current` is expected to fail here, and the failing run's own artifact is what
the re-baseline must be taken from.

Do not take it from a local run in this checkout. This checkout produces ≈208 `textDocument/inlayHint`
keys that differ from `origin/main` and that CI does not reproduce — measured twice, once as the
two-arm cancellation described above and once directly, where CI's value equals `origin/main`'s for
all 208 while a local artifact's matched neither. The cause is not established; the best-supported
hypothesis is the tsgo **platform build**, since the pinned `~7.0.0-dev.20260703.1` resolves to seven
different binaries and this machine runs `darwin-arm64` against CI's `linux-x64`. That is consistent
with the churn being inlayHint-only (those payloads carry tsgo-rendered type text), with it being
independent of which rsvelte fix is in the arm, and with CI and `origin/main` agreeing exactly. It
is not measured. Tracked as #4479.

### The corpus half is not measured on this PR, so the dispatch run is required

Measured on this PR's own file list, with both halves read because they answer different questions:

```
$ node scripts/ci/corpus-compat-job-filter.mjs --changed-files <this PR's files>
lsp-fixtures-current=true
lsp-corpus=true
lsp-ratchet=false          <- no compatibility/ or scripts/compat-lsp/ path in the diff

corpus-compat.yml, lsp-corpus:
  if: (schedule || workflow_dispatch || lsp-ratchet == 'true') && lsp-corpus != 'false'
```

On a `pull_request` the event is neither `schedule` nor `workflow_dispatch` and `lsp-ratchet` is
false, so the first disjunct fails and **the 16 real-world shards do not run here** — even though
the filter's blast radius says `lsp-corpus=true`. The fixture gate does run, and will fail until
the re-baseline lands.

That asymmetry is the reason the `workflow_dispatch` run is not optional. This PR changes the
projection, so corpus-scope keys can move; nothing on the PR would measure that, and the first run
to see it would be the nightly — on `main`, after merge. That is the shape #4465 records for the
`#4407` submodule bump, where an unmeasured population entered the gate and every ratchet-touching
PR afterwards failed on it.

### DIFF
--- a/.changeset/component-name-source-range.md
+++ b/.changeset/component-name-source-range.md
@@ -1,0 +1,19 @@
+---
+'@rsvelte/compiler': patch
+'@rsvelte/svelte2tsx': patch
+'@rsvelte/svelte-check': patch
+---
+
+fix(svelte2tsx): give a component's tag name its own source range
+
+`handle_component` interpolated the tag name into the
+`__sveltets_2_ensureComponent(<name>)` header with `format!`, so the name
+reached the shadow as generated text with no map segment and
+`textDocument/hover` on a `<Comp />` tag answered `null`. Upstream pushes
+`[nodeNameStart, nodeNameEnd]` — a source range — into that exact position
+(`InlineComponent.ts:101-110`), and `<svelte:component this={…}>` is the same
+shape with `[node.expression.start, node.expression.end]`.
+
+Both now go through `segs_push_src`, so the name and the `this` expression reach
+the shadow as unedited source chunks. The generated TSX is unchanged; only the
+map moves.
--- a/.changeset/svelte-specifier-hover.md
+++ b/.changeset/svelte-specifier-hover.md
@@ -1,0 +1,19 @@
+---
+'@rsvelte/language-server': patch
+---
+
+fix(language-server): hover a `.svelte` import specifier
+
+The overlay rewrites `'./Other.svelte'` to `'./Other.svelte.tsx'` so tsgo can
+resolve it, and registered those four inserted bytes in the same
+`generated_ranges` list as the `Ωignore` regions. `is_generated_range` is an
+intersection test and `map_generated_range` bails on it, so every tsgo response
+whose range covers the specifier — the hover, and any range-carrying feature
+anchored there — was discarded before it was mapped, and the client saw `null`.
+
+The two kinds are not the same thing: an `Ωignore` region is whole synthetic
+text with no source, while a `.tsx` insertion is four synthetic bytes inside a
+token the user wrote. The insertions move to their own list, consulted for a
+*position* as before and not for a *range*; a range lying wholly inside an
+insertion still returns `None`, because both its endpoints fail the position
+check.
--- a/crates/rsvelte_language_server/src/tsgo_overlay.rs
+++ b/crates/rsvelte_language_server/src/tsgo_overlay.rs
@@ -179,6 +179,10 @@ struct ShadowState {
     source_map: Option<String>,
     tokens: Vec<MappingToken>,
     generated_ranges: Vec<std::ops::Range<usize>>,
+    /// The `.tsx` bytes inserted into a `.svelte` module specifier. Unlike an
+    /// `Ωignore` region these sit *inside* a real token, so a range that spans
+    /// one is still the user's own text and must map.
+    specifier_insertions: Vec<std::ops::Range<usize>>,
     identity: bool,
     plain_insertions: Vec<(usize, std::ops::Range<usize>)>,
     /// Byte offset in `source_text` that generated offset 0 corresponds to,
@@ -190,6 +194,16 @@ struct ShadowState {
     /// `None` for an `identity` entry, whose text is the user's own: a
     /// `function $$render()` they wrote is a real declaration, not ours.
     render_return_type: Option<usize>,
+}
+
+impl ShadowState {
+    /// Whether a shadow byte offset carries no source text of its own.
+    fn is_generated_offset(&self, offset: usize) -> bool {
+        self.generated_ranges
+            .iter()
+            .chain(&self.specifier_insertions)
+            .any(|range| range.contains(&offset))
+    }
 }
 
 /// Workspace-scoped diskless overlay used by the tsgo LSP proxy.
@@ -390,12 +404,11 @@ impl TsgoOverlay {
             text: generated_text,
             version,
         };
-        let mut generated_ranges = ignored_ranges(&document.text);
-        generated_ranges.extend(
-            import_insertions
-                .iter()
-                .map(|(_, generated)| generated.clone()),
-        );
+        let generated_ranges = ignored_ranges(&document.text);
+        let specifier_insertions = import_insertions
+            .iter()
+            .map(|(_, generated)| generated.clone())
+            .collect();
         let state = ShadowState {
             source_path: source_path.clone(),
             shadow_path: shadow_path.clone(),
@@ -414,6 +427,7 @@ impl TsgoOverlay {
             tokens,
             generated_ranges,
             render_return_type: render_return_type_offset(&document.text),
+            specifier_insertions,
             identity: false,
             plain_insertions: Vec::new(),
             fragment_offset: 0,
@@ -468,6 +482,7 @@ impl TsgoOverlay {
             source_map: None,
             generated_ranges: Vec::new(),
             render_return_type: None,
+            specifier_insertions: Vec::new(),
             identity: true,
             plain_insertions: Vec::new(),
             fragment_offset,
@@ -531,6 +546,7 @@ impl TsgoOverlay {
             source_map: None,
             generated_ranges: Vec::new(),
             render_return_type: None,
+            specifier_insertions: Vec::new(),
             identity: true,
             plain_insertions,
             fragment_offset: 0,
@@ -812,11 +828,7 @@ impl TsgoOverlay {
                     .position(&entry.source_text, source_offset),
             );
         }
-        if entry
-            .generated_ranges
-            .iter()
-            .any(|range| range.contains(&generated_offset))
-        {
+        if entry.is_generated_offset(generated_offset) {
             return None;
         }
 
@@ -868,10 +880,7 @@ impl TsgoOverlay {
             return false;
         };
         let offset = utf8_offset(&entry.document.text, position);
-        entry
-            .generated_ranges
-            .iter()
-            .any(|range| range.contains(&offset))
+        entry.is_generated_offset(offset)
     }
 
     /// Whether a tsgo position is the return-type slot of the generated
@@ -891,7 +900,10 @@ impl TsgoOverlay {
         utf8_offset(&entry.document.text, position) == offset
     }
 
-    /// Whether any byte of a tsgo range intersects generated-code markers.
+    /// Whether any byte of a tsgo range intersects an `Ωignore` region.
+    ///
+    /// A `.tsx` specifier insertion is deliberately not counted: it sits inside
+    /// a token the user wrote, so the enclosing range is theirs and must map.
     #[must_use]
     pub fn is_generated_range(&self, shadow_path: &Path, range: Range) -> bool {
         let Some(source_path) = self.source_for_shadow(shadow_path) else {
@@ -3056,6 +3068,57 @@ mod tests {
         let position = utf8_position(&shadow.text, ignored);
         assert!(overlay.is_generated_position(&shadow_path, position));
         assert_eq!(overlay.map_generated_position(&shadow_path, position), None);
+    }
+
+    #[test]
+    fn a_svelte_specifier_range_maps_back_across_its_tsx_insertion() {
+        let workspace = TestWorkspace::new("specifier-range");
+        let other = workspace.0.join("Other.svelte");
+        write(&other, "<p>other</p>\n");
+        let app = workspace.0.join("App.svelte");
+        let source = "<script>import Other from './Other.svelte';</script>\n<Other />\n";
+        write(&app, source);
+        let overlay = TsgoOverlay::build(&workspace.0, None).unwrap();
+        let shadow_path = overlay.shadow_dir.join("App.svelte.tsx");
+        let shadow = overlay.shadow_for_source(&app).unwrap();
+
+        let rewritten = "'./Other.svelte.tsx'";
+        let start = shadow.text.find(rewritten).unwrap();
+        let insertion = shadow.text.find(".tsx'").unwrap();
+
+        // The whole specifier: the user's own token, four synthetic bytes long.
+        let quoted = Range::new(
+            utf8_position(&shadow.text, start),
+            utf8_position(&shadow.text, start + rewritten.len()),
+        );
+        assert!(!overlay.is_generated_range(&shadow_path, quoted));
+        let mapped = overlay.map_generated_range(&shadow_path, quoted).unwrap();
+        let mapped_text = {
+            let index = crate::text::LineIndex::new(source);
+            let from = index.offset(source, mapped.start);
+            let to = index.offset(source, mapped.end);
+            source[from..to].to_string()
+        };
+        assert_eq!(mapped_text, "'./Other.svelte'");
+
+        // A range lying wholly inside the insertion has no source text at all,
+        // so narrowing the range rule must not let it through.
+        let synthetic = Range::new(
+            utf8_position(&shadow.text, insertion),
+            utf8_position(&shadow.text, insertion + 4),
+        );
+        assert_eq!(overlay.map_generated_range(&shadow_path, synthetic), None);
+        assert!(overlay.is_generated_position(&shadow_path, synthetic.start));
+
+        // An `Ωignore` region is the other kind and still rejects a range that
+        // merely touches it.
+        let ignored = shadow.text.find(IGNORE_START).unwrap();
+        let touching = Range::new(
+            utf8_position(&shadow.text, ignored.saturating_sub(2)),
+            utf8_position(&shadow.text, ignored + IGNORE_START.len() + 1),
+        );
+        assert!(overlay.is_generated_range(&shadow_path, touching));
+        assert_eq!(overlay.map_generated_range(&shadow_path, touching), None);
     }
 
     #[cfg(unix)]
--- a/crates/rsvelte_projection/src/svelte2tsx/template/nodes/inline_component.rs
+++ b/crates/rsvelte_projection/src/svelte2tsx/template/nodes/inline_component.rs
@@ -7,6 +7,18 @@ fn source_offset(value: usize) -> u32 {
     u32::try_from(value).expect("template source offsets are represented as u32")
 }
 
+/// The source range of a component's tag name, located the way upstream locates
+/// it (`InlineComponent.ts:103`, `str.original.indexOf(node.name, node.start)`).
+fn component_name_range(comp: &Component, source: &str) -> Option<(u32, u32)> {
+    let start = comp.start as usize;
+    let offset = source.get(start..)?.find(comp.name.as_str())?;
+    let name_start = start + offset;
+    Some((
+        source_offset(name_start),
+        source_offset(name_start + comp.name.len()),
+    ))
+}
+
 use crate::ast::template::{
     Attribute, Component, SvelteComponentElement, SvelteElement, TemplateNode,
 };
@@ -26,7 +38,9 @@ use crate::svelte2tsx::template::attributes::let_::{
 };
 use crate::svelte2tsx::template::attributes::spread::format_spread_attribute_segments;
 use crate::svelte2tsx::template::ctx::{Counter, ElementOpenerCommentIndex};
-use crate::svelte2tsx::template::segs::{Seg, bake_out_of_order_src, emit_segmented_overwrite};
+use crate::svelte2tsx::template::segs::{
+    Seg, bake_out_of_order_src, emit_segmented_overwrite, segs_push_lit, segs_push_src,
+};
 use crate::svelte2tsx::template::utils::expr::{
     extend_expr_end_with_ts_postfix, get_binding_lhs_text, get_expression_end_stripping_ts,
     get_expression_range, get_expression_text, get_set_binding_ranges,
@@ -315,30 +329,35 @@ pub fn handle_component(
     // is appended (as ignore-wrapped statements) for every non-`bind:this`
     // binding on a component.
     let component_bind_suffix = build_component_bind_suffix(&comp.attributes, source, &inst_var);
-    let (header_lit, trailer_lit) = if needs_instance {
+    let (header_pre, header_post, trailer_lit) = if needs_instance {
         let on_calls = if has_events {
             build_on_calls(&inst_var, &on_directives, source)
         } else {
             String::new()
         };
         (
+            format!("{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent("),
             format!(
-                "{}{{ const {}C = __sveltets_2_ensureComponent({}); const {} = new {}C({{ target: __sveltets_2_any(), props: {{",
-                block_indent, inst_var, comp.name, inst_var, inst_var,
+                "); const {inst_var} = new {inst_var}C({{ target: __sveltets_2_any(), props: {{"
             ),
             format!("}}}});{component_bind_suffix}{on_calls}"),
         )
     } else {
         (
-            format!(
-                "{}{{ const {}C = __sveltets_2_ensureComponent({}); new {}C({{ target: __sveltets_2_any(), props: {{",
-                block_indent, inst_var, comp.name, inst_var,
-            ),
+            format!("{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent("),
+            format!("); new {inst_var}C({{ target: __sveltets_2_any(), props: {{"),
             "}});".to_string(),
         )
     };
-    let mut opener_segs: Vec<Seg> = Vec::with_capacity(attr_segs.len() + 2);
-    opener_segs.push(Seg::Lit(header_lit));
+    let mut opener_segs: Vec<Seg> = Vec::with_capacity(attr_segs.len() + 4);
+    opener_segs.push(Seg::Lit(header_pre));
+    // The name is the `__sveltets_2_ensureComponent` argument and upstream gives
+    // it its own source range, so hover on the tag name resolves through it.
+    match component_name_range(comp, source) {
+        Some((start, end)) => segs_push_src(&mut opener_segs, start, end),
+        None => segs_push_lit(&mut opener_segs, comp.name.as_str()),
+    }
+    segs_push_lit(&mut opener_segs, &header_post);
     opener_segs.extend(attr_segs);
     if !use_snippet_props {
         // The snippet-prop path leaves the `props: { … ` object literal open so
@@ -703,30 +722,39 @@ pub fn handle_svelte_component(
     } else {
         " ".repeat(scomp_spacing.before_block)
     };
-    let (header_lit, trailer_lit) = if needs_inst {
+    let (header_pre, header_post, trailer_lit) = if needs_inst {
         let on_calls = if has_events {
             build_on_calls(&inst_var, &on_directives, source)
         } else {
             String::new()
         };
         (
+            format!("{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent("),
             format!(
-                "{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent({expr_text}); const {inst_var} = new {inst_var}C({{ target: __sveltets_2_any(), props: {{"
+                "); const {inst_var} = new {inst_var}C({{ target: __sveltets_2_any(), props: {{"
             ),
             format!("}}}});{component_bind_suffix}{on_calls}"),
         )
     } else {
         (
-            format!(
-                "{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent({expr_text}); new {inst_var}C({{ target: __sveltets_2_any(), props: {{"
-            ),
+            format!("{block_indent}{{ const {inst_var}C = __sveltets_2_ensureComponent("),
+            format!("); new {inst_var}C({{ target: __sveltets_2_any(), props: {{"),
             "}});".to_string(),
         )
     };
     // The snippet-props path keeps the props object open so the demoted
     // `{#snippet}` children can be moved inside it.
-    let mut opener: Vec<Seg> = Vec::with_capacity(attr_segs.len() + 3);
-    opener.push(Seg::Lit(header_lit));
+    let mut opener: Vec<Seg> = Vec::with_capacity(attr_segs.len() + 4);
+    opener.push(Seg::Lit(header_pre));
+    // `this={expr}` is upstream's `[node.expression.start, node.expression.end]`
+    // entry, so the expression reaches the shadow as an unedited source chunk.
+    match get_expression_range(&comp.expression) {
+        Some((start, end)) if !expr_text.is_empty() => {
+            segs_push_src(&mut opener, start, end);
+        }
+        _ => segs_push_lit(&mut opener, expr_text),
+    }
+    segs_push_lit(&mut opener, &header_post);
     opener.extend(attr_segs);
     if !use_snippet_props {
         opener.push(Seg::Lit(trailer_lit.clone()));
--- a/crates/rsvelte_projection/tests/svelte2tsx_component_name_map_4097.rs
+++ b/crates/rsvelte_projection/tests/svelte2tsx_component_name_map_4097.rs
@@ -1,0 +1,145 @@
+//! A component's tag name was interpolated into the
+//! `__sveltets_2_ensureComponent(<name>)` header as generated text, so the name
+//! reached the shadow with no map segment and `textDocument/hover` on a
+//! `<Comp />` tag answered `null` where official answers. Upstream pushes
+//! `[nodeNameStart, nodeNameEnd]` — a source RANGE — into the same position
+//! (`InlineComponent.ts:101-110`), and `this={…}` on `<svelte:component>` is
+//! `[node.expression.start, node.expression.end]` at `:102`.
+//!
+//! The generated TSX is unchanged, so neither the 253 svelte2tsx fixtures nor
+//! the map gate (which asserts well-formedness, not equality) can see this.
+//! The attribute-value cells already pass on `main` — they are regression
+//! guards, not discriminating cells.
+
+use rsvelte_projection::svelte2tsx::{Svelte2TsxOptions, svelte2tsx};
+use std::collections::HashSet;
+
+struct Projected {
+    code: String,
+    map: sourcemap::SourceMap,
+}
+
+fn project(source: &str) -> Projected {
+    let result = svelte2tsx(
+        source,
+        Svelte2TsxOptions {
+            filename: "A.svelte".to_string(),
+            is_ts_file: true,
+            ..Default::default()
+        },
+    )
+    .expect("svelte2tsx");
+    let map = sourcemap::SourceMap::from_slice(result.map.as_deref().expect("map").as_bytes())
+        .expect("valid source map");
+    Projected {
+        code: result.code,
+        map,
+    }
+}
+
+fn mapped_source_positions(projected: &Projected) -> HashSet<(u32, u32)> {
+    projected
+        .map
+        .tokens()
+        .map(|token| (token.get_src_line(), token.get_src_col()))
+        .collect()
+}
+
+/// `(line, first column)` of `needle` in `text`.
+fn locate(text: &str, needle: &str) -> (u32, u32) {
+    let line = text
+        .lines()
+        .position(|line| line.contains(needle))
+        .unwrap_or_else(|| panic!("needle {needle:?} not found in\n{text}"));
+    let column = text.lines().nth(line).unwrap().find(needle).unwrap();
+    (line as u32, column as u32)
+}
+
+fn span_of(source: &str, marker: &str, name: &str) -> (u32, u32, u32) {
+    // `marker.find(name)` takes the FIRST occurrence, so a name that also occurs
+    // inside the attribute/directive name silently points the assertion at text
+    // that is never mapped (`style:color={col}` finds the `col` of `color`).
+    assert_eq!(
+        marker.matches(name).count(),
+        1,
+        "{name:?} occurs {} times in {marker:?}; pick a name unique within the marker",
+        marker.matches(name).count(),
+    );
+    let (line, column) = locate(source, marker);
+    let start = column + (marker.find(name).expect("name inside marker") as u32);
+    (line, start, start + name.len() as u32)
+}
+
+/// Every character of `name` inside `marker` carries a map segment.
+fn assert_mapped(source: &str, marker: &str, name: &str) {
+    let projected = project(source);
+    let mapped = mapped_source_positions(&projected);
+    let (line, start, end) = span_of(source, marker, name);
+    for c in start..end {
+        assert!(
+            mapped.contains(&(line, c)),
+            "{marker}: source {line}:{c} carries no map segment\n--- generated ---\n{}",
+            projected.code,
+        );
+    }
+}
+
+/// No character of `name` inside `marker` carries a map segment.
+fn assert_unmapped(source: &str, marker: &str, name: &str) {
+    let projected = project(source);
+    let mapped = mapped_source_positions(&projected);
+    let (line, start, end) = span_of(source, marker, name);
+    for c in start..end {
+        assert!(
+            !mapped.contains(&(line, c)),
+            "{marker}: source {line}:{c} unexpectedly carries a map segment\n--- generated ---\n{}",
+            projected.code,
+        );
+    }
+}
+
+const HEAD: &str =
+    "<script lang=\"ts\">\n\timport Comp from './Comp.svelte';\n\tlet hue = 'red';\n</script>\n\n";
+
+fn with_head(body: &str) -> String {
+    format!("{HEAD}{body}\n")
+}
+
+#[test]
+fn a_component_tag_name_carries_its_own_map_segments() {
+    assert_mapped(&with_head("<Comp title={hue} />"), "<Comp title=", "Comp");
+}
+
+#[test]
+fn a_component_tag_name_is_mapped_when_the_tag_has_no_attributes() {
+    assert_mapped(&with_head("<Comp />"), "<Comp />", "Comp");
+}
+
+#[test]
+fn a_svelte_component_this_expression_carries_its_own_map_segments() {
+    assert_mapped(
+        &with_head("<svelte:component this={Comp} title={hue} />"),
+        "this={Comp}",
+        "Comp",
+    );
+}
+
+/// Pre-registered as already passing on `main`: the attribute value in the same
+/// start tag is segmented by the opener work this builds on, so a green here
+/// says the name change did not un-map its neighbours.
+#[test]
+fn an_attribute_value_beside_the_name_stays_mapped() {
+    assert_mapped(&with_head("<Comp title={hue} />"), "title={hue}", "hue");
+}
+
+/// `<svelte:self>` is lowered to `__sveltets_2_createComponentAny` and upstream
+/// pushes no name range for it, so the source text of the tag name must stay
+/// unmapped — the negative direction of the same check.
+#[test]
+fn a_svelte_self_tag_name_carries_no_map_segment() {
+    assert_unmapped(
+        &with_head("<svelte:self title={hue} />"),
+        "<svelte:self ",
+        "svelte:self",
+    );
+}


baseballyama:
`Language server` and `LSP fixture parity current` are red, and `Tests` is red only as their rollup (`LANGUAGE_SERVER: failure`, every other shard `success`). Both LSP jobs report the identical verdict, which is expected — they run the same `verify.mjs --suites fixtures,upstream-features,upstream-testfiles` invocation.

**This is not a plain re-baseline.** The raw counts are 284 NEW / 286 stale, but the counts are not the quantity that decides anything. Stripping the `[...]` brackets to compare divergence *identities* rather than key strings:

```
re-keyed (same divergence, new hash)   228
retired                                 58
GENUINELY NEW                           56   <- these need a cause, not a baseline
```

The 56, by method and divergence kind:

```
  22  textDocument/inlayHint    extra-rsvelte
  16  textDocument/diagnostic   value-mismatch
  16  textDocument/inlayHint    value-mismatch
   2  textDocument/inlayHint    missing-rsvelte
```

over 21 carrier files, the largest being `upstream-features/component-invalid/input.svelte` (8 differential + 8 expected).

## Why the cause is not yet attributed

This branch carries **two commits and two changesets** — `fix(svelte2tsx): give a component's tag name its own source range` and `fix(language-server): a .tsx specifier insertion is not a generated region` — deliberately, since the PR body states they are two defects sharing one symptom. Both make previously-unmappable positions map, so both predict "more things map, so more hints are emitted", and the 56 have **two candidate causes with one measurement between them**.

The carrier list does not separate them either: it holds component fixtures (`component-invalid`, `find-component-references-parent`, `component-events-completion`) *and* import fixtures (`updateimports`, `organize-imports-group`, `remove-imports-inline`, `emptytext-importer`, `unresolvedimport2`). One artifact cannot say which commit produced which key; that needs one arm per commit.

## The hypothesis, stated as one

40 of the 56 are `inlayHint`, and #4464 records that rsvelte's `inlayHint` path ports **none** of upstream's `InlayHintProvider` filters — including `isSvelte2tsxFunctionHints`, which drops any hint whose containing expression starts with `$$_`, i.e. exactly the generated component constructors this PR gives real source ranges to. So the likely shape is: this fix is correct, and it makes an already-missing filter reachable — GATES.md 39b, where repairing one divergence exposes its children.

That is a hypothesis with a cheap test attached, and it is **not** evidence yet. If it holds, landing #4464 first should absorb most of the 22 `extra-rsvelte` and the 16 `inlayHint value-mismatch` before this PR is measured again, which is both the correct order and the one that keeps each PR to one cause.

## Next

1. Build the two arms (one commit each) and attribute the 56 — a count of 56 with two possible producers is not a finding.
2. If the inlayHint majority is #4464's missing filter, land #4464 first and re-measure rather than baselining it here.
3. Whatever remains after that gets re-baselined with a per-cluster justification, not as a block.

Note for whoever picks this up: **do not build the stale or NEW list from the job log.** Both are printed through `slice(0, SHOW)` with `SHOW = 30` and no "and N more" line (#4484), so at 284 and 286 the log shows 30 of each. These figures come from the run's `lsp-current-fixtures` artifact, recomputed with the gate's own exported `selectKnownForScope`, and validated by reproducing the gate's published 284/286 exactly.

baseballyama:
## Decision: this is blocked on #4464, not re-baselined here

40 of the 56 genuinely-new divergences are `textDocument/inlayHint` (22 `extra-rsvelte`, 16 `value-mismatch`, 2 `missing-rsvelte`), and that is #4464's class rather than this PR's.

**Evidence, graded.**

*Structural argument from code (not a measurement).* Upstream's `InlayHintProvider.ts:66-74` drops six classes of hint that exist only because svelte2tsx generated them. rsvelte ports none of them — searching every filter name across `crates/`:

```
  isInGeneratedCode / is_in_generated_code    9   <- all in rsvelte_check/svelte_check/mapper.rs
  isSvelte2tsxFunctionHints                   0
  isGeneratedVariableTypeHint                 0
  isGeneratedAsyncFunctionReturnType          0
  isGeneratedFunctionReturnType               0
  checkGeneratedFunctionHintWithSource        0
  startsWithIgnoredPosition                   0
```

The one non-zero is on svelte-check's diagnostic mapper, not the language server. `command grep -n -i inlay crates/rsvelte_language_server/src/server.rs` is two lines — the capability at `:247` and the routing at `:711`, where `textDocument/inlayHint` sits in the **generic** `forward_tsgo_request` list. So it is 0 of 6: every hint tsgo produces at a mappable position is returned. (Positive control: the same search shape finds 15 lines for `textDocument/hover`.)

*Measurement, from #4464, taken on `main` before this branch existed.* Against both servers on a ten-line component, rsvelte answers 7 hints where official answers 0, and the issue attributes each of the seven to a named upstream rule — including `": __sveltets_2_IsomorphicComponent<…>"` on the generated `$$_pmoC0C` declaration and its `type:` / `options:` parameter hints. That is the `extra-rsvelte` shape. It also records labels with their property names elided, which is the `value-mismatch` shape.

*Why this PR makes them reachable.* Upstream maps a component's tag name — `InlineComponent.ts:101-109` pushes `[nodeNameStart, nodeNameEnd]` into `__sveltets_2_ensureComponent(...)` — so the mapping this PR adds is faithful to upstream, and upstream compensates **downstream**, in the provider's filters. rsvelte had neither half. Adding the mapping without the filters is GATES.md 39b: repairing one divergence makes its children reachable. The carrier list agrees, with `upstream-features/component-invalid/input.svelte` the largest at 8 differential + 8 expected.

**So the ordering is #4464 first, then re-measure this.** Three independent reasons, none of which needs the attribution below:

1. One issue, one PR. Porting six filters is #4464's stated job.
2. Baselining the 40 here would need a justification, and the only available one is "deliberate" — which this repository forbids for a behaviour nothing implements. Pinning them would assert *we choose not to close this* where the truth is *we have not built this*.
3. If the filters absorb them, the residue is 16 `textDocument/diagnostic value-mismatch` — small, single-cause, and decidable on its own.

**What I have NOT measured, stated as such.** This branch carries two commits with two changesets, and I have not attributed the 56 between them. The ordering above does not need it: whichever commit produced them, if the filters absorb them the question dissolves, and if they do not, the residue gets attributed then — at which point the two commits are already separate and one arm each is cheap. Recording it so nobody reads the mechanism story above as having closed it. The 16 re-keyed/retired figures are not affected either way: 228 re-keyed and 58 retired are hash churn and genuine retirements respectively.

Converting to draft so this stops consuming CI on a measurement that has to be retaken after #4464 regardless.

baseballyama:
## Correction: the mechanism above is wrong for 16 of the 40

I attributed all 40 new `inlayHint` divergences to upstream's six generated-hint filters. Printing the keys instead of summarising them says otherwise:

```
differential:.../organize-imports-group.svelte|textDocument/inlayHint|/:value-mismatch[official=74234e98afe7,rsvelte=bcdc926e19c2]
differential:.../remove-imports-inline.svelte |textDocument/inlayHint|/:value-mismatch[official=74234e98afe7,rsvelte=bcdc926e19c2]
differential:.../importing.svelte             |textDocument/inlayHint|/:value-mismatch[official=74234e98afe7,rsvelte=c5be6f017f53]
```

`74234e98afe7` is `sha256(JSON.stringify(null)).slice(0,12)` — #4464 records it, and `[]` is `4f53cda18c2b`, which appears **0** times here. All 16 `inlayHint` value-mismatches carry it and no other key does (16 of 16, and the null-digest census over the whole 56 is exactly 16).

So official did not return an elided label. **Official returned `null`** — which `getInlayHints` does when `areInlayHintsEnabled(userPreferences)` is false. A filter reduces a list; it never turns one into `null`. The corrected split of the 40:

| n | shape | class |
|---:|---|---|
| 22 | `extra-rsvelte` | the six filters — as I said |
| 16 | `value-mismatch`, `official=null` | **not the filters** |
| 2 | `missing-rsvelte` | unattributed |

**What the 16 actually are is not yet settled, and two mechanisms predict the identical key.**

*(a) rsvelte ignores the preference.* `TsgoPreferences::default()` (`tsgo_client.rs:47-70`) hard-codes every inlay category to `enabled`, and that literal is the only occurrence of `inlayHints` in the whole language server — `includeInlayParameterNameHints` and `includeInlayVariableTypeHints` are **0** hits each. So rsvelte answers hints whatever the client asked for, while official honours the setting and returns `null`.

*(b) this PR made the request succeed.* #4464 §2 records that a range starting `0:0` makes rsvelte fail to map and answer the unmapped result, and `suites.mjs:120` sends `0:0`. If `0:0` did not map on these files before, rsvelte answered `null` and *accidentally agreed with official*; giving component names and specifier insertions real ranges could make `0:0` map, and rsvelte then answers a list where official still says `null`.

Both produce `official=null, rsvelte=<list>`. Adding inputs cannot separate them — it needs the inside. The discriminator is one probe on a `main`-built server: ask `organize-imports-group.svelte` for `inlayHint` at the gate's own range and see whether `main` answers `null` or a list. `main` answering `null` is (b); `main` answering a list means these keys existed before and my identity partition is wrong somewhere. That binary is building now.

**Nothing about the routing changes.** Both (a) and (b) are #4464's, both are on `main` today, and neither is baselineable here. What does change is the fix's shape: **the six filters alone would not absorb these 16**, so "land #4464 and re-measure" needs #4464 to cover the preference plumbing as well — which is why it is worth knowing before the code is written rather than after the gate stays red.

I am leaving the 2 `missing-rsvelte` explicitly unattributed rather than folding them into either bucket.

baseballyama:
## Measured: both my hypotheses were wrong, and the real account explains the retired set too

I built `main`'s language server (`sha256` prefix `f2849953e7f22a75740f3623`) and drove both servers over stdio against the real fixtures with the gate's **own** `initializationOptions` (copied from `verify.mjs:292`), asking `textDocument/inlayHint` over the gate's own `0:0 → endPosition` range.

```
### code-actions/organize-imports-group.svelte   (plain <script>)
  official inlay : null            rsvelte inlay : null
### call-hierarchy/outgoing-component.svelte     (lang="ts")
  official inlay : list len=3      rsvelte inlay : null
```

`documentSymbol` answered on all four legs, so a `null` here is a measurement and not a dead server, and `null` / `[]` / no-response are printed as three different strings rather than through a serializer that has a value for `null`.

**Calibration against cases the gate has already ruled on** — both directions:

| fixture | main's ratchet | my probe |
|---|---|---|
| `outgoing-component` | `value-mismatch[official=5f838b566f5b,**rsvelte=74234e98afe7**]` | official list, **rsvelte null** ✓ |
| `organize-imports-group` | **0 inlayHint keys** (the sides agree) | **null / null** ✓ |

So on `main` rsvelte answers **`null`** on these files — #4464 §2, the `0:0` range failing to map — and this PR makes the request map, so rsvelte starts answering a real list. That is neither of the mechanisms I proposed: (a) had the residual right and the transition wrong, (b) had the transition right and I could not tell which side moved.

### The account, and it predicts the retired set

Going `null → list` on 19 files does three things at once, and the numbers are exact:

| | carriers | keys | why |
|---|---:|---:|---|
| retires `value-mismatch[rsvelte=null]` | 11 | 22 | rsvelte now answers where it used to say `null` |
| retires `missing-rsvelte` | 2 | 4 | rsvelte now has hints it previously lacked |
| **gains** `extra-rsvelte` | 11 | 22 | `lang="ts"`: official had a list, rsvelte's is unfiltered → **#4464 §1** |
| **gains** `value-mismatch[official=null]` | 8 | 16 | plain `<script>`: official returns `null`, rsvelte answers anyway → **#4464 §3** |

The linking checks, including the cell that must be empty:

```
carriers that retired an rsvelte=null key : 11
carriers that gained an extra-rsvelte key : 11
intersection (same file, null -> list)    : 11
gained extra but retired no null          :  0
carriers that gained official=null keys   :  8
 ...of which retired an rsvelte=null key  :  0   <- must be 0: these AGREED before, so had nothing to retire
```

That last row is the discriminating one. The plain-`<script>` files had **no key at all** on `main` — `null` on both sides is agreement — which is why they appear as genuinely new rather than re-keyed, and it is why the `lang="ts"` / plain-`js` split is 11/11 and 8/8 with no counterexample.

The language split is the gate's configuration, not a compiler fact: `verify.mjs:292` gives **`typescript`** all six inlay categories and gives **`javascript`** none, so upstream's `areInlayHintsEnabled` is false for a plain `<script>` and it returns `null`. rsvelte's `TsgoPreferences::default()` hard-codes all six on and reads no client setting (`includeInlayParameterNameHints`: **0** hits in the server), so it cannot answer `null` for that reason.

### What this changes

**This PR is not introducing divergences. It is un-masking #4464 §1 and §3 by partially fixing §2** — #4464's own text says "the two defects hide each other", and this measures a third edge of that. Net inlayHint movement is **−26 +38 = +12** keys, every one with a named cause.

One thing I am *not* claiming: that §2's masking is universal. `main`'s ratchet has 130 `extra-rsvelte` inlayHint keys, so rsvelte does answer lists on plenty of files today; 58 carry `rsvelte=null`. The masking is file-specific, and my probe covers 2 files, calibrated.

**Consequence for #4464's shape, which is the actionable part:** §2 must not land alone. Fixing the clamp everywhere would un-mask §1 and §3 across the whole corpus at once, and this PR is a 19-file preview of what that looks like. The order that keeps each measurement meaningful is §1 + §3 first — both are no-ops while §2 masks them — and §2 last.

baseballyama:
## Parked deliberately, and the reason is not the failing checks

I assumed this was independent of #4488 because it edits no ratchet file. That is wrong, and the distinction is worth writing down: **#4488 changes a ratchet's contents, this changes a ratchet's measurement.** Giving a component's tag name its own source range moves the projection, the projection moves every LSP response derived from it, and the gate reports:

```
[lsp-verify] 284 NEW divergence(s)
[lsp-verify] 286 stale ratchet entry/entries
```

So this needs an LSP re-baseline of ~570 entries in the same file #4488 is rewriting. Two PRs re-baselining one ratchet from different bases is the FALSE-SHRINK shape, and it does not announce itself — each is self-consistent on its own branch. This waits for #4488 to merge, then re-baselines from the tree that results.

Read the 284 as the expected direction rather than as damage: this is `GATES.md` 39b, where fixing one divergence makes its neighbours comparable and they enrol. Whether all 284 are that, or some are regressions, is **UNMEASURED** and is the first thing the re-baseline has to split — printing both directions separately, not a net.

## The other failure is real and is not the ratchet

`Tests` is the rollup, and its shards read:

```
UNIT: success   RUNTIME: success   BULK: success   SAFETY: success
PROBE: success  CORPUS: success    SERVER: failure
```

One shard genuinely failed — this is not the cancelled-shard case where the rollup goes red over shards that never ran, and there is a real log (984 KB / 1.03 MB / 20 KB on the three failing jobs), so it is not a vanished runner either. `SERVER` has to be diagnosed on its own; it is a defect in this change, not a consequence of the ratchet move.

## Also on this branch: two changesets, one of which is not mine

`.changeset/` here carries both `component-name-source-range.md` and `svelte-specifier-hover.md`. The second belongs to a different change. That is a live instance of #4486 — `check-core-consumer-changesets` reads the whole directory, so another PR's changeset satisfies this PR's requirement — and it means the changeset gate currently proves nothing about this branch. Worth resolving here whichever way #4486 goes.

## Staleness, for the record

The newest check on this branch started `03:15:01Z`; `main` moved at `05:58:31Z` (#4450) and `06:00:00Z` (#4491). Every verdict above describes a superseded tree. The three failures are still worth acting on — a red does not become false by ageing — but the *absence* of other failures says nothing about today's `main`.

baseballyama:
> **Superseded in part — see the correction below.** The bucket counts here stand, but the
> closing claim that the 11 `lang="ts"` carriers are the work is wrong: they already diverged
> before this PR and only changed shape. The true appearances are the other 8.

## Triage of the `284 NEW / 286 stale`: 230 are re-keys, 68 retire, and 11 carriers are real work

The gate's log reports `284 NEW divergence(s)` and `286 stale`, which reads as a catastrophe and is
mostly churn. Bucketed by **identity** rather than by key, on run `34303105707`'s
`lsp-current-fixtures` artifact:

| bucket | keys |
|---|---|
| re-keyed (same divergence, new payload) | **230** |
| genuinely retired | **68** |
| genuinely new | **44** |

So this PR **retires more than it adds**, and the headline number is 5x the real movement.

### Method, because two things make the obvious reading wrong

**The log cannot answer this.** It prints 30 of 284 and 30 of 286 with no "and N more" — #4484,
whose fix is written and held. The first 30 are not a sample, so nothing extrapolates from them;
this is the artifact's full `current` list diffed against `compatibility/lsp-known-failures.json`
at this branch's own head.

**A key carries two content hashes, and folding only one of them over-reports.** The
`[count=..,hash=..]` payload is the obvious one. The other is the `@diagnostic-<hash>` **item id
inside the pointer** — so a change that moves a range re-identifies the item, and one moved
diagnostic is reported as a retirement plus a new divergence. Folding both takes the "genuinely
new" diagnostics from 16 to 4. Confirmed rather than assumed: the four newly-identified items on
`upstream-features/component-invalid` all carry the **same** official message (TS2345,
`typeof DoesntWork`) in `diagnosticDetails`.

Scope control: the filter keeps **2092 / 2092** artifact keys. Its first version kept 2082 —
`measuredIds` stores ids without the `differential:` / `aggregate:` prefix, and `initialize` is a
synthesized unit that is absent from `measuredIds` by construction. Before that was fixed the
reconstruction reported *every* key as new (2092 / 0), which is the key-space-mismatch signature
and not a result.

My reconstruction reads `286 NEW / 298 stale` against the gate's `284 / 286`. I cannot reproduce
the gate's scope rule exactly, so treat the buckets' **margins** as the claim, not their last digit.

### The part that is actual work: 40 keys over 19 carriers

The 44 genuinely-new keys are 4 diagnostic `message:value-mismatch` and **40 `inlayHint`
`extra-rsvelte` over 19 carriers**. That direction is expected from this change — declassifying the
`.tsx` specifier insertion as a generated region lets hints through positions that were previously
filtered.

**#4488 does not absorb them all.** It silences hints on components without `lang="ts"`; the split
is **8 plain-JS (absorbed) and 11 `lang="ts"` (not)**:

```
lang="ts", NOT absorbed by #4488
  call-hierarchy/outgoing-component            completions/component-events-completion
  code-actions/codeaction-custom-fix-all-component6   completions/component-props-completion-rune
  definition/definition-rune                   diagnostics/different-ts-service
  diagnostics/different-ts-service/different-ts-service    diagnostics/unresolvedimport2
  find-component-references-parent             find-component-references-parent2
  rename/rename2
```

So the next action on this PR is **not** a re-baseline. It is to attribute those 11 carriers —
either they are hints upstream also emits (in which case the divergence is elsewhere) or they are
generated-position hints that one of #4464's remaining unported filters should drop, in which case
they land with that work and not here.

The 68 retirements are `hover` 12, `diagnostic` 30, `inlayHint` 26 and are the change doing what it
says.

baseballyama:
## Correction: the 11 carriers I named as the work are refinements, and the real appearances are the other 8

My triage above split the new `inlayHint` keys by whether #4488 absorbs them and called the 11
`lang="ts"` carriers "actual work". That is backwards, and the check that found it is one I should
have run before posting: **do these carriers diverge on `inlayHint` before this PR?**

They do — all 11, in a local pre-#4483 arm and in the committed ratchet alike. So they were never
newly-diverging. What changed is the divergence's **shape**:

```
/:value-mismatch   ->  :extra-rsvelte     20 units
/:value-mismatch   ->  :missing-rsvelte    2 units
```

That is the whole result no longer differing: the two sides now align item-by-item and the residue
is a set difference. It is a **refinement, not a regression** — `GATES.md` 39b's adding direction,
where repairing one divergence makes its children comparable.

Re-bucketing by unit `(id, method, phase)` with the shape dropped, so a shape change cannot be
counted once as a retirement and once as an appearance:

| | units |
|---|---|
| shape refinement (present on both sides) | **22** — all `inlayHint` |
| true appearance (nothing retired there) | **20** |
| true disappearance | **32** |
| re-keyed (payload only) | 230 keys |

**So this PR retires 32 units, adds 20, and refines 22.**

And the 20 true appearances are the *opposite* set from the one I named:

* **16** are `inlayHint /:value-mismatch` on the **8 plain-JS carriers** — `organize-imports-group`,
  `remove-imports-inline`, `codelens/importing`, `emptytext-importer`, `find-file-references-parent`,
  `hover/hoverinfo`, `rename/rename4`, `update-imports/updateimports`. Every one is a component
  without `lang="ts"`, which is #4488's and #4494's population, not this PR's.
* **4** are `component-invalid` diagnostic `message:value-mismatch` (differential and expected).

### Why the first version was wrong

I partitioned the new keys by a property of the *carriers* (`lang="ts"`) and read the partition as
if it separated real from absorbed. It separates neither: it happens to correlate with which
carriers already diverged, because a plain-JS component emits no hints at all here today. The
discriminating question was never about the carrier — it is whether the same `(id, method, phase)`
also appears on the retired side, and that is one set intersection I did not compute.

The mechanism reading in the commit message stands and predicted the refinement correctly: a
request **range** spanning a `.tsx` specifier insertion used to be rejected outright, so `inlayHint`
returned nothing on any file importing a `.svelte` component. Nineteen of the moved carriers import
one.

**Revised next action.** Nothing here needs attributing to this PR. The 16 belong with #4488, and
the 4 `component-invalid` diagnostics are the only thing left that is arguably this change's own —
worth one look at whether the message differs or only the item's range moved, since
`diagnosticDetails` shows the four re-identified items on that file carrying the *same* official
TS2345 text.

baseballyama:
## The 4 `component-invalid` diagnostics are #4498, not this PR — triage is now closed

The correction above left one open item: whether those four `message:value-mismatch` keys are a real
message difference or only a moved item range. Answered from the artifact's `diagnosticDetails`.

It is a real message difference, and it is **pre-existing**. Upstream's `adjustIfNecessary`
(`plugins/typescript-go/features/DiagnosticsProvider.ts:946-981`) appends Svelte-specific guidance to
a TS2345 whose message mentions `ConstructorOfATypedSvelteComponent`; rsvelte's proxy ports none of
it — 0 occurrences in `crates/rsvelte_language_server/src`. On this carrier the two messages agree
for 385 characters and rsvelte's is a **strict prefix** of official's 547.

Filed as #4498 with the family enumerated from upstream's function rather than from this carrier (the
other two arms are unported too and have no carrier in this shard, so their reach is recorded as
UNMEASURED rather than zero).

Why it appeared in this PR's run without being this PR's defect: improving range fidelity let the
differ pair the items on that file, so their `message` fields were compared for the first time.
`GATES.md` 39b's adding direction again — the same mechanism as the 22 `inlayHint` shape refinements.

**Final accounting for this PR**, by unit `(id, method, phase)`:

| | units | attribution |
|---|---|---|
| true disappearance | 32 | this PR |
| shape refinement | 22 | this PR, `value-mismatch` -> item-comparable |
| true appearance | 16 | #4488 / #4494 (plain-JS carriers) |
| true appearance | 4 | **#4498** |
| re-keyed (payload only) | 230 keys | neither |

Nothing in the run is left unattributed, and nothing is attributed to a defect this PR introduces.
The remaining blocker is the ratchet re-baseline, which wants #4488 merged first so the plain-JS
appearances are not baselined in.

baseballyama:
## Correcting my own partition, and a merge-order constraint this PR creates

Two things, both re-derived from the artifact rather than recalled.

### 1. The partition I gave earlier does not sum, and is superseded

I previously split this PR's ratchet delta as 230 re-keyed / 32 disappearances / 22 shape
refinements / 16 + 4 appearances. Those numbers do not add up to the declared totals -- 230+32
is 262 against 286 stale, and 230+22+16+4 is 272 against 284 new -- and a partition that does
not sum is the recorded fingerprint of a figure carried in prose rather than computed. It was
carried in prose. Recomputed from the run's own key lists:

| | count |
|---|---|
| re-keyed (same key, different payload) | 228 |
| true appearances | 56 |
| true disappearances | 58 |

`228 + 56 = 284` and `228 + 58 = 286`, which is the check the old numbers failed.

One method note that is worth more than the correction. Three independent hash levels live in
one ratchet key -- the trailing `[count=,hash=]` payload, the `@diagnostic-<hash>` item id
*inside* the pointer, and the shape suffix -- and folding only the payload can over-report
movement. On this population it does not: folding the payload alone and folding all three give
the identical 228/56/58. That is a measurement about this delta, not a general licence.

The true appearances split by method as `inlayHint 40` (22 `:extra-rsvelte`, 16
`/:value-mismatch`, 2 `:missing-rsvelte`) and `diagnostic 16`, over 21 carrier files.

### 2. The 16 diagnostic appearances are #4498's defect, and they order the two PRs

All 16 sit on `upstream-features/component-invalid/input.svelte`, and they are not something
this PR breaks -- they are GATES.md 39b's adding direction, stated as clearly as I have seen it.

On `main` that file diverges as a **whole-list** mismatch, `/items:extra-rsvelte[count=13]` and
`/items:missing-rsvelte[count=5]`. While the lists diverge as wholes the gate never pairs
individual items, so **no message text is ever compared**. This PR narrows them to
`extra 9 / missing 1`; four items pair for the first time; and their messages are compared for
the first time, giving 4 items x opened/`phase=edit` x `differential`/`expected` = 16 keys. All
four carry the same `official=32c129f586d1, rsvelte=224effe9a18b`, so it is one systematic
difference seen four times.

Read out of this PR's own CI log, the split is exact:

```
4 official WITH-suffix
4 rsvelte  NO-suffix
```

and the suffix is the "Possible causes:" paragraph upstream's `adjustIfNecessary` appends
(#4498). The current ratchet holds 8 `component-invalid` keys and **0** of them are message
keys, which is the positive control that these really are new comparisons rather than re-keys.

**The constraint.** If this PR re-baselines with those 16 enrolled and #4498 merges afterwards,
the 16 go stale and the two-sided ratchet turns `main` red -- visible on neither PR, since each
merge ref holds only the other's absence. So #4498 lands first, or the two re-baseline together.
I am holding #4498 committed and unpushed rather than racing it.

### Status

This PR's three red checks are two: `Tests` is a rollup whose only non-success is
`LANGUAGE_SERVER: failure` (`BULK/FMT_CORPUS/OXC_FORMATTER_PROBE/RUNTIME/THREAD_SAFETY/UNIT` all
success), so it is not independent evidence. The real failures are `Language server` and
`LSP fixture parity current`, both the re-baseline requirement above.

The branch is also 7 commits behind `main`, and one of those -- #4450, `anchor an attribute
key's opening quote` -- moves svelte2tsx source ranges, the same key space this PR moves. The
re-baseline therefore has to happen on a rebased branch, or it enrols entries measured against
a tree that no longer exists. I am deliberately not rebasing yet: a force-push re-runs ~60 jobs,
and doing it once together with the re-baseline spends one CI cycle instead of two.

baseballyama:
## All 56 true appearances, attributed — and one of them is UNMEASURED rather than zero

Following on from the corrected partition (228 re-keyed / 56 appearances / 58 disappearances),
here is the whole appearance set with a cause each. It sums, which is the check the earlier
version failed.

| n | class | attribution |
|---|---|---|
| 16 | `diagnostic` on `component-invalid` | #4498 / #4499 |
| 24 | `inlayHint`, rsvelte `null` → array, 11 files | #4488 effect **UNMEASURED** |
| 16 | `inlayHint`, official `null`, 8 files | #4494 |
| **56** | | |

The two recurring digests are worth writing down because they carry the whole reading and they
are one line to reproduce: `digest(null) = 74234e98afe7` and `digest([]) = 4f53cda18c2b`. They
are not interchangeable — `InlayHintProvider.ts:42` returns `null` when the provider is
**disabled** and `[]` when it ran and found nothing.

**The 16 `inlayHint` with `official=null` are #4494, not a defect here.** All 16 carry
`official=74234e98afe7` — uniform, no exceptions — so official's provider was disabled on those
8 files. That is the gate's own configuration gap: `ls-config.ts:558-561` picks the preference
bucket by `scriptKind`, everything that is not TS/TSX lands in `javascript`, and `verify.mjs`
sends no `javascript.*`, so `areInlayHintsEnabled` is false. These 8 are plain-JS components.
Nothing in this PR causes them; they become visible because rsvelte now answers where it
previously returned `null`.

**The 24 with `rsvelte=null` before are the honest gap.** Their stale keys read
`official=[] (9), <array> (2)` against `rsvelte=null`, and after this PR rsvelte returns an
array, so the comparison deepens into `:extra-rsvelte[count=N]`. The tempting sentence is "#4488
will filter those", and I am not writing it, because it is not measurable from what I have.

I checked directly rather than assumed. All 11 files **are** inlayHint carriers in #4488's own
arm, in both the base and the fix arm (11/11) — so they are inside its measured population and
it clears none of them. But in that arm rsvelte returned `null` on those files, and a `null` has
no hints to filter, so the arm cannot say what #4488 does to hints that did not exist in it. The
arm and this PR are measurements of two different trees, and the question falls in the gap
between them. It is UNMEASURED, which is a different claim from zero, and the answer arrives free
once #4488 merges and this branch rebases.

For contrast, the 8 `official=null` files carry **0** inlayHint keys in either arm of #4488 —
so those really are outside its reach, and that one is measured rather than inferred.

### Method note: `comm` mispaired 26 keys, exactly as documented

Comparing two key sets, `comm -23` reported 26 keys present in one file and absent from the
other. With `Set`s the answer is **0 in both directions**. The ids contain `$`
(`upstream-features/$$events-usage/...`) and the two `sort`s disagreed on collating order, which
AGENTS.md records as `comm`/`join` silently mispairing rows whose inputs are not in one order.

The reduction returned a well-formed, plausible 26. What exposed it was printing the differing
keys instead of the count: the first one printed was byte-identical to a line visible in the
file it was supposedly missing from. A count cannot show that and a printed value cannot hide it.

baseballyama:
## Correcting the method note in my previous comment: it was not `comm`'s collating order

I attributed a 26-key phantom difference to `comm` mispairing rows whose inputs are not in one
collating order, citing the `$` in `upstream-features/$$events-usage/...`. That is a real
documented hazard and it is **not what happened here**. Retracting the mechanism, because a wrong
cause is worse than no cause — it sends the next reader to `LC_ALL=C`, which fixes nothing.

The actual cause was **duplicate lines in my own extraction**, and the control separates the two
cleanly:

```
raw extracted NEW lines            56
distinct after stripping a colon   30
lines ending in ':'                26
duplicates after stripping         26

comm -23 with deduped input                0
comm -23 with deduped input, LC_ALL=C      0     <- collating order changes nothing
```

`comm` computes a set difference only over **unique** sorted input. With each key present twice
on one side and once on the other, it reports the extra copies as unmatched — 26 of them. The
`LC_ALL=C` row is the control that kills the locale hypothesis: if collation were the cause, that
line would not read 0.

Where the duplicates came from is `verify.mjs:1105-1106`:

```js
for (const entry of added.slice(0, SHOW)) console.error(`  ${entry}`);   // 30 keys
for (const entry of added.slice(0, SHOW)) { ... }                        // detail dump, re-prints the key with a trailing ':'
```

so the log carries each shown key twice, once bare and once colon-suffixed, and my parser's
`.trim()` kept the colon and counted them as two distinct strings.

**Nothing downstream moves.** The partition (228 / 56 / 58), the attribution table, and the
`digest(null)` / `digest([])` reading were all computed with `Set`s in node, which is
duplicate-safe and locale-free, and they are unchanged. What changes is only the sentence
explaining why the earlier `comm` run disagreed with them.

Two things worth keeping. The rule I actually followed and that worked was printing the differing
keys rather than the count — the first one printed was byte-identical to a line visible in the
file it was "missing" from, which is what said the difference was not real. The rule I then broke
was the next one: I took the first documented hazard that fit the symptom and did not run the
control that separates it from the alternative. `LC_ALL=C` was one command away the whole time.

Incidentally this **confirms** #4484 rather than complicating it: the cap really is `SHOW = 30`
for both lists (`verify.mjs:61,1105,1118`), there is no "and N more" marker anywhere in the log
(grepped: 0 occurrences), and my apparent "56 NEW" was my parser rather than a second cap.

baseballyama:
## The rebase requirement, with a number instead of a procedure

I said earlier that this branch should be rebased before re-baselining because "a baseline is a
measurement of a tree". That is a rule; here is the measurement, and it came out of two
derivations of one quantity disagreeing.

Computing the post-merge ratchet size two ways:

```
arithmetic:  23870 + 284 - 286  =  23868
set union:   recomputed          =  23920
                                    ------
                                    gap 52
```

A gap is evidence about method before it is evidence about the world, so I chased it:

```
main ratchet                                   23870
this branch's base tree (16e21f3a5) ratchet    24126     (256 more)

of this PR's 286 STALE, present in MAIN         234
                        already absent           52   <- retired by intervening merges
of this PR's 284 NEW,   already in MAIN            0

23870 - 234 + 284 = 23920                      reconciles
```

So **52 of the 286 entries this PR reports as stale have already been retired on `main`**, by the
seven merges since its base — chiefly #4450, `anchor an attribute key's opening quote`, which
moves the same svelte2tsx source ranges this PR does. The 256-entry difference between the two
ratchets is the size of what has happened underneath this branch.

That converts the rebase from housekeeping into a correctness precondition. Re-baselining from a
dispatch run on the current branch would write a file measured against a tree that no longer
exists, and 52 of its removals are already-done work being re-reported. `--update-baseline`
deletes every baseline id it did not measure, so the direction of that error is deletion, not a
visible conflict.

The order is therefore fixed and each step's reason is different:

1. **#4499 merges** — otherwise the 16 newly-paired `component-invalid` message keys enrol here
   and go stale the moment it lands, reddening `main` where neither PR can see it.
2. **#4488 merges** — it also settles the UNMEASURED row in my attribution table, since its arm
   could not answer what it does to hints that returned `null` in it.
3. **Rebase this branch onto that `main`** — for the 52 above.
4. **Then** dispatch `Corpus Compat` against the rebased branch for the 17-artifact union
   (`lsp-current-fixtures` + `lsp-current-corpus-{0..15}`), because `verify.mjs:54-59` refuses
   `--update-baseline` and `mergeCurrentArtifacts` (`artifacts.mjs:102-147`) throws on anything
   short of exactly 17 at one revision.

One trap for step 4 that is easy to miss because it is in a different file:
`known-failures-md-check.mjs` gates the LSP ratchet's **declared count** and three partitions that
must sum to it (key kind, request phase, repository). A net change of -2 still moves the declared
count, so all four have to be recomputed **from the JSON** rather than by arithmetic on the
existing prose — which is the same mistake, one artifact over, as the partition I had to correct
above.

baseballyama:
**Retracting the merge-order reason I gave for #4499, and replacing it with a measured one for #4488.**

I wrote that #4499 must merge before this branch re-baselines because *"if #4483 re-baselines those 16 in and this PR merges after, the 16 go stale and the two-sided ratchet turns `main` red."* The 16 was recalled from an earlier triage of **attributable appearances**, which is a different population from *keys this branch adds*. Measured against `origin/main` (`71cebe04d`, unmoved):

```
message-valued keys on main   (pointer matches /message:)        6
  control: keys whose pointer names `items`                    878
of this branch's 388 additions, message-valued                   0
```

`diff.mjs:20-21` keys a diagnostic on `digest([code, source, range.start])` — **message is not in the identity** — so a message-text change cannot create or destroy `extra-rsvelte`/`missing-rsvelte` set membership. It can only move a `…/message:value-mismatch` key, and this branch adds none. **The mechanism I named has no carrier here.** (My first count of message keys was 78, which was `/message/` matching *file paths* like `message-actions.svelte`; anchoring the pattern to the pointer gives 6.)

The ordering constraint that *is* real is with **#4488**, and it is larger than the one I retracted:

```
main    23870
#4488   23768   = 23870 − 116 + 14
#4483   24136   = 23870 − 122 + 388
  of #4488's 116 retirements, still listed by this branch : 116  (all)
  of #4488's 14  additions,   absent from this branch     :  14  (all)
  overlap of the two retirement sets                      :   0
```

Both reconcile exactly and the retirement sets are disjoint, so this is not a shared-cause coincidence: merging this branch's ratchet as it stands after #4488 would re-list 116 entries that already pass and drop 14 that are needed — 130 keys, failing the two-sided check in **both** directions, on `main`, where neither PR's merge ref can see it.

So the order stands and the reason changes: **#4488 → rebase → re-baseline**, because the 17-artifact regeneration is what makes those 130 come out right. #4499 remains ahead of the re-baseline as ordinary hygiene rather than as a fix for a named failure.

One thing left unmeasured rather than assumed: all 6 message-valued keys are fixture-suite (`differential:`/`expected:upstream-*`), which runs on every PR, and #4499's run was green — but that run reports `1956/1972 compared, 16 skipped` and `no regressions … in measured scope`, and the official server projected only `365/423` components (58 fail, 13.7%). Whether those 6 were inside the measured scope is not readable from the log, so "#4499 retires 0 of the 6" is **not** established here. It does not affect the order.

baseballyama:
**Correcting the comment above: my retraction over-corrected, and the #4499 ordering constraint survives at a smaller size.**

I checked this branch's **additions** for message-valued keys, found 0, and concluded the mechanism had no carrier. Additions are not the whole population — what a branch **retains** from `main` is equally exposed, and I did not look:

```
the 6 message-valued keys on main, against this branch's ratchet
  retired   differential:upstream-features/checkjs-nostrict/component-props-ts   (x2)
  retired   expected:upstream-features/checkjs-nostrict/component-props-ts       (x2)
  RETAINED  differential:upstream-testfiles/completions/string-completion.svelte (x2)
                                                       retained 2  /  dropped 4
```

So if #4499 makes `string-completion.svelte`'s `@diagnostic-47abf651aebb/message:value-mismatch` pass, and this branch re-baselines on a tree without it, those **2** keys are listed-but-passing the moment #4499 merges — two-sided ratchet, red on `main`, invisible on both PRs. Not 16, and not 0.

**The order is unchanged; only its justification and its size are.** Re-baseline last, after #4499, because the 17-artifact regeneration on the final tree is what makes both the 130 (#4488) and these 2 come out right.

Whether #4499 actually moves those 2 remains unmeasured — its run was green, but reports `1956/1972 compared, 16 skipped` and `no regressions … in measured scope`, with the official server projecting only `365/423` components. The safe order costs nothing and does not depend on resolving it.

Recording the shape, because it is the second time in this thread that a correction went further than the evidence: the first number (16) was recalled from a different population, and the correction (0) was measured on a population I chose — additions — which was *also* not the one the question is about. A retraction is a claim and takes the same scrutiny as what it retracts; mine ran in the direction I already believed, which is the direction that does not prompt a re-check.

Status: **#4488 is merged** (`c7ecaf69e`) — 61/61, all 16 LSP shards SUCCESS, both LSP ratchet jobs SUCCESS. #4499 is rebased onto it (`eca708659` → `5deaa5c4f`) because every one of its checks had started before that merge and both PRs edit `server.rs`; the rebase was clean and its symbols and hook site survive intact. This branch still needs its rebase onto `c7ecaf69e` before the dispatch.

baseballyama:
**Third correction, and the previous two share its cause. This branch does not edit the ratchet at all.**

```
merge base 32ea33526   ratchet 24136
zz4483     (branch)    ratchet 24136      <- identical to its merge base
origin/main c7ecaf69e  ratchet 23768

files the branch changes (merge-base..branch):
  .changeset/component-name-source-range.md
  .changeset/svelte-specifier-hover.md
  crates/rsvelte_language_server/src/tsgo_overlay.rs
  crates/rsvelte_projection/src/svelte2tsx/template/nodes/inline_component.rs
  crates/rsvelte_projection/tests/svelte2tsx_component_name_map_4097.rs
```

Every number I published about this branch's ratchet was produced by `git show origin/main:<file>` against `git show zz4483:<file>` — **two trees compared directly, with the difference attributed to the branch**. It is main's movement: eleven commits have landed since `32ea33526`, several shrinking the ratchet. So:

- *"#4483 retires 122 and adds 388"* — no. The branch retires and adds nothing; that was main moving.
- *"116 of #4488's retirements are still listed by #4483, all 14 additions absent — a 130-key incompatibility, retirement sets disjoint"* — the observation is true and the framing is wrong. There is no conflict between two sets of edits. The branch simply carries a pre-#4488 ratchet because it has not been rebased.
- The earlier `23870` for main is also stale now: #4488 took it to **23768**.

**What survives is the operational rule, and it no longer needs any of that arithmetic**: rebase, then re-baseline last, on the final tree. That was the conclusion under all three wrong justifications, which is exactly why none of them got checked.

The pattern is one mistake three times: *comparing a branch against current `main` and reading the difference as the branch's doing.* `git diff --name-only <merge-base> <branch>` answers "what did this branch change" in one command, and I did not run it until now — after publishing a retraction and a correction built on the wrong reference. AGENTS.md records the two-dot form of this (`main` moving shows up in your own diff with the sign reversed); the form I hit is subtler because `git show <rev>:<path>` looks like it is reading *the branch's file* rather than one endpoint of an undeclared range.

Sequence, unchanged and now simply stated:

1. #4499 merges (it is rebased onto `c7ecaf69e` and green so far).
2. This branch rebases onto that main — which is what brings it from 24136 to 23768.
3. Dispatch `Corpus Compat` on the rebased branch for the 17-artifact union.
4. `merge-current.mjs --update-baseline`, then recompute the gated numbers in `known-failures-md-check.mjs` **from the JSON**, not from any figure in this thread.

baseballyama:
Rebased onto `62410db75` (post-#4499). No conflicts; `9017853fd` → `c9da24915`, two commits, same content.

**Why the rebase had to come first, and why the red before it was expected.** This branch was cut before #4499, and its own `Compiler parity`-era ratchet reading is a measurement of a tree that has moved. #4499's PR body pre-registers that *this* PR is what makes four `component-invalid` diagnostics pair for the first time, which is when the message comparison it fixes goes live — so re-baselining here before that merged would have enrolled sixteen keys that go stale on `main` the moment it did.

**What the pre-rebase run actually said, and what I could not read from it.** `LSP fixture parity current` reported `284 NEW / 286 stale`. The samples are same id, same method, same `count=`, different `hash=` — re-keys, which is the expected shape of a change that moves diagnostic *positions* without changing how many there are. I am **not** quoting a breakdown of the 284/286: the gate prints at most 30 of each and says so in its own output, so a table built from the log describes a display cap rather than a population. The full sets live in `ratchet.removed` / `ratchet.added` in the run's artifact.

One thing the 30 *do* support, because a sample can carry a presence claim even when it cannot carry a proportion: a whole `extra-rsvelte` diagnostic set retires on the fixture named `ignore-generated-code`, which is half (b) working.

**Next, in order.**

1. This push's CI establishes that the rebased tree builds and that nothing outside the ratchet regressed. `lsp-corpus` stays skipped here — the branch touches neither `compatibility/lsp-known-failures*.json` nor `scripts/compat-lsp/**`, so the `lsp-ratchet` hatch is closed and the sixteen real-world shards do not run.
2. `workflow_dispatch` of `Corpus Compat` on this branch for the complete 17-artifact union, re-checking `git merge-base --is-ancestor origin/main HEAD` immediately before rather than trusting this comment.
3. `merge-current.mjs --update-baseline` over that union. `verify.mjs:59` refuses a direct `--update-baseline`, and the refusal is the right one here: unlike #4464, where the edit was a deletion of forty keys named by a measurement, this PR moves keys in both directions and the artifact union is the only thing that can say which.

Both directions get reported separately when it lands, not a net.

Changesets check clean on the rebased tree: `@rsvelte/compiler`, `@rsvelte/svelte2tsx`, `@rsvelte/svelte-check` and `@rsvelte/language-server` are all named.

baseballyama:
## The `LSP fixture parity current` red is this PR's, and it is a re-key rather than a regression

Scope first: this is red **on this PR only**. `main` at `62410db75` is unaffected — this branch
changes behaviour and has not been re-baselined yet, which is the state that check reports.

### What the log shows, and what it does not

`LSP fixture parity current` reports `230 NEW` and `268 stale`, then prints **30 keys per
section** — a display cap with no marker in that code path. Building a shape table from those 60
lines gives 24/6/6, which is a *presence* result over 13% and 11% of the two populations and not a
proportion. The section headers state the real denominators, which is the only reason the cap is
visible at all.

### Measured over the full population

Reconstructed from the run's own `lsp-current-fixtures` artifact
(`projectRevision 9d98cd403b`, the merge ref) against the ratchet blob at that same revision, so
the arm and the ratchet come from one tree. Calibration: the reconstruction reproduces the gate's
own verdict exactly — `NEW=230, STALE=268` — before being asked anything the gate has not ruled on.

| bucket | n |
|---|---|
| re-keyed (same `file\|method\|pointer:shape`, different digest) | **206** |
| genuinely new | **24** |
| genuinely retired | **62** |

1976 in-suite baseline keys → 1938 current, net **−38**.

So 90% of the movement is one mechanism restating itself: same key, same `count=`, different
`hash=`, i.e. the diagnostic *text* changed because the component name did. That is what this PR
does. The 30-key sample would have put the retired bucket at 6 where it is 62 — the direction was
right and the magnitude was not.

### The 24 genuinely-new keys are all one method

Every one is `textDocument/inlayHint` (22 `extra-rsvelte`, 2 `missing-rsvelte`), on
`call-hierarchy/outgoing-component`, `codeaction-custom-fix-all-component6`,
`component-events-completion`, `component-props-completion-rune`, `definition-rune`,
`different-ts-service` (×2 paths), `unresolvedimport2`, `find-component-references-parent{,2}`
and `rename/rename2`. Four of them carry `count=1` with the identical digest `6712cb870407`.

That is #4505's class, which has a consequence for ordering rather than for correctness here:
re-baselining this branch **before** #4505 merges would enrol inlay-hint keys that #4505 then
moves again. Whether #4505 retires any of these 24 or merely re-keys them is **UNMEASURED** — a
`count=1` key would retire if that one hint is the `$$render` return type and a `count=7` key
would only re-key, and I have not run the two fixes together.

### Consequence

The re-baseline stays sequenced after #4505, and it needs the complete 17-artifact union from a
`workflow_dispatch` of `Corpus Compat` on this branch — the fixture artifact above covers 3 of the
17 and cannot shrink the ratchet. I will re-check `git merge-base --is-ancestor origin/main HEAD`
immediately before that dispatch rather than trusting this comment, and report both directions
separately when it lands rather than a net.

baseballyama:
Addendum: `Language server` has since gone red too, and it is the **same** cause, not a second one. Its failing step is #26 "Run LSP differential fixture and upstream gates" — the same `verify.mjs` invocation — reporting the identical `230 NEW` / `268 stale`. Every Rust test in that job passes (`7 passed; 0 failed` twice). Two red check names, one artifact; the count above stands unchanged.

baseballyama:
Second addendum, and this closes the set: `Tests` has gone red as well, and it is the same cause a third time. It is a rollup — its only failing step is "Verify all test jobs passed", whose output is `BULK: success`, `UNIT: success`, then `Test job(s) FAILED: language-server.` All three `Test (ubuntu-latest, N)` shards are `completed/success`, and the `language-server` job it reads fails at step 26, the same `verify.mjs` reporting the same `230 NEW` / `268 stale`.

So: three red check names, one artifact, zero additional problems. Per this repo's own rule about reading a rollup — the check name and the conclusion together are not enough — the step env is what says so, and it says the shards ran and passed.

baseballyama:
**The red here is a re-key, not a regression — and I had been calling it "expected" without having measured it. Correcting that.**

`LSP fixture parity current` reports **230 NEW divergence(s)** and **268 stale ratchet entries**. Two numbers that size a regression, so they deserve the classification rather than a label.

**Discriminator.** The identical job on #4505 — same `main`, no change from this branch — is `COMPLETED/SUCCESS`. So the movement is this PR's, not the environment's and not a `main` breakage.

**What the movement is.** Both lists are display-capped at 30, so this is a sample and I am reporting it as one. Stripping only the `hash=` digest from each key and intersecting the two printed sets:

```
printed NEW keys that also appear in printed STALE : 22 / 30
printed NEW with no stale twin                     :  8
printed STALE with no new twin                     :  8
```

22 of 30 are the *same unit, method, phase, class and count* under a different digest — e.g. `$$events-usage|textDocument/diagnostic|/items:extra-rsvelte[count=6]` leaves with `hash=83797e47dcfe` and returns with `hash=9c5a93d6b4e5`. That is exactly what giving a component's tag name its own source range should do: the divergence is unchanged in shape and its *positions* moved, so the value digest moves and the key re-keys.

The 8 that do not pair are the interesting half, and they run the right way:

- `component-invalid`: `extra-rsvelte[count=12]` **and** `missing-rsvelte[count=4]` retire, `extra-rsvelte[count=8]` arrives — 12 extra + 4 missing → 8 extra, i.e. the four previously-missing items are now produced and four spurious ones are gone.
- `fixtures/html-smoke-hover-uppercase|textDocument/hover|0:2|/:value-mismatch` retires outright, both phases.
- `sveltekit-autotypings` / `sveltekit-autotypings-arrow` arrive as `extra-rsvelte[count=1]`, newly comparable. These are the `data` prop type (`{ data: any }` here against `{ data: { exists: boolean } }` official) — a kit-autotyping difference that has nothing to do with tag ranges and was previously masked by a coarser key. GATES 39b: fixing a whole-list mismatch makes its children comparable, so a NEW key there is expected rather than a regression.

**Sequencing, also now measured.** I had said this PR waits on #4505 "because it moves the ratchet". That is true and I had not checked it: #4505 is `38+ / 78-` on `compatibility/lsp-known-failures.json` and `48+ / 168-` on `lsp-mechanisms.json`. Two PRs re-baselining the same two files is the ordering hazard, so this one rebases after #4505 lands and re-baselines on top of its result — not before.

**What is still unmeasured**: the other 200 NEW and 238 stale entries. The 30-entry cap means the pairing rate above cannot be extrapolated; the full sets come from a `workflow_dispatch` run on the rebased branch, and the re-baseline will report both directions separately over the whole population rather than a sample.

baseballyama:
The three red checks are **one** failure seen three times — `Tests` is an honest rollup whose
only non-success is `LANGUAGE_SERVER: failure`, and `LSP fixture parity current` and
`Language server` are the same gate. So there is one thing to explain: `230 NEW / 268 stale`.

I reconstructed the gate from its own artifact rather than reading the log, which caps its
listing at 30. The reconstruction had to be checked before it could be read: my first attempt
scoped `added` against the *restricted* ratchet and got `534 / 232`. `verify.mjs:1090-1091`
computes `added` against the **full** ratchet and `removed` against `selectKnownForScope`, and
with the gate's own rule it reproduces CI exactly:

```
RECONSTRUCTION  added=230  removed=268
CI REPORTED     added=230  removed=268
```

### What the 230 actually are

```
identity                          added  removed  in both   new-only  retired-only
unit|method|pointer|kind            230      268      206         24            62
```

**206 of the 230 are re-keys** — same unit, same method, same pointer, same kind, different
`count=`/`hash=`. Under a set difference a re-key is indistinguishable from a removal plus an
addition, which is the whole reason the raw pair reads as alarming.

### The 24 that are not re-keys are the gate becoming *able* to compare

All 24 are `textDocument/inlayHint`, 12 units x 2 phases. Taking the loudest one, the ratchet
and the current run side by side:

```
before:  …|textDocument/inlayHint|/:value-mismatch[official=103bb86e95dc, rsvelte=74234e98afe7]
after:   …|textDocument/inlayHint|:extra-rsvelte[count=16, hash=135f8cb33c21]
         …|textDocument/inlayHint|:missing-rsvelte[count=3,  hash=d164043c3503]
```

`74234e98afe7` is not an opaque hash — re-implementing `diff.mjs:3-4`'s digest independently,
`digest(null) = 74234e98afe7`. So **rsvelte was answering `null`** for that file's inlay hints,
which the gate can only record as a whole-response `value-mismatch` at the root pointer. After
this PR rsvelte answers a real array, and the comparison becomes per-hint — which *creates*
`extra-` and `missing-` keys where there was one coarse key before.

That is [`GATES.md` 39b](compatibility/GATES.md#gate-coverage)'s adding direction: fixing a
divergence makes its children reachable, and the new keys are expected rather than a
regression. The same shape shows in `documentSymbol` on that unit, where `missing-rsvelte`
goes 7 → 6 while `extra-rsvelte` goes 11 → 13.

### What this does not yet establish

The artifact I read was measured at merge ref `9d98cd403b`, and `main` has moved four times
since. So this says the 230 are not regressions **in that tree**; it does not license
re-baselining in the current one. Next step is a rebase onto `main` and a fresh full run, and
only then `--update-baseline` — the ratchet is shrink-only and `--update-baseline` enrols
whatever it is shown, so the check has to be made against the tree the baseline will describe.

I am holding that run until #4511 settles rather than competing for the job ceiling.

baseballyama:
Rebased onto `main` at `47c5bc5f5`. The conflict was in `tsgo_overlay.rs` and it is a union, not a
choice: `main` has since gained `render_return_type` on the same struct, so both fields, both
initializers in all three constructors, and both methods stay. The only thing dropped was a
duplicated first line of `is_generated_range`'s doc comment.

Verified on the rebased tree, not on the pre-rebase one: `cargo fmt --all -- --check`, then
`cargo clippy --workspace --all-targets --all-features -- -D warnings`, then
`cargo test --release -p rsvelte_language_server -p rsvelte_projection` — **109 suites, 0 failed**.

The first run of that test command reported 5 failures in `upstream_fixture_manifest`, all
spelling `fixture root is missing: …/submodules/language-tools/…`. That was the worktree, not the
change: a linked worktree initializes no submodules, so the fixture roots did not exist. After
`git submodule update --init submodules/language-tools` the same command is green. Recording it
because an absent submodule shows up only as a smaller denominator, and here it showed up as a
red that looks like a real one.

baseballyama:
## レビュー(コードオーナー視点)

**総合判定: close 推奨** — #4611(+ #4609)の分割ペアを採用し、本 PR は閉じるのが妥当と考えます。

### 重複判定
本 PR は「#4609(LSP overlay) + #4611(タグ名マップ) + `<svelte:component this={}>` 対応」の合体版で、
2 コミットが #4609 / #4611 と 1 対 1 対応します。テストファイル名
`crates/rsvelte_projection/tests/svelte2tsx_component_name_map_4097.rs` まで同一で、両方はマージできません。

本 PR が優れている点もあります(テストが 5 セルで否定コントロールと `span_of` の一意性 assert を持つ、
changeset が 4 パッケージすべてを明示、LSP ratchet 再ベースライン手順を本文に詳述)。
それでも close を推奨するのは、**本 PR だけが持つ `<svelte:component>` 対応に実測回帰があるため**です。
テストと本文の資産は #4611 に移植してください。

---

### 【重大】`<svelte:component>` で `this=` より前の属性がマップを失う

使い捨て worktree で `inline_component.rs` 1 ファイルだけを差し替えた 3 アーム比較。
オラクルは corpus ゲート自身が使う `submodules/language-tools/packages/svelte2tsx/index.js`
(`scripts/compat-corpus/svelte2tsx-compile.mjs:39` の `OFFICIAL`、svelte 5.57.0)。
生 `mappings` を VLQ デコードし、テンプレート行のマップ済みソース列を比較:

**入力: `<svelte:component title={hue} this={Comp} />`**

| アーム | マップされたソース列 |
|---|---|
| official | `0, 18..29, 36..40, 42, 44` |
| main | `0, 18,19,20,21,22,23, 25,26,27,28, 44` |
| **#4483** | `0, ` **`36,37,38,39,40`** `, 44` |

`title={hue}`(列 18-28)が全消失。**main も official も持つ 11 列**を失います。
`this={Comp}` を先に書いた `<svelte:component this={Comp} title={hue} />` では消失しません(列 24-28 を獲得)ので、
**属性順に依存した実欠陥**です。

なお `<Comp title={hue} />`(#4097 本来の対象)では #4611 と同一の正しい出力になります。回帰は
`<svelte:component>` 経路に限定されています。

#### 機構
1. `crates/rsvelte_projection/src/svelte2tsx/template/nodes/inline_component.rs:751-757` が
   式の `Seg::Src` を `opener.extend(attr_segs)` より**前**に push する
2. `this` はパース時に attributes から抜かれる —
   `crates/rsvelte_core/src/compiler/phases/1_parse/state/element.rs:744-748`
   (`extract_this_attribute` / `remove_first_this_attribute`)。
   よって `this=` より前にある属性の `Seg::Src` は、**後ろの位置に、より小さいオフセットで**並ぶ
3. `crates/rsvelte_projection/src/svelte2tsx/template/segs.rs:101-118` の `bake_out_of_order_src` が
   `Seg::Src(s, e) if s >= last_end` を満たさない要素をリテラルへ焼き込む → 先行属性のマッピング喪失

#### 上流との差
上流は焼き込まず `str.move()` で並べ替えます(`node-utils.ts:79-102` の `moves` / `sortedMoves`)。
rsvelte にはその移植が無く、`bake_out_of_order_src` は「諦める」代替実装です。
したがって `<svelte:component>` 対応を通すには、先に
**(a) `bake_out_of_order_src` の制約を解く(= `str.move()` 相当を移植する)**か、
**(b) 「`this=` より前に属性がある場合は `Src` を諦める」条件を明示的に入れる**かのどちらかが前提になります。
これは #4097 とは独立した設計課題なので、別 PR に切り出すのが適切です。

---

### 【要記録】列 5 の余剰セグメント(#4611 と共通)
両 PR とも official には無い**列 5(タグ名直後の空白 1 文字)**にセグメントを出します。
上流は `node-utils.ts:56-71` の `transform()` で保持範囲を 1 文字伸ばし、次のリテラルでその 1 文字を上書きするガードを持ちます。
rsvelte 側の対応装置は `Seg::LitOpen`(`segs.rs:31-35`)ですが、**どちらの PR も未移植**です。
実害は小さいと見積もっていますが LSP 応答レベルでは未実測なので、別 issue として記録してください(詳細は #4611 のコメント)。

---

### 提案する進め方
1. 本 PR のテスト 5 セル(`<svelte:component>` セルを除く)と本文の LSP ratchet 再ベースライン手順を #4611 に移植
2. #4609 → #4611 の順で landing(両者は inlayHint の NEW キーが重複するため、後続は rebase 後に再測定)
3. `<svelte:component>` のタグ名マップは、`bake_out_of_order_src` / `str.move()` の課題として別 issue + 別 PR
4. 列 5 の `LitOpen` ガード未移植を別 issue で記録
5. 本 PR を close

baseballyama:
Superseded by #4609 and #4611, which split this PR's two mechanisms into one PR each:

- **(a) the component tag name had no source range** → #4611 (`s2tx/4097-component-name-map`)
- **(b) the `.tsx` specifier insertion was classified as generated code** → #4609 (`lsp/4097-hover-svelte-specifier`)

Together they close #4097; separately neither does, which is why neither links it. Keeping three
branches for two mechanisms costs a full LSP shard set per push against a 20-job ceiling, and the
split pair is what the #4649 / #4650 review was written against.

Closing this one. Nothing here is lost — reopen if a third mechanism turns up that neither of the
two carries.

Issue #4484, opened 2026-09-09
lsp: `verify.mjs` truncates its NEW and stale lists at 30 with no "and N more", so a log-sourced re-baseline is silently partial


baseballyama:
`verify.mjs` prints its verdict lists through `slice(0, SHOW)` with `SHOW = 30` (`scripts/compat-lsp/verify.mjs:61`) and **no "… and N more" line** in either block:

```js
if (added.length) {
  console.error(`\n[lsp-verify] ${added.length} NEW divergence(s):`);
  for (const entry of added.slice(0, SHOW)) console.error(`  ${entry}`);   // :1105
  ...
}
if (removed.length) {
  console.error(`\n[lsp-verify] ${removed.length} stale ratchet entry/entries:`);
  for (const entry of removed.slice(0, SHOW)) console.error(`  ${entry}`); // :1118
}
```

`--show` appears **0 times** in `.github/workflows/` (positive control: `compat-lsp` appears on 8 lines there, so the grep reaches those files), so 30 is what CI runs.

## Why this is not just a display choice

This is the recorded "a list truncated for DISPLAY becomes a population when something downstream reads it" shape. The log is the obvious source for two operations:

- **re-baselining a stale ratchet** — take the printed stale entries and remove them. Above 30 that silently removes a subset, and the *next* run goes red on exactly the entries the cap hid, which reads as a fresh regression rather than as an incomplete edit.
- **diagnosing a regression** — a run with more than 30 NEW divergences prints 30, and the 31st is indistinguishable from absent.

The header line carries the true count, so the output is self-*contradicting* rather than silent — but only for a reader who compares the header against the number of lines they collected, which is not what "copy the list" looks like.

## Measured

On run [34298166242](https://github.com/baseballyama/rsvelte/actions/runs/34298166242) (7 of 16 corpus shards reported at time of writing), the per-shard stale counts are `8, 10, 16, 16, 18, 20, 22`. Every one is under the cap, so **the hazard did not bite on this run** — that is the population, not the design.

## The uncapped source, and its one gap

`--write-current` uploads `lsp-current-corpus-<shard>.json` with `if: always()`, so the artifact survives the failing verify. It holds `current` but **not** `removed` — `report.ratchet = { current, added, removed }` (`:1093`) goes to the report file, which is not uploaded. So a consumer has to recompute:

```js
const selected = selectKnownForScope(known, a.suites, a.repos, a.shard);
[...]

PR #4488, opened 2026-09-09
fix(language-server): an inlay hint the client did not ask for is not a hint (#4464)

baseballyama:
Closes part of #4464 — its §3, which the issue did not have when it was filed.

## The defect

`getInlayHints` returns `null` before asking TypeScript for anything when `areInlayHintsEnabled(userPreferences)` is false (`InlayHintProvider.ts:38-43`), and `ls-config.ts:515-522` maps each of the six categories from the client's config with **no fallback** — so a client that sends no `inlayHints` has all six undefined and gets `null`.

rsvelte read none of them. The only occurrence of `inlayHints` in the language server was a literal in `TsgoPreferences::default()` that turned all six on, so hints were produced whatever the client asked for and `null` was not a reachable answer:

```
includeInlayParameterNameHints   0 hits
includeInlayVariableTypeHints    0 hits
inlayHints                       1 hit   <- the hard-coded default
```

The config is picked by the **document's own script kind**, not the shadow's extension: `getUserPreferences` takes `'typescript'` for `ScriptKind.TS`/`TSX` and `'javascript'` otherwise (`LSAndTSDocResolver.ts:339-342`). rsvelte could not have got this right from the shadow, because every shadow is a `.svelte.tsx` and reads as TypeScript. `is_typescript_component` already answers it the way upstream's `DocumentSnapshot` does, and `component_reference_code_lens_enabled` already makes the sibling decision, so this follows that shape.

## Measurement

Both servers driven over stdio with the LSP gate's own `initializationOptions` (`verify.mjs:292`), which gives `typescript` all six categories and `javascript` none. Arms are distinct binaries (`f2849953e7f22a75740f3623` before, `9a821dda92b6aa022ccd7815` after).

| component | official | before | after |
|---|---|---|---|
| `codeaction-checkJs.svelte` (plain `<script>`) | `null` | list | **`null`** |
| `organize-imports-error.svelte` (plain) | `null` | list | **`null`** |
| `another-ref-format-date.svelte` (`lang="ts"`) | list 1 | list | **list 2** |
| `codeaction-const-reassign.svelte` (`lang="ts"`) | list 2 | list | **list 5** |

The `lang="ts"` rows are the load-bearing half — a rule reading the six categories off rsvelte's *merged* values, or keying on the shadow's extension, disables hints there too and still turns the plain rows green. That those rows stay lists **larger than official's** is also correct: that is §1, the six missing generated-hint filters, which this PR does not touch.

`null` / `[]` / no-response are printed as three different strings by the probe rather than through a serializer that has a value for `null`, and `documentSymbol` answers on every leg, so a `null` here is a measurement and not a dead server.

### The instrument was wrong first, and its failure is why this has a control

The first run of this probe reported the `lang="ts"` rows as `null` too, which reads as "the fix is over-broad". It was the probe: it answered `workspace/configuration` with the whole configuration object for every item, where the gate resolves each item's dot-path `section` (`verify.mjs:333`). Every section landed in the wrong shape, so hints were disabled for the *right answer for the wrong reason* on the plain rows and the wrong answer on the TS rows. Corrected, the four cells separate two-sided. Recording it because a symmetric filter failure reads as agreement, and the plain-`js` half looked correct throughout.

## Why the defaults are removed rather than kept

Upstream forwards exactly what the client asked for. Keeping rsvelte's defaults alongside the new check would mean a client that enables one category gets six — a second divergence in the shape of a fix.

## Ratchet — pre-registered

I expect this to **retire** inlayHint entries and add none. `main`'s ratchet carries 98 `official=74234e98afe7` (the digest of `null`) inlayHint keys over 49 carriers, and **0 of those 49 are `lang="ts"`** (35 plain-`js`, 6 with no `<script>`, 7 in other suites) — which is the shape the mechanism predicts, since a TS component's hints stay enabled. The re-baseline was taken from CI's artifact, not the job log, because both the NEW and stale lists print through `slice(0, SHOW)` with `SHOW = 30` and no "and N more" line (#4484) — and the stale list was 116, over the cap.

This also corrects something I wrote on #4464: I said §1 and §3 are "no-ops while §2 masks them, so they can land without moving the ratchet at all". That is too strong for §3 — the masking is file-specific (`main` has 130 `extra-rsvelte` inlayHint keys, so plenty of files do get lists today), and on those files §3 is not masked and should retire the entry.

## Tests

`inlay_hint_preference_tests`, four cases, expectations taken from the oracle rather than transcribed. Pinned: an absent category is off; the gate's own two sections; `parameterNames` is a string enum (`'literals'`/`'all'` on, `'none'` off, `true` off) while the other five are booleans; any single category is enough; a later layer overrides an earlier one.

Ablation control: forcing the predicate to `true` reddens 3 of the 4, and restoring leaves the tree byte-identical. The 4th asserts only positives and stays green under an always-on bug — which is why the other three exist.

Denominator: `cargo test -p rsvelte_language_server`, all 9 targets, rc=0. The first run failed 5 cases in `upstream_fixture_manifest` with `fixture root is missing: .../submodules/language-tools/...` — this worktree had no submodules, since `git worktree add` initializes none and an uninitialized submodule answers `rev-parse HEAD` with the **superproject's** HEAD rather than erroring. After `--init --depth 1 --force`, all 9 pass.

---

## Re-baseline: the prediction above was half right, and the half it missed is the interesting one

Taken from run `34310065086`, artifact `lsp-current-fixtures`, `projectRevision` `6b9c092b4c88`
(this PR's merge ref). Its `compatibility/lsp-known-failures.json` blob `36ae901df` is identical at
the merge ref, at this PR's base and at `origin/main`, so `main` moving does not bear on the
measurement.

**Both directions, separately:**

| | `differential:` | `expected:` | total |
|---|---|---|---|
| retired | 98 | 18 | **116** |
| added | 0 | 14 | **14** |

The prediction said *98 retired, none added*. The 98 landed exactly — same count, same class, and
`0 of 49 carriers are lang="ts"` held. The "none added" was wrong, and it was wrong because the
prediction reasoned about **one oracle**. This ratchet carries two for the same request:
`differential:` compares rsvelte to the live official server, and `expected:` compares it to
upstream's committed snapshot (`verify.mjs:743-756`). I varied the first and held the second fixed
without noticing it was there.

**The 14 additions are not new carriers.** Measured over the added and removed sets keyed on
`(kind, id, method, phase)`: **0 carriers gained a divergence they did not already have**, and 98
were cleared outright (negative control: a fabricated carrier is reported as gained). The seven
carriers' `expected:` divergence changed *shape*, from `:extra-rsvelte` / `:missing-rsvelte` to
`/:value-mismatch`, which the key treats as new because the class is in the key by design.

**Their cause is the gate's own configuration, filed as #4494.** `verify.mjs:292-311` enables every
`typescript.inlayHints.*` preference and sends no `javascript.inlayHints`; upstream selects the
preference set by script kind (`LSAndTSDocResolver.ts:339-343`), so a component with no `lang="ts"`
reads the absent `javascript` set, `areInlayHintsEnabled` is false, and the official server returns
`null` before computing a hint. After this PR rsvelte answers `null` there too — which is why all
nine non-TS fixtures now hold **zero** `differential:` keys while all nine TS fixtures still hold
live ones (positive control: 170 `differential:` inlayHint keys remain in the run).

Independent confirmation, from a number I did not use to build the hypothesis: the run logs
`typescript-inlay-hints: 14/18 upstream snapshots reproduced`. `calibrationView`
(`normalize.mjs:81`) rewrites official's `null` to `[]` when the snapshot is an array; five of the
nine non-TS snapshots are literally `[]` and four hold a real list. 9 TS + 5 = 14, and the
unreproduced 4 are exactly `action`, `animation`, `reactive-block`, `snippet.v5`.

**One note for whoever re-baselines this ratchet next.** I wrote both files in their on-disk
convention — tab indent, code-unit key order. `merge-current.mjs:49-50` sorts the sidecar with
`localeCompare`, which disagrees with what is committed (it orders `Navbar` before `NavBrand`; the
file has the reverse), so the next `--update-baseline` will reorder ~20k sidecar lines for reasons
unrelated to whatever that PR changes. I did not fix it here because a 20k-line reorder around a
-102 change is unreadable.


baseballyama:
## Merge ordering against #4450, measured rather than assumed

#4450 also edits `compatibility/lsp-known-failures.json`, so whichever of the two merges second
holds a verdict measured on a tree without the other — the stale-green hazard, which does not
announce itself. Measured against #4450's own merge base (`16e21f3a5`):

| | #4450 | #4488 |
|---|---|---|
| ratchet keys removed | 376 | 116 |
| ratchet keys added | 120 | 14 |
| removed by **both** (double-delete) | \-\- | **0** |
| keys #4488 adds that #4450 removes | \-\- | **0** |
| removals/additions touching `inlayHint` | **0 / 0** | 116 / 14 |

#4450's keys are `definition` (240), `diagnostic` (104+88), `documentSymbol` (30) and `completion`
(2); #4488's are `inlayHint` only. Positive control on the predicate: it finds 116 of 116 in my own
removal set, so the two zeros are a measurement and not a dead filter.

So the two are disjoint by key **and** by method, and no three-way merge can produce a ratchet
either PR would disagree with. The stronger point is that #4450's own run is evidence about the
interaction I would otherwise have to guess at: if its projection change moved inlay-hint positions,
its ratchet diff would carry `inlayHint` churn, and it carries none. That holds after #4488 too,
since #4488 only *removes* hints on components without `lang="ts"` and leaves the TS components —
where the remaining live `differential:` inlayHint keys are — untouched.

Conclusion: this PR's verdict does not need re-running if #4450 lands first. I would not have
claimed that from "the files merge cleanly", which is the test that does not answer the question.

baseballyama:
## Merge note: #4450 and this PR **do** conflict, and that is the safe outcome

Correcting the comment above in one respect. The two are disjoint by ratchet *key* and by *method*,
which is what makes a combined ratchet correct — but they are not disjoint by *line*. Trial-merged
with `git merge-tree` at their shared base `16e21f3a5` (git 2.33 here, so not the `--write-tree`
form; `128` from that flag is "unknown rev", not a conflict):

```
7 conflicts:  KNOWN-FAILURES.md 3,  lsp-known-failures.json 3,  lsp-mechanisms.json 1
```

The `KNOWN-FAILURES.md` ones are both PRs editing the **same gated numbers** — `contains 24024` vs
`contains 23870`, and the two partition lines. That is the good case: the numbers are gated, both
sides moved them, so git stops instead of silently keeping one side's stale figure.

Resolution is arithmetic, not judgement. The JSONs take the **union of both deletions and both
additions**, which is correct precisely because no key is removed by both (0 double-deletes) and no
key one adds is one the other removes. Computed from the union:

| | entries |
|---|---|
| merge-base `16e21f3a5` | 24126 |
| #4488 alone (−116 +14) | 24024 |
| #4450 alone (−376 +120) | 23870 |
| **both merged** | **23768** |

The three gated numbers in `KNOWN-FAILURES.md` after the merge — to be recomputed from the merged
JSON rather than taken from here, but stated so a wrong resolution is visible:

```
contains 23768 entries
by key kind:      `21792 + 1646 + 330`
by request phase: `11889 + 11879`
```

The phase addends differ by **10**, which is the invariant the section's own prose asserts (the
session-level `initialize` keys, which run once per session rather than once per phase). A botched
union would not preserve it, so that difference is the cheap check on the resolution.

Issue #4489, opened 2026-09-09
A trailing instance-script comment is dropped when a top-level {#snippet} precedes the script in output order


baseballyama:
## Symptom

A comment at the end of the instance script is dropped when the template holds a top-level
`{#snippet}`. Official keeps it, printed between `const` and the declarator name.

```svelte
<script>
	let c = 1;
	let p = Promise.resolve(1);
	// x
</script>
{#snippet s()}<b>{c}</b>{/snippet}
```

```js
// official (submodules/svelte/packages/svelte/src/compiler/index.js)
export default function C($$anchor) {
	const // x
	s = ($$anchor) => { … };

	let c = 1;
	let p = Promise.resolve(1);
}

// rsvelte — the comment is gone
export default function C($$anchor) {
	const s = ($$anchor) => { … };

	let c = 1;
	let p = Promise.resolve(1);
}
```

Reproduces on `client` and `client-dev`, with and without a `{@render}` — 4 of the 32 cells of
the #4480 grid.

## This is NOT a missing `loc`, and that was measured rather than assumed

Upstream reuses the **source AST** identifier verbatim (`SnippetBlock.js:82`,
`b.const(node.expression, snippet)`), so acorn's own `loc` rides along, which is what made
"rsvelte does not stamp it" the obvious hypothesis. I implemented exactly that — a
`const_decl_anchored` builder taking `node.expression.start()` — and it moved **0 of 32** grid
cells.

A counter at `to_oxc::comment_anchor` said the anchor was produced and rejected
(`anchor=Some(76)`, `last_region_source=None`), and a second counter at the chunk appender named
the cause. Ordering of the two events, one line each, same run:

```
plain:   [chunk] region=(78,123) source_offset=Some(10)   THEN  [ca] anchor=67 last_region_source=Some(10) -> CLAIMED
head:    [chunk] region=(47,92)  source_offset=Some(10)   THEN  [ca] anchor=67 last_region_source=Some(10) -> CLAIMED
snippet: [ca] anchor=76 last_region_source=None -> rejected  THEN  [chunk] region=(92,137)
```

## Mechanism: buffer order against source order

esrap flushes a pending comment by **source offset**, against a monotonically advancing
`comment_index` — the `_` visitor's `if (node.loc) flush_comments_until(context, null,
node.loc.start, true)`. The generated program's statement order need not follow source order, and
a top-level `{#snippet}` declaration is hoisted **above** the instance script's own statements.
[...]

Issue #4481, opened 2026-09-09
Trailing script comment attaches to the wrong node in generated calls (svelte:window/body/document, {#key}, {@html})


baseballyama:
## Summary

A trailing `// comment` at the end of an instance script is kept by both compilers but **attached to a different node**: official places it inside the generated call, as the leading comment of the argument it precedes in source order; rsvelte defers it past the whole statement, or into a later argument's parameter list. Because the comment then forces a different line-breaking decision, one comment produces 8-18 differing lines.

Third of three directions found in one grid. Siblings: #4475 (rsvelte **keeps** what official drops, `{#if}`) and #4480 (rsvelte **drops** what official keeps, `{#snippet}` / `<svelte:head>`). Filed separately because the direction differs; a fix for one is not evidence about the others.

## Repro

```svelte
<script>
	let c = 1;
	// x
</script>

<svelte:window onresize={() => c} />
```

official:

```js
	$.event(
		'resize',
		$.window,
		// x
		() => c
	);
```

rsvelte:

```js
	$.event('resize', $.window, () => c);
	// x
```

`{#key}` is the same defect one argument over — official writes the comment into the **second** argument's parameter list, rsvelte into the **third**'s:

```js
	// official
	$.key(
		node,
		(// x
		) => c,
		($$anchor) => { … }
	);

	// rsvelte
	$.key(node, () => c, ($$anchor// x
	) => {
```

and `{@html c}` is the same shape again (official `$.html(node, (// x` newline `) => c);`, rsvelte `$.html(node, () => c);` with the comment two lines below).

### The shared grid

Sixteen components, all with the same instance script — `let c = …;` then a trailing `// x` — under sixteen different templates. Every cell is also compiled with the comment removed, so a divergence is attributed to the comment only when the comment-free pair is byte-identical.

| cell | template | official `// x` | rsvelte `// x` | bare `{` block | diff **with** | diff **without** | class |
|---|---|---|---|---|---|---|---|
| a_if | `{#if c}` | 0 | 1 | 1 | 2 | 0 | **KEEPS** |
| o_const | `{#if true}{@const d = c}…{/if}` | 0 | 1 | 1 | 2 | 0 | **KEEPS** |
| g_snip | `{#snippet s()}` | 1 | 0 | 0 | 5 | 0 | **DROPS** |
| p_two | element + `{#snippet}` | 1 | 0 | 0 | 5 | 0 | **DROPS** |
[...]

PR #4490, opened 2026-09-09
fix(client): anchor $.head's callee on the tag name, as upstream does (#4480)

### LINKED ISSUE #4480: Trailing script comment is dropped from client output when the template has a {#snippet} or <svelte:head>
## Summary

A trailing `// comment` at the end of an instance script is **dropped entirely** from rsvelte's client output where the official compiler keeps it, whenever the template contains a `{#snippet}` or a `<svelte:head>`. Output parses and runs identically; the divergence is bytes only, so only the output-equality gates can see it.

Sibling of #4475, which is the **opposite** direction (rsvelte keeps a comment official drops, on a `{#if}`). Filed separately because they share a subsystem and not a direction: a fix for one is not evidence about the other, and #4475's own carrier correlation (a bare block statement in the output) is 0 of 3 here.

## Repro

```svelte
<script>
	let c = 1;
	// x
</script>

<svelte:head><title>{c}</title></svelte:head>
```

official (`generate: "client"`):

```js
	var head = $.head(($$anchor) => { … });
	// x
```

rsvelte: the `// x` line is absent. Whole diff is that one line.

The `{#snippet}` carrier is the same loss wearing upstream's odd placement — official emits

```js
	const // x
	s = ($$anchor) => {
```

and rsvelte emits `const s = ($$anchor) => {`.

### The shared grid

Sixteen components, all with the same instance script — `let c = …;` then a trailing `// x` — under sixteen different templates. Every cell is also compiled with the comment removed, so a divergence is attributed to the comment only when the comment-free pair is byte-identical.

| cell | template | official `// x` | rsvelte `// x` | bare `{` block | diff **with** | diff **without** | class |
| a_if | `{#if c}` | 0 | 1 | 1 | 2 | 0 | **KEEPS** |
| o_const | `{#if true}{@const d = c}…{/if}` | 0 | 1 | 1 | 2 | 0 | **KEEPS** |
| g_snip | `{#snippet s()}` | 1 | 0 | 0 | 5 | 0 | **DROPS** |
| p_two | element + `{#snippet}` | 1 | 0 | 0 | 5 | 0 | **DROPS** |
| h_head | `<svelte:head>` | 1 | 0 | 0 | 2 | 0 | **DROPS** |
| c_key | `{#key c}` | 1 | 1 | 0 | 18 | 0 | **MISPLACES** |
| j_win | `<svelte:window onresize>` | 1 | 1 | 0 | 10 | 0 | **MISPLACES** |
| k_body | `<svelte:body onclick>` | 1 | 1 | 0 | 10 | 0 | **MISPLACES** |
| m_doc | `<svelte:document onclick>` | 1 | 1 | 0 | 10 | 0 | **MISPLACES** |
| l_html | `{@html c}` | 1 | 1 | 0 | 8 | 0 | **MISPLACES** |
| f_each | `{#each c as v}` | 1 | 1 | 0 | 25 | **6** | contaminated |
| b_await | `{#await c}` | 1 | 1 | 0 | 0 | 0 | control |
| d_elem | `<b>{c}</b>` | 1 | 1 | 0 | 0 | 0 | control |
| e_none | *(no template)* | 1 | 1 | 0 | 0 | 0 | control |
| i_bound | `<input bind:value={c} />` | 1 | 1 | 0 | 0 | 0 | control |
| n_opts | `<svelte:options runes={false} />` | 1 | 1 | 0 | 0 | 0 | control |

Five cells agree byte-for-byte and are live controls (each one *does* carry the comment through, on both sides). `f_each` is excluded: with the comment stripped it still differs by 6 lines (`$.only_child(b, true)` against `$.child(b, true)`), so its 25 is two mechanisms added together — and it is the largest number in the table, i.e. the cell most likely to be picked up first.

Three of the sixteen cells are in this class: `g_snip`, `p_two`, `h_head`. `p_two` is `<b>{c}</b>` followed by a `{#snippet}`, so the snippet carries it even when it is not the first thing in the template.

## Reach

**Not measured for this mechanism specifically.** What is measured is the containing population: over the collected corpus, **116 `(id, target)` pairs diverge and become byte-equal when the trailing script comment is removed** (284 carriers of the shape screened; 88 diverge for reasons the comment does not explain; 30 threw). Those 116 have not been partitioned across this issue, #4475 and the misplacement sibling — that partition is the measurement that would size all three, and it needs one sweep rather than three.

## Where it goes

The comment cursor in `crates/rsvelte_esrap/src/printer.rs` — `flush_comments_until` / `reset_comment_index` / the 16 `flush_leading` call sites — and what `js_ast/to_oxc.rs` hands it. `JsBlockStatement` carries no span field, so `to_oxc::consumed` derives one from the comment-buffer region a subtree consumed; a node whose consumed region starts *after* a pending comment leaves that comment to some later flush, and a node whose reset empties the cursor loses it. #4475's instrumented trace (`#[track_caller]` on both printer functions plus `consumed`, printing `to`/`node_start`, `has_loc`, `loc_base`, the cursor and the next pending comment) reproduces on these cells unchanged and is the instrument to start from.

## What a fix must not break

The five control cells above are byte-equal today **and carry the comment through on both sides**, so they are not vacuous: `{#await}`, a plain element, a component with no template construct at all, `<input bind:value>`, and `<svelte:options runes={false} />`. Any fix has to leave all five at zero.

baseballyama:
Closes #4480 (the `<svelte:head>` half; the `{#snippet}` half is a different producer and is
tracked in #4489 — see below).

## The defect

A comment at the end of the instance script is dropped when the template holds a `<svelte:head>`.

```svelte
<script>
	let c = 1;
	// x
</script>
<svelte:head><title>{c}</title></svelte:head>
```

Official keeps `// x` and rsvelte drops it, on `client` and `client-dev`.

## Cause

esrap's `_` visitor flushes every pending comment at the first printed node that carries a source
`loc` (`languages/ts/index.js:961`, `if (node.loc) flush_comments_until(context, null,
node.loc.start, true)`). Upstream stamps `$.head`'s callee with the tag name's own source position
so that node is the flush point:

```js
// packages/svelte/src/compiler/phases/3-transform/client/visitors/SvelteHead.js:17
b.stmt(b.call(b.id('$.head', node.name_loc), …))
```

rsvelte built the same call unanchored, so nothing after the script chunk carried a comment-space
position and the comment was never flushed. The fix passes `node.name_loc` through a new
`b::stmt_anchored`.

**Upstream's loc-carrying builder sites are a closed set, not a work log.** Every `_loc` use under
`phases/3-transform` is `SvelteSelf.js:11`, `BindDirective.js:56,58`, `SvelteHead.js:17`,
`TitleElement.js:23`, `shared/component.js:277,278`, `shared/fragment.js:116`, `Component.js:10`,
`Fragment.js:89`, and all of them take `name_loc` from the parser
(`1-parse/state/element.js:186/203/660/691`). The enumeration here is the grammar.

## Why the `{#snippet}` half is not in this PR

`SnippetBlock.js:82` reuses the **source AST** identifier verbatim (`b.const(node.expression,
snippet)`), so acorn's own `loc` rides along — which makes "rsvelte does not stamp it" the obvious
hypothesis. I implemented it (a `const_decl_anchored` taking `node.expression.start()`) and it
moved **0 of 32** grid cells. A counter at the port said the anchor was produced and *rejected*,
and a counter at the chunk appender named why:

```
plain:   [chunk] region=(78,123) src=Some(10)   THEN  [ca] anchor=67 last_region_source=Some(10) -> CLAIMED
head:    [chunk] region=(47,92)  src=Some(10)   THEN  [ca] anchor=67 last_region_source=Some(10) -> CLAIMED
snippet: [ca] anchor=76 last_region_source=None -> rejected  THEN  [chunk] region=(92,137)
```

A top-level `{#snippet}` declaration is hoisted above the instance script's statements; esrap
flushes by *source offset* while rsvelte's comment buffer is built lazily in *walk order*, so at
the snippet's claim the script chunk is not in the buffer yet. That is a model change, not a
builder edit. Full write-up, with 13 candidate carriers of which 3 confirmed divergent, in #4489.

## Measurement

**Grid** — 16 templates × 2 dev modes, expectations generated from the pinned official compiler
per run, each cell also compiled with the comment removed so a divergence counts only when the
comment-free pair is byte-equal:

| arm | EQ | DIFF |
|---|---:|---:|
| base | 24 | 8 |
| this PR | **26** | 6 |

No cell moved the wrong way. The 6 remaining are `snippet`/`snippet_rnd` (#4489) and `boundary`
(a separate residue, 0 real-world carriers). Oracle liveness: 13 keeps / 3 drops, so the grid is
two-sided rather than measuring one answer.

**Arm identity** — `base 9b50b1c4c07eff34`, `head 72082a9b1eb1e009`, distinct; two-sided probe
run before the result was read: `head 0 → 1` (contains the fix) and `if_ 0 → 0`, `plain 1 → 1`
(lacks what it should lack).

**Corpus sweep** — two arms, one process per arm, over the 104 declared corpus sources read from
the primary checkout (this worktree has 117 of 119 submodules empty; `submodules/vize` is the one
empty directory there and is not a corpus source).

```
files=37397  live=141662  dead=7926   (identical on both arms)
rows=149588  distinctKeys=149588      (no key collapse)
MOVED = 6
stage 2, raw bytes against the oracle:
  base   {"MISMATCH": 6}
  this   {"match": 6}
  match -> MISMATCH : 0
```

The 6 are three real-world files × `client` and `client-dev`:
`skeleton/.../themes/create/+page.svelte`, `svelte-bits/.../PixelBlastDemo.svelte`,
`svelte-commerce/.../my/addresses/[id]/+page.svelte`. `known-failures.client.json` and
`known-failures.client-dev.json` are both empty and no server unit moved, so there is no
re-baseline. Only 6 rows changed bytes at all, so no other entry's verdict can have moved either.

The sweep was measured at `a605c9d55`; the branch now sits on `68d3b8e5b`, whose single
intervening commit touches `AGENTS.md` and `.claude/skills/perf-loop/SKILL.md` and no compiler
code — a complement, so the sweep holds for the tree CI scores.

## Test

`crates/rsvelte_core/tests/svelte_head_comment_anchor.rs`. The corpus output gates cannot hold
this class — `ast_equiv_batch` runs with `CommentPolicy::Ignore` and `verify.mjs` normalizes with
oxfmt plus blank-line stripping, so a comment-placement divergence is erased twice and a
`pattern-corpus` repro would be green on both arms.

Expectations are read off the pinned official compiler rather than typed, and the grid asserts its
own two-sidedness on both axes (a cell expecting 0, a cell expecting > 0, and a cell whose dev and
prod answers differ). Ablation control: replacing `b::stmt_anchored` with `b::stmt` reddens the
suite naming `cell=head`; restoring leaves the tree byte-identical.

### DIFF
--- a/.changeset/head-call-carries-its-tag-name-loc.md
+++ b/.changeset/head-call-carries-its-tag-name-loc.md
@@ -1,0 +1,12 @@
+---
+'@rsvelte/compiler': patch
+---
+
+anchor `$.head`'s callee on the tag name, as upstream does
+
+Upstream stamps the call's callee with the tag name's own source position
+(`b.id('$.head', node.name_loc)`), which is where esrap flushes a comment left
+pending at the end of the instance script. rsvelte built the call unanchored, so
+the comment was dropped where official keeps it. `<svelte:head>` now passes
+`node.name_loc` through a `b::stmt_anchored`. The `{#snippet}` half of the same
+symptom is a different mechanism and is tracked separately (#4489).
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/svelte_head.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/visitors/svelte_head.rs
@@ -45,13 +45,17 @@ pub fn svelte_head(node: &SvelteElement, context: &mut ComponentContext) {
     // Build the head call: $.head('hash', ($$anchor) => { ... })
     let content_fn = b::arrow_block(vec![b::id_pattern("$$anchor")], content_block.body);
 
-    let head_call = b::stmt(
+    // Upstream stamps the callee with the tag name's own source position
+    // (`b.id('$.head', node.name_loc)`), which is where esrap flushes a comment
+    // left pending by an earlier chunk.
+    let head_call = b::stmt_anchored(
         &context.arena,
         b::call(
             &context.arena,
             b::member_path(&context.arena, "$.head"),
             vec![b::string(&hash), content_fn],
         ),
+        node.name_loc.as_ref().map(|loc| loc.start.character),
     );
 
     context.state.init.push(head_call);
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/builders.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/builders.rs
@@ -1511,6 +1511,16 @@ pub fn stmt(arena: &JsArena, expression: JsExpr) -> JsStatement {
     })
 }
 
+/// `expr;` whose callee carries the original-source offset upstream stamps on it
+/// (`b.id('$.head', node.name_loc)`), which is where esrap flushes comments left
+/// over from an earlier chunk.
+pub fn stmt_anchored(arena: &JsArena, expression: JsExpr, anchor: Option<u32>) -> JsStatement {
+    JsStatement::Expression(JsExpressionStatement {
+        expression: arena.alloc_expr(expression),
+        comment_anchor: anchor,
+    })
+}
+
 /// Create a return statement.
 pub fn return_stmt(arena: &JsArena, argument: Option<JsExpr>) -> JsStatement {
     JsStatement::Return(JsReturnStatement {
--- a/crates/rsvelte_core/tests/svelte_head_comment_anchor.rs
+++ b/crates/rsvelte_core/tests/svelte_head_comment_anchor.rs
@@ -1,0 +1,89 @@
+//! Upstream stamps `$.head`'s callee with the tag name's own source position
+//! (`b.id('$.head', node.name_loc)`), so esrap flushes a comment left pending at
+//! the end of the instance script there. rsvelte built the call unanchored, and
+//! the comment was dropped.
+//!
+//! Expectations are read off the pinned official compiler
+//! (`submodules/svelte/.../src/compiler/index.js`) rather than typed, and pin
+//! *whether the comment survives* per target so unrelated codegen movement does
+//! not re-pin this class silently. The corpus output gates cannot hold it:
+//! `ast_equiv_batch` compares with `CommentPolicy::Ignore` and `verify.mjs`
+//! normalizes with oxfmt, so both arms score equivalent.
+
+use oxc_allocator::Allocator;
+use oxc_parser::Parser;
+use oxc_span::SourceType;
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+
+/// `(name, template, comments kept with dev off, comments kept with dev on)`.
+const CELLS: &[(&str, &str, usize, usize)] = &[
+    (
+        "head",
+        "<svelte:head><title>{c}</title></svelte:head>",
+        1,
+        1,
+    ),
+    ("plain", "<b>{c}</b>", 1, 1),
+    ("if_", "{#if c}<b>{c}</b>{/if}", 0, 0),
+    ("each", "{#each [1] as n}<b>{c}{n}</b>{/each}", 1, 1),
+    ("key", "{#key c}<b>{c}</b>{/key}", 1, 1),
+    ("await_", "{#await p}<b>{c}</b>{/await}", 1, 1),
+    ("html", "{@html c}", 1, 1),
+    ("const_", "{#if c}{@const d = c}<b>{d}</b>{/if}", 0, 0),
+    ("window", "<svelte:window on:resize={() => c} />", 1, 0),
+    ("bind", "<input bind:value={c} />", 1, 1),
+    ("empty", "", 1, 1),
+];
+
+const HEAD: &str = "<script>\n\tlet c = 1;\n\tlet p = Promise.resolve(1);\n\t// x\n</script>\n";
+
+fn compile_cell(template: &str, dev: bool) -> String {
+    let source = format!("{HEAD}{template}\n");
+    let output = compile(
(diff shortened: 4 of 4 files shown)


baseballyama:
Re-keyed the grid on **where** the comment lands rather than **whether** it survives, because a
count key cannot see a misplacement and #4481 is exactly that class. The key is the line holding
`// x` plus the previous and next non-blank lines.

Under that key, over the same 16 templates x 2 dev modes:

```
base   (9b50b1c4c07eff34)   EQ 15 / DIFF 17
#4490  (72082a9b1eb1e009)   EQ 17 / DIFF 15

set difference of the DIFF sets:  head/false, head/true   — and nothing else
```

So the fix moves exactly the two cells it claims and no third cell's placement changed, which the
count key could not have told me: it scored the same grid 24 → 26 and calls nine placement
divergences EQ. The 15 that remain are #4489 (`snippet`, `snippet_rnd`), #4481 (`each`, `key`,
`html`, `window`, `document`, `body`) and the `boundary` residue.

The corpus sweep already reported `MOVED = 6` over 149,588 rows, all `MISMATCH -> match`; this is
the same claim on the generated-shape side rather than the byte side, and it is the one that rules
out "the comment now survives but lands somewhere else".

One correction to my own PR body while I am here: I described the grid's 6 remaining DIFFs under
the count key. Under the placement key it is 15, and that is the honest number for "cells where
this family still diverges" — the count key's 6 is a smaller number because the key is weaker, not
because fewer cells diverge.

PR #4496, opened 2026-09-09
fix(release): a changeset tick that another PR earned says so (#4486)

### LINKED ISSUE #4486: release: `check-core-consumer-changesets` reads the whole `.changeset/` directory, so another PR's changeset satisfies the requirement
`check-core-consumer-changesets.mjs` prints a line that reads as a statement about the PR:

```
✓ @rsvelte/compiler (touched: crates/rsvelte_core/src/)
```

Its two inputs are not symmetric. The **touched** side is `git diff base...HEAD` — this PR's files. The **changeset** side is a directory read over `.changeset/`, which is *every pending changeset in the tree*, including ones already on `main` from other PRs. So the requirement can be satisfied by somebody else's changeset, and the check still prints `✓`.

Confirmed by ablation on a real PR: moving that PR's own changeset out of `.changeset/` and re-running **still passes**, because nine changesets already on `main` name `@rsvelte/compiler`.

## Measured

Against `origin/main` at `16e21f3a5`, 13 pending changesets, for the 7 packages any rule can require:

| | packages |
|---|---|
| satisfiable by someone else's changeset | `@rsvelte/compiler` (9), `@rsvelte/language-server` (4), `@rsvelte/svelte-check` (3), `@rsvelte/svelte2tsx` (3) |
| discriminating | `@rsvelte/capi`, `@rsvelte/fmt`, `@rsvelte/vite-plugin-svelte-native` |

**3 of 7 discriminating today.**

The discriminating power is therefore a function of the release backlog, and it is weakest exactly when the backlog is largest — when the most PRs are landing. It is fully armed only immediately after a release, and it is *never* armed for `@rsvelte/compiler`, which is what the busiest rule in the file (`crates/rsvelte_core/src/` → `['@rsvelte/compiler']`) requires.

## What this does and does not break

The release outcome is still correct in the ordinary case: if any pending changeset bumps `@rsvelte/compiler`, a fix in `rsvelte_core` ships inside that bump. What the green does **not** establish is what the printed line appears to say — that *this PR* carries a changeset.

The narrow real hole is a changeset that satisfies the check and is then **withdrawn before the release**. The case where that bites is `crates/rsvelte_projection/src/svelte2tsx/`, whose rule requires all three of `@rsvelte/compiler`, `@rsvelte/svelte2tsx` and `@rsvelte/svelte-check` — precisely because `@rsvelte/svelte-check` is a self-contained native binary with **no cascade edge**, as the script's own comment explains. `@rsvelte/svelte-check` is in the satisfied-by-someone-else column right now, so that rule's load-bearing third requirement is currently unarmed.

## Suggested fix

Restrict the changeset side to the same commit range the touched side uses — the changesets **added by this PR** (`git diff --name-only --diff-filter=A base...HEAD -- .changeset/`) — rather than reading the directory. Keeping the directory read as a secondary, separately-labelled line would preserve the "will this ship at all" answer without letting it stand in for the per-PR one.

## Method notes

Two instrument failures on the way to this number, both worth having in the issue because each returned a well-formed wrong answer:

- The first ablation was performed on the **working-tree** file while the checker reads the **directory**, and the changeset was already committed. With nine other changesets also present there were two independent reasons for a non-firing control, and neither could be separated from the result — only reading the script's two input sources did.
- A first count of the satisfied column returned **7 of 7 discriminating**, because the extraction regex matched only double-quoted package names. Changesets on `main` use single quotes (`'@rsvelte/language-server': patch`); the PR's own used double, so the same regex worked there and failed on `main`. It was caught because a second party's count disagreed, and because 13 changesets yielding **0** package names is a shape that cannot occur. The corrected extractor accepts either quote style and carries a positive control (13 of 13 changesets name at least one `@rsvelte` package).

baseballyama:
Closes #4486.

`check-core-consumer-changesets` computes its verdict from **every pending changeset in the tree**, and that is deliberate — its own comment says so, and the question it answers ("will the changed code ship in some bump") is the right one for a release gate. What was wrong is the line it printed, which reads as the *other* question:

```
✓ @rsvelte/compiler  (touched: crates/rsvelte_core/src/)
```

That tick can be earned by a changeset somebody else added. **The verdict is unchanged by this PR** — no new failures, no policy change — only the report now says which question each tick answered.

## What it prints now

Forced end-to-end with a temporary commit touching `crates/rsvelte_core/src/`, then reverted:

```
Shared-core changes require these packages to be named in a changeset:
  ✓ @rsvelte/compiler  (touched: crates/rsvelte_core/src/)  — NOT named by this PR;
    satisfied by 12 pending changesets already in the tree (attribute-key-quote-anchor.md, …)

Note: 1 requirement is met only by changeset(s) this PR did not add. The release is
still correct while those stay pending, but withdrawing one before the release would
leave this change with no version bump and no CHANGELOG entry — and this check would
not have said so.
```

## Re-measured, because the issue's numbers were taken against an older tree

At `5ad043008`, 16 pending changesets (the issue measured 13 at `16e21f3a5`), for the 7 packages any rule can require:

| | packages |
|---|---|
| satisfiable by someone else's changeset | `@rsvelte/compiler` (**12**), `@rsvelte/language-server` (5), `@rsvelte/svelte-check` (4), `@rsvelte/svelte2tsx` (4) |
| discriminating | `@rsvelte/capi`, `@rsvelte/fmt`, `@rsvelte/vite-plugin-svelte-native` |

Still **3 of 7**, and the mechanism has got worse rather than better — the discriminating power is a function of the release backlog, so it is weakest exactly when the most PRs are landing. Positive control on the re-measurement: 16 of 16 changesets parse to at least one `@rsvelte` package.

## `packagesIn` is extracted so it can be pinned

The issue records that a first count of this read **7 of 7 discriminating**, because the extractor matched only double-quoted names while changesets on `main` use single quotes. That failure mode returns a plausible **empty set** rather than an error, so it reads as "this changeset names nothing". Both spellings occur here, so both are now pinned, along with an unquoted one.

## Controls, and one of them was dead when written

Three new checks, each ablated and required to go red:

- **double-quotes-only** (the actual #4486 bug) → 3 checks fail. ✅
- **scan the whole file instead of the frontmatter** → **green on the first attempt.** The `frontmatter only` cell used `mentions @rsvelte/fmt: in prose` as its body line, and the regex anchors at `^`, so scanning the whole file changed nothing. The cell was named for a property it did not test. Re-written with a body line that itself looks like a frontmatter entry, the ablation goes red. ✅
- **every pending changeset names a package** — a live control on the tree, so a reader matching none of them cannot pass quietly.

After each ablation the file was restored and `cmp` confirmed byte-identical, with a clean re-run at exit 0.

The end-to-end check was worth its cost for the same reason: on this branch the script exits 0 saying *"No shared-code source touched"*, so the new code never runs. A green here proves nothing about it without forcing a rule to fire.

## Scope

`scripts/release/` only; no package source, so no changeset (and no rule prefix matches, so the guard does not ask for one). Node-only — no cargo, no ratchet, and it does not trigger the LSP corpus job.

### DIFF
--- a/scripts/release/check-core-consumer-changesets.mjs
+++ b/scripts/release/check-core-consumer-changesets.mjs
@@ -170,20 +170,58 @@ function changedFiles(base) {
   return out ? out.split('\n').filter(Boolean) : [];
 }
 
-// Names in the frontmatter of every pending changeset (working-tree state — the
-// set the Release workflow will consume), not just ones added in this PR.
+// Package names in one changeset's frontmatter. Both quote styles occur in this
+// repository, and a double-quoted-only reader returns a plausible empty set
+// rather than an error (#4486).
+export function packagesIn(text) {
+  const named = new Set();
+  const m = text.match(/^---\r?\n([\s\S]*?)\r?\n---/);
+  if (!m) return named;
+  for (const line of m[1].split('\n')) {
+    const pkg = line.match(/^\s*["']?(@[^"':]+)["']?\s*:/);
+    if (pkg) named.add(pkg[1]);
+  }
+  return named;
+}
+
+// Every pending changeset (working-tree state — the set the Release workflow
+// will consume), mapped to the files naming it. This is the set the verdict is
+// computed from, and it deliberately includes changesets already on main:
+// the question it answers is "will the changed code ship in some bump", not
+// "does this PR carry a changeset".
 function namedPackages() {
   const dir = path.join(repoRoot, '.changeset');
-  const named = new Set();
+  const named = new Map();
   for (const file of readdirSync(dir)) {
     if (!file.endsWith('.md') || file === 'README.md') continue;
-    const text = readFileSync(path.join(dir, file), 'utf8');
-    const m = text.match(/^---\r?\n([\s\S]*?)\r?\n---/);
-    if (!m) continue;
-    for (const line of m[1].split('\n')) {
-      const pkg = line.match(/^\s*["']?(@[^"':]+)["']?\s*:/);
-      if (pkg) named.add(pkg[1]);
+    for (const pkg of packagesIn(readFileSync(path.join(dir, file), 'utf8'))) {
+      if (!named.has(pkg)) named.set(pkg, []);
+      named.get(pkg).push(file);
     }
+  }
+  return named;
+}
+
+// The changesets this PR itself adds or edits. Separate from the set above
+// because the two answer different questions and only this one is a statement
+// about the pull request.
+function changesetsTouchedHere(base) {
+  let out = '';
+  try {
+    out = sh(`git diff --name-only --diff-filter=AM ${base}...HEAD -- .changeset`);
+  } catch {
+    return new Set();
+  }
+  const named = new Set();
+  for (const rel of out ? out.split('\n').filter(Boolean) : []) {
+    if (!rel.endsWith('.md') || rel.endsWith('README.md')) continue;
+    let text;
+    try {
+      text = readFileSync(path.join(repoRoot, rel), 'utf8');
+    } catch {
+      continue; // added then removed again within the range
+    }
+    for (const pkg of packagesIn(text)) named.add(pkg);
   }
   return named;
 }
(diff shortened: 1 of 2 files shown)


baseballyama:
## Both branches of the new code have now been run

The end-to-end probe in the description exercised one branch. A tick has two outcomes and only the borrowed one had ever executed, so here is the other, forced the same way (temporary commit touching `crates/rsvelte_core/src/` **and** adding a changeset naming the required package, then reverted):

```
Shared-core changes require these packages to be named in a changeset:
  ✓ @rsvelte/compiler  (touched: crates/rsvelte_core/src/)  — named by this PR
All required consumer packages are named. ✓
```

Note the summary paragraph is correctly **absent** here — it is emitted only when `borrowed > 0`, so a PR that carries its own changeset sees no new output at all. That is the case worth checking, because it is every well-formed PR, and a gate that grew a paragraph for them would be noise on the common path.

After both probes: `git diff --numstat origin/main` is the two script files and nothing else.

## One degradation direction, stated

`changesetsTouchedHere` wraps its `git diff` in a try/catch returning an empty set. If the range is unavailable — a shallow checkout, say — every satisfied package prints `NOT named by this PR` and the note fires. That is noisy and it is the safe direction: it can over-warn, never under-warn, and it cannot change the verdict, which is still computed from `namedPackages()` exactly as before. The existing `changedFiles()` makes the same `base...HEAD` call without a guard and works in CI today, so this path is not expected to be reached.

PR #4499, opened 2026-09-09
fix(lsp): append the Svelte guidance upstream adds to two TypeScript diagnostics (#4498)

### LINKED ISSUE #4498: lsp: the tsgo diagnostic proxy ports none of `adjustIfNecessary`, so a component-type error loses its "Possible causes" guidance
## What

Upstream's tsgo diagnostics provider post-processes a TypeScript diagnostic before returning it.
`adjustIfNecessary` (`plugins/typescript-go/features/DiagnosticsProvider.ts:946-981`, and its twin
at `plugins/typescript/features/DiagnosticsProvider.ts:423-443`) rewrites the **message text** of two
diagnostic codes, and `swapDiagRangeStartEndIfNecessary` beside it normalises an inverted range.

rsvelte's proxy ports none of it. `crates/rsvelte_language_server/src` contains **0** occurrences of
either augmentation's text.

```ts
// upstream, both providers
if (diagnostic.code === 2345 && diagnostic.message.includes('ConstructorOfATypedSvelteComponent')) {
    return { ...diagnostic, message: diagnostic.message +
        '\n\nPossible causes:\n' +
        '- You use the instance type of a component where you should use the constructor type\n' +
        '- Type definitions are missing for this Svelte Component. ' + ... };
}
if (diagnostic.code === 1184) {   // "Modifiers cannot appear here."
    return { ...diagnostic, message: diagnostic.message +
        '\nIf this is a declare statement, move it into <script context="module">..</script>' };
}
```

This is a straight port rather than a design question: rsvelte's architecture mirrors the
`typescript-go` provider, which carries the same function.

## Measured

Artifact `lsp-current-fixtures` from run `34303105707` (the fixtures shard), reading the gate's own
`diagnosticDetails`, which stores both sides' diagnostic items for every diverging key.

| arm | official | rsvelte | carriers |
|---|---|---|---|
| 2345 + `ConstructorOfATypedSvelteComponent` -> `Possible causes:` | present | **absent** | `upstream-features/component-invalid/input.svelte` |
| 1184 -> `declare statement` hint | no item with code 1184 in this shard | — | **UNMEASURED** |
| `swapDiagRangeStartEndIfNecessary` | no item with `start > end` in this shard | — | **UNMEASURED** |

Only the first arm has a carrier here, so only the first is measured. The other two are unported by
inspection and their reach is unknown — not zero.

On the one measured carrier the two messages agree for 385 characters and then rsvelte simply stops:
official is 547 characters, rsvelte's is a **strict prefix** of it.

```
common prefix ends: ...dSvelteComponent': $$prop_def, $$events_def, $$slot_def, $on
official continues: "\n\nPossible causes:\n- You use the instance type of a component where ..."
rsvelte  continues: ""
```

## How it surfaced, and why it is not #4483

It is pre-existing; nothing about it is new. It became *visible* in #4483's gate run because that
change improves range fidelity, so the differ began pairing diagnostic items on this file and could
compare their `message` fields for the first time — `GATES.md` 39b's adding direction. Before that
the divergence was coarser and the message never got compared.

Two consequences worth stating:

* The four `component-invalid` `message:value-mismatch` keys in #4483's run are **this** defect, not
  that PR's, and should not be attributed to it.
* A gate can hold a defect for as long as its comparison stays coarse. This one was reachable by the
  fixtures shard the whole time and no key expressed it.

## Note on the denominator

Counting raw items in `diagnosticDetails` gives "96 of 614 official items carry the augmentation
against 0 of 1110 rsvelte items". That 96 **over-counts**: the sidecar stores the full item list once
per diverging key, so one file's items are repeated across its keys. The honest figure for this
shard is **one carrier file**, and the meaningful zero is rsvelte's — 0 occurrences of the string in
the source, which is a property of the port and not of the population.

baseballyama:
Closes #4498.

Upstream rewrites a diagnostic's message before returning it — `adjustIfNecessary`,
`typescript-go/features/DiagnosticsProvider.ts:946` — and rsvelte forwarded tsgo's text
unchanged. Two arms: a TS2345 whose message mentions `ConstructorOfATypedSvelteComponent`
gains a "Possible causes:" paragraph naming the instance-vs-constructor mistake, and TS1184
(`Modifiers cannot appear here.`) gains a line pointing at `<script context="module">`.

## The carrier is measured, not asserted

Read out of **#4483's own CI log** rather than from a run of mine — the gate's own output is a
better oracle than a local arm, and it cost no build:

```
upstream-features/component-invalid/input.svelte, 4 diagnostics
  4 official WITH-suffix
  4 rsvelte  NO-suffix
```

Official's text ends with exactly `COMPONENT_CONSTRUCTOR_HINT`, trailing space included. The
trailing space is not a transcription slip: upstream concatenates a Svelte-3.31
`SvelteComponentTyped` paragraph after it and selects `''` when `isSvelte5Plus`, so on the only
target this port has, the message ends with the separator still attached.

The 1184 arm has **0** occurrences in that same log, against 48 of `Possible causes` as the live
control. So it is ported by inspection with a *measured* absence of carriers rather than an
assumed one.

## What this PR should move: 0 ratchet keys

Pre-registered before pushing, and this is a correction — my first prediction was 8 re-keys and
it was wrong. `diff.mjs:20-21` decides it by reading:

```js
if (method === "textDocument/diagnostic" && pointer.endsWith("/items")) {
  return `diagnostic-${digest([value.code, value.source, value.range?.start])}`;
}
```

Identity is `(code, source, range.start)` — message is not in it — and the `extra`/`missing` set
digest is over those identity keys, not over values. So a message change moves neither pairing
nor any set hash. Three legs:

1. Pairing and set hashes exclude `message`, per the code above.
2. The whole ratchet holds exactly **2** `message:value-mismatch` keys of any method, both on
   `upstream-testfiles/completions/string-completion.svelte`, whose source is `abc('@')` against
   `function abc(str: 'hi' | '@hi' | '~hi')` — a TS2345 with no
   `ConstructorOfATypedSvelteComponent`, which the guard declines. It is the live corpus case for
   `only_a_component_constructor_2345_is_adjusted`.
3. On `main`, `component-invalid`'s items do not pair (`extra 13 / missing 5`), so no message
   there is compared until #4483 lands.

Can it *break* a currently-matching diagnostic? Structurally no: official appends the suffix
unconditionally for that code-and-substring, so a paired carrier could only be matching today if
rsvelte already emitted the suffix, which it never has. Every such pair would already be listed,
and neither listed key is that shape.

If the LSP gate goes red here, the argument is wrong somewhere and I would rather learn it from
the gate.

## What a green gate here does NOT establish

Stated plainly because it is the thing most likely to be misread later. The prediction above is
0 keys **on `main`**, and the risk is not that it fails — it is that it succeeds and the success
is later cited as validation of the fix.

Nothing in the ratchet distinguishes *0 because the carrier is unobservable until #4483 makes
those four diagnostics pair* from *0 because the change does nothing at all*. Both produce an
identical green. So a green `Language server` here confirms that this PR is inert on the current
tree, which is what it was predicted to be, and it is not evidence that the transformation is
correct.

The evidence for correctness is the five-row table above — in particular the second row, which is
the gate's own digest over the whole diagnostic object rather than a restatement of the string
equality, and the negative control that dropping one trailing space from a 162-character constant
breaks the match. Those were measured against the oracle's real output. The green was not.

The check that *would* validate this against the gate arrives when #4483 lands, since that is
what makes the carrier observable at all.

## Merge order: this lands before #4483 re-baselines

#4483 narrows `component-invalid` to `extra 9 / missing 1`; four items pair for the first time;
their messages are compared for the first time; and 16 keys appear (4 items × opened/`phase=edit`
× `differential`/`expected`), all carrying the identical
`official=32c129f586d1, rsvelte=224effe9a18b`. That is GATES.md 39b's adding direction, not
something either PR breaks — but if #4483 re-baselines those 16 in and this PR merges after, the
16 go stale and the two-sided ratchet turns **`main`** red, visible on neither PR because each
merge ref holds only the other's absence.

## Tests, and what they do not cover

Four unit tests, all passing; `cargo test -p rsvelte_language_server --lib` reads
380 passed / 0 failed / 1 ignored.

Two ablations, because "4 tests pass" is not a coverage claim:

- **Drop the trailing space from the constant** → exactly one test fails, and it is
  `the_constructor_hint_is_byte_identical_to_the_official_server`. The other three are written in
  terms of the constant, so they are self-consistent and cannot catch a wrong one — the
  port-vs-port oracle problem inside a single test file. Only the independently-pinned test can.
- **Remove the call site, keep the function** → **381 passed, 0 failed**. The function can go
  dead with the suite green. `fill_diagnostic_tags` beside it has the identical gap. What covers
  the wiring is the LSP fixtures gate's `component-invalid` carrier, which is this change's
  falsifiable prediction rather than a claim of coverage.

## Not ported

Upstream calls `swapDiagRangeStartEndIfNecessary` on the following line. rsvelte has that swap in
`map_preprocessed_diagnostic` — the preprocessor source-map path — and the crate's only
`std::mem::swap` is that one, so the tsgo proxy path has none. Whether that path can produce an
inverted range is unmeasured. It is a range defect rather than a message one and is deliberately
left out.

One incidental worth recording: the oracle carries **two** copies of this function,
`typescript/features/` (706 lines, `adjustIfNecessary` at :423) and `typescript-go/features/`
(1281 lines, at :946). They are byte-identical over that function, so the strings do not depend
on which plugin a run selects — but the line number does, and rsvelte proxies tsgo. I nearly
"corrected" the right citation to the wrong port.

### DIFF
--- a/.changeset/diagnostic-message-adjustments.md
+++ b/.changeset/diagnostic-message-adjustments.md
@@ -1,0 +1,7 @@
+---
+"@rsvelte/language-server": patch
+---
+
+Append the Svelte-specific guidance upstream adds to two TypeScript diagnostics: a component
+constructor-type error (TS2345 mentioning `ConstructorOfATypedSvelteComponent`) and
+`Modifiers cannot appear here.` (TS1184).
--- a/crates/rsvelte_language_server/src/server.rs
+++ b/crates/rsvelte_language_server/src/server.rs
@@ -3400,6 +3400,7 @@ impl Server {
                     }
                     if method == "textDocument/diagnostic" {
                         fill_diagnostic_tags(result);
+                        adjust_diagnostic_messages(result);
                     }
                     if let Some(fallback) = fallback_result {
                         merge_tsgo_result(&method, result, fallback);
@@ -4305,6 +4306,59 @@ const DEPRECATED_CODES: [i64; 2] = [6385, 6387];
 /// `reportsDeprecated` off the TypeScript API instead, and its `getDiagnosticTag`
 /// returns `[]` rather than `undefined`, so every diagnostic it emits carries the
 /// key. The code decides both flags, so the LSP surface can reproduce them.
+/// Upstream rewrites two diagnostics' message text before returning them
+/// (`typescript-go/features/DiagnosticsProvider.ts:946-981`). Svelte 5 is the
+/// only target here, so the `isSvelte5Plus` arm's `SvelteComponentTyped`
+/// paragraph is never appended -- which is why the first suffix ends in a space.
+const COMPONENT_CONSTRUCTOR_HINT: &str = "\n\nPossible causes:\n\
+    - You use the instance type of a component where you should use the constructor type\n\
+    - Type definitions are missing for this Svelte Component. ";
+
+const DECLARE_STATEMENT_HINT: &str =
+    "\nIf this is a declare statement, move it into <script context=\"module\">..</script>";
+
+fn adjust_diagnostic_messages(result: &mut serde_json::Value) {
+    fn adjust_items(value: &mut serde_json::Value) {
+        let Some(items) = value
+            .get_mut("items")
+            .and_then(serde_json::Value::as_array_mut)
+        else {
+            return;
+        };
+        for item in items {
+            let Some(object) = item.as_object_mut() else {
+                continue;
+            };
+            let code = object.get("code").and_then(serde_json::Value::as_i64);
+            let Some(message) = object.get("message").and_then(serde_json::Value::as_str) else {
+                continue;
+            };
+            let suffix = match code {
+                Some(2345) if message.contains("ConstructorOfATypedSvelteComponent") => {
+                    COMPONENT_CONSTRUCTOR_HINT
+                }
+                Some(1184) => DECLARE_STATEMENT_HINT,
+                _ => continue,
+            };
+            if message.ends_with(suffix) {
+                continue;
+            }
+            let adjusted = format!("{message}{suffix}");
+            object.insert("message".to_string(), serde_json::Value::String(adjusted));
+        }
+    }
+
+    adjust_items(result);
+    if let Some(related) = result
+        .get_mut("relatedDocuments")
+        .and_then(serde_json::Value::as_object_mut)
+    {
+        for report in related.values_mut() {
+            adjust_items(report);
+        }
+    }
+}
+
 fn fill_diagnostic_tags(result: &mut serde_json::Value) {
     fn tag_items(value: &mut serde_json::Value) {
         let Some(items) = value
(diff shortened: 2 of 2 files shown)


baseballyama:
## The fix reproduces official's bytes on the real carrier — verified without running the gate

A peer retracted a session's worth of correct measurement today for being attached to the wrong
issue, and the caution they drew from it applies straight to this PR: their candidate fix scored
20/20 on a synthetic grid and then **inverted** on a 288-cell sweep (`EQ 264 → 24`), with **0**
corpus carriers to catch it. This PR is in the same structural position — I predict it moves 0
ratchet keys on `main`, which means the gate cannot currently validate it either.

So I checked the thing that actually matters, and it was checkable from #4483's CI log with no
build and no gate run. Not "does my constant match upstream's literal" (which the unit test
already pins) but **does applying my transformation to rsvelte's real output produce official's
real output**:

```
official 2345-Constructor items: 4    rsvelte: 4
constant length: 162

rsvelte.message + HINT === official.message           4/4
digest(fixed rsvelte item) === digest(official item)  4/4

NEG (trailing space dropped) matches official          false   <- must be false
NEG (unfixed rsvelte)        matches official          false   <- must be false

paired items still differing on message AFTER the fix   0
```

The digest is the gate's own — `sha256(JSON.stringify(v)).slice(0,12)` from `diff.mjs` — so the
second row is not a paraphrase of the first: it is the whole diagnostic object hashing equal
under the comparison the ratchet actually performs, not just the message string being equal.

Both negative controls fire, and the first one is the discriminating one: dropping a single
trailing space from a 162-character constant breaks the match. That is the same byte the unit
test pins, now confirmed against the oracle's output rather than against my reading of upstream's
source.

One line in that output that I am **not** claiming: `item-identity multisets equal: false`. That
is true and has nothing to do with this change — on #4483's tree the lists still differ
(`extra 9 / missing 1`), so the full identity multisets cannot match. The relevant row is the
last one: of the items that *do* pair, zero still differ on message once the fix is applied.

This does not make the 0-key prediction for `main` any less falsifiable, and `Language server`
and `LSP fixture parity current` are still the checks that settle it. What it does mean is that
if those come back green, green is the *predicted* answer rather than an absence of evidence —
which is the distinction that would have caught the peer's 264.

baseballyama:
## Prediction settled: 0 keys, and the gate was live while it said so

Both decisive checks are green, and the gate's own summary is the useful part rather than the
check name:

```
[lsp-verify] 423 cases; 1956/1972 compared, 1301 divergent requests / 2078 divergent fields, 16 skipped
[lsp-verify] no regressions (2078 known divergences in measured scope)

LSP fixture parity current   SUCCESS
Language server              SUCCESS
```

Zero `NEW divergence` and zero `stale ratchet entry` lines in a 238 KB log carrying 31
`lsp-verify` lines — so the absence is a result and not a job that failed to run. The corrected
prediction was **0 ratchet keys**, against my retracted first guess of 8 re-keys, and 0 is what
it moved.

The run's own liveness is worth reading off the same output, because "nothing moved" and "nothing
was compared" are the failure pair this gate can produce: 1956 of 1972 units compared, 2078
divergent fields still being scored, and the oracle calibration above its floor on all three
suites (typescript-diagnostics 88/92, folding-range 13/15, inlay-hints 14/18). The official
server projected 365/423 components with svelte 5.57.0, resolved from a path the log prints. This
was not an empty measurement.

**What this does not establish** — restating it here because a comment is what gets quoted, and
the PR body now carries it too. Nothing in the ratchet separates *0 because the carrier is
unobservable until #4483 pairs those four diagnostics* from *0 because the change does nothing*.
Both print this identical green. So this confirms the PR is inert on the current tree, which is
exactly what it was predicted to be, and it is **not** evidence that the transformation is
correct.

The evidence for correctness remains the five-row table above: `rsvelte.message + HINT ===
official.message` 4/4, `digest(fixed) === digest(official)` 4/4 under the gate's own digest, and
two negative controls including the one where dropping a single trailing space from a
162-character constant breaks the match. Those were measured against the oracle's real output.
The gate-level check that would validate this arrives with #4483, since that is what makes the
carrier observable at all.

Remaining here: `Compiler parity` and `Formatter parity`, neither of which this change can reach.

Issue #4494, opened 2026-09-09
LSP gate sends no `javascript.inlayHints`, so the `expected:` inlay-hint oracle is unsatisfiable for components without `lang="ts"`


baseballyama:
## What

`scripts/compat-lsp/verify.mjs:292-311` builds the `initializationOptions.configuration` the gate
serves to both servers. It enables every `typescript.inlayHints.*` preference and sends **no
`javascript.inlayHints` key at all**.

Upstream picks the preference set by script kind, not by file extension
(`LSAndTSDocResolver.ts:339-343`):

```ts
const configLang =
    tsDoc.scriptKind === ts.ScriptKind.TS || tsDoc.scriptKind === ts.ScriptKind.TSX
        ? 'typescript'
        : 'javascript';
```

A `.svelte` component with no `lang="ts"` is JS, so it reads `javascript.inlayHints.*`, which the
gate never sends. `areInlayHintsEnabled` (`InlayHintProvider.ts:92-102`) is then false and
`getInlayHints` returns `null` at line 43 — before any hint is computed.

So for every non-TS component the official server answers `null`, and the `expected:` oracle
(upstream's committed `expectedv2.json`, recorded by upstream's own harness with hints on) can
never be satisfied — by either server.

## Measured

Run `34310065086`, artifact `lsp-current-fixtures`, `projectRevision` `6b9c092b4c88` (the merge ref
of #4488, whose `compatibility/lsp-known-failures.json` blob `36ae901df` is identical at the merge
ref, at its base and at `origin/main`, so `main` moving does not bear on this).

* The 18 upstream inlay-hint fixtures split **9 with `lang="ts"` and 9 without**:
  * non-TS: `$$props`, `$$restProps`, `$$slots`, `$store`, `action`, `animation`, `reactive-block`,
    `snippet.v5`, `transition`
  * TS: `component-handler`, `disable-argument-name-match`, `disable-type-name-match`, `each`,
    `element-handler`, `function-call-in-props`, `property-declaration`, `reactive-varable`,
    `union-type-props`
* With #4488 applied (rsvelte honours the client's preferences), **all 9 non-TS fixtures carry zero
  `differential:` inlayHint keys** — official and rsvelte agree — while **all 9 TS fixtures still
  carry live `differential:` keys**. Positive control: 170 `differential:` inlayHint keys remain in
  the run overall, so the instrument is live.
[...]

PR #4505, opened 2026-09-09
fix(lsp): drop the inlay hint on the generated `$$render` return type (#4464)

### LINKED ISSUE #4464: lsp: `textDocument/inlayHint` ports none of `InlayHintProvider`'s filters, and a range starting at 0:0 answers `[]`
`textDocument/inlayHint` is in `server.rs:711`'s generic `forward_tsgo_request` list, so
tsgo's raw hints are mapped back by the generic response mapper and returned unfiltered.
Upstream does not do that: `InlayHintProvider.ts:66-74` drops six classes of hint that only
exist because svelte2tsx generated them. rsvelte ports **none** of the six —
`command grep -rn 'inlay' crates/rsvelte_language_server/src/` returns exactly three lines
(`server.rs:247` capability, `server.rs:711` routing, `tsgo_client.rs:56` config) and
`tsgo_response.rs` has no inlay-hint handling at all.

There are two defects here and they point in opposite directions. Both are reachable on
`main` today.

## 1. Generated hints are not filtered

Measured over the protocol against both servers (`svelte-language-server` from
`submodules/language-tools`, and `target/release/rsvelte-language-server` built from
`main` at the time, `sha256` prefix `376ed9d15699f0bd`). Ten-line component:

```svelte
<script lang="ts">
	import Comp from './Comp.svelte';
	const n = 1;
	function take(howMany: number) { return howMany; }
	take(3);
</script>

<Comp a={n} />
<Comp />
```

Request range `5:0 - 9:0` — the markup only:

```
official (0): []
rsvelte  (7): [
  {"at":"0:1","label":": { : Record<string,never>; : {}; : string; : {}; : {} }","kind":1},
  {"at":"7:0","label":": __sveltets_2_IsomorphicComponent<{ ?: number | undefined },…>","kind":1},
  {"at":"7:0","label":"type:","kind":2},
  {"at":"7:0","label":"options:","kind":2},
  {"at":"8:0","label":": __sveltets_2_IsomorphicComponent<…>","kind":1},
  {"at":"8:0","label":"type:","kind":2},
  {"at":"8:0","label":"options:","kind":2}
]
```

Every one of the seven maps onto a rule upstream has:

| rsvelte hint | upstream rule that drops it |
|---|---|
| `0:1  ": { : Record<string,never>; … }"` | `inlayHint.position !== renderFunctionReturnTypeLocation` |
| `7:0  ": __sveltets_2_IsomorphicComponent<…>"` | `isGeneratedVariableTypeHint` — the declaration is `$$_pmoC0C`, and the rule drops a `Type` hint whose closest containing `VariableDeclaration` has a name starting `$$` |
| `7:0  "type:"`, `7:0  "options:"` | `isSvelte2tsxFunctionHints` — `expressionText.startsWith('$$_')`; these are the parameters of `new $$_pmoC0C({ target: …, props: … })` |

The label text is separately wrong: `{ ?: number \| undefined }` and
`{ : Record<string,never>; : {}; … }` have their property names elided, so even a hint that
survived filtering would render unusably. That is a third thing, not covered by the six rules.

## 2. A request whose range starts at `0:0` returns `[]`

Same file, same binaries, moving one endpoint at a time:

| range | official | rsvelte |
|---|---:|---:|
| `0:0 - 9:0` | 2 hints | **`[]`** |
| `0:0 - 8:8` | 2 hints | **`[]`** |
| `0:0 - 4:10` | 2 hints | **`[]`** |
| `0:1 - 9:0` | 2 hints | 9 hints |
| `0:18 - 9:0` | 2 hints | 9 hints |
| `1:0 - 4:10` | 2 hints | 3 hints |

The **start** alone decides it and the end is irrelevant; it holds for `lang="ts"` and for a
plain `<script>` alike. Upstream clamps an unmappable start instead of failing the request —
`convertToTargetTextSpan` is `generatedStartOffset.line < 0 ? 0 : snapshot.offsetAt(...)`.
rsvelte's `map_request` fails on the position and `forward_tsgo_request` answers with the
unmapped result.

An editor asks for the visible range, which normally starts at `0:0`, so **the practical
symptom today is that a Svelte file gets no inlay hints at all** — the spurious hints of §1
are what a client sees only when it asks for a sub-range. The two defects hide each other.

## Size, and why the LSP fixture ratchet understates it

`compatibility/lsp-known-failures.json` currently carries **314 `textDocument/inlayHint`
keys over 143 distinct files** — 130 `:extra-rsvelte`, 28 `:missing-rsvelte`, 58 both-lists
value mismatches, and 98 where official answered `null`. That is the standing cost of the
missing provider.

Two cautions for whoever picks this up.

`74234e98afe7` in a key is the digest of **`null`**, not of `[]`
(`sha256(JSON.stringify(v)).slice(0,12)`; `[]` is `4f53cda18c2b`). Official returns `null`
from `getInlayHints` when `areInlayHintsEnabled(userPreferences)` is false, so a key of that
shape is not evidence that official found no hints — and the gate's official arm answers
`null` on files where the probe above has it answering with hints, which means the gate's
configuration and a real client's are not the same. **Do not size §1 from the ratchet.**

And the gate sends `0:0` as its range start (`scripts/compat-lsp/suites.mjs:120`) while still
recording rsvelte lists, so the gate is not a witness for §2 either. The probe is.

## Where a fix goes

`isInGeneratedCode` is the one rule rsvelte partly has (the generic mapper drops a position
inside an `Ωignore` region). The other five need to identify the *enclosing call or
declaration* in the shadow, which upstream reads off a `ts.SourceFile` — rsvelte has no TS
AST for the shadow, since tsgo is a separate process. Two shapes worth pricing before coding:
record the synthesized call-argument and `$$_` declaration positions in `ShadowState` as the
overlay writes them (rsvelte emits those headers, so it need not re-derive them), or ask tsgo
for the enclosing node. A text scan over the shadow is the option to avoid; `AGENTS.md`
records several defects of exactly that shape.

§2 is small and independent: clamp an unmappable range start to the start of the document,
matching `convertToTargetTextSpan`.

## Repro

`zz-inlay-4097.mjs` — drives both servers over stdio, writes the workspace to a temp dir,
takes the range from `ZZ_RANGE=l,c,l,c` and the script flavour from `ZZ_PLAIN=1`. The
`take(howMany)` call in the script is a positive control: it must produce two hints on
official, and it did on every row of the table above, so a `0` from rsvelte is a measurement
rather than a dead probe.

baseballyama:
Fixes #4464 (in part — see *What this does not close*).

Upstream's `InlayHintProvider` filters tsgo's hints against the **generated** TSX before mapping
them back to the component, and one of its predicates is the return-type slot of the `$$render`
function svelte2tsx emits. rsvelte forwarded tsgo's hints unfiltered, so every `lang="ts"`
component reported a return-type hint on a function the user never wrote.

## Measured, two arms, one tree

One worktree, one official server, one `CARGO_TARGET_DIR`, so the only thing that differs between
the arms is the rsvelte binary.

```
                        raw removed   re-keys   REAL retired   REAL new
base arm vs ratchet         114         114          0            0
fix  arm vs ratchet         116          76         40            0
```

The base arm's `0 / 0` is the live control, and it is a stronger one than a hash comparison: an arm
that reproduces the checked-in ratchet exactly *is* today's `main` by the only property this
measurement depends on. The two binaries additionally differ by `sha256`, and the official arm is
byte-identical across both legs.

All 40 are `textDocument/inlayHint`; **0** keys move in any other method, in either direction.

**Carriers: 17, and all 17 are `lang="ts"`** — resolved by reading the suite roots out of
`suites.mjs` rather than guessing them, with `UNRESOLVED` printed as its own bucket. That bucket
read 14 under a guessed root before the roots were read from the harness, which is the live control
that it can be non-zero. The split is consistent with #4494: a component without `lang="ts"` gets
no hints from either server today, so it has nothing to filter.

## The re-baseline is a hand edit, and that is the right operation here

`verify.mjs:59` refuses `--update-baseline` outright — the baseline can only be *derived* from the
complete 17-artifact union that `merge-current.mjs` merges. Deleting 40 keys named by a measurement
is a different operation from re-deriving the file: it cannot enrol an entry that was never
measured, which is the failure mode that refusal exists to prevent. And the two-sided ratchet
checks it exactly — remove too few and the surplus reports as stale, too many and they report as
NEW — so CI is the oracle for the edit rather than my arithmetic.

Two controls on the edit itself:

* the writer was verified against the artifact before it was used. Re-serializing each file with no
  edit at all reproduces it **byte for byte**, including the `—` escape that plain
  `JSON.stringify` does not emit. Without that check the diff would have carried a silent
  one-character normalization in `$schema-note`.
* the diff is **0 added lines**: `0/40` in `lsp-known-failures.json` and `0/120` in
  `lsp-mechanisms.json`, the sidecar edited in the same commit because `lsp-mechanisms-check`
  requires a bijection. `node scripts/ci/lsp-mechanisms-check.mjs` exits 0 at 23728/23728.

## What this does not close

#4464 §1 lists six unported filters; this is one of them. #4494 is the reason the other five are
hard to measure today — half the upstream inlay-hint corpus emits no hints from either server
because the gate sends no `javascript.inlayHints`, so those fixtures are inert rather than passing.
Landing this before that is deliberate: the 40 above are measured under today's configuration, and
#4494 changes what the configuration measures.


baseballyama:
## Two commits added: a doc-count fix I owed, and a scan hazard I found auditing my own patch

Tip is now `bb5b28ab4`, current with `main` at `62410db75`.

### 1. The four red checks were one cause, and it was mine

`Compiler parity`, `Shape matrix parity (generated inputs)`, `Corpus verify baseline-flag
contract` and `LSP real-world parity (stable shard 1/16)` all failed at the **same step** —
"Verify known-failures docs match the JSON ratchets" — which several jobs run as a precondition.
Reading the four names suggests four problems; reading the failing *step* gives one. The 16 LSP
shards were aborting at step 3, so none of the ~950 job-minutes had been spent yet.

I deleted 40 keys from `lsp-known-failures.json` and left three gated numbers in
`compatibility/KNOWN-FAILURES.md` at their old values. Recomputed from the JSON rather than by
subtracting 40 from the prose:

| gated line | before | after |
|---|---|---|
| declared entries | 23768 | **23728** |
| partition by key kind | `21792 + 1646 + 330` | `21792 + 1612 + 324` |
| partition by request phase | `11889 + 11879` | `11869 + 11859` |

Cross-checks: the key-kind deltas are −34 `differential:` and −6 `expected:`, summing to the 40
retired, with the corpus addend unmoved — consistent with all 40 carriers being fixture/upstream.
The phase addends each drop 20, which is the 20 opened/edit pairs, and their difference stays 10 as
the prose asserts. `known-failures-md-check` failed before the edit and passes after
(`50 declared ratchets across 36 docs match their JSON; 34 cluster partitions add up`), which is
the two-sided evidence. `lsp-mechanisms-check` passes at 23728/23728; `attribution-check` is red on
`main` too (the standing DoD backlog, 3 problems in fmt/lsp/parse-ast) and its attributed count is
unchanged at 1228, so none of the 40 carried a target. The doc's "Ten entries sit under
`differential:fixtures/capabilities|initialize|`" still measures 10.

One census I ran here was worthless and is worth naming: "do any retired keys appear in the doc"
returned `102 of 40` — impossible, because the doc's bare `` `differential:` `` citation matches
every key three times. Its positive control had also returned 0. The checks above replace it.

### 2. `render_return_type_offset` was a raw first-occurrence scan

Auditing my own fix against the #2986/#2987/#2988 class. It located the generated header with
`text.find(";function $$render() {")`. svelte2tsx splices a hoisted props type **verbatim** into
`;type $$ComponentProps = …` *above* that header, so a string-literal type forges the needle ahead
of the real one. Measured on the real emitter rather than argued — one component,
`let { a }: { a: ";function $$render() {" } = $props()`, projects to:

```
///<reference types="svelte" />
;
;type $$ComponentProps =  { a: ";function $$render() {" };function $$render() {
```

2 occurrences; `render_return_type_offset` answered `Some(86)`, inside the user's string, where the
real header's slot is at 111. That both suppresses a legitimate hint at 86 and fails to suppress
the generated one.

The header is emitted **bare**, not inside `Ωignore` markers (`create_render_function.rs:95`), so
restricting the scan to generated regions is not available; and neither `find` nor `rfind` is right
in general, because a props type precedes the header and an instance-script string follows it. So
the fix declines when the needle is ambiguous, which degrades exactly to the behaviour before this
filter existed.

Reach: **0 of 33,170** corpus components carry the needle, control 6,813 carry `function ` — latent,
not shipped, and with no corpus gate either way, so the guard is a unit test. Three earlier
attempts at that census were wrong and none of them by a small margin: `$$` in a double-quoted
shell string expanded to the PID, then `$render` was the wrong name (`$$render`), then the same
grep run from this worktree read 77 instead of 6,813 because its submodules are unpopulated.

Controls on the test: it asserts the forgery is **still spliced** (count == 2) before asserting the
offset, so it cannot pass vacuously if svelte2tsx stops hoisting; ablating the two-line guard turns
it red with `left: Some(86) right: None` — the observed offset, not a generic failure — and
restoring leaves the file byte-identical. 27 tests in `tsgo_overlay::tests`, both the new test and
the pre-existing header test named in the output. The first run of this used a `render_return`
filter, which does not match the new test's name and silently ran 1 test; the denominator is the
module, not the guess.

Residue, stated rather than implied: when the needle is ambiguous the filter is simply lost, and
the principled fix is for svelte2tsx to report the offset instead of the overlay recovering it from
text. That is a change in another crate and is not in this PR.

baseballyama:
## The first re-baseline was half of one, and the half it missed was invisible to how I measured it

Tip `0ccad0f02`. `LSP fixture parity current` reported `38 NEW / 38 stale` on `bb5b28ab4`, and every pair is the same `file|method|pointer:shape` with `count=N` -> `count=N-1`. That is the fix working: I retired the 40 keys where the `$$render` return-type hint was a file's **only** extra hint, and missed the 38 where it was one of several, because there the key survives with a new digest.

It was missed for a specific reason worth naming. My local arm delta stripped the digest bracket before comparing sets — and stripping is exactly what makes a re-key indistinguishable from no change. It reported `RETIRED=42 NEW=0`, which is a true statement about stripped keys and says nothing about the population the gate compares.

Measured properly from the run's own `lsp-current-fixtures` artifact (`projectRevision 4b452aeb08`) against the ratchet blob at that same revision, calibrated by reproducing the gate's own `38 / 38` first:

| bucket | n |
|---|---|
| re-keyed | **38** |
| genuinely new | 0 |
| genuinely retired | 0 |

The log prints only 30 of each (`verify.mjs:1105/1117`, `slice(0, SHOW)`, no truncation marker on that path), so 8 of each were not in it and the listing could not have been used as the input.

Mechanism sets are taken from the artifact rather than carried across, and that gives an independent corroboration of the fix rather than just bookkeeping: **10 of the 38 move `inlay-hint-set-extra-mixed` -> `inlay-hint-set-extra-parameter`**. Dropping a *Type* hint from a file whose remaining extras are all *Parameter* hints is precisely what removing the render-return-type slot does, and the classifier reaches that from the payload without knowing anything about the patch.

Totals are unchanged at 23728 (38 out, 38 in, all `differential:`, 19 per phase), so the gated declaration and both partitions stay put; `known-failures-md-check` and `lsp-mechanisms-check` both pass.

**Pre-registered:** the next run of `LSP fixture parity current` must report `0 NEW` and `0 stale`. Simulating the new ratchet against the same artifact gives exactly that, and the artifact is the arm this fix produced — so if the run disagrees, the disagreement is about the tree, not about this arithmetic.

Issue #4512, opened 2026-09-09
A non-destructured `$props()` declaration with an inline comment hoists that comment and drops what follows it (client)


baseballyama:
## Symptom

On the **client** only, a comment written after a `$props()` declaration on the
declaration's own line is deleted from the output — but only when the declaration
also carries a comment of its own. Both trailing kinds are lost.

```svelte
<script>
	let p = /* c */ $props(); // a b
	let q = 1;
	console.log(q);
</script>
<i>{p.a}</i>
```

official (client) keeps both; rsvelte emits `/* c */` and drops `// a b`.

## Measured

Four cells, `client` and `server`, counting each comment's occurrences in the
output. official first, then `main` at `458b132ea` (post-#4509):

| cell | target | official | main |
|---|---|---|---|
| props comment + trailing `//` | client | `// a b`=1 | **0** |
| props comment + trailing `//` | server | `// a b`=1 | 1 |
| props comment + trailing `/* */` | client | `/* t */`=1 | **0** |
| props comment + trailing `/* */` | server | `/* t */`=1 | 1 |
| **no** props comment + trailing `//` | client | `// a b`=1 | 1 |
| props comment, trailing on the **next** line | client | `// a b`=1 | 1 |

So it is a two-term conjunction — a comment inside the declaration **and** a
comment after it on the same line — and it is client-only. The last two rows are
the controls that say which term is which: removing either one restores the
comment.

`/* c */` itself is emitted once in every rsvelte row, so this is not the
re-emission path failing; the declaration's own comment survives and its
neighbour does not.

## Cause (not isolated)

The client rewrites the declaration as a whole **unit**, and the unit is the
line. #4503 was the same shape with a *statement* after the `;`, and #4506's
guard requires the `;` to be the last **code** byte on the line — a trailing
comment is not a code byte, so the guard passes and the rewrite takes the
comment with it. That attribution is read off the guard and the four cells above,
not isolated with an arm of its own.

## Reach

**0 carriers** in the collected corpus. Scanning 32,666 `.svelte` files for a
line containing `$props(` with a comment on both sides of it: 4 lines carry a
leading comment only, 1 carries a trailing comment only, **0 carry both** — the
two non-zero counts are the live controls saying the scan works. So this is a
unit-test-only defect today, in the same way #3026's shape was.

## Not fixed by #4510

Measured on that branch's arm (`7338cafd252c5efa`): every row above is unchanged.
[...]

PR #4513, opened 2026-09-09
fix(client): end a `$props()` declaration at its code `;` and count comments instead of testing substrings (#4510)

### LINKED ISSUE #4510: A `$props()` comment is dropped by two raw scans whose errors cancel (`;`-in-comment truncation + a substring re-emission gate)
## Symptom

Two of the four raw text scans in the `$props()` comment path are wrong, in opposite
directions, and **the error of one is what hides the error of the other**. Repairing
either alone measures worse than repairing neither.

### (a) The `;` scan truncates the region at a `;` written inside a comment

`props_declaration_comments` (`client/mod.rs:3264`) ends the declaration's region at
`raw[props..].find(';')`. When a comment near the declaration contains a `;`, the
region stops inside that comment and the extracted "comment" is a **fragment**:

```
source   // Header chrome is store-editable theme content; the literals below are only the
extracted// Header chrome is store-editable theme content;
```

### (b) The re-emission gate is a substring test

`client/mod.rs:1459` decides whether the declaration's comment still needs emitting
with `transformed_script.contains(comment.as_str())`. A comment whose text also occurs
inside a string, a template literal, a longer comment — **or which is a prefix of the
comment it was truncated from** — reads as "already emitted" and is suppressed.

So (b) suppresses (a)'s fragments, and today's output is right on those files for the
wrong reason.

## Measured, three arms against the oracle

Base `f47de20348511bc1`; corpus 33,578 files x {client, client-dev} = 67,156 cells,
63,324 live; stage 2 compares each moved unit's raw output to official.

| arm | change | MOVED | match -> MISMATCH | MISMATCH -> match |
|---|---|---:|---:|---:|
| `a33a5cdf078732de` | fix (b) only: compare against the script's real comments | 32 | **19** | 0 |
| `040e1a34b91ccb62` | fix (b) and make the `;` scan code-aware | 4* | **4** | 0 |
| `b8084573760a5c89` | fix the `$props()` anchor scan only (#4493) | 0 | 0 | 0 |

\* the (b)+(`;`) arm's 4 are a different set: threlte's `Svg` and `ContactShadows` on
both client targets. Every `;` in those two instance scripts sits inside a string or a
template literal, so a code-aware `;` scan finds **none**, the region runs to the end of
the script, and unrelated comments are swept into the declaration — each file gained a
stray `// svelte-ignore non_reactive_update`. That attribution is derived from the two
files' bytes plus the two scans' definitions, not isolated with an arm of its own.

Constructed cells that (b) alone fixes, and that this issue keeps as its acceptance
criteria (counting `/* c */` in client output, official first):

| cell | official | base |
|---|---:|---:|
| `let p = /* c */ $props();` + `const s = "/* c */";` | 2 | 1 |
| `let p = /* c */ $props();` + `` const s = `x/* c */y`; `` | 2 | 1 |
| `let p = /* c */ $props();` + `let b = 1; // see /* c */ below` | 2 | 1 |
| `let p = /* c */ $props();` + `let b = 1; /* c */` | 2 | 1 |
| `let p = /* c */ $props();` + `let b = 1; /* d */` (control) | 1,1 | 1,1 |

The fourth row is not a decoy at all — the other occurrence is a real comment with the
same text — so the gate's premise is wrong on ordinary code too. The question it needs
to answer is whether **this** occurrence survived the transform, which is positional;
the props comment already carries a source offset.

## What a fix has to do

Both together, in one change:

- end the region at the declaration statement's end — the first **code** `;`, or the end
  of the line carrying `$props(` when there is none (ASI), which bounds it in both
  semicolon styles;
- replace the substring presence test with one that identifies the occurrence.

## What no gate can see

Both outputs parse and are AST-equivalent, and the corpus output gate's
`ast_equiv_batch` runs with `CommentPolicy::Ignore`, so a comment-only divergence is
rescued on both sides. The 19 and 4 above are **raw-byte** stage-2 counts, which are
stricter than the gate: they are regression *candidates*, and the reason they are
reported as such is that a stricter reconstruction's non-zero is a candidate list, not
a finding. What is not a candidate is the direction — 0 units improve under either
partial repair.

## Related

- #4493 / #4509 (the anchor scan; fixed separately, and the only one of the four that is not entangled — it moves 0 corpus cells)
- #4448 (the comment's slot is wrong in every one of these cells, independently of all four scans)
- #4501 (`trimmed.is_empty()` suppresses this whole re-emission loop)

baseballyama:
Closes #4510.

## The two scans

`props_declaration_comments` ended the `$props()` declaration's region at the
first **raw** `;`. A `;` written inside the declaration's own comment therefore
truncated it, and the extracted "comment" was an unterminated prefix — which
`main` emits:

```svelte
<script>
	let p = /* c */ $props(/* ; */);
	let q = 1;
	console.log(q);
</script>
<i>{p.a}</i>
```

```js
	$.push($$props, true);
	/* c */
	/* ;
	let p = $.rest_props($$props, rest_excludes);
```

acorn: `Unterminated comment (10:1)`. official emits `/* ; */` intact. Two
characters, and **zero carriers in the collected corpus** — the corpus never
holding it is the argument for the cell, not a caveat about it.

The re-emission gate asked `transformed_script.contains(comment)`, so a comment
whose text also occurs in a string, a template literal, a longer comment, or a
second time as its own comment read as "already emitted". A truncated fragment
is also a substring of the comment it came from, which is why the two scans had
to be repaired together: **fixing either alone measures worse than fixing
neither** (#4510's table records 19 and 4 regressing units for the two partial
arms).

## The fix

- the region ends at whichever comes first of the first **code** `;` and the end
  of the line carrying `$props(`. A code-aware scan alone finds none in a
  semicolon-free source and then runs to the end of the script — threlte's `Svg`
  and `ContactShadows` are that shape, their only `;` being inside a template
  literal;
- presence is a **count over comment tokens** rather than a substring test:
  re-emit exactly the occurrences the transform dropped.

## Measured

Corpus, 33,578 files × {client, client-dev} = 67,156 cells, 63,324 live, key
sets identical, base `b8084573760a5c89` (= `main`) against `7338cafd252c5efa`:

```
MOVED = 214 over 107 files

MISMATCH -> match     169
match -> MISMATCH       0
MISMATCH -> MISMATCH   45     (other mechanisms, unrelated to this change)
```

Oracle is `submodules/svelte` at this branch's own pin, `7bc0a70fe` /
`VERSION 5.57.0`. The first run of this comparison used a checkout parked
thirteen commits back, whose pin is `56a036f4c` / `5.56.10`; the two oracles
disagree on **48 of the 214** moved units, which moved the improvement count
(136 → 169) and left the regression count at 0 either way. Corrected below in a
comment rather than silently.

Raw bytes, which is **stricter** than the gate (`ast_equiv_batch` runs with
`CommentPolicy::Ignore`), so the zero on the regressing side is free: a zero
under a stricter comparison is a zero under the gate. Both output ratchets
(`known-failures.client.json`, `known-failures.client-dev.json`) are empty, so
there is nothing to re-baseline.

Generated shape matrix run locally on the same arm: 7,714 cases, 30,424
comparisons, `match 26096 / error-parity 4328`, every other verdict **0**.

## Controls

Nine cells, every expected value taken by compiling that exact cell through
`submodules/svelte`'s source compiler rather than inferred from a neighbour.
Ablating each half reddens a **disjoint** set — the raw `;` scan fails the two
truncation cells, the substring gate the four decoy cells — and restoring is
byte-identical.

Three arms with distinct hashes: `b8084573760a5c89` base, `d6054abf11068502`
gate-only, `7338cafd252c5efa` gate+region. The first four cells written could not
separate the last two; the `$props(/* ; */)` cell is what does, and it reports
`/* ; */` × 0 on the gate-only arm against × 1 on the shipping one.

The doc comment records which cells fail under neither ablation and why, so the
two negative controls and the semicolon-free cell are not read as coverage they
do not have.

## Not in scope

The `$props(); // a b` shape — a comment after the declaration on its own line
being deleted on the client when the declaration also carries one — is a
different stage and is filed as #4512, measured unchanged on this branch's arm.

### DIFF
--- a/.changeset/props-comment-scan-pair.md
+++ b/.changeset/props-comment-scan-pair.md
@@ -1,0 +1,5 @@
+---
+'@rsvelte/compiler': patch
+---
+
+Keep a `$props()` declaration's comment whose text also occurs in a string, a template literal or another comment, and stop truncating the declaration at a `;` written inside its own comment — which emitted an unterminated `/*` fragment
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -99,8 +99,8 @@ use std::sync::LazyLock;
 
 use crate::compiler::phases::phase1_parse::parser::is_js_whitespace;
 use crate::compiler::phases::phase3_transform::shared::js_scan::{
-    after_keyword, after_keywords, code_bytes, contains_identifier, find_code, find_rune_code,
-    is_ident_byte, skip_opaque,
+    after_keyword, after_keywords, code_bytes, code_bytes_from, contains_identifier, find_code,
+    find_rune_code, is_ident_byte, skip_opaque,
 };
 use crate::compiler::phases::phase3_transform::shared::rune_shadow;
 use crate::compiler::phases::phase3_transform::shared::substring::Substring;
@@ -1455,8 +1455,20 @@ pub(crate) fn transform_client(
                 content.start + (text.len() - text.trim_start().len()) as u32
             });
         if !trimmed.is_empty() {
+            // A comment's text also occurs in a string, inside a longer comment,
+            // and sometimes a second time as its own comment, so presence is a
+            // count of comment tokens rather than a substring test: re-emit
+            // exactly the occurrences the transform dropped.
+            let mut live = script_comment_texts(&transformed_script);
+            let source_comments = script_comment_texts(&content.raw);
             for (offset, comment) in &props_comments {
-                if !transformed_script.contains(comment.as_str()) {
+                let kept = live.iter().filter(|text| *text == comment.as_str()).count();
+                let had = source_comments
+                    .iter()
+                    .filter(|text| *text == comment.as_str())
+                    .count();
+                if kept < had {
+                    live.push(comment.to_string());
                     component_body.push(JsStatement::RawMapped {
                         code: comment.clone(),
                         source_offset: content.start + *offset,
@@ -3259,6 +3271,14 @@ fn script_raw_statement(
     }
 }
 
+/// The text of every comment token in a script, in source order.
+fn script_comment_texts(script: &str) -> Vec<String> {
+    crate::compiler::phases::phase3_transform::server::transform_script::extract_comments_from_snippet_with_pos(script)
+        .into_iter()
+        .map(|(_, comment)| comment)
+        .collect()
+}
+
 /// Comments inside the `$props()` declaration survive upstream's lowering even
 /// though the declaration itself is removed from the component body.
 fn props_declaration_comments(raw: &str) -> Vec<(u32, CompactString)> {
@@ -3268,14 +3288,17 @@ fn props_declaration_comments(raw: &str) -> Vec<(u32, CompactString)> {
     let Some(props) = find_rune_code(bytes, b"$props(") else {
         return Vec::new();
     };
-    // The `;` scan stays raw: a code-aware one finds none in a semicolon-free
-    // source and the region then runs to the end of the script.
     let Some(start) = raw[..props].rfind_sub("let") else {
         return Vec::new();
     };
-    let end = raw[props..]
-        .find(';')
-        .map_or(raw.len(), |end| props + end + 1);
+    // Whichever bound comes first: a raw `;` scan stops inside a comment that
+    // contains one, and a code-aware scan finds none in a semicolon-free source
+    // and then runs to the end of the script.
+    let semi = code_bytes_from(bytes, props)
+        .find(|&(_, byte)| byte == b';')
+        .map_or(raw.len(), |(at, _)| at + 1);
+    let line_end = raw[props..].find('\n').map_or(raw.len(), |at| props + at);
+    let end = semi.min(line_end);
     crate::compiler::phases::phase3_transform::server::transform_script::extract_comments_from_snippet_with_pos(&raw[start..end])
         .into_iter()
         .map(|(offset, comment)| ((start + offset) as u32, comment.into()))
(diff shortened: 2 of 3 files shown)


baseballyama:
## Correction: the first stage-2 run used the wrong oracle

The commit message on this branch says `136 MISMATCH -> match` and `78 still
diverging`. Both numbers were measured against **`submodules/svelte` at
`56a036f4c` / `VERSION 5.56.10`**, which is the pin of a checkout parked thirteen
commits behind `main`, not this branch's pin. Re-run against the branch's own
oracle — `7bc0a70fe` / `VERSION 5.57.0`, the tree `git ls-tree origin/main
submodules/svelte` names:

```
oracle 5.56.10 vs 5.57.0 differ on 48 of 214 moved units

MISMATCH -> match     169     (was reported as 136)
match -> MISMATCH       0     (unchanged)
still diverging        45     (was reported as 78)
```

The load-bearing claim survives: **0 units regress**, under either oracle. What
moved is the credit side — the change fixes 169 of the 214 units it touches, not
136 — and the residue, which is 45 units carrying mechanisms unrelated to this
change.

The nine test cells were re-derived against the correct oracle too and **every
expected value is unchanged**, so nothing in `props_comment_scan_pair_4510.rs`
encodes 5.56.10's behaviour. The generated shape matrix was run inside the branch
worktree, which carries the correct pin, so its `0` verdicts stand as reported.

### How it happened, since the mechanism is reusable

Corpus *sources* have to be read from a checkout whose submodules are populated,
and this branch's worktree has 102 of 104 empty — so the sweep ran from the
primary checkout. The *oracle* is a submodule too, and it came along with it.
`AGENTS.md` records three things answering to "the official compiler" and a
fourth axis, the submodule pin; this is that fourth axis, reached by a rule about
where to read *inputs* silently deciding where the *oracle* came from.

`VERSION` did separate them here (5.56.10 against 5.57.0) and I did not read it.
What surfaced it was an impossible combination rather than a suspicion: a
`$.child` / `$.only_child` divergence on a file that is a manifest entry, while
`known-failures.client.json` is `[]` — a byte divergence the gate cannot be green
on. Checking `ls-tree` took one command after that.

The source trees have the same problem in principle — the primary checkout's
`layerchart`, `shadcn-svelte` and `svelte.dev` are modified relative to its pin —
but both arms of this sweep read the identical bytes, so `MOVED` and the
direction are unaffected; only the oracle leg was wrong, and only for the 48
units where the two compilers differ.

PR #4520, opened 2026-09-09
fix(language-server): apply InlayHintProvider's generated-code filters

baseballyama:
Closes part of #4464. #4505 landed one of the nine `InlayHintProvider` filters; this adds five
more — the four AST pre-filters and `isInGeneratedCode` — into the seam #4505 already created.

### Where the filters run

Upstream applies six pre-filters at `InlayHintProvider.ts:69-74` on **shadow** offsets and three
post-filters at `:86-88` on original-document positions. This PR is the pre-filter half, in
upstream's order, on shadow offsets. Positions are mapped afterwards, as upstream does.

`ShadowNodes` parses the shadow once per request with oxc — not once per hint, matching upstream's
single `SourceFile` — and answers the four structural questions. A shadow oxc rejects yields **no
tree rather than an empty one**, so a filter never runs against a tree it could not build:
upstream's `SourceFile` is best-effort where oxc is not, and an empty tree would silently answer
"not generated" for everything.

### Two things that had to be measured, not read

**`getTypeAnnotationPosition` is one field, not a token scan.** Upstream is
`findChildOfKind(decl, CloseParenToken)?.end ?? decl.parameters.end` and oxc exposes no token
stream. It does not need one — `FormalParameters.span.end` *is* the close paren's end where there
is one, and is the `parameters.end` fallback for a bare-identifier arrow. 11 arrow shapes on the
pinned oxc 0.146, all equal, ablated both ways.

**The outermost/innermost distinction is unreachable for these four predicates.** My first test for
it passed — *and so did its ablation to "greatest start"*, because in the cell I wrote only one
node ever contained the offset. It is replaced by a property test (`findClosestContainingNode`'s
result is always inside `findContainingNode`'s, over 154 matched offsets) and a shortcut-vs-faithful
comparison over 912 pairs with 0 disagreements. The doc comment records that the two rules cannot
be told apart here, which is a different claim from the one the passing test looked like it made.

### Measurement

Both arms run locally so the local-vs-CI digest offset cancels; these are before/after deltas.

```
inlayHint units diverging   212 -> 152      hints in excess of official   420 -> 154  (-63.3%)
  fully retired  60      count grew     0   <- must be 0
  count reduced  36      newly present  0   <- must be 0
  unchanged     116
```

Raw keys: 1976 -> 1916, 96 retired / 36 appeared / 0 re-keys, every one `textDocument/inlayHint`
`extra-rsvelte`. **Zero `missing-rsvelte` on any method** — that is the over-filter control: a
filter that dropped a hint official emits would appear there and nowhere else.

The 36 "appeared" keys are the same units re-keyed because their `count=` fell, which is why the
numbers above are retired *units* rather than moved digests; under a set difference a re-key is
indistinguishable from a removal plus an addition.

The two arms' reports agree on every population and oracle field — `oracleCalibration` byte-identical
at 115/125 with the same 10 named misses, `compared` 1956, `skipped` 16, `transportTimeouts` 0 —
and differ only in `divergent` and `divergentFields`.

### Verified on the tree this actually lands on

The work was developed before #4511 settled, so the delta was re-derived against #4511's head
rather than carried across a merge commit. On top of `fd6eee4b1` it is exactly **5 files,
677+/17-**: `tsgo_inlay_hints.rs` (new), `tsgo_overlay.rs` (+48), `tsgo_response.rs` (+28),
`lib.rs` (+1), `Cargo.toml` (+5 -- it pulls `oxc_allocator`/`ast`/`ast_visit`/`parser`/`span`, so
the lockfile moves).

The 17 deletions are **not** a revert of #4511's consolidation. `render_return_type_offset` moves
from above `occurrences` to below `is_in_generated_code`; checked as a multiset, removed lines not
present among added lines = **0**, and the function's body hashes identically in both trees
(`d46874f2cb0a`, with sha1-of-empty printed beside it as the control, because the first attempt at
that check hit a shell modifier and returned the empty hash from *both* arms).

`cargo check -p rsvelte_language_server --all-targets` exits 0 and
`cargo test -p rsvelte_language_server --lib` reports **396 passed, 0 failed, 1 ignored of 397**,
with all 8 new tests named in the output rather than inferred from a count.

### Ratchet, and why this PR needs a dispatch rather than a green CI

`compatibility/lsp-known-failures.json` is byte-identical across this branch, #4511 and `main`
(blob `e362a9da9`, 23,728 entries) -- the 40-entry shrink in the history came from #4505 and was
inherited, so **this change's own effect on the ratchet is unmeasured, not zero**.

It cannot be measured by this PR's CI. `corpus-compat-job-filter.mjs:157` derives `lsp-ratchet`
from a diff touching `lsp-known-failures*.json` or `scripts/compat-lsp/**`, and this PR touches
neither, so `lsp-ratchet=false` and `lsp-corpus` does not run -- confirmed on the filter and
observed live, the same job going SKIPPED on #4511. Meanwhile `verify.mjs:1090-1091` scopes the
ratchet's two directions differently: `added` against the full ratchet, `removed` against the run's
own suites. So a fixture-only run correctly declines to call corpus entries stale, and nothing else
evaluates them either until this is on `main`. That matters here specifically, because this change
*retires* divergences, and retiring is what makes a listed entry stale.

The sequencing is therefore: dispatch `corpus-compat.yml` on this branch for the full 17-artifact
set, re-baseline from it, and commit the result -- which flips `lsp-ratchet` true and makes this
PR's own `lsp-corpus` run verify the re-baseline. Recorded as GATES row 36.

### Still open on #4464

The three post-filters. 7 and 8 (`line >= 0`, `character >= 0`) are upstream's sentinel handling for
a failed mapping and are satisfied structurally here (the mapper drops unmappable positions;
`lsp_types::Position` is unsigned). Filter 9, `checkGeneratedFunctionHintWithSource`
(`InlayHintProvider.ts:323`), is not written, so #4464 stays open. Read rather than assumed, it is
three branches and each needs something this PR does not have: a position in `moduleScriptInfo` is
always kept; a position in `scriptInfo` is kept only if the **source** text from that offset
`trimStart()`s to `$:`; and everything else goes through `getAttributeContextAtPosition`, keeping a
hint only for a `name` containing `:` that is an `on:` with no matching attribute on the element, or
one of `in`/`out`/`animate`/`transition`/`use`. So it needs source-document script-tag ranges, the
source text at a mapped position, and element/attribute context — none of which the shadow-offset
seam this PR builds can supply, which is why it is a separate change rather than a sixth filter
here.


baseballyama:
## `LSP fixture parity current` is red, and the direction is measured: 0 regressions, 188 spurious hints removed

The failure is the expected re-key, not a regression — but "30 NEW divergences" cannot say that on its own, because **the count is inside the ratchet key**, so lowering a spurious-hint count retires one key and creates another. Paired properly, from the run's own artifact:

| | entries | spurious rsvelte-only hints |
|---|---|---|
| re-keyed with a **lower** count | 30 | 204 → 72 (**−132**) |
| **retired outright** (no current key at all) | 20 | 56 → 0 (**−56**) |
| re-keyed with a **higher** count | **0** | — |
| brand-new base key | **0** | — |

**188 spurious hints removed across 20 fixture files; nothing moved the wrong way.** `30 re-keyed + 20 retired = 50 stale`, which is the arithmetic the run reports.

Method, so it can be rechecked: `added = current ∖ known` against the **full** ratchet (what `verify.mjs:1090` does), then each added key's `[count=N,hash=…]` bracket stripped to a base key and paired with its committed sibling. The pairing is 1:1 — 0 added keys have more than one known sibling — and every one of the 30 matched a sibling, which is the positive control that the bracket regex fired at all (had it not, all 30 would have read `BRAND-NEW`).

Artifact `lsp-current-fixtures` from run `34363022891`, `projectRevision ff5153546a1c8a6688dd8fdce3773bb9739e490b` — the merge ref, so this is measured on this branch merged with `main`, not on the branch alone.

### One thing worth flagging about the log

The job log declares `50 stale ratchet entry/entries:` and then **prints 30 of them**. A reader working from the log alone sees 30 NEW against 30 printed stale, concludes a clean 1:1 re-key, and misses all 20 retirements — which are the better half of the result. The cap is a display choice and correct as such; it is only a hazard because the number beside it is the real one. The artifact is the population; the log is a reading of it.

### Why this cannot be fixed by re-baselining here

`verify.mjs:59` refuses a direct `--update-baseline`:

> `direct --update-baseline is disabled; merge the complete fixture artifact and 16 corpus artifacts with merge-current.mjs --update-baseline`

and `lsp-corpus` is **skipped on this PR** — its `if:` requires a `schedule`/`workflow_dispatch` event or `lsp-ratchet == true`, and `lsp-ratchet` keys on a diff touching `compatibility/lsp-known-failures*.json` or `scripts/compat-lsp/**`. This PR changes LSP *behaviour* and neither of those paths, so the gate that would observe it is exactly the one that does not run. That asymmetry is now written up as GATES row 36 in #4519.

The sequence, which needs capacity this account does not currently have (job census at the time of writing: **20 in-progress against the 20 ceiling, 4 queued**):

1. `gh workflow run corpus-compat.yml --ref lsp/4464-inlay-generated-code`
2. merge the resulting 17 artifacts with `merge-current.mjs --update-baseline`
3. commit the re-baselined ratchet — which flips `lsp-ratchet` true, so **this PR's own `lsp-corpus` then runs and verifies the re-baseline** rather than taking it on trust

Holding until the queue drains rather than dispatching 16 hour-long shards into a saturated pool that my own two open PRs are already waiting in.

baseballyama:
### The three reds are one cause, and **no test failed**

`Language server` and `Tests` joined `LSP fixture parity current`. They are the same defect, not three:

- **`Language server`** — 33 steps, one failure: step 26, *"Run LSP differential fixture and upstream gates"*. That is the same gate as `LSP fixture parity current`, run from `ci.yml` instead of `corpus-compat.yml`. Steps 1–25 all `success`.
- **`Tests`** — a rollup that ran for **8 seconds** and whose only failing step is *"Verify all test jobs passed"*. Its own log:

  ```
  BULK: success
  UNIT: success
  LANGUAGE_SERVER: failure
  ##[error]Test job(s) FAILED: language-server.
  ```

So the Rust suites both pass and `Tests` is red purely because `Language server` is in its `needs:`. Enumerating the run's jobs confirms it — exactly two non-success, non-skipped jobs, and one of them is the rollup itself.

Worth stating plainly because the check *name* is misleading here: **"Tests" being red is not evidence that a test broke.** A rollup reports its inputs, and reading the name rather than the step env would have sent someone looking for a regression in the 8 new `tsgo_inlay_hints.rs` tests, which pass.

All three clear together when the ratchet is re-baselined; none of them is a second problem.

baseballyama:
Drive-by diagnosis from rsvelte-70, not a takeover — the branch is still checked out in your worktree and I have not touched it.

All three reds (`Language server`, `LSP fixture parity current`, and the `Tests` roll-up, which only reports `language-server`) are **one** cause, and it is the ratchet, not the change:

```
[lsp-verify] 50 stale ratchet entry/entries:
  differential:upstream-features/component-handler/input.svelte|textDocument/inlayHint|:extra-rsvelte[count=14,hash=434eb454e6ad]
  ...
  expected:upstream-features/component-handler/input.svelte|textDocument/inlayHint|:extra-rsvelte[count=6,hash=b18a4dc5c897]
```

The key carries `count=` and a hash, so applying the filters re-keys every `inlayHint` entry it touches: `component-handler` goes `count=14 -> count=6`, `each` `2 -> 1`, `element-handler` `5 -> 1`, `function-call-in-props` `4 -> 2`. The old keys are then stale and the new ones are new, and the ratchet is two-sided, so both halves fail CI. That is the change working.

Per `AGENTS.md`, the PR that moves entries re-baselines in the same PR, and the re-baseline is a measurement of a **tree** — so rebase onto `main` *before* `--update-baseline`, not after. `main` has moved a fair way since `9ff66b2d5` (2026-09-09): #4573, #4574, #4576 landed today and all three touch client codegen, though none of them touches the projection.

Two things worth checking while re-keying, both from the repo's own rules rather than from anything I measured here:

- The three buckets when you diff the baseline are **re-keyed / genuinely new / retired**, and a re-key is not a fix. Stripping the `[count=…,hash=…]` bracket from both sides before diffing separates them; otherwise 50 stale + N new reads as a much bigger move than it is.
- `count=14 -> 6` and `count=5 -> 1` are the filters removing hints. Worth one cell confirming the *remaining* ones are the ones upstream keeps, rather than only that fewer survive — a filter that over-rejects shrinks the count the same way.

I have not run the LSP suite, so none of the above is measured — it is read off the job log at `runs/34363022891` and `runs/34363022896`.

If you have moved on from this branch, say so (or just release the worktree) and I will rebase it, re-baseline, and take it to merge.

baseballyama:
## Re-baselined from a full-population run on this branch

The two-sided ratchet requires the PR that fixes entries to re-baseline in the same PR, and
`merge-current.mjs` refuses anything short of the complete 17-artifact union — so this comes from
a `corpus-compat` **workflow_dispatch on this branch**
([run 34662510788](https://github.com/baseballyama/rsvelte/actions/runs/34662510788)), which is
the path the workflow's own comment prescribes. All 16 `lsp-corpus` shards concluded `success`;
every one of the 17 artifacts carries `projectRevision 2d68837e7`, `configurationId lsp-diff-v12`
and `language-tools 2cfcc15b4`, so the union is one arm rather than a mixture.

```
[lsp-merge] 23708 current, 30 new, 50 stale     <- before --update-baseline
[lsp-merge] 23708 current,  0 new,  0 stale     <- after
```

**The ratchet moves by exactly -20, and every one of the 80 diff lines is
`textDocument/inlayHint`** — checked as `grep -vc textDocument/inlayHint` over the added and
removed lines, which is `0`. The 30 new keys are old ones re-keyed downward, which is what a
filter that removes hints looks like through a `count=`-bearing key:

| file | before | after |
|---|---|---|
| `upstream-features/component-handler` | `count=14` | `count=6` |
| `upstream-features/element-handler` | `count=5` | `count=1` |
| `upstream-testfiles/completions/importcompletions-text` | `count=10` | `count=1` |
| `upstream-features/function-call-in-props` | `count=4` | `count=2` |
| `upstream-features/reactive-varable` | `count=2` | *(retires)* |

Declared counts in `KNOWN-FAILURES.md` are re-derived from the JSON rather than adjusted by
-20: total `23708`, key kind `21792 + 1594 + 322` (aggregate / differential / expected), request
phase `11859 + 11849`. `known-failures-md-check` passes.

## One thing in this diff is not this PR's

`lsp-mechanisms.json` shrinks by ~6,900 label references. That is **not** the inlay-hint change:
6,451 of the 23,678 entries shared between the two versions carry different mechanism labels, and
the ones I sampled are `completion-item-set-missing-html-close-tag` and
`projection-origin-range` on **corpus** cases — nothing an inlay-hint filter touches. The sidecar
can only be rewritten by a full-population run, and those happen on schedule or dispatch, so it
has been carrying labels from before a number of merged commits. This dispatch is the first run to
observe them.

Flagging it rather than letting it read as part of the fix: the `-20` on the ratchet is the part
this PR is responsible for, and it is the part that is fully attributed above.

Issue #4311, opened 2026-09-04
fmt: `value_sequence` builds `RawExpr` with `src: None`, so its build-time `broken_width` is never re-checked against the real print indent (52% of the break population)


baseballyama:
Second of three quantities in the `Doc::RawExpr` break branch (see #4309 and #4310). This one is not a disagreement between two values — it is a value nothing ever re-checks.

`markup/value_sequence.rs:159-163` builds its interpolations as `Doc::Group([Doc::RawExpr { flat, broken, src: None }])`. The `broken` lines come from `format_attribute_value_expression_at_width(inner, &opts_sq, broken_width)`, where `broken_width` is derived at doc-build time from the builder's own running column estimate `col`.

Because `src` is `None`, the printer's rebuild is unreachable (`doc.rs:185-190` needs `src.as_deref()`), so:

- `fits` and the printer read the **same** value — the fits/print split of #4309 does not exist on this path;
- and the build-time `broken_width` is compared against the real print indent **never**.

`collapse/doc_build.rs:290` is the other port and always sets `src: Some`, so this is a property of `value_sequence` alone.

## Exposure

Same run as #4309 — `compatibility/sources`, 33,911 `.svelte` files (`-type f`), one `rsvelte-fmt .` under `scripts/fixtures/fmt-corpus.oxfmtrc.json` with `--oxfmt-bin`, 28,739 files rewritten.

| `Doc::RawExpr` print branch | n |
|---|---:|
| `flat-mode` | 17,859 |
| `unbreakable` | 75 |
| **break, `src: None` (this issue)** | **1,444** |
| break, `src: Some` | 1,310 |

**1,444 of the 2,754 break prints — 52.4% — are on the path that cannot re-check.** The break total 2,754 has now reproduced across three separate instrumented runs.

`src: Some` rebuilds never failed (0 of 1,310), so the earlier run's `norebuild` bucket was entirely this port. That zero is a derivation, not a clean measurement: the marker used fires on the `src: None` path too, so its count is `srcNone + srcSome-failures`, and it equalled `srcNone` exactly.

## Symptom, with the reason it is not yet a defect

Instrumenting the printed lines against the real indent (first line from the live `pos`, continuations from `unit.columns() * ind`):

| port | break prints | prints with >=1 line over the print width | over-long lines | worst excess |
|---|---:|---:|---:|---:|
| `src: None` | 1,444 | **471 (32.6%)** | 571 | 630 cols |
[...]

Issue #4309, opened 2026-09-04
fmt: `fits` charges a breakable `RawExpr` by the doc-build `broken[0]` while the printer rebuilds it, and `fits` has no `ind` to rebuild with


baseballyama:
`doc.rs`'s printer and its `fits` measurement disagree about what a breakable `Doc::RawExpr` costs, and only one of them is allowed to ask the real question.

`RawExprSource`'s own doc comment states the rule (`doc.rs:17-20`):

> The broken form's break points depend on the columns left after the printed line's indent, which no doc-build caller knows, so the printer rebuilds them from this.

The **printer** does exactly that (`doc.rs:185-190`): it calls `broken_lines(s, width - unit.columns() * ind)` and falls back to the stored `broken` only when `src` is `None` or the rebuild returns `None`.

**`fits` does not** (`doc.rs:334`). It charges a breakable `RawExpr` by `broken[0]` — the vector built at doc-build time, which `collapse/doc_build.rs:288` produces at `usize::MAX`, i.e. the *column-unaware* shape its own comment calls "all a doc builder can know". And `fits` could not rebuild even if it wanted to: its signature is `fn fits(remaining: isize, rest_stack: &[Cmd], next: &[Doc], unit: IndentUnit)` — **there is no `ind`**, so the budget the printer uses is not a quantity `fits` has.

So the value that decides whether a group breaks and the value that is printed are two different strings.

## Measured

`compatibility/sources`, **33,911** `.svelte` files (`-type f`; a plain `find -name '*.svelte'` returns 33,914 because three *directories* carry that suffix), staged outside the repo and formatted by one `rsvelte-fmt .` invocation under the gate's own `scripts/fixtures/fmt-corpus.oxfmtrc.json` with `--oxfmt-bin`. Liveness: **28,739 files were rewritten**, so the run reached the corpus rather than reporting a zero from an empty population.

`Doc::RawExpr` print events, by branch:

| branch | n |
|---|---:|
| `flat-mode` | 17,859 |
| `unbreakable` (`broken.len() <= 1`) | 75 |
| break, `src: None` (`markup/value_sequence.rs:161`) | 1,444 |
| break, `src: Some` (`collapse/doc_build.rs:290`) | 1,310 |
| **total** | **20,688** |

Of the 1,310 that rebuild, the rebuilt first line differs from the stored one in **19**:

| | n |
|---|---:|
| rebuilt first line == stored | 1,288 |
[...]

PR #4539, opened 2026-09-10
fix(fmt): a broken content mustache is rebuilt at the column it starts and against what follows it

### LINKED ISSUE #4310: fmt: the print-time `RawExpr` rebuild picks a break the oracle does not, and the strongest carrier of the `fits` staleness is actually this
Third of three quantities that live in the same `Doc::RawExpr` break branch and are easy to mistake for each other. This one is the least measured and it is the one a carrier file points at first.

When the printer breaks a `RawExpr` that carries a `src`, it rebuilds the broken form at `width - unit.columns() * ind` (`doc.rs:185-190`) and prints those lines. That budget is computed correctly from the real indent — the issue is that the break it produces is not the break the oracle produces.

## The carrier

`adventurelog/frontend/src/routes/profile/[uuid]/+page.svelte`, a listed `fmt-known-failures.json` entry. Its first differing line against the oracle is at 955-958, which is the interpolation itself:

```
oracle
                                Best happened at {categoryData.record_holders
                                  .max_distance.location_name ||
                                  "this location"}

rsvelte
                                Best happened at {categoryData.record_holders.max_distance
                                  .location_name || "this location"}
```

Instrumented at print time, the same site records:

```
ind=16  stored="{categoryData.record_holders.max_distance.location_name ||"   (58 cols)
        rebuilt="{categoryData.record_holders.max_distance"                   (41 cols)
```

So rsvelte printed the **rebuilt** first line, 41 columns. The printer did its job: it rebuilt at the real indent, and it emitted what the rebuild produced. The oracle breaks one level deeper again, at 28 columns.

**This entry is therefore not the `fits` staleness issue** (#4309). It was the strongest-looking carrier of that mechanism — it is one of the 19 events where stored and rebuilt disagree — and reading the mechanism off the carrier gets it wrong: fixing `fits` moves nothing here, because the value `fits` mis-charged is not the value that was printed.

## What is not measured

**The population.** One carrier is one carrier. How many of the 1,310 `src: Some` break prints disagree with the oracle about *where* to break is unknown, and it cannot be read off the numbers already collected:

- the 19 stored-vs-rebuilt divergences measure `fits` against the printer, not the printer against the oracle;
- an overflow check on the printed lines cannot see this either, because the rsvelte break above does not overflow — it is a legal layout that the oracle simply does not choose.

Measuring it needs the oracle in the loop: for each `src: Some` break print, compare the emitted lines against `compatibility/fmt/oracle`'s text at the same site. Locating "the same site" is the hard half, and it must not be done by brace-matching the mustache out of the source — a `{` inside a string literal makes that instrument report spans that are not expressions.

## Why it is filed with n=1

An issue that says "population unmeasured" is what the next person measuring the population reads first. Left only in a message it is invisible, and the mechanism is one that carrier evidence actively misattributes — which is the reason to write it down before anyone re-derives the wrong owner from the same file.

## The set of three

1. #4309 — `fits` reads the doc-build `broken[0]` while the printer rebuilds. Measured: 19 of 1,310, direction uniform (overcharge, 10-64 columns). Output effect unmeasured.
2. `markup/value_sequence.rs:161` builds with `src: None`, so the printer *cannot* rebuild and the build-time `broken_width` is never re-checked against the real indent. Exposure 1,444 of 2,754.
3. This one: the rebuild's own budget produces a break the oracle does not choose. Carrier 1, population unmeasured.

They share a code path and separate cleanly by which value is wrong: the one `fits` read, the one nobody re-checked, or the one that was correctly rebuilt and still disagrees.

baseballyama:
Closes #4310.

## Mechanism

When the printer breaks a content-level mustache (`Doc::RawExpr` with a source), it rebuilt the broken form at `width − indent`, the budget of its continuation lines, and printed that. The first line was therefore laid out as if it started at the indent, and the last as if nothing followed the `}`. prettier measures every JS group of a mustache in place: the first against the column the `{` sits at, after the prose before it on the line, and the last against the closing brace and any text glued to it, up to the next break opportunity. So on the issue's carrier the oracle prints

```
Best happened at {categoryData.record_holders
  .max_distance.location_name ||
  "this location"}
```

(`.record_holders` is the last member that fits at column 49) where rsvelte kept `.max_distance` on the first line and overflowed it.

## Fix

- `broken_lines_at` takes the first-line offset (`pos − indent + {prefix`) and the last-line suffix. The suffix is `}` plus `rest_to_break`, the columns the pending print commands occupy before their first line break in the modes they were pushed with — `fits` is now that measurement as a predicate (`consumed_to_break`), so the two cannot disagree.
- The forced-break cap is `offset + flat + suffix − 1`, one column short of the whole flat line, instead of `flat − 1` of the continuation width — the old cap squashed the first line below its own offset once the offset was charged.
- The same measurement exposed two approximations in `format_expr_core_layout` that had been invisible while nothing charged small values: an offset or suffix too narrow for its placeholder comment (`/**/` needs 5 and 6 columns) was charged by narrowing the width, i.e. to **every** line. `({visit.activities` lost a column on its first line to the `)` behind it, and a call's argument line lost one to the `{` before the call. Both are charged exactly now: the unoffset (unsuffixed) layout stands whenever its first (last) line still has room, because only a measurement starting on (reaching) that line sees the charge; otherwise the narrowest comment stands in. The #4119 unit test that had pinned the narrowing form at offset 3 is updated to the exact form.

The `fits` side of this family is #4309 (stacked on this PR); `value_sequence`'s `src: None` port is #4311.

## Measured

Two arms over all **33,644** corpus components through the fmt gate's own pipeline (`fmt.mjs --actual` + `fmt-verify.mjs`, oracle tree at `svelte@56a036f4c` / `svelte.dev@fa53d1446` / oxfmt 0.64.0; base arm = #4535's binary, sha256 `c2329df4…`, fix arm `1b212ee2…`):

| | n |
|---|---|
| outputs moved | 33 |
| `MISMATCH -> match` (retired) | **27** |
| `match -> MISMATCH` (regressed) | **0** |
| still mismatching, moved | 6 (differing lines 11→6, 12→4, 8→4, 7→3, 7→3, 8→4) |
| moved while matching in both arms | 0 |

All 27 are `breaks-later` (rsvelte kept more of the mustache on its first line than the oracle); by diff shape 20 placement, 3 line-breaks-only, 4 token. Three intermediate arms were measured on the way: the first (offset + suffix, old cap) retired 12 and regressed 31, the second (new cap) retired 26 and regressed 4, the third (exact narrow suffix) 27 / 1, and the exact narrow offset closes the last. Each regression class was opened rather than tallied — the four `({visit` cells and the call-argument cell above are the ones that named the two approximations. The pre-existing local `NEW` on `svelte.dev/…/docs/[topic]/[...path]/+layout.svelte` and the stale `print/samples/comment/output.svelte` are properties of the main checkout's corpus pin and appear identically on both arms.

Cells: `crates/rsvelte_formatter/tests/broken_mustache_in_place_4310.rs` — 8 oracle-generated cells: the carrier (`{` at column 49), the same expression at a shallow indent (breaks *later*, the other direction), text glued to `}`, `({visit.activities` ending exactly at the width and one column over, a call whose argument line ends exactly at the width, and two controls that must not move; on the base arm 3 of the first 5 diverge, here 8/8, and the oracle's output is a fixed point for all 8. `expression/tests.rs` pins the narrow-offset and narrow-suffix charges on their own.

## Ratchet

`fmt-known-failures.json` 480 → 453 (the 27 above, all previously `unattributed` in `fmt-mechanisms.json`); `KNOWN-FAILURES.md` count, cluster / mechanism / diff-shape partitions and a paragraph for the retirement. `known-failures-md-check`, `attribution-check --gate-known` and `check-core-consumer-changesets` pass locally.

Repro: `compatibility/pattern-corpus/issues/broken-mustache-in-place.svelte` (+ `.md`), committed in the oracle's form; it diverges on `main` and matches here.

https://claude.ai/code/session_01YQAqKJWyJNQss6cDCTnt8U


baseballyama:
`Formatter parity` is red for the *good* reason — the two-sided ratchet, not a regression.

```
included  33839
matched   33362
failed      452
❌ 1 baseline entries already PASS — the ratchet is stale.
  - layerchart/docs/src/examples/components/Chart/facet-click.svelte
```

Zero new failures; this fix made one listed entry pass, so the baseline has to be re-cut in this same PR:

    node scripts/compat-corpus/fmt-verify.mjs --update-baseline

The entry lives at `compatibility/fmt-known-failures.json:182` and `compatibility/fmt-mechanisms.json:206` (`"unattributed"`), and `known-failures-md-check.mjs` gates the declared count and the partition lines in `KNOWN-FAILURES.md`, so those move too.

Separately, the run logs a NOTICE (not a failure) that an excluded id now matches the oracle: `svelte/packages/svelte/tests/runtime-legacy/samples/block-expression-assign/main.svelte`.

Issue #4516, opened 2026-09-09
A trailing comment on the last instance-script statement is emitted after the lowered call, not inside it


baseballyama:
The comment trailing the last statement of an instance script belongs **inside**
the call that statement lowers to; rsvelte closes the call and puts it after the
`;`. 23 `(id, target)` pairs over 13 corpus files.

```svelte
<script>
	let a = $state(1); // c
</script>
<C bind:value={a} />
```

```js
// official (client)          // rsvelte
let a = $.state(1 // c        let a = $.state(1); // c
);
```

`client-dev` is the same one line deeper: official `let a = $.tag($.state(1 // c`
⏎ `), 'a');`, rsvelte `let a = $.tag($.state(1), 'a'); // c`.

## Both arms

Oracle is the recorded submodule pin `7bc0a70fe` (`VERSION 5.57.0`) via
`submodules/svelte/packages/svelte/src/compiler/index.js`, the source path the gates
use. rsvelte arm is `origin/main` `b3747891b`, `sha256 0efd4c92d23b93cb`.

## Eight cells, six live controls

| cell | template | trailing? | verdict |
|---|---|---|---|
| `comp_bind` | `<C bind:value={a} />` | `// c` on the declaration line | **DIFF** |
| `comp_prop` | `<C value={a} />` | same | EQ |
| `comp_bare` | `<C />` | same | EQ |
| `comp_plainlet` | `<C value={a} />`, `let a = 1` (no rune) | same | EQ |
| `CTL_element` | `<p>{a}</p>` | same | EQ |
| `CTL_no_comment` | `<C value={a} />` | — | EQ |
| `CTL_no_template` | *(none)* | same | EQ |
| `CTL_own_line` | `<C value={a} />` | comment on its **own** line | DIFF — a *different* mechanism, see below |

Every EQ row carries the comment through on both sides, so they are live. Both
targets agree on all eight.

## The enabling condition is plural, and this is one sufficient form

`bind:` on a component is sufficient and is not necessary. The upstream fixture
`runtime-runes/samples/effect-tracking-unowned/main.svelte` reproduces it with no
component at all — a bare read statement:

```svelte
<script>
	let n = $state(1);
	let t = $derived(n + 1);
	t; // read derived in a non-tracking context
</script>
<p>{t}</p>
```

```js
// official                                              // rsvelte
$.get(t // read derived in a non-tracking context        $.get(t); // read derived …
);
```

[...]

Issue #4500, opened 2026-09-09
A block comment flushed before the template root always gets a newline; upstream keeps it inline when </script> and the first element share a line


baseballyama:
### What

A block comment in the instance script that is flushed before the template-root
declaration is **always** followed by a newline in rsvelte's client output.
Upstream emits it **inline** when no source newline separates the comment's end
from the template root's first element.

```svelte
<script>/* c */ let { a } = $props();</script><b>{a}</b>
```

```js
// official                       // rsvelte
var /* c */ b = root();           var /* c */
                                  b = root();
```

### Both arms

Oracle `submodules/svelte/packages/svelte/src/compiler/index.js` at the recorded
pin `7bc0a70fe` (`VERSION 5.57.0`; the checked-out submodule matches the pin in
both the primary checkout and the measuring worktree, so the arm is identified by
the pin rather than by the version string). `generate: 'client'`, `dev: false`.
Verdict is full `js.code` byte equality.

### The axis is the TEMPLATE's source line, not the script's

Crossing leading newlines inside the script against a newline before the first
template element — a falsifiable prediction of the mechanism, run after it was
written down:

| `nlScript` | `nlTpl` | official | rsvelte |
|---:|---:|---|---|
| 0 | 0 | INLINE | NEWLINE — **DIFF** |
| 0 | 1 | NEWLINE | NEWLINE — EQ |
| 1 | 0 | INLINE | NEWLINE — **DIFF** |
| 1 | 1 | NEWLINE | NEWLINE — EQ |

4/4. Upstream's decision is independent of everything inside the script and is
decided by `</script>` vs `<b>` sharing a source line. That is a plain
`has_newline_between(comment.end, template_root.start)` in **source**
coordinates.

### Why rsvelte cannot currently take that branch

`rsvelte_esrap`'s `flush_comments_until` (`printer.rs:1188`) asks
`!cmt.block || self.has_newline_between(cmt.end, to)`. Instrumented over a 4x5
grid (comment position x script context): **`has_newline_between` is `true` on
19 of 19 block-comment flushes**, in EQ and DIFF cells alike, so the `else if pad`
space branch is dead.

The cause is that `to` is a **comment-space** offset. `parse_raw_statements`
(`to_oxc.rs:2035-2036`) pushes each chunk's text into `synth.source` followed by
`'\n'`, so the comment always sits at the end of its buffer line and
`has_newline_between(cmt.end, to)` is structurally `true`. It carries no
[...]

PR #4559, opened 2026-09-10
fix(client): the comment coordinate boundary clears the source, so a source offset cannot read as comment space

### LINKED ISSUE #4521: loc_base is not an upper bound on source offsets, so a real source position can sit inside a comment-space range
### What

`loc_base` is meant to be the boundary between comment-space and source coordinates — every
real source offset below it, every synth-buffer offset above. It is computed as
`synth.max_span + 2` (`to_oxc.rs:198`), where `max_span` is the running max of whatever the
probe pass happens to call `note_span` on: chunk **text lengths** (`:2011`, `:2020`), some
statement ends, and some `RawMapped` source offsets.

Nothing makes that an upper bound on the source offsets the printer will actually resolve. A
template element's identifier span is one that is not noted, so on a file whose template is
larger than its largest chunk, a genuine source offset sits **above** `loc_base` and is
indistinguishable from a synth-buffer offset by the only test either consumer has.

### Measured

`submodules/svelte-eslint-parser/tests/fixtures/parser/ast/svelte5/ts-event03-type-output.svelte`.
Source offset **375** is `<button` at line 8 column 1. The client map maps that element's
identifier and, on `main`, reports 8:1 — correct, and correct *by accident*: `map_position`'s
`loc_map` lookup misses and falls through `None => offset`, so the identity is returned.

#4466's fix registers the comment-buffer regions in `loc_map`. Those regions begin at
`synth.cursor()`, which is at least `loc_base - 1`, so on this file a range **contains 375** and
the offset is translated as if it were comment space: `s8:1 -> s12:42`, inside an unrelated
comment. Three segments move that way while two others are repaired to official's exact
positions, so the file nets −1 of 9 identifier-judged segments in a change that is +846 across
the corpus.

The guard at `:202` — `if synth.max_span >= loc_base { note_fallback("loc-base"); return None; }`
— cannot see it, because it re-reads the same under-counting `max_span`.

### The obvious repair costs comment fidelity, which is why this is its own issue

Variant measured (`sha256 4c4e3814f3ae1832`): have `open_source_region_parts` and
`open_island_region` `note_span` their region's **source end** before the `enabled` gate, so the
probe pass raises `loc_base` above them. On the 73 real-world components whose map moves:

| arm | identifier-correct | out-of-range | files better/worse |
|---|---:|---:|---:|
| `main` | 12,965 | 732 | — |
| #4466 fix (`loc_map` only) | 13,811 | 12 | 69 / 1 |
| + `note_span` the region end | 13,814 | 12 | **70 / 0** |

Strictly better on the map, and it is **not shippable as a source-map change**, because raising
`loc_base` moves `has_loc`, which is what drives comment flushing. Over 33,623 real-world
components the `loc_map`-only fix moves `js.code` on **0**; the `note_span` variant moves it on
**1** — the same fixture — and the direction is wrong:

```
official     emits 2 comments on that `var`
main         emits 4
note_span    emits 0
```

So the boundary repair is a comment-cursor change wearing a source-map hat, and it belongs with
#4489 / #4492 / #4500 / #4516 / #4517 rather than in front of them.

### Why `max_span` cannot be patched locally

Adding the missing `note_span` calls one at a time reproduces the enumeration hazard the repo
already records: the list would come from the offsets someone hit, not from the offsets the
printer can resolve. The quantity that is actually an upper bound is the **source length**, and
`program_to_oxc_with_islands` does not receive it — it takes a `JsProgram`, an arena, an
allocator and the islands. Threading it in is a signature change with its own blast radius, and
it is the only version of this that closes rather than narrows.

### Controls

- The `js.code` sweep is two-armed over 33,623 live components with an independently verified
  arm pair (`f788b5a09c2c5cd7` vs `d7b83710b0de76cb` vs `4c4e3814f3ae1832`, distinct `sha256`).
- Its first run reported `js.map MOVED = 0` for every file, which was a fabricated zero:
  the NAPI `compile` returns `js.map` as an **object**, and the harness hashed `String(map)`,
  i.e. `"[object Object]"` for every input. Caught by a positive control (three files whose maps
  are known to move) plus a negative one (a file known not to), not by re-reading the code.
- `sourcemaps_gate` is green on the shipped half: 0 out-of-range of 1,622, 58/58 byte-identical
  outputs, 808/808 official segments, 0 known failures.

baseballyama:
Closes #4521.

## What was wrong

`loc_base` is the boundary between the two coordinate spaces the client printer works in — below it an
offset is a real source position carrying no comment-space location, at or above it an offset indexes
the synthetic comment buffer. It was `synth.max_span + 2`, and `max_span` is the running max of
whatever the probe pass happens to `note_span`: chunk text lengths, some statement ends, some
`RawMapped` source offsets. Nothing made that an upper bound on the source offsets the printer
actually resolves, so on a file whose template outgrew its largest chunk a genuine source offset sat
**above** `loc_base` and was indistinguishable from a buffer offset by the only test either consumer
has (`has_loc`).

The guard at `to_oxc.rs` could not see it because it re-reads the same under-counting `max_span`.

## The fix

The quantity that genuinely bounds every source offset is the **source length**, and
`program_to_oxc_with_islands` did not receive it. It does now, and `loc_base` is
`max(max_span, source_len) + 2`. `program_to_oxc`, `print_module_program` (the `compileModule` path,
both client and server) and the devtools sink thread it through; the synthetic-program unit tests pass
`0`, which is what "this program has no original source" means.

Adding the missing `note_span` calls one at a time was the alternative and is the enumeration hazard
this repo already records — the list would come from the offsets someone hit, not from the offsets the
printer can resolve.

## Measured, two arms

Arms are distinct binaries (`sha256 048d8c2e5603…` base, `bec6ddc9579a…` head) and are separated
behaviourally, not by label: an env-gated probe present in **both** builds prints `source_len=` only on
the head arm (`loc_base=56 max_span=54` vs `loc_base=57 max_span=54 source_len=55`). The probe is not
in the shipped diff.

Population: 34,930 real-world `.svelte` files under `submodules/`, `generate: client`, `dev: false`.
`rows == distinct keys == 34,930` on both arms, 1,269 throw identically on both. `js.map` is an
**object** across NAPI, so it is hashed as `JSON.stringify`, with the runtime type printed
(`TYPES code=string map=object`) before any hash is believed.

### `js.code` — one file moves, and it moves toward official

`submodules/svelte-eslint-parser/tests/fixtures/parser/ast/svelte5/ts-event03-type-output.svelte`,
the file #4521 names. Differing lines against official (`submodules/svelte` at the pinned
`7bc0a70fe`, `VERSION 5.57.0`): **18 → 11**. The region official writes as

```js
	var button = $.first_child(fragment);

	var // e: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement; }
	// e.currentTarget: EventTarget & HTMLButtonElement
	input = $.sibling(button, 2);
```

is now byte-identical. On `main` all four comments are parked on `button` — the wrong declarator — and
`var button` and `var input` are in the wrong order. The head arm drops two of the four (official puts
them inside the `$.delegated('input', …)` arrow, which is #4481's shape); the two it keeps are on the
node official keeps them on. So the arm with *fewer* comments is the one that places them correctly,
which a count-keyed comparison would have scored the other way.

### `js.map` — 1,016 files move; segments paired by generated position

`js.code` is byte-identical between the arms on 1,015 of those, so segments pair exactly. Verdicts are
oracle-free properties of (map, source, generated text): a segment is `ID-CORRECT` when the generated
position starts an identifier and the original position starts the same identifier, `OUT-OF-RANGE`
when the original column exceeds that source line's length.

```
paired-segments=460655  MOVED=3381
  1912  OUT-OF-RANGE -> OUT-OF-RANGE
  1070  NEW-IN-HEAD
   926  non-identifier-site -> non-identifier-site
   256  id-wrong -> ID-CORRECT
   169  non-identifier-site -> OUT-OF-RANGE
   117  id-wrong -> id-wrong
     1  ID-CORRECT -> ID-CORRECT
```

**256 repaired, 0 regressed** on the identifier verdict.

### The 169, with their raw values

Out-of-range rises by 169, and that number is a leak becoming visible rather than a leak being created.
Every one of them is a comment-space offset that reaches the map untranslated (`map_position`'s
`None => offset` fallback); raising `loc_base` moves where the untranslated value lands. Printing the
underlying values rather than tallying the bucket:

```
gen 150:6  "\t\t\t\t\t\t}"   base -> src 112:0   "        }}>"
gen 648:2  "\t}"             base -> src 1400:23 "ansform: uppercase;"     <- a CSS line
gen  14:1  "};"              base -> src 15:3    "</script>"
gen  39:2  "\t}"             base -> src 103:1   "    .popup-title {"      <- a CSS line
gen 124:3  "\t\t\t},"        base -> src 169:14  "h: fit-content;"         <- a CSS line
gen  51:2  "\t\t},"          base -> src 20:17   inside a template literal
```

The base positions are wrong in every case — they are in range by accident, pointing at CSS or template
text with no relation to the generated brace. The head positions are past the source end and therefore
countable. Fixing the fallback is #4466's subject, not this one; the measurement above is posted there.

### What this does NOT move

The five open sibling comment defects, run through both arms and against official:

| cell | base == official | head == official | base == head |
|---|---|---|---|
| #4500 `<script>/* c */ let { a } = $props();</script><b>{a}</b>` | false | false | **true** |
| #4448 `let p = /* c */ $props();` | false | false | **true** |
| #4516 `let a = $state(1); // c` + `<C bind:value={a} />` | false | false | **true** |
| #4481 trailing `// x` + `<svelte:window>` | false | false | **true** |
| #4489 trailing `// x` + top-level `{#snippet}` | false | false | **true** |

All five are live divergences under both arms, so this is a negative result with its own liveness
control, not a silent zero. This PR fixes the boundary and nothing downstream of it.

## Pin

`a_declared_source_length_lifts_the_comment_boundary_over_every_source_offset` in `to_oxc.rs`'s test
module: a program with a noted identifier span at 1,000 and a declared source length of 1,200 must have
`loc_base > 1_200`. Negative control — deleting `.max(source_len)` fails **that test and only that
test**, with its sibling
`generated_identifier_source_span_does_not_raise_comment_boundary` (source length `0`) still passing,
so the pin reads the new invariant rather than the module.

### DIFF
--- a/.changeset/loc-base-source-upper-bound.md
+++ b/.changeset/loc-base-source-upper-bound.md
@@ -1,0 +1,28 @@
+---
+"@rsvelte/compiler": patch
+---
+
+client: the comment coordinate boundary clears the source, so a source offset cannot read as comment space
+
+`loc_base` separates the two coordinate spaces the client printer works in:
+below it, an offset is a real source position with no comment-space location;
+at or above it, an offset indexes the synthetic comment buffer. It was computed
+as `synth.max_span + 2`, and `max_span` is the running max of whatever the probe
+pass happened to note — chunk text lengths, some statement ends, some
+`RawMapped` source offsets. Nothing made that an upper bound on the source
+offsets the printer actually resolves, so on a file whose template outgrew its
+largest chunk a genuine source offset sat above `loc_base` and was
+indistinguishable from a buffer offset by the only test either consumer has
+(#4521).
+
+The quantity that *is* an upper bound is the source length, and
+`program_to_oxc_with_islands` now takes it. `has_loc` therefore means what it
+says, and the printer stops treating a template identifier's position as a
+comment-space one.
+
+Over 34,930 real-world components, `js.code` moves on one file and moves toward
+official (18 differing lines to 11; the `var … = $.sibling(…)` region becomes
+byte-identical, where before four comments were parked on the wrong declarator).
+Pairing map segments by generated position across 1,016 files whose maps move:
+256 segments go from pointing at the wrong identifier to the right one, and none
+goes the other way.
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/client/mod.rs
@@ -403,7 +403,7 @@ pub fn transform_client_module(
         }
     }
 
-    print_module_program(body, &header)
+    print_module_program(body, &header, source.len() as u32)
 }
 
 /// Print a `.svelte.(js|ts)` module body the way upstream's `client_module` /
@@ -434,6 +434,7 @@ fn expand_module_inspect_holes(code: String) -> String {
 pub(crate) fn print_module_program(
     body: Vec<JsStatement>,
     header: &str,
+    source_len: u32,
 ) -> Result<String, TransformError> {
     use super::js_ast::codegen::generate;
 
@@ -443,8 +444,8 @@ pub(crate) fn print_module_program(
     };
     let arena = super::js_ast::arena::JsArena::new();
     let alloc = oxc_allocator::Allocator::default();
-    if let Some(code) =
-        super::js_ast::to_oxc::program_to_oxc(&program, &arena, &alloc).map(|converted| {
+    if let Some(code) = super::js_ast::to_oxc::program_to_oxc(&program, &arena, &alloc, source_len)
+        .map(|converted| {
             let print_opts = rsvelte_esrap::PrintOptions::default().with_unlocated_program(true);
             match &converted.comment_source {
                 Some(cs) => {
@@ -2807,6 +2808,7 @@ pub(crate) fn transform_client(
                     &context.arena,
                     &alloc,
                     &ast_islands,
+                    source.len() as u32,
                 )
             })
             .map(|converted| {
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/codegen.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/codegen.rs
@@ -3560,7 +3560,7 @@ mod tests {
 
         let handwritten = generate(&prog, &arena).unwrap();
         let allocator = oxc_allocator::Allocator::default();
-        let converted = super::super::to_oxc::program_to_oxc(&prog, &arena, &allocator).unwrap();
+        let converted = super::super::to_oxc::program_to_oxc(&prog, &arena, &allocator, 0).unwrap();
         let esrap = rsvelte_esrap::print(&converted.program, "");
 
         for (printer, code) in [("handwritten", handwritten), ("oxc/esrap", esrap)] {
--- a/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/to_oxc.rs
+++ b/crates/rsvelte_core/src/compiler/phases/3_transform/js_ast/to_oxc.rs
@@ -179,23 +179,32 @@ pub fn program_to_oxc<'a>(
     program: &JsProgram,
     arena: &JsArena,
     allocator: &'a oxc_allocator::Allocator,
+    source_len: u32,
 ) -> Option<Converted<'a>> {
-    program_to_oxc_with_islands(program, arena, allocator, &[])
+    program_to_oxc_with_islands(program, arena, allocator, &[], source_len)
 }
 
 /// Convert an IR program while inserting retained source ASTs directly.
+/// `source_len` is the byte length of the original file this program was
+/// generated from, and it is the only quantity here that is genuinely an upper
+/// bound on the source offsets the printer can be asked to resolve. `max_span`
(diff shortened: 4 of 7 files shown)


baseballyama:
Closing unmerged: the change is measured and correct in principle (1 `js.code` file moves and moves toward official, 256 map segments repaired / 0 regressed over 34,930 components), but it regresses `a_script_comment_maps_before_the_template_expression_it_precedes`, which is oracle-confirmed. On that shape `max_span=41` against a 104-byte source, so `main`'s `loc_base=43` puts every source offset above the boundary — and that misreading is what currently anchors the comment inside `$.set_text(…)`. Raising the boundary removes the anchor with the misreading.

Full measurement, including the 0-of-4,756 carrier count for the regression and its positive control, is on https://github.com/baseballyama/rsvelte/issues/4521#issuecomment-5619855318. The branch `fix/4521-loc-base` stays pushed for whoever lands this together with the source-space anchor.

Issue #4448, opened 2026-09-08
A comment before `$props()` in a non-destructured declaration is emitted leading, not trailing


baseballyama:
A comment between `=` and `$props()` in a **non-destructured** props declaration is emitted on the line before the declaration; upstream emits it trailing the rewritten declaration.

### Repro

```svelte
<script>
	let p = /* c */ $props();
</script>
<i>{p.a}</i>
```

### Both arms

Oracle is the recorded submodule pin `7bc0a70fe` (`VERSION 5.57.0`), via `submodules/svelte/packages/svelte/src/compiler/index.js` — the source path the gates use. `generate: 'client'`.

```js
// official
export default function R($$anchor, $$props) {
	$.push($$props, true);

	let p = $.rest_props($$props, rest_excludes); /* c */
	var i = root();
	...
```

```js
// rsvelte
export default function R($$anchor, $$props) {
	$.push($$props, true);

	/* c */
	let p = $.rest_props($$props, rest_excludes);

	var i = root();
	...
```

Two differences: the comment moves from trailing to leading, and a blank line appears after the declaration.

### Why it was not already recorded

It is invisible to a count-keyed comparison — **both sides emit exactly one copy** — and it survived a 22-cell grid I built for #4398 because that grid keyed on the number of copies. It only appeared when the same cells were re-keyed on byte equality. It is also invisible to every corpus gate: `scripts/compat-corpus/verify.mjs:607-611` records that `ast_equiv_batch` is called with no arguments, so `CommentPolicy::Ignore` applies and a divergence living only in comments is byte-different, AST-equivalent, and scored a pass for every entry and every target.

### Not the same defect as #4398

#4398 is a comment inside a **destructured** pattern whose declaration the transform removes outright. Here the declaration is **rewritten** (`$.rest_props(...)`), the comment rides along in the replacement text, and what diverges is which side of the statement the printer puts it on. #4447 fixes #4398 and leaves this cell `DIFF` on both arms — deliberately, and the PR records it.

### Reach

Unmeasured. The 337-carrier census I ran for #4398 selects on `let {` … `$props()`, so a non-destructured declaration is outside its population by construction; I have not counted this shape's carriers and am not guessing at one.

baseballyama:
Reach measured, and it is **0 real-world carriers** — plus the axis turns out to have a sibling with a different symptom.

### The count

[...]

PR #4563, opened 2026-09-10
fix(client): a removed `$props()` declaration no longer takes its comments with it

### LINKED ISSUE #4501: A value-position comment on a destructured $props() is dropped when the script has no statement on another line (no TypeScript involved)
### What

A block comment in the **value position** of a destructured `$props()`
declaration is **dropped entirely** on the client when the instance script holds
no statement on any other line. No TypeScript is involved.

```svelte
<script>let { a } = /* c */ $props();</script><i>{a}</i>
```

```js
var /* c */                var i = root();
i = root();
```

`o=1 r=0`.

### Both arms

Oracle `submodules/svelte/packages/svelte/src/compiler/index.js` at the recorded
pin `7bc0a70fe` (`VERSION 5.57.0`; the checked-out submodule matches the pin in
the primary checkout and in the measuring worktree). `generate: 'client'`,
`dev: false`. `o` / `r` count the comment in `js.code`.

### The rule is a two-term conjunction, and both terms have live controls

| # | script body | rest? | stmt on another line? | r | |
|---|---|---|---|---:|---|
| 1 | `let { a } = /* c */ $props();` | | | 0 | **DROPPED** |
| 2 | `const s = 1;`⏎`let { a } = /* c */ $props();` | | ✓ | 1 | kept |
| 3 | `let { a } = /* c */ $props();`⏎`const t = 2;` | | ✓ | 1 | kept |
| 4 | both of the above | | ✓ | 1 | kept |
| 5 | ⏎`let { a } = /* c */ $props();` | | | 0 | **DROPPED** |
| 6 | `let { a } = /* c */ $props();`⏎ | | | 0 | **DROPPED** |
| 7 | `let { a } = /* c */ $props(); const t = 2;` | | | 0 | **DROPPED** |
| 8 | `let { a } = /* c */ $props(); let s = $state(0);` | | | 0 | **DROPPED** |
| 9 | `let { a, ...z } = /* c */ $props();` | ✓ | | 1 | kept |

**DROPPED iff there is no rest element and no statement on a line other than the
declaration's.** All nine cells agree.

Rows 5/6 and 7/8 are what make the terms separable: 5 and 6 add a newline with no
statement and still drop, so it is not "any newline"; 7 and 8 add a surviving
statement on the *same* line and still drop, so it is not "any statement". Row 9
holds everything fixed and adds a rest element, which is enough on its own.

### Two hypotheses this falsifies, since both are the obvious first guess

**`trimmed.is_empty()`** at `3_transform/client/mod.rs:1457`, which skips the
`props_declaration_comments` re-emission loop when the transformed script is
empty. Rows 7 and 8 leave a surviving statement and still drop, so the simple
form is wrong.

**"no newline after the comment"**, which fits rows 1/3/7 and is killed by rows 5
and 6. I had this written down as the axis before running them.

What survives both is line-structure-sensitive, which is the shape this repository
keeps finding in `content.raw` scans rather than in AST work.

### Reach: 0

Screen over the primary checkout's populated submodules, **35,723**
`.svelte`/`.svelte.(js|ts)` files. Control: 8,711 files mention `$props(`. The
screen's own regexes carry a two-sided control — 5 constructed positives match
and 3 constructed negatives (no comment / comment after / comment leading the
declaration) match neither.

| shape | files |
|---|---:|
| non-destructured + comment before `$props()` — this is **#4448**'s shape | 0 |
| destructured, no rest, + comment — **this issue** | 0 |
| destructured with rest, + comment | 0 |

The first row also fills the gap #4448 records as *"Reach: Unmeasured"*.

No corpus gate can hold any of the three: `ast_equiv_batch` is called with no
arguments, so `CommentPolicy::Ignore` applies and a comment-only divergence is
scored a pass. A unit test is the only possible guard.

### Relation to the neighbouring issues

**Not #4448**, which is non-destructured and is a *placement* defect — the comment
survives and moves from trailing to leading. Here it is gone. Row 9 above shows the
placement defect appearing at `DIFF` once a rest element keeps the comment alive,
so #4448's symptom is what this one degenerates from.

**Not #4398** as that issue is written: its repro and its control table hold the
comment *inside an erased inline TS annotation* and claim that removing any one of
{destructure, inline annotation, `$props()`} restores the count. These cells have
no annotation at all, so they are outside that table's domain rather than a
counterexample to it.

The record is ambiguous on this point and it is worth resolving before either is
fixed: **#4448** describes #4398 as *"a comment inside a destructured pattern whose
declaration the transform removes outright"*, and by that description these cells
are in scope, since a destructured `$props()` with no rest is also removed
outright. Whoever fixes #4398 should run row 1 above: if it goes to `r=1` the two
share a mechanism and this issue closes with it; if it does not, they do not.

baseballyama:
Closes #4501.

## What was wrong

A block comment in the value position of a destructured `$props()` declaration was dropped from
the client output when that declaration was the instance script's **only** statement.

The transform removes the declaration, so `transformed_script` is empty; the loop that puts a
dropped declaration comment back (`client/mod.rs`) sits inside `if !trimmed.is_empty()`, so it
never runs. Nothing else carries the comment, and it is gone.

## The fix

Re-enter the dropped comments as the script's own text when the transform emptied it. That is
what puts them in the comment buffer, and the template root's declarator — built by
`b::var_decl_anchored`, which already stamps it with the source offset upstream anchors to —
picks them up. The re-emission loop below then sees them live (`kept == had`) and does not push
a second copy, so the two paths do not fight.

Emitting them as a `RawMapped` statement instead, which is what the existing loop does, puts the
comment on its own line before the first surviving statement. That is a different position from
upstream's, and it is why the fix is upstream of the loop rather than inside it.

## Measured

The issue's nine cells plus two controls, three arms — official (`submodules/svelte`, `VERSION
5.57.0`), `main`, and this branch — verdict full `js.code` byte equality:

| cell | official vs `main` | official vs head | `main` vs head |
|---|---|---|---|
| 1 `let { a } = /* c */ $props();` | DIFF | DIFF | **MOVED** |
| 2 statement before | DIFF | DIFF | same |
| 3 statement after | EQ | EQ | same |
| 4 both | DIFF | DIFF | same |
| 5 leading newline | DIFF | DIFF | **MOVED** |
| 6 trailing newline | DIFF | **EQ** | **MOVED** |
| 7 same-line `const` | DIFF | DIFF | same |
| 8 same-line `$state` | DIFF | DIFF | same |
| 9 rest element | DIFF | DIFF | same |
| n1 comment-only script (#4500's repro) | DIFF | DIFF | same |
| n2 no comment | EQ | EQ | same |

Three cells move and all move one way: **no EQ cell is lost**, cell 6 becomes byte-equal, and
cells 1 and 5 go from the comment being absent to

```js
	var /* c */
	i = root();
```

where official writes `var /* c */ i = root();`. The remaining difference is the line break after
the block comment, which is **#4500** — n1 is #4500's own repro and is byte-identical between
`main` and this branch, so this PR neither fixes nor worsens it.

The issue's own rule needs correcting on one point: it lists cells 7 and 8 as dropped, and they
are not dropped on `main` today. #4528 landed between the filing and now. The live rule is
"dropped iff the declaration is the only statement and there is no rest element", which is cells
1, 5 and 6 — exactly the three this PR moves. The issue's count key also could not see that the
*kept* cells put the comment in the wrong place; the table above is byte equality, which can.

### Not moved

`#4448`, `#4500` and `#4516`'s repros are byte-identical between `main` and this branch, and all
three remain DIFF against official under **both** arms — so this is a negative result with its
own liveness control rather than a silent zero.

## Pins

Two tests in `client/tests.rs`. The first asserts cell 6's **whole** output against the oracle's
own bytes for that input. The second asserts `var /* c */` for cells 1 and 5 — the declarator
position — rather than `contains("/* c */")`, which the statement-shaped re-emission this
replaces would also satisfy.

A repro cannot live in `compatibility/pattern-corpus/`: a comment-only divergence is
AST-equivalent, so the corpus scores it a pass on both arms, and the issue's screen over 35,723
files finds **0** instances of the shape anyway. It is recorded as a no-repro doc under the
convention #4556 added.

### DIFF
--- a/.changeset/props-only-script-comment.md
+++ b/.changeset/props-only-script-comment.md
@@ -1,0 +1,18 @@
+---
+"@rsvelte/compiler": patch
+---
+
+client: a `$props()` declaration the transform removes outright no longer takes its comments with it
+
+When the destructured `$props()` declaration is the script's only statement, the
+transform empties the instance script, and the re-emission loop that puts a
+dropped declaration comment back is inside `if !trimmed.is_empty()`. The comment
+was therefore dropped from the client output entirely. Re-entering the comments
+as the script's own text puts them in the comment buffer, where the template
+root's declarator — which already carries the source anchor upstream stamps on
+it — picks them up, so the output is `var /* c */ i = root();` the way upstream
+writes it.
+
+The line break rsvelte still adds after a block comment in this position when
+`</script>` and the first element share a source line is #4500 and is unchanged
+here.
--- a/compatibility/KNOWN-FAILURES.md
+++ b/compatibility/KNOWN-FAILURES.md
@@ -7280,10 +7280,11 @@ over the `pattern/issues/` prefix alone, with the AST rescue removed. Its unit i
 `(id, target)`, folded to one entry per id carrying a `details` array naming each diverging
 target, so an id that diverges on all four targets is one ratchet entry rather than four.
 
-**Current baseline**, measured by the enrolling CI run (`Corpus Compat` 34430947366, tree
-`7b270ecbc`, 668 `pattern/issues/` manifest entries per target): 11 distinct ids, 30 `(id, target)`
-pairs, filed as `pattern-exact-known-failures.client.json`, 6 entries;
-`pattern-exact-known-failures.client-dev.json`, 6 entries;
+**Current baseline**, from the enrolling CI run (`Corpus Compat` 34430947366, tree
+`7b270ecbc`, 668 `pattern/issues/` manifest entries per target) less the two client entries of
+`008-comment-props-destructure.svelte`, which #4563 retired: 11 distinct ids, 28 `(id, target)`
+pairs, filed as `pattern-exact-known-failures.client.json`, 5 entries;
+`pattern-exact-known-failures.client-dev.json`, 5 entries;
 `pattern-exact-known-failures.server.json`, 9 entries; and
 `pattern-exact-known-failures.server-dev.json`, 9 entries. The two client targets and the two dev
 targets agree entry for entry, which is what makes the server/client split the real axis here.
@@ -7322,7 +7323,7 @@ same comment on both sides in a different place.**
 
 | id | targets | direction |
 |---|---|---|
-| `008-comment-props-destructure.svelte` | all 4 | **misplaced.** Client wants `var /* ) scanner payload */ div = root();` and rsvelte drops it; server puts the same comment after `= $$props;` where official has none. One comment, two verdicts — a relocation, not a loss. |
+| `008-comment-props-destructure.svelte` | server, server-dev | **rsvelte emits extra**, on the server only: the comment lands after `= $$props;` where official has none. The client half — official writing `var /* ) scanner payload */ div = root();` where rsvelte dropped the comment — was the enrolling baseline's other verdict and retired in #4563, so the entry that once read as a relocation is now a one-sided extra. |
 | `3515-props-default-line-comment.svelte` | server, server-dev | **rsvelte emits extra.** `let { a = 1 } = $$props; // initializer` against official's bare statement. |
 | `3515-props-default-multiline-comment.svelte` | all 4 | **mixed.** Server is the same extra `/* initializer` as the line-comment sibling; client is an indentation difference *inside* a kept block comment (`\t\t` vs `\t\t\t\t`), which is the continuation-line half of the same divergence. |
 | `3515-props-plain-line-comment.svelte` | server, server-dev | **rsvelte emits extra**, as the `default` sibling. |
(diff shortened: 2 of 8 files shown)


baseballyama:
Pushed a second commit: the `#4373` pin (`the_blocked_hosts_are_pinned_rather_than_matched`) was designed to fail when one of its rows is fixed, and this change fixes two of them. `INLINE_TWO` and `INLINE_ONE` move into `cells()`, whose expected values are the oracle's on both targets.

That is a second issue this closes as reported: **#4398**'s own repro (`let { a }: { /* c */ a?: number } = $props();` as the script's only statement) goes from DIFF to **byte-equal** with official on the client. The variant with another statement beside the declaration does not move — the comment is present with the right run order and lands before the surviving statement instead of on `var i = root()` — so #4398 stays open for that half; measurement posted there.

Local: all 114 `rsvelte_core` test targets matching comment/props/erase pass (`Running` lines == 114 == the target count), plus the `--lib` suite at 1974 passed / 0 failed.

PR #4568, opened 2026-09-10
fix(compiler): an erased annotation's comment keeps the line it shared with the initializer

### LINKED ISSUE #4397: A one-line erased TS annotation's two comment copies land on two lines where upstream keeps one

baseballyama:
Closes #4397.

## What was wrong

A comment left behind by an erased TypeScript **annotation** is flushed at the initializer's start,
and the flush wrote `'\n'` after it unconditionally. Upstream's printer does not: esrap keeps a
leading comment inline when it shared the anchored node's source line and breaks only where the
source broke. So a one-line annotation came out on three lines where official keeps one.

```svelte
<script lang="ts">
	let a: { /* c */ b: number } = { b: 1 };
</script>
```

```js
// official                              // rsvelte, before
let a = /* c */ /* c */ { b: 1 };        let a = /* c */
                                         /* c */
                                         { b: 1 };
```

## The axis

Not "one-line annotations" and not "one-line initializers": the separator is the **text between the
comment's end and the flush point**. The oracle says so on three cells that hold the annotation's
shape fixed and move the newline:

| source | official |
|---|---|
| `let a: { /* c */ b: number } = { b: 1 };` | inline |
| `let a: {`⏎`/* c */ b: number } = { b: 1 };` — newline **before** | inline |
| `let a: { /* c */ b: number } = {`⏎`b: 1`⏎`};` — multi-line initializer | inline |
| `let a: { /* c */ b: number`⏎`} = { b: 1 };` — newline **after** | breaks |

The last row is the live negative control: it is EQ on `main` and must stay EQ.

## The fix

`emit_pending_comment` takes the separator instead of hard-coding `'\n'`. The two in-place call
sites — a whole erased *statement*, which has no flush point — keep `'\n'`, which is what holds the
`type` alias and `interface` cells from #4397's own table unchanged. The flush site inside
`push_range_flushing_pending` reads the source between the comment and the flush point. A `//`
comment always breaks, since a space there would swallow the rest of the line.

The separator is one byte either way, so every `CopiedSourceChunk` and `reemitted_comment_outputs`
offset is unchanged and the projection is untouched.

## Measured

Nine cells, verdict full `js.code` byte equality against official (`submodules/svelte`, pin
`7bc0a70fe`, `VERSION 5.57.0`), three arms — official, `main` at `9b86d816d` (`sha256`
`1695d13114c7…`) and this branch (`b703fe8fdead…`), separate binaries.

| cell | main | this branch |
|---|---|---|
| `let a: { /* c */ b: number } = { b: 1 };` — the reported one | DIFF | **EQ** |
| newline **before** the comment | DIFF | **EQ** |
| multi-line initializer | DIFF | **EQ** |
| newline **after** the comment (live negative) | EQ | EQ |
| multi-line annotation / `type` alias / `interface` / `//` comment | EQ | EQ |
| annotation on an **uninitialized** declarator (#4396, a different mechanism) | DIFF | DIFF |

Identical on the **client and the server**, which is what a change in the shared erasure should
look like — and 78 neighbouring cells across six sibling grids (#4396, #4501, #4448, #3515, #4453
and the `$props()` newline axis, both targets) are byte-identical between the two arms. The
#4397 grid moving six of its own cells is that zero's liveness control.

## Pins

`crates/rsvelte_core/tests/erased_annotation_comment_line_4397.rs`. No repro can live in
`compatibility/pattern-corpus/`: `ast_equiv_batch` runs with `CommentPolicy::Ignore`, so a
comment-only divergence scores `match` on every target, and #4397 records a source-shape screen over
33,890 corpus files selecting 0 carriers.

### DIFF
--- a/.changeset/erased-annotation-comment-separator.md
+++ b/.changeset/erased-annotation-comment-separator.md
@@ -1,0 +1,26 @@
+---
+"@rsvelte/compiler": patch
+---
+
+compiler: an erased annotation's comment keeps the line it shared with the initializer
+
+A comment left behind by an erased TypeScript annotation is flushed at the
+initializer's start, and the flush wrote a newline after it unconditionally.
+Upstream's printer does not: esrap keeps a leading comment inline when it shared
+the anchored node's source line and breaks only where the source broke. So a
+one-line annotation came out on three lines where official keeps one (#4397).
+
+The separator is now read off the text between the comment's end and the flush
+point — not off the shape of the annotation and not off the shape of the
+initializer. A newline *before* the comment still prints inline, a multi-line
+initializer still prints inline, and a newline *after* the comment still breaks.
+A `//` comment always breaks, since a space there would swallow the rest of the
+line. The two in-place call sites — a whole erased *statement*, which has no
+flush point — keep the newline they had, which is what leaves the `type` alias
+and `interface` cells unchanged.
+
+Measured on nine cells against Svelte 5.57.0: three go from divergent to
+byte-equal on **both** the client and the server, none goes the other way, and
+78 neighbouring cells across six sibling grids are byte-identical between the
+arms. The separator is one byte either way, so every source-projection offset is
+unchanged.
--- a/compatibility/pattern-corpus/issues/4397-erased-annotation-comment-line.md
+++ b/compatibility/pattern-corpus/issues/4397-erased-annotation-comment-line.md
@@ -1,0 +1,19 @@
+# a one-line erased annotation's comment printed on its own line
+
+**Issue:** [#4397](https://github.com/baseballyama/rsvelte/issues/4397)
+**Repro:** none — `crates/rsvelte_core/tests/erased_annotation_comment_line_4397.rs`
+
+The flush that puts an erased annotation's comment back at the initializer wrote
+a newline after it unconditionally. Upstream keeps it inline when the comment
+shared the initializer's source line.
+
+The axis is the text between the comment's end and the flush point: a newline
+*before* the comment (`let a: {`⏎`/* c */ b: number } = { b: 1 };`) still prints
+inline, and a multi-line initializer still prints inline. Only a newline between
+the two breaks the line.
+
+No repro can live here, for the same reason as
+`4453-comment-in-removed-props-pattern.md`: `ast_equiv_batch` runs with
+`CommentPolicy::Ignore`, so a comment-only divergence is a pass on both arms.
+#4397 also records a source-shape screen over 33,890 corpus `.svelte` files that
+selects 0 carriers, so no corpus growth is the instrument for this either.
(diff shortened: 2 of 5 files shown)


baseballyama:
`Test (ubuntu-latest, 2)` was `erased_annotation_comment.rs::the_two_residual_shapes_are_pinned_rather_than_matched` — the two-sided pin firing as designed: the one-line annotation it recorded as a divergence now agrees with upstream. Moved into that file's matched cells with the oracle's own text, and the pin (renamed, since it holds one shape now) keeps only #4396's.

Re-ran locally the 71 `rsvelte_core` suites whose fixtures carry both `lang="ts"` and a comment — the population this change can reach — plus the two directly affected: `RC=0`, 71 `Running tests/` lines.

Issue #4517, opened 2026-09-09
A comment at the end of the instance script is dropped when the first template statement is a component call with children (`has_loc` reads a synthesized span as "before every comment")


baseballyama:
## Symptom

```svelte
<script>
	// c
</script>
<X>t</X>
```

Official emits the comment ahead of the component call; rsvelte emits nothing.

```
==== ts_simple  <X />  DIFF
   1 | import 'svelte/internal/disclose-version';     | import 'svelte/internal/disclose-version';
   2 | import 'svelte/internal/flags/legacy';         | import 'svelte/internal/flags/legacy';
   3 | import * as $ from 'svelte/internal/client';   | import * as $ from 'svelte/internal/client';
   4 |                                                | 
   5 | export default function C($$anchor) {          | export default function C($$anchor) {
*  6 | 	// c                                          | 	X($$anchor, {});
*  7 | 	X($$anchor, {});                              | 	// c
   8 | }                                              | }
==== ts_child   <X>t</X>  DIFF
   1 | import 'svelte/internal/disclose-version';     | import 'svelte/internal/disclose-version';
   2 | import 'svelte/internal/flags/legacy';         | import 'svelte/internal/flags/legacy';
   3 | import * as $ from 'svelte/internal/client';   | import * as $ from 'svelte/internal/client';
   4 |                                                | 
   5 | export default function C($$anchor) {          | export default function C($$anchor) {
*  6 | 	// c                                          | 	X($$anchor, {
*  7 | 	X($$anchor, {                                 | 		children: ($$anchor, $$slotProps) => {
*  8 | 		children: ($$anchor, $$slotProps) => {       | 			$.next();
*  9 | 			$.next();                                   | 
* 10 |                                                | 			var text = $.text('t');
* 11 | 			var text = $.text('t');                     | 
* 12 |                                                | 			$.append($$anchor, text);
* 13 | 			$.append($$anchor, text);                   | 		},
* 14 | 		},                                           | 		$$slots: { default: true }
* 15 | 		$$slots: { default: true }                   | 	});
* 16 | 	});                                           | }
* 17 | }                                              | 
```

The self-closing case (upper block) is the same defect with a different outcome: the leading
flush is skipped there too, and the comment survives only because the body's end flush picks it
up — so it comes out **after** the call. Adding children destroys it outright.

## Grid

12 cells against `submodules/svelte` at the `origin/main` pin (`7bc0a70fe`, `VERSION 5.57.0`),
[...]

Issue #4492, opened 2026-09-09
A trailing instance-script comment is kept inside <svelte:boundary>'s arrow parameter list, where official drops it


baseballyama:
## Symptom

A comment at the end of the instance script is **kept** by rsvelte and **dropped** by official
when the template holds a `<svelte:boundary>`, and rsvelte glues it into the arrow's parameter
list:

```svelte
<script>
	let c = 1;
	// x
</script>
<svelte:boundary><b>{c}</b></svelte:boundary>
```

```js
// official — the comment is absent
	$.boundary(node, {}, ($$anchor) => {

// rsvelte
	$.boundary(node, {}, ($$anchor// x
	) => {
```

Reproduces on `client` and `client-dev`. Both cells are attributable: with the comment removed the
two outputs are byte-equal, so the comment is the whole difference. Reproduces identically on the
pre-#4490 arm (`9b50b1c4c07eff34`) and the #4490 arm (`72082a9b1eb1e009`), so it is not a side
effect of that PR.

## Direction and family

This is the **KEEPS** direction — the same as #4475, where rsvelte kept a comment official drops at
a `b.block([…])` brace. The carrier here is not the block but the **arrow's parameter list**, which
is the same landing site `{#key}` uses in #4481. So of the three directions in this family:

| direction | issue | this cell |
|---|---|---|
| rsvelte keeps what official drops | #4475 | **yes** |
| rsvelte drops what official keeps | #4480 / #4489 | no |
| both keep, different node | #4481 | no |

Instrumenting `flush_comments_until` on the sibling cells (posted to #4481) showed the shared
machine: rsvelte holds instance-script comments in a buffer whose offsets sit **above** every real
source offset, so a script comment can only be flushed at a node carrying a comment-space anchor,
never at a source-positioned node inside a generated call. Here that anchored node is inside
`$.boundary`'s arrow, and upstream's equivalent flush point does not exist because upstream's
`b.arrow` carries no `loc` at all.

## Reach: zero measured carriers

Over the collected corpus sources read from the primary checkout, `.svelte` files only:

```
files containing <svelte:boundary>     198
official rejection codes over them     OK 103, experimental_async 95
compared units (103 files x 2 dev)     206      both-reject 190      one-side reject 0
byte-different units                     8
units carrying the signature             0
```

The 95 `experimental_async` rejections are **not** a limitation of this probe: `targets.mjs` sets
no `experimental` option either, so the corpus gates reject the same 95 files. The population here
[...]

Issue #4466, opened 2026-09-08
The client source map's out-of-range segments are mostly a second mechanism: 94% sit in files that never enter split coordinates


baseballyama:
### What

#4454 reports that the client source map points past the end of a source line on real
components. #4455 accounts for one mechanism behind that — a chunk whose text contains
`$effect` never claims its `loc_map` region, so its comment-space offsets reach the map
untranslated. Measured against the population, **that mechanism is 0.7% of it**. This
issue is for the other 94%, so that it has a number before the #4455 fix lands and its
residue gets read as the fix underperforming.

### Measurement

`compatibility/pattern-corpus`, `generate: client`, `dev: false`, clean-`main` arm
`cc-base2.node` (`sha256 a98d3c23…`). 1,260 files compared, 121 skipped (one side
throws), 3 excluded for containing U+2028/U+2029 (method note below). Buckets are
`$effect` substring present × a `//` or `/*` inside any `<script>`:

| bucket | files | files with any | segments oob / total | share | median file share | max |
|---|---:|---:|---:|---:|---:|---:|
| `$effect` + script comment | 3 | 3 | 41 / 108 | 38.0% | 18.2% | 51.6% |
| `$effect` only | 32 | 4 | 9 / 4348 | 0.2% | 0.0% | 2.0% |
| script comment only | 89 | 8 | 291 / 6401 | 4.5% | 0.0% | 47.2% |
| **neither** | **1136** | 141 | **5675 / 100085** | **5.7%** | 0.0% | **69.9%** |

official: **0 files, 0 of 92,252 segments** — the negative control at this scale.

So the `neither` bucket carries 5,675 of 6,016 out-of-range segments, and one of its
files reaches 69.9%.

### It is not #4455's mechanism, and the evidence is the mechanism quantity

#4455's mechanism needs the printer to be in split coordinates, which the printer
reports directly (`loc_base` and the `loc_map` range count). Probed on the five worst
`neither` carriers, with a positive control that must print a line:

```
issues/3103-reactive-array-rest.svelte                  (no split_coords line)
issues/3071-rune-argument-class-in-component.svelte     (no split_coords line)
adversarial/legacy/legacy-reactive-ordering.svelte      (no split_coords line)
adversarial/legacy/reactive-destructuring.svelte        (no split_coords line)
adversarial/legacy/reactive-ordering.svelte             (no split_coords line)
control: a $effect + comment file           split_coords base=53 ranges=0 line_starts=5
```

[...]

Issue #4454, opened 2026-09-08
The client source map points past the end of a source line on 14.6% of sampled corpus components, and the gate that owns the predicate reads 0 on 29 fixtures


baseballyama:
### What

On real components, a large fraction of the client source map's segments point to a **column past the end of the source line** they name. `sourcemaps_gate.rs` already owns this exact predicate and reads **0** — on its 29 fixtures.

`crates/rsvelte_core/tests/sourcemaps_gate.rs:469-472` defines it:

> Number of segments whose original position lies outside its source … its line is past the last line of the source, or its column is past the end of that line (column == line length is legal: it addresses the line terminator).

Measured with the same convention (UTF-16 code units, `column == length` legal), `generate: client`, `dev: false`, arm `cc-clean-main.node` @ `b003aae56` (`sha256 d5cb72ce…`), oracle `VERSION 5.57.0`:

| file | official | rsvelte |
|---|---|---|
| `svelte-maplibre-gl/…/layers/CustomLayer.svelte` | 0 / 304 | **360 / 476** |
| `svelte-maplibre-gl/…/layers/RawLayer.svelte` | 0 / 818 | **1030 / 1168** |
| `immich/web/…/AdaptiveImage.svelte` | 0 / 1100 | **1062 / 1474** |
| `bits-ui/…/context-menu.svelte` | 0 / 116 | 0 / 92 |

Every one is a **column** overflow; the line index is in range in all of them. One concrete segment, on an arm where the generated code is byte-identical to official's:

```
generated 12:1   official -> source 39:1        (a real position)
                 rsvelte  -> source 88:173      (line 88 of an 88-line file whose longest line is 86 columns)
```

Both compilers declare the identical `sourcesContent`, byte for byte and line for line, so this is a map pointing outside the source it names rather than a disagreement about what the source is.

### Reach

Deterministic stride sample of the collected corpus, 300 of 33,890 ids, 288 compared (12 throw on one side):

```
official : 0 / 48,346 segments out of range     0 of 288 files with any
rsvelte  : 4,228 / 51,429 segments out of range  42 of 288 files with any   (14.6% of files, 8.2% of segments)
```

Official's 0 across 288 files is the negative control at scale: the detector does not manufacture out-of-range on a correct map.

### Controls

[...]

PR #4577, opened 2026-09-11
fix(sourcemap): do not emit a segment whose offset is past the source

baseballyama:
Cuts the out-of-range half of #4466 (and of #4454). Does **not** close either, and does **not** fix the mechanism — see *This is the tail, not the class*.

## What was wrong

`RestoreRawMappedSpans::visit_span` writes nothing when `copied_spans` cannot translate an offset, which leaves the node's span in **chunk** coordinates — an offset into the transformed script text, not into the `.svelte` source. Its own comment says so:

> Keep that node in chunk coordinates instead of constructing an invalid OXC span.

Nothing downstream can tell that apart from a source offset, so the map named a position no source line can hold:

```
issues/3103-reactive-array-rest.svelte   309 of 442 segments (69.9%)
   gen 33:10 -> src 12:2   (the file has 11 lines; line 12 is empty)
```

## Why the fix is not "unlocate the span"

Setting the span to `SPAN` when it cannot be mapped takes out-of-range to 0 — and moves `js.code` on **ten** `pattern-corpus` files, every one of them away from official (`FIXED 0 / BROKEN 10`): stray `;;` statements appear and `$.user_effect(/* fx */ () => {})` loses its comment. The span is load-bearing for comment placement. Measured, not assumed; that arm is what said the fix has to be elsewhere.

So the rejection is at the two **map emission** sites, which is output-neutral by construction:

- `command::Driver`, `EventKind::LocationOffset` — the site that produces almost every segment;
- `Printer::map_position` — the `Location { line, column }` path.

Both reject an offset past the length of the text the map resolves against (`map_source` for a split print, `source` otherwise).

Both are load-bearing, which is why neither is dead code:

| arm | out-of-range |
|---|---|
| `main` | 5,891 |
| `Driver` guard only | 1,637 |
| `map_position` guard only | 4,461 |
| both | **77** |

## Measured

`compatibility/pattern-corpus`, 1,387 files, `generate: client`, `dev: false`. Arms are separate `compile_one` binaries built from `main` (`47c5bc5f5`) and this branch; `sha256` `8ea0f979…` / `9f911b5a…`. Columns compared in **bytes**, which is what rsvelte emits.

| arm | segments | out-of-range |
|---|---:|---:|
| `main` | 111,276 | 5,891 (5.3%) |
| this branch | 105,462 | **77 (0.1%)** |
| official, for scale | 92,600 | — |

Official's column is not directly comparable: it emits UTF-16 columns, so a byte-length predicate is the wrong yardstick for it. It is here only to say that 105,462 is not a collapse of coverage — the unlocate arm dropped to 91,444, *below* official, which is the other reason it was rejected.

**`js.code` MOVED = 0** over all 2,774 client+server corpus units.

## This is the tail, not the class

#4466's own second comment establishes that out-of-range is a **filter artefact**: a segment is
out of range exactly when its wrong source position happens to clear a line end, so which
segments qualify is decided by where the file ends. The class underneath is "the segment claims
the wrong source position", and its mechanism is `parse_chunk`'s comment-free path returning
**chunk-local** spans that `map_position` then resolves against the source line table. That is
not touched here, and #4466 explicitly reserves it because making those spans unlocatable also
moves `has_loc` and so moves comment flushing (#4489 / #4492 / #4500 / #4516 / #4517).

So the question this branch has to answer is not "does out-of-range go down" — it must, by
construction — but **does dropping that tail cost any correct segment**. Re-derived here rather
than cited, on the same 1,263 live files (124 skipped: one side throws, or the file carries
U+2028/U+2029). Unit: a segment whose *generated* position starts an identifier; verdict is
whether the *source* position it claims starts the same identifier.

| arm | segments | identifier-judged | correct | out-of-range |
|---|---:|---:|---:|---:|
| official | 92,476 | 39,729 | 33,602 (**84.6%**) | 0 |
| `main` | 111,152 | 48,318 | 27,518 (**57.0%**) | 5,891 |
| this branch | 105,338 | 45,752 | **27,518** (60.1%) | 77 |

**`correct` is identical to the byte.** The branch removes 2,566 identifier-judged segments and
every one of them was already wrong; the other ~3,250 dropped segments are not
identifier-judged. Official's 84.6% is the metric's own floor — esrap legitimately anchors some
nodes at a parent position — so 57.0 → 60.1 is a move toward it and not a claim of correctness.

The cost is that the cheapest instrument for the class gets quieter: out-of-range was the number
#4466 was sized with. The identifier-correct rate above is the one to watch instead, and it is
unmoved in the numerator, which is the point.

## What is left in this branch's own scope

The 77 are a different shape: an **exclusive end** landing exactly one past the last byte of a line, which is what an exclusive end is. Rejecting those would need `offset >= len` and would drop every legitimate end-of-file end. `3071-rune-argument-class-in-component.svelte`'s remaining one is that, and it is pinned as a recorded value rather than driven to zero.

#4466 also asks whether the *surviving* segments are otherwise correct, and #4454 measures a different population (`compatibility/sources`, not materialised here). Both stay open.

## The gate does not see this

`sourcemaps_gate.rs` has an out-of-range budget, and it is **empty before and after** — its population is the upstream `sourcemaps` fixtures, and the carriers are legacy `$:` and rune-argument components that live only in `pattern-corpus`. No re-baseline; `compatibility/sourcemap-known-failures.json` is unchanged.

## Pins

`crates/rsvelte_core/tests/map_offset_past_source_4466.rs`: the three carriers #4466 named as recorded values (0 / 1 / 0, against 309 / 63 / 276 before), and a corpus-wide ceiling with its denominator asserted. Positive control: removing both guards fails both tests and prints the pre-fix numbers exactly (`309 of 442`, `5891 of 111332`).

`compile_one` gains `--map`, the instrument this was measured with.

### DIFF
--- a/.changeset/map-offset-past-source.md
+++ b/.changeset/map-offset-past-source.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Stop emitting client source-map segments whose source offset is past the end of the source: an untranslated chunk coordinate resolved to a position no source line can hold.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2662,7 +2662,7 @@ dependencies = [
 
 [[package]]
 name = "rsvelte_esrap"
-version = "0.10.41"
+version = "0.10.42"
 dependencies = [
  "compact_str",
  "memchr",
--- a/compatibility/pattern-corpus/issues/4466-map-offset-past-source.md
+++ b/compatibility/pattern-corpus/issues/4466-map-offset-past-source.md
@@ -1,0 +1,22 @@
+# client source-map segments pointing past the end of a source line
+
+**Issue:** [#4466](https://github.com/baseballyama/rsvelte/issues/4466)
+**Repro:** none — `crates/rsvelte_core/tests/map_offset_past_source_4466.rs`
+
+`RestoreRawMappedSpans::visit_span` returns without writing when a chunk offset
+has no `copied_spans` translation, which leaves the node's span in **chunk**
+coordinates — an offset into the transformed script text. Nothing downstream can
+tell that apart from a source offset, so the map named a position no source line
+can hold (69.9% of one carrier's segments).
+
+Leaving the span located is deliberate: unlocating it instead changes `js.code`
+on ten `pattern-corpus` files, all of them away from official. So the rejection
+is at the two **map emission** sites (`Driver::LocationOffset` and
+`Printer::map_position`), both of which now drop an offset past the mapped
+source's length. `js.code` is byte-identical over the whole corpus.
+
+Both sites are load-bearing: the `Driver` alone leaves 1,637 out-of-range
+segments and `map_position` alone leaves 4,461, against 77 with both.
+
+No repro can live here: the corpus gate compares `js.code`, which this does not
+change.
--- a/crates/rsvelte_core/Cargo.toml
+++ b/crates/rsvelte_core/Cargo.toml
@@ -94,7 +94,7 @@ oxc_ast = { version = "0.146", features = ["serialize"] }
 oxc_span = "0.146"
 oxc_diagnostics = "0.146"
 oxc_codegen = "0.146"
-rsvelte_esrap = { version = "=0.10.41", path = "../rsvelte_esrap" }
+rsvelte_esrap = { version = "=0.10.42", path = "../rsvelte_esrap" }
 oxc_syntax = "0.146"
 oxc_semantic = "0.146"
 
--- a/crates/rsvelte_core/tests/map_offset_past_source_4466.rs
+++ b/crates/rsvelte_core/tests/map_offset_past_source_4466.rs
@@ -1,0 +1,167 @@
+//! A generated node whose span `RestoreRawMappedSpans` could not translate back
+//! keeps its **chunk** coordinate — an offset into the transformed script text,
+//! not into the `.svelte` source. Read as a source position it resolves to a
+//! line that cannot hold it, so the map points past the end of a source line
+//! (#4466). The two emission sites reject such an offset instead of naming a
+//! position that does not exist.
+//!
+//! `sourcemaps_gate.rs` does not see this: its population is the upstream
+//! `sourcemaps` fixtures, whose out-of-range budget is empty both before and
+//! after. The carriers are legacy `$:` and rune-argument components, which live
+//! in `compatibility/pattern-corpus`.
+//!
+//! Columns here are compared in **bytes**, because that is what rsvelte emits
+//! (`offset - line_start`). Official emits UTF-16 columns, so this predicate is
+//! not the one to judge official's map with.
+
+use rsvelte_core::{CompileOptions, GenerateMode, compile};
+use std::path::{Path, PathBuf};
+
+fn map_of(path: &Path) -> Option<(String, String)> {
+    let source = std::fs::read_to_string(path).ok()?;
+    let result = compile(
+        &source,
+        CompileOptions {
+            filename: Some(path.to_string_lossy().into_owned()),
+            generate: GenerateMode::Client,
+            ..Default::default()
+        },
+    )
+    .ok()?;
+    Some((source, result.js.map?))
+}
+
+/// `(segments, out_of_range)` for one map, where out-of-range means the source
+/// column is past the last byte of the source line the segment names.
+fn count(source: &str, map_json: &str) -> (usize, usize) {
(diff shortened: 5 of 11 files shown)


baseballyama:
## The first push was red on three comment-placement tests, and they are one cause

`Test (ubuntu-latest, 1)` and `(ubuntu-latest, 2)` failed on three unrelated-looking suites:

| suite | assertion |
|---|---|
| `svelte_head_comment_anchor` | `cell=each`, `left: 0 right: 1` |
| `component_call_comment_anchor_4529` | `the_dynamic_component_wrapper_claims_no_anchor` |
| `template_body_own_line_comment` | `official emits it once, got 0` |

All three are a comment that stopped being emitted, and all three come from where I put the
guard: **inside `map_position`**. That function has two kinds of consumer, and only one of them
is the source map —

```
 902  offset_to_line_col          -> PLACEMENT (9 callers: 1106 1112 1117 1132 1136 1137 2443 2444 2854)
1220  flush_comments_before_source-> PLACEMENT
1225  flush_comments_before_source-> PLACEMENT
1228  flush_comments_before_source-> PLACEMENT
1033 1062 1066 1113 2584 4156 4180 -> ctx.location(), i.e. a map segment
```

Returning `None` to a placement caller does not drop a segment, it drops the comment. The
pattern-corpus measurement could not see it: it compares `js.code` and the cells that move are
`svelte:head` + `{#each}`, `<svelte:component>` and a comment above an `{#if}`, none of which
are in that corpus.

## What changed

The rejection now lives in `map_segment` / `map_segment_end`, which only the seven
`ctx.location()` sites call. `map_position` and `offset_to_line_col` are byte-for-byte the
behaviour they have on `main`, so placement is untouched by construction.

The three `ctx.location()` sites that also choose *text* (`write_source_keyword`'s
keyword/suffix split at 1033, the `function`/`async function` match at 2584, `KeywordCursor` at
1113) write the same bytes on both arms — the `None` branch of each concatenates exactly what
the `Some` branch writes in pieces — so the narrowed guard is map-only by construction, not
merely by measurement.

baseballyama:
## Re-measured on the narrowed guard

Arms are separate binaries, `sha256`:
`8ecf0c67287e84f7…` `main` @ `47c5bc5f5` (fully green), `de2164595ab94323…` this branch.
Oracle is `submodules/svelte` (`VERSION 5.57.0`), the source path the gates use.
Population is `compatibility/pattern-corpus`, 1,387 files.

**Out-of-range segments** (`column > line length`, UTF-8 bytes, `column == length` legal):

| arm | segments | out-of-range |
|---|---|---|
| official | 92,600 | 12 (0.0%) |
| `main` | 111,276 | **5,891** (5.3%) |
| this branch | 105,466 | **81** (0.1%) |

The three carriers the test pins, unchanged from the first push: `3103-reactive-array-rest`
309 → **0**, `3071-rune-argument-class-in-component` 63 → **1**,
`legacy-reactive-ordering` 276 → **0**.

**The control that says nothing correct was thrown away.** Judging every segment whose
*generated* position starts an identifier against whether its *source* position starts the
same one, over the 1,263 files live under all three arms:

| arm | segments | judged | correct | rate | out-of-range |
|---|---|---|---|---|---|
| official | 92,476 | 39,729 | 33,602 | 84.6% | 0 |
| `main` | 111,152 | 48,318 | **27,518** | 57.0% | 5,891 |
| this branch | 105,342 | 45,754 | **27,518** | 60.1% | 81 |

`correct` is **identical** on both arms — 5,810 segments disappear and not one of them was
right. The rate moves only because the denominator shrinks.

**`js.code` is untouched**: sweeping `main` against this branch over the same corpus on both
targets, `units same=2532 MOVED=0 both-threw=242`.

**Local suites**: 15 of 15 green, including all three that were red on the first push —
`svelte_head_comment_anchor`, `component_call_comment_anchor_4529`,
`template_body_own_line_comment`, plus `sourcemaps_gate`, `map_offset_past_source_4466`,
`map_region_locmap_4466`, `instance_script_dropped_comment_4517`,
`arrow_argument_comment_anchor_4481`, `client_script_comments_pin`,
`trailing_script_comment_2727`, `split_coordinates`, `esrap_samples`,
`keyword_anchor_fidelity`, `dual_run_gate`, `sequence_measure_residue`. The `Running tests/`
count is 15, which is the denominator I asked for.

One instrument note: the first run of the out-of-range sweep keyed its aggregate on the
binary's **basename**, and both arms are named `compile_one`, so the two collided into one
bucket and printed a single plausible row (`216742 segments, 5972 out-of-range`). The numbers
above are from arms copied to distinct names.

PR #4578, opened 2026-09-12
fix(compiler): float an uninitialized declarator's erased-annotation comment forward

### LINKED ISSUE #4396: A comment in an erased TS annotation on an UNINITIALIZED declarator is dropped; upstream floats it to the next node

baseballyama:
Closes #4396.

## What was wrong

A declarator with **no initializer** ends at its identifier, so a comment left behind by its
erased TypeScript annotation has nothing to attach to. rsvelte dropped it; upstream flushes it
at the next located node.

```svelte
<script lang="ts">
	let a: {
		/* c */
		b: number;
	} | null;
	let n = 1;
</script>
<i>{a}{n}</i>
```

```js
// official                    // rsvelte (before)
let a;                         let a;

/* c */                        let n = 1;
/* c */
let n = 1;
```

Upstream prints it twice because acorn-typescript speculates over the annotation; rsvelte already
reproduces that duplication for the **initialized** shape (`let a = /* c */` ⏎ `/* c */` ⏎ `null;`
is byte-equal on `main`), so the machinery was present and only this branch was `None`.

## Why it goes on the following statement and not at the removal point

This is the shape #4395 deliberately left alone, and the reason is still live: the client's legacy
state lowering is line-oriented, so a comment inside the declaration's own line turns
`let stats = $.mutable_source(0)` into `let stats;` — text that parses, runs, and is no longer
reactive. Putting it on its own line ahead of the **next** statement is both upstream's placement
and safe; `the_following_legacy_declaration_keeps_its_initializer` pins exactly that pair.

## What this does NOT close

When the declarator ends its statement list, upstream carries the comment out of the script
entirely, onto the template's own first statement. No source-range rewrite can express that, so
that shape still drops and is pinned as such rather than left unstated.

## Measured

Arms are separate binaries: `sha256 088df89eb48cab8f…` `main` at `b75886891`,
`8411f1d796cc5c9f…` this branch. Oracle is `submodules/svelte` (`VERSION 5.57.0`), the source path
the gates use. Verdict is full `js.code` byte equality.

| cell | target | main | this branch |
|---|---|---|---|
| block comment, statement follows | client | DIFF | **EQ** |
| block comment, statement follows | server | DIFF | **EQ** |
| line comment, statement follows | client | DIFF | **EQ** |
| line comment, statement follows | server | DIFF | **EQ** |
| initialized declarator (control) | client | EQ | EQ |
| initialized declarator (control) | server | EQ | EQ |
| comment outside the annotation (control) | client | EQ | EQ |
| comment outside the annotation (control) | server | EQ | EQ |
| declarator ends its list | client | DIFF | DIFF — the half above |
| declarator ends its list | server | DIFF | DIFF — the half above |

**Nothing else moves.** `compatibility/pattern-corpus`, both targets: `units same=2532 MOVED=0
both-threw=242`.

**Real-world reach, screened rather than pattern-matched.** 34,933 `.svelte` files across the
populated submodules; 22,816 have a TS instance script and 1,235 hold an uninitialized annotated
declarator (both marginals live, so a small carrier count is a property of the population).
Sweeping those 1,235 on both targets: `same=2380 MOVED=2 both-threw=88`, and the two moved units
are one file —
`submodules/adventurelog/frontend/src/routes/profile/[uuid]/+page.svelte`, whose `// Legacy fields`
this restores. Its region now matches the oracle line for line, **including** the
`let stats = $.mutable_source()` beside it that #4395's hazard would have emptied. The file stays
DIFF overall under both arms for an unrelated CSS-hash divergence, so it is a `both_diff` unit
whose comment region converged — which is why the verdict above is per-region on that file and
byte-equality on the grid.

## Pins

`crates/rsvelte_core/tests/uninit_annotation_comment_4396.rs`, 7 tests: the four moved cells, the
`$.mutable_source` hazard, the initialized control, and the still-dropping half.

A repro cannot live in `compatibility/pattern-corpus/`: `ast_equiv_batch` runs with
`CommentPolicy::Ignore`, so a comment-only divergence scores `match` on both arms.

## Provenance

The approach is a re-derivation of `42aed0b3e`, an unpushed commit from four days ago that sat 98
commits behind `main` in another worktree. Its diff no longer applies — `types.rs` has since grown
the `pending` / `flush_at` machinery this rides on — so the hook points, the tests and the
measurements here are new; what carried over is the mechanism and the #4395 hazard it recorded.

### DIFF
--- a/.changeset/uninit-annotation-comment-forward.md
+++ b/.changeset/uninit-annotation-comment-forward.md
@@ -1,0 +1,5 @@
+---
+"@rsvelte/compiler": patch
+---
+
+Float an uninitialized declarator's erased-annotation comment onto the following statement instead of dropping it, which is where upstream flushes it.
--- a/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
+++ b/crates/rsvelte_core/src/compiler/phases/2_analyze/types.rs
@@ -669,6 +669,7 @@ fn strip_typescript_from_program_impl(
     // projection-less one while the client reads the other.
     let repeat_regions = collect_speculative_type_head_regions(program);
     let declarator_annotations = collect_declarator_annotations(program);
+    let uninit_float_targets = collect_uninit_annotation_float_targets(program);
 
     // Text-based fallback: strip `declare global { ... }`, `declare module ... { ... }`,
     // and `declare namespace ... { ... }` blocks. These may not always be parsed as
@@ -789,10 +790,18 @@ fn strip_typescript_from_program_impl(
                 .iter()
                 .find(|(from, _)| *from == *remove_start)
                 .map(|(_, init_start)| *init_start);
-            // `Some(None)` is an uninitialized declarator: upstream drops the
-            // comment onto the next located node, so this one is not ours to print.
+            // `Some(None)` is an uninitialized declarator: its declaration ends at
+            // the identifier, so upstream flushes the comment at the NEXT located
+            // node instead. Printing it here would put it after the `;`, where the
+            // client's legacy state lowering stops scanning (#4395). Float it to
+            // the following statement when there is one; when the declarator ends
+            // the list upstream carries it out of the script entirely, which no
+            // source-range rewrite can express, so that shape still drops (#4396).
             let flush_at = match declarator_annotation {
-                Some(None) => None,
+                Some(None) => uninit_float_targets
+                    .iter()
+                    .find(|(from, _)| *from == *remove_start)
+                    .map(|(_, at)| Some(*at)),
                 Some(Some(init_start)) => Some(Some(init_start)),
                 None => Some(None),
             };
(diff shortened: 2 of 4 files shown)


baseballyama:
## The one red was the residue pin, failing because the residue is gone

`Test (ubuntu-latest, 2)` failed on a single assertion:

```
crates/rsvelte_core/tests/repeated_type_head_comment_runs_4373.rs:275
  left: "c d c d"
 right: ""
```

`the_blocked_hosts_are_pinned_rather_than_matched` existed to write down that rsvelte answered
`""` where the oracle answers `c d c d`, and its own doc comment names #4396 as the owner. This
branch's float retires that cell, so the pin fails in the direction a shrink-only residue test is
supposed to fail in.

Re-derived from the oracle rather than from the retired pin's comment — `submodules/svelte` at
`7bc0a70fe64d`, `VERSION 5.57.0`, `dev: false`, which is the pin the branch, `origin/main` and the
checkout the oracle was run from all three carry:

```
oracle client: c d c d
oracle server: c d c d
```

The cell moves into `cells()` instead of being deleted, so it keeps the same per-target assertion
every other head in the table has.

## The shard, not the four suites I had run

The first local reproduction reported six failures, five of which were
`Fixtures not found for Svelte commit: 325cf6297181` — a linked worktree initializes no
submodules, so `rev-parse` returned the **superproject's** HEAD and `css` / `validator` /
`preprocess` / `compiler_fixtures` / `dual_run_gate` were measuring an absent denominator rather
than this change. After `git submodule update --init submodules/svelte` and staging the generated
fixtures, the whole shard is live:

```
Summary [ 19.186s] 1732 tests run: 1732 passed, 2 skipped
```

That is the denominator CI uses for this job, not the four suites I had scoped the earlier run to
— `repeated_type_head_comment_runs_4373` was not among them, which is why the shard found it and I
had not.

PR #4580, opened 2026-09-13
fix(cargo): mark every submodule update=none so git dependencies fetch

kosayoda:
Cargo initialises the submodules of a git dependency recursively and has no option not to. The corpus now includes projects whose own nested submodules are private (skeletonlabs/skeleton-plus-premium) or SSH-only (git@github.com:basharovV/portable-vocal-remover.git), so depending on rsvelte_core as a git dependency fails at checkout on every commit since the corpus grew in July:

    Updating git submodule `git@github.com:basharovV/portable-vocal-remover.git`
    git@github.com: Permission denied (publickey).
    error: failed to get `rsvelte_core` as a dependency of package ...

#1861 fixed the same failure for one submodule by marking vize `update = none`, which Cargo honours. This applies that to all of them: nothing under submodules/ is needed to build the crates, and rsvelte_core's build.rs already tolerates the svelte submodule being absent. A git dependency now checks out only the repository itself, which also stops downstream users cloning all of sveltejs/svelte as a side effect.

`git submodule update --init <path>` skips submodules marked `update = none`, so every explicit checkout in the workflows, package scripts, helper scripts and docs gains `--checkout`, which restores the previous behaviour for them. CI already checks out submodules by explicit path everywhere (`submodules: false` plus a selective `git submodule update`), so no job changes what it checks out.


baseballyama:
Thank you so much!

baseballyama (approved):

PR #4581, opened 2026-09-13
fix(sourcemap): build the comment coordinate space above the source's last byte

### LINKED ISSUE #4521: loc_base is not an upper bound on source offsets, so a real source position can sit inside a comment-space range

baseballyama:
Closes #4521.

## What was wrong

`loc_base` is the boundary between real source offsets and the comment coordinate space, and it
was `synth.max_span + 2`. `max_span` is a running max over whatever the probe pass happened to
call `note_span` on — chunk text lengths, some statement ends, some `RawMapped` source offsets.
Nothing makes that an upper bound on the source offsets the printer will actually resolve. A
template element's identifier span is not among them, so on a component whose template reaches
past the largest noted span a genuine source offset sat **above** the boundary and was translated
as if it were a comment-space offset. The guard at `to_oxc.rs:202` cannot see it: it re-reads the
same under-counting `max_span`.

#4521 also explains why the obvious repairs do not close it. Adding the missing `note_span` calls
one at a time builds the list out of the offsets someone hit, not out of the offsets the printer
can resolve — and the issue measured a variant that raises `max_span` from region ends, which was
strictly better on the map and moved a component's comment output the **wrong** way (`2 → 0` where
official emits 2).

The quantity that *is* an upper bound is the source's length. It is threaded into
`program_to_oxc{,_with_islands}` and the boundary becomes `max(max_span, source_len) + 2`.

## The carrier

`submodules/svelte-eslint-parser/.../svelte5/ts-event03-type-output.svelte`, the component the
issue measured. All three segments whose generated position starts the identifier `button`:

| arm | source position | what is there |
|---|---|---|
| `origin/main` `7588fdc12` | `11:42` | `ventTarget &` — inside `HTMLButtonElement;`'s comment |
| this branch | `7:1` | `button oncli` — the tag name |
| official | `7:1` | the same three, same positions |

## Measurement

Arms: a pristine `git worktree --detach` at `origin/main` `7588fdc12` and this tree, distinct
`sha256`. Oracle `submodules/svelte` `7bc0a70fe64d` / `VERSION 5.57.0`.

Sweep over every `.svelte` file under `submodules/` — 36,552 units, 35,148 live (1,404 refused by
**both** arms with identical exit codes, 0 asymmetric):

**`js.code` moved on 1 file.** It is the same file, and the same count, that the issue's
`note_span` variant moved — but the direction is the other one: differing lines against official
`18 → 11`.

**`js.map` moved on 740 files.** Per file, on (identifier-correct, out-of-range):

```
better 307    worse 0    equal 433
```

On the 554 of those whose `js.code` is byte-identical to official's — where generated positions
line up and the two maps can be compared as **sets** rather than judged:

| | `origin/main` | this branch |
|---|---:|---:|
| official's segments reproduced (of 213,944) | 117,733 | **119,343** |
| segments emitted that official does not have | 27,220 | **26,287** |

Both directions improve, which a change that merely emitted more segments could not do.

`sourcemaps_gate` is green and `compatibility/sourcemap-known-failures.json` is untouched. All
three test shards green locally (1600 / 1802 / 1571, 0 failed).

## Controls

The unit test's positive control is its own neighbour:
`generated_identifier_source_span_does_not_raise_comment_boundary` runs the same program with
`source_len = 0` and still requires the boundary to stay *below* the identifier registered at
1,000, so the source length is the only thing that moves it.

Ablating `.max(source_len)` fails the unit test and `the_button_tag_maps_to_its_own_name`, and
leaves `the_input_tag_still_maps_to_its_own_name` green — `<input>` starts at a *larger* source
offset than `<button>` and was already mapped correctly under the old rule, so a change that
pointed every template identifier at one place would satisfy the first test and fail that one.
`the_extractor_is_live` keeps an empty `all()` from being what makes either pass.


baseballyama:
Closing this without merging. The change does what #4521 predicted on the map, and it pays the
price #4521 predicted on comments — and I now have one fact the issue did not have.

## What it fixes (unchanged from the description)

`origin/main` maps the carrier's `button` to `src 11:42`, inside a comment; this maps all three to
`src 7:1`, where the oracle puts them. Over the 554 corpus components whose `js.code` is
byte-identical to official's, official segments reproduced go 117,733 → 119,343 and segments
official does not have go 27,220 → 26,287.

## What it breaks

`cargo test -p rsvelte_core --lib` fails on exactly one test:

```
compiler::phases::phase3_transform::client::tests::a_script_comment_maps_before_the_template_expression_it_precedes
```

```svelte
<script>
  // "data" prop contains property "result"
  let { data } = $props();
</script>
{data.result}
```

`origin/main` is **byte-identical to the oracle** here — the comment sits inside `$.set_text(`,
on its own line before `$$props.data.result`. On this branch it lands after `$.pop()`, at the end
of the component body.

## The new fact: it is not the `has_loc` gate

`loc_base` answers two questions — "may this span carry comments" (`has_loc`) and "is this offset
in the comment buffer" (the `loc_map` lookup). The obvious repair is to split them: keep the gate
at the bound the probe measured and put the buffer above the source's last byte. I built that
(`comment_gate` on `Converted`, `print_split` taking it instead of `loc_base`) and re-measured:

| | buffer raised | buffer raised + gate decoupled |
|---|---:|---:|
| `js.code` MOVED over 35,148 live components | 1 | 1 |
| `js.map` MOVED | 740 | 740 |
| the `--lib` test above | fails | fails |

Byte-identical outcomes. So the comment placement does not depend on the gate — it depends on
**where the buffer sits**, because the comment cursor orders comment offsets against node spans in
one integer space, and lifting the buffer above the source makes every real source offset sort
before every comment. Splitting the number is not enough; the two spaces have to stop being one
ordered domain, which is the model change #4489 also wants.

## And a note on how this got to CI

The corpus sweep says `js.code` MOVED = **1** over 35,148 live components. That is a true number
and it is not a safety margin — the shape that carries the class occurs 0 times in the corpus and
lives in a unit test. `scripts/ci/run-test-shard.sh` partitions **integration** binaries only (its
own comment says the `--lib` tests run in the separate `test-unit` job), so three green local
shards structurally cannot see a `--lib` failure. I read the small corpus number as coverage; the
denominator that mattered was `--lib`.

Branch kept for the next attempt.

PR #4588, opened 2026-09-13
test(check): pin the #4583 epilogue guard end to end

### LINKED ISSUE #4583: svelte-check: false-positive `Cannot find name '$$ComponentProps'` for un-destructured, un-annotated `$props()`
### Summary

`@rsvelte/svelte-check` reports a false-positive `TS2304: Cannot find name '$$ComponentProps'` for any component that calls `$props()` **without destructuring and without a type annotation**. Official `svelte-check` reports no error for the same file.

The diagnostic is always pinned to `<lines_in_file + 1>:1`, which is the tell: it originates in generated overlay text that has no source-map coverage, and the greatest-lower-bound lookup pins it past the end of the author's source.

Version: `@rsvelte/svelte-check@0.5.28` (darwin-arm64), compared against `svelte-check@4.7.6` / `svelte2tsx@0.7.45` / `svelte@5.56.3`.

### Reproduction

`Repro.svelte`:

```svelte
<script lang="ts">
  const props = $props();
  void props;
</script>

<p>hi</p>
```

- `rsvelte-check` → 1 error: `Cannot find name '$$ComponentProps'` at `Repro.svelte:7:1` (the file is 6 lines long)
- `svelte-check` → 0 errors

Adding either destructuring (`const { a } = $props()`) or an annotation (`const props: { a: number } = $props()`) makes the error disappear.

In a real project this showed up as `Cannot find name '$$ComponentProps'` at `assessments/+page.svelte:190:1` (189-line file) and `tracking/+page.svelte:395:1` (394-line file).

### This is not a projection bug

My first guess was that `create_runes_props_str` emitted `$$ComponentProps` without emitting the accompanying `type $$ComponentProps = …`. That is **not** the case — rsvelte's overlay is bit-for-bit correct here. Upstream `svelte2tsx` emits exactly the same thing for this input:

```
///<reference types="svelte" />
;function $$render() {

  const props = $props();
  void props;
;
async () => {

 { svelteHTML.createElement("p", {});  }
};
return { props: {} as any as $$ComponentProps, exports: {}, bindings: __sveltets_$$bindings(''), slots: {}, events: {} }}
const Repro__SvelteComponent_ = __sveltets_2_fn_component($$render());
```

Note there is no `type $$ComponentProps` declaration in the upstream output either. So the reference genuinely is undeclared in both, and `tsc`/`tsgo` genuinely does report `TS2304` on both overlays.

### Root cause: the diagnostic filter, not the emission

Official `svelte-check` is clean because it **drops** this diagnostic. `MagicString` gives purely-inserted text a source-less segment, so `originalPositionFor` returns null, the mapped range comes back with a negative line, and `hasNoNegativeLines` discards it (`svelte-check@4.7.6`, `dist/src/index.js:122166`):

```js
function hasNoNegativeLines(diagnostic) {
    return diagnostic.range.start.line >= 0 && diagnostic.range.end.line >= 0;
}
```

rsvelte reimplements that filter as `EntryMap::is_inserted_text` in `crates/rsvelte_check/src/svelte_check/mapper.rs:55`, and its doc comment states the intent explicitly ("Official svelte-check drops it instead (`hasNoNegativeLines`) … We mirror that exactly").

Because rsvelte's map has no source-less segment to key off, the function infers an insertion from its *ends*: it requires a mapped segment on both sides of the gap, on the same generated line, pointing at the same source position.

```rust
fn is_inserted_text(&self, dst_line: u32, dst_col: u32) -> bool {
    let start = self
        .segments
        .partition_point(|s| (s.dst_line, s.dst_col) <= (dst_line, dst_col));
    let Some(prev) = start.checked_sub(1).map(|i| self.segments[i]) else {
        return false;
    };
    // Exactly on a segment: the position is mapped, gap or not.
    if prev.dst_line != dst_line || prev.dst_col == dst_col {
        return false;
    }
    let Some(next) = self.segments.get(start) else {
        return false;   // <-- trailing insertion falls through here
    };
    next.dst_line == dst_line && next.src.is_some() && next.src == prev.src
}
```

That two-sided requirement cannot hold for a **trailing** insertion, because there is no following segment. The `return { props: … }` epilogue is exactly such a trailing insertion — it sits on a generated line the source map does not cover at all. For the repro above:

```
epilogue at 0-based generated line 10, $$ComponentProps at col 29
segments on that line: undefined      (mappings has only 10 lines; generated code has 14)
```

So `is_inserted_text` returns `false`, the diagnostic is not filtered, and `map.lookup_token(10, 29)` greatest-lower-bound-matches the last mapped token — the end of the author's source — producing `<lines + 1>:1`.

### Suggested fix

In `is_inserted_text`, treat "no following segment on this generated line" as inserted text rather than as mapped. Equivalently (and closer to the upstream semantics it is mirroring): drop any diagnostic whose generated line falls beyond the last line the source map covers, since such a position is by construction generated-only.

The `#728` comment in that file rightly warns that an over-aggressive filter here turns real type errors into false negatives, so the narrow "generated line is past the end of the map" form seems safer than loosening the two-sided check in general.

Happy to open a PR with the fix plus a unit test alongside the existing `injected_kit_annotation_is_inserted_text_but_the_prop_is_not` case (`mapper.rs:691`) if that direction looks right to you.

polvallverdu:
Follow-up to #4584, which fixed #4583 (a diagnostic on the generated `$$render` epilogue was pinned past the end of the author's file — a 6-line component reported an error at `7:1`).

I opened this originally with a fix of my own; #4584 landed first and covers the reported bug, so I dropped my mapper change and kept only the test. **This PR is test-only — no behaviour change.**

## Why another test

#4584's unit test builds the `EntryMap` and looks the position up itself, so the map and the boundary are fixed by construction. That cannot notice the overlay's *layout* moving underneath it: an epilogue that starts carrying segments, or authored code that stops carrying them, shifts the boundary without touching `is_inserted_text`, and the unit test keeps passing while the user-visible bug returns.

This drives the real path instead — `materialize_overlay` → `map_tsgo_diagnostics` — and derives both positions from the emitted shadow:

- `a_diagnostic_on_the_generated_epilogue_is_dropped` — a `TS2304` at `$$ComponentProps` in the generated text produces no diagnostic.
- `a_diagnostic_on_authored_code_still_maps_into_the_component` — the same diagnostic at the authored `void props;` maps to the `.svelte` file, at a line within the component. A guard that swallowed the whole overlay would pass the first assertion and hide every real type error.

## Verification

Load-bearing: with #4584's trailing-insertion check disabled, `a_diagnostic_on_the_generated_epilogue_is_dropped` fails; the authored-code test still passes.

I also confirmed #4584 fixes the original report end to end on a minimal SvelteKit route (6-line component, `--tsgo`): `@rsvelte/svelte-check@0.5.28` reports `1 error` at `+page.svelte:7:1`, and `main` reports `0 errors` while keeping the genuine `state_referenced_locally` warning at `3:8` and a real `TS2322` at `4:9`.

While checking whether the guard should be broader than "past the last mapped line", I scanned the emitted maps of several component shapes (`$props()` with and without an interface, `$state`/`$derived`/`$effect`, `bind:`, `{#each}`/`{#if}`/`{#snippet}`, `<script module>`, `<svelte:window>`, `<slot>`). The only *interior* unmapped line carrying code was inside an `Ωignore` region, which `is_in_generated_code` already drops before the map is consulted — so the narrower trailing-only guard in #4584 is sufficient, and I'm not proposing to widen it.

`cargo fmt --all --check`, `cargo clippy -p rsvelte_check --all-targets` and the full `rsvelte_check` suite are clean on Rust 1.97.


baseballyama:
Thank you so much!

baseballyama (approved):

PR #4594, opened 2026-09-14
fix(parse): a binding pattern's TS annotation reaches the public AST

baseballyama:
`catch (e: unknown)`, `{#each xs as x: T}` and `{@const x: T = …}` all dropped
the annotation from the public AST. OXC keeps it *beside* the pattern —
`CatchParameter` and `VariableDeclarator` each carry their own
`type_annotation` — and three ports read only `&param.pattern` / `&decl.id`.

## Two readers, two node shapes

Upstream has two, and the port has to carry which one it is:

| upstream reader | used by | node |
|---|---|---|
| `read_pattern` → `read_type_annotation` (`1-parse/read/context.js`) | `{#each … as x: T}`, `{:then v: T}`, `{:catch e: T}`, `{@const x: T = …}` | built by hand: starts at the **pattern's end** (before the whitespace around the `:`), **no `loc`**, and only a destructured pattern takes its `end` |
| `read_declaration` | `{const x: T = …}` / `{let x: T = …}` declaration tags | comes from the parser, so it has its own span **and** a `loc` |
| acorn-typescript | a `catch` parameter inside a `<script>` | the annotation is on the pattern and the pattern's `end` covers it |

Attaching the hand-built shape at a declaration tag turns one divergence into
another — `typeAnnotation.loc` missing on
`print/samples/declaration-tag-division` — which is how the split was found;
the first arm looked right on the two `{@const}` carriers and only the
declaration-tag cell said otherwise.

## Measurement

Sweep over the whole corpus (34,087 components, official compiler from
`submodules/svelte` at `7bc0a70fe` / 5.57.0) for an `Identifier` official gives
a `typeAnnotation` and rsvelte does not:

```
before  read=34087 threw=480 carriers=138     178 CatchClause.param
                                                4 VariableDeclarator.id
                                                1 EachBlock.context
after   read=34087 threw=480 carriers=0
```

- `parse-ast` ratchet 165 → **163**, retiring exactly `legacy::` and
  `modern::Identifier.typeAnnotation#missing`; **0 new**, **0 re-clustered**.
- Corpus output gate: `✅ no regressions`, warnings/errors/parseability all
  identical (arm `660fcb141`, `dirty:false`).
- `-p rsvelte_core --lib --test parser_fixtures --test compiler_fixtures --test
  print --test ssr --test validator --test css --test sourcemaps_gate`, run
  `--no-fail-fast` — 8 `Running` lines, 0 failures.

The pattern-corpus repro was measured **with the file in `manifest.json`**
(component entries 34,087 → 34,088): the corpus population is that manifest, not
a directory walk, so copying a file into `compatibility/sources` leaves the
denominator — and the verdict — untouched.

Part of #4133.


baseballyama:
Two follow-ups pushed onto this PR.

**`4133-binding-pattern-type-annotation.svelte.md`** — the pattern-corpus docs guard requires a
sibling describing what a repro pins; it named the missing file, which is the only way to learn
the rule from outside CI.

**`upstream_issues/4133-svelte-each-expression-loc-keeps-the-swallowed-ts-assertion.md`** — a
second #4133 cluster, measured while the gate ran, and it is upstream's rather than ours. In a TS
component `{#each a ?? [] as x}` reaches acorn-typescript with the whole template visible, so
`[] as x` parses as a `TSAsExpression`; `tag.js` unwraps it and writes `expression.end = end`
without moving `expression.loc`.

The check is internal to official — `loc.end.column` against the column computed from that same
node's `end` offset — so rsvelte is not one of the arms:

```
corpus 34,087 components, official 5.57.0 (submodules/svelte @ 7bc0a70fe)
carriers 108 files / 139 each-expressions
official loc.end.column disagrees with its own end offset: 139
rsvelte  loc.end.column disagrees with its own end offset:   0
  135  LogicalExpression
    4  ConditionalExpression
```

A six-cell grid pins the mechanism rather than the symptom: the three cells that agree fail it in
three different ways — no `lang="ts"` (nothing for `as` to parse as an assertion), parentheses
(the expression ends before the `as`), and a bare identifier (no operand for `as <ctx>` to bind
into).

139 and the split 135/4 are exactly the four ratchet keys the attribution row claims
(`legacy::`/`modern::` x `LogicalExpression#span` 104 per axis, `ConditionalExpression#span` 4),
so the row is not covering a second mechanism underneath. `parse-ast` attribution goes 15 of 165
to 19 of 163.

PR #4609, opened 2026-09-14
fix(lsp): a hover on a .svelte module specifier is not generated code

baseballyama:
First of the two mechanisms in #4097 (does not close it — the component-tag-name
cell is a separate defect in `rsvelte_projection`, measured below).

## Mechanism, measured

`rewrite_plain_svelte_imports` appends `.tsx` to every `.svelte` module
specifier so tsgo resolves the shadow, and `TsgoOverlay` recorded those four
bytes in `generated_ranges` — the same list as the `Ωignore` regions.
`is_generated_range` rejects any range that **overlaps** that list, and
`map_response` turns a failed range into `Value::Null` for the whole response.

Probe on the issue's own repro (instrumentation reverted; the run is the
evidence, the grep was not):

```
before: specifier shadow range 2:34-2:54  generated=true   mapped=None
after : specifier shadow range 2:34-2:54  generated=false  mapped=1:35-1:51
```

`1:35–1:51` is exactly `'./Other.svelte'` in the source — the range official
answers with.

## Why the distinction is the right one

An `Ωignore` region is generated code with no source text behind it, so a range
inside it has nothing to map to. Four bytes inserted *inside* a source-backed
token are not that: `'./X.svelte.tsx'` has real source on both sides of them.
So the suffixes move to their own list, which
- still makes a **position** inside `.tsx` unmappable, and
- still rejects a **range** that lies entirely inside one,

while a range that merely spans one maps through the exact map as before. The
plain-`.ts` document path already behaved this way (it shifts offsets and never
rejects), which is why the issue's `'./other'` control passes on both servers.

## Controls

- New test `a_svelte_module_specifier_maps_back_without_the_appended_tsx`
  asserts the mapped range **and** the two negative cases (the suffix range is
  generated; a position inside it is generated).
- Positive control: widening the new containment test back to an overlap test
  makes it fail at the first assertion, then restored — green.
- `cargo test -p rsvelte_language_server`: 452 pass. The 5 failures in
  `upstream_fixture_manifest` are this worktree's un-populated
  `submodules/language-tools`, identical before the change.

---

## Rebase onto #4616 (2026-09-20)

Rebased onto `04ec5dc7e`; the branch was 51 commits behind. That surfaced a real conflict with
#4616 rather than the ratchet churn I expected:
`an_inlay_hint_viewport_starting_at_the_first_character_is_clamped_not_dropped` failed, because
a `0:0` viewport is dropped for every non-inlayHint method **only** by the guard at
`tsgo_response.rs:588`, and that guard read the same predicate `map_generated_range` reads.

Ablation, same worktree and target dir, one file swapped: with `main`'s `tsgo_overlay.rs` the
test passes (`1 passed`), with this branch's it fails — so the cause is this change, not
something upstream of it.

The two callers want opposite rules for the four appended bytes:

| caller | direction | question | rule |
|---|---|---|---|
| `map_generated_range` | shadow → source (a response) | is this range generated? | containment — `'./X.svelte.tsx'` is the source specifier plus four bytes, so **no** |
| `tsgo_response.rs:588` | source → shadow (a request) | does this range touch generated text? | intersection — an editor viewport running through those bytes does, so **yes** |

`is_generated_range` keeps the first meaning and `touches_generated_range` is the second, with
the request guard calling it. An `Ωignore` region answers both the same way, exactly as before.
The specifier range is the one case where the two answers differ, and the test pins both.

`cargo test --release -p rsvelte_language_server`: `--lib` 409/409. The one failing target,
`upstream_fixture_manifest`, reports `fixture root is missing: .../submodules/language-tools/...`
— a linked worktree initializes no submodules.


baseballyama:
## レビュー(コードオーナー視点)

**総合判定: 要修正**

### 設計
`.svelte` → `.svelte.tsx` の付加 4 バイトを `import_suffix_ranges` として `generated_ranges` から分離し、
`is_generated_range` のみ suffix について overlap → containment に緩和。
上流には `.tsx` 付加自体が無い(`module-loader.ts` 経由)ので、ミラーすべき上流アルゴリズムは存在せず、
**rsvelte 固有の設計判断**です。理屈自体は通っています。

### 【重大】影響範囲が hover に限定されていない
タイトルは hover ですが、`is_generated_range` は `tsgo_overlay.rs:853`(`map_generated_range` の先頭ガード)
から呼ばれ、`tsgo_response.rs:553,555` 経由で **全メソッドのレスポンス Range 写像**を通ります。
本文の scope をこの実態に合わせてください。

### 【重大】ラチェット未更新で CI レッド、かつ移動が双方向
実測 `46 NEW / 48 stale`。NEW の中身は hover ではなく `documentSymbol` と多数の `inlayHint` です。
特に `find-component-references-parent.svelte` は stale `extra=11/missing=7` → NEW `extra=13/missing=6` と
**双方向に移動**しており、単純な改善ではありません。
CLAUDE.md の「fix が over-rejection を止めたとき、その拒否が何を隠していたかを問う」に該当します。
再ベースラインの前に、この双方向移動が容認できるものかを判断してください。

### 【中】#4616 と再ベースラインが干渉する
NEW キーが #4616 と重複します(`outgoing-component.svelte|textDocument/inlayHint|:extra-rsvelte[count=1,hash=6712cb870407]` 等、ハッシュまで同一)。
2 本が同じ inlayHint 経路を独立に開通させているため、同時投入は不可。順番を決めて片方を rebase 後に再測定してください。

### 【軽微 / 要確認】
- 新しい containment 判定 `suffix.start <= start && end <= suffix.end` は `generated_ranges` 側の厳密 overlap と
  非対称で、幅 0 Range が suffix 側だけ「生成」判定になります。意図した非対称かご確認ください。
- 終端だけが suffix 内に落ちる Range(`'./X.svelte.ts'` 相当)は containment を外れて通過したあと
  `map_generated_position` が `None` を返し、レスポンス全体が null になります。
  tsgo が実際にその形を返すかは**未検証**です。

baseballyama:
Re-baselined from Corpus Compat run [35501368277](https://github.com/baseballyama/rsvelte/actions/runs/35501368277), dispatched on this branch at `4e9700c76`.

**Artifacts** — all 17 (16 corpus shards + fixtures), every one stamping `projectRevision: 4e9700c76`, `counts.transportTimeouts` 0 on all of them, `compared` 205,512–285,102 per shard (3,876,570 total; the varying column is the live control).

**Ratchet** — 23,790 → 23,784:

- 6 retirements: `upstream-features/invalid-import/input.svelte` and `upstream-testfiles/diagnostics/unresolvedimport2.svelte` `textDocument/diagnostic` `missing-rsvelte`, initial and edit phases, plus their two `expected:` siblings.
- 4 re-keys on `find-component-references-parent.svelte` `textDocument/documentSymbol` (extra 11 → 13, missing 7 → 6, both phases). A re-key is not a resolution.

**What the ratchet cannot see** — an `aggregate:` key carries no count or digest, so it cannot express a change of composition. The sidecar records 1,281 corpus entries whose mechanism set moved:

| | entries |
|---|---:|
| stopped carrying `rsvelte-empty` | 84 (54 hover, 30 definition) |
| `textDocument/completion` gaining text-edit labels | 1,193 |

The 84 are the fix on the real-world corpus: the response path now maps back a range that merely *touches* an import suffix. The completion movement is the other half of the same predicate split sitting on the request path; it is a change of which entries carry existing labels, not a new class — **225 mechanism labels before, 225 after, none added or removed**.

Docs: `KNOWN-FAILURES.md`'s declared count and both partitions re-derived from the JSON (`11897 + 11887`, `21792 + 1680 + 312`); `known-failures-md-check`, `attribution-check --gate-known` and `check:lsp-mechanisms` all pass.

PR #4615, opened 2026-09-14
fix(client): `{@html}`'s thunk carries the anchor its `{#key}` sibling reaches by recovery

### LINKED ISSUE #4481: Trailing script comment attaches to the wrong node in generated calls (svelte:window/body/document, {#key}, {@html})

baseballyama:
Closes #4481 for the `{@html}` carrier.

## What

`{@html expr}` lowers to `$.html(node, () => expr)`. Upstream builds that thunk
with `b.thunk`, so the **arrow** carries no `loc` while the expression it wraps
keeps the one it had in source. esrap runs an arrow's parameter sequence `until`
the body's start (`languages/ts/index.js`), so a comment still pending from the
instance script is claimed by the empty parameter list:

```js
// official                         // rsvelte before
$.html(node, (// x                  $.html(node, () => c);
) => c);                            …
                                    // x
```

rsvelte built the thunk with no anchor at all, so the located pass had no site
before the end of the component body and wrote the comment there.

`b::thunk_anchored` is `b::thunk` plus `set_expr_comment_anchor` on the body,
mirroring the existing `stmt_anchored` / `var_decl_anchored` / `const_decl_anchored`
family. `to_oxc::comment_anchor` turns the source offset into a comment-space
position, and only when the anchor really does follow the pending comment in
source order — so a `{@html}` that precedes the script's comments keeps none.

It runs `unthunk` **first** and anchors only if an arrow survives. The first
version built the arrow directly and skipped that: the corpus sweep scored
0 fixes / 35 regressions, every one of them `() => snippet()` where `b::thunk`
collapses to `snippet`. Nothing about comments — the builder had simply stopped
being `b::thunk`.

## Why the `{#key}` sibling already passed

`{#key}` builds the same `b::thunk` and also has no anchor, but it lands
correctly through `print_split`'s **recovery** pass — the second print, armed
only when the first pass *drops* a comment. Instrumented, `{#key}` prints four
arrows (two passes) and `{@html}` one (a single pass): a comment the first pass
**misplaces** rather than drops leaves the recovery disarmed. That asymmetry is
why the two cells of one issue behaved differently.

## Measurement

Fourteen-cell grid, each cell also compiled with the comment removed so a
divergence is attributed to the comment only when the comment-free pair is
byte-identical (oracle `submodules/svelte` @ `7bc0a70fe`, `VERSION 5.57.0`).
Differing lines vs. the oracle, before -> after:

| cell | before | after |
|---|---|---|
| `l_html` — `{@html c}` | **5** | **0** |
| `html_then_elem` — `{@html c}` then an element | **4** | **0** |
| `html_twice` — two `{@html}` in one component | **7** | **0** |
| `html_in_if` — `{@html}` inside `{#if}` | 0 | 0 |
| `c_key`, `c_key2`, `j_win`, `k_body`, `k_body2`, `m_doc`, `d_elem`, `b_await` | 0 | 0 |
| `l_html_svg` — 3-arg `$.html`, the root `var` claims it | 0 | 0 |

Three cells move DIFF -> EQ, none moves the other way. `html_in_if` is the
negative control that matters: the shape is present and it already agreed, so
"a `{@html}` is involved" is not the predicate — the anchor only fires where the
pending comment precedes the tag in source order.

`z4516` is in the directory as a neighbour and diverges on both arms
(`diffWith=13 / diffWithout=1`); it is #4516's carrier, unaffected here.

### Reach

Screen over the corpus (33,988 `.svelte` files): **452** use `{@html}`, and **2**
of those end their instance `<script>` with a comment. Neither of the two is a
live carrier — one is byte-equal to the oracle already and the other differs only
in the filename-derived CSS hash. Both marginals are live, so the zero is a
property of the population, not of the screen. That is why the sweep below shows
no fixes: the shape exists in the grid and not in the corpus, and the unit test
is the only possible guard.

### Corpus sweep

33,988 `.svelte` units per target x 4 targets, both arms built from `19f97ffc3`
(`corpus_hash` base `ca5403904d7ce0ac`, head `8e0bae6b3b47a17e`), row counts
equal on every leg (`rows=33988`, eight legs):

```
MOVED client = 0   client-dev = 0   server = 0   server-dev = 0
```

Zero bytes move anywhere in the corpus, which is what the reach screen predicted
and is the whole of the regression evidence. The *first* version of the builder
was measured the same way and scored MOVED 54 / 35 regressions, so the
instrument does report a moved set when there is one.

## Guard

`crates/rsvelte_core/tests/html_tag_comment_anchor_4481.rs`, 7 tests, every
expected string generated from the oracle. A `compatibility/pattern-corpus`
repro would be inert here: `verify.mjs` compares through `ast_equiv_batch` under
`CommentPolicy::Ignore`, so a comment-only divergence is byte-different,
AST-equivalent and scored a pass on both arms. Ablating the two source hunks and
keeping the test turns it red (recorded in the PR discussion).


baseballyama:
## Red control for the guard

Ablation: `b::thunk_anchored(&context.arena, html_expr, node.expression.start())`
in `html_tag.rs` replaced by `b::thunk(&context.arena, html_expr)` (the builder
left in place, so only the call site moves), same release profile, test file
untouched:

```
test result: FAILED. 2 passed; 5 failed
failures:
    the_comment_is_claimed_by_the_thunks_parameter_list
    it_is_not_deferred_past_the_statement
    a_block_comment_reaches_the_same_place
    a_compound_expression_lands_the_same_way
    the_dev_target_reaches_the_same_place
```

Tree restored, `git status --porcelain` empty afterwards.

The two that stay green under ablation are worth naming rather than counting as
guard: `no_pending_comment_leaves_the_call_unchanged` is the negative control and
is *supposed* to be insensitive, and `the_comment_survives_exactly_once` is an
occurrence count — a comment that moved is still present exactly once, so that
key cannot see this defect at all. Five tests carry the assertion; the other two
bound it.

Issue #4614, opened 2026-09-14
lsp: every corpus shard records transport timeouts, and the artifact a poisoned run leaves behind is indistinguishable from a clean one


baseballyama:
### What

Every `lsp-corpus` shard job records `counts.transportTimeouts > 0`, and
`verify.mjs:1113-1117` throws on that with

> A timeout is compared as a transport error, so it changes this run's keys and
> the next run would disagree; raise `--request-timeout-ms` rather than
> baselining the result

The artifact is written **before** the throw, on purpose, so a poisoned run
leaves behind a well-formed `lsp-current-corpus-N.json`. Anything that consumes
those artifacts — `merge-current.mjs --update-baseline` in particular — cannot
tell them from a clean run's. That is how #4391's re-baseline (PR #4597) was
built on one; the counts were only noticed after the fact.

### Measured

Dispatch run `34829458289` on `fix/4391-per-document-tsconfig`, all 16 corpus
shards + fixtures, `--concurrency 64`, `--request-timeout-ms` at its default
`180000`. `counts.transportTimeouts` per artifact:

| shard | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 8 | 9 | 11 | 12 | 13 | 14 | 15 | fixtures |
|---|--|--|--|--|--|--|--|--|--|--|--|--|--|--|--|
| timeouts | 62 | 64 | 64 | 64 | 64 | 60 | 64 | 80 | 64 | 64 | 80 | 64 | 64 | 57 | 0 |

915 over 14 shards (shards 7 and 10 not in the set I downloaded). The `fixtures`
suite, which does not use the corpus servers, is 0.

The PR run `34843334540` reproduces it on the shards that have finished so far
(1, 6, 9), each with 64.

### Why "slow runner" does not fit

The counts sit at or near the job's own `--concurrency 64`, and one shard in the
PR run finished with **zero**. A uniform slowdown would produce a spread
proportional to the shard's request count (205k-285k compared per shard); instead
14 shards land in 57-80. The shape that fits is *one* window of in-flight
requests being lost — a server that stops answering once per shard — not every
request being 180s slow. I have not identified which server or why, and the
artifact does not record how far past the deadline each request went, only that
it passed it.

### Why it matters beyond a red check

`compatibility/lsp-known-failures.json` is a two-sided ratchet. A key created by
a transport error does not reproduce, so the next clean run reports it as
"listed but now passing" and goes red — attributed to whatever change happens to
[...]

PR #4597, opened 2026-09-14
fix(lsp): a subdirectory's own tsconfig applies to the documents beneath it

baseballyama:
Closes #4391.

## The defect

The tsgo overlay built **one** project from the tsconfig at the workspace root.
`svelte-language-server` resolves a project **per document** — the nearest ancestor
`tsconfig.json`, then `jsconfig.json`, without leaving the workspace or crossing a
`node_modules` (`plugins/typescript/utils.ts:146-167`, keyed and created per config in
`service.ts:172-201`). So in any repository whose subdirectories carry their own config, every
option those configs declare was invisible.

## Both arms on #4391's repro

`<ws>/tsconfig.json` (no `paths`), `<ws>/docs/tsconfig.json` (`"@/*": ["./src/*"]`),
`<ws>/docs/src/{util.ts,Widget.svelte,App.svelte}`; `textDocument/hover` at each use site and
`textDocument/diagnostic`. Oracle: `submodules/language-tools` at `2cfcc15b`, built; tsgo
`7.0.0-dev.20260703.1`.

| import | official | rsvelte before | rsvelte after |
|---|---|---|---|
| `@/util` (`.ts`) | `(alias) const aliasTs: 1` | **`null`** | `(alias) const helper: 1` |
| `@/Widget.svelte` | `(alias) const aliasSvelte: 1` | **`null`** | `(alias) const fromWidget: 1` |
| `./Widget.svelte` (relative) | `(alias) const relSvelte: 1` | `(alias) const fromWidget: 1` | `(alias) const fromWidget: 1` |
| diagnostics | `[]` | **`1:35 [2307] Cannot find module '@/util'`** | `[]` |

The relative row is the live positive control: it holds in all three columns, so the server and
the overlay were working and only the aliases were unreachable. rsvelte names the *original*
export where official names the local alias — that residue is present on the relative control
too, in both columns, so it is a separate divergence and is not claimed here.

## The change

- `nearest_tsconfig` ports `findTsConfigPath`: nearest `tsconfig.json` then `jsconfig.json`,
  stopping at the workspace boundary and at a `node_modules`.
- `TsgoRuntime::ensure_overlay_for` builds the owning project **on demand** before one is picked,
  which is upstream's own laziness — a repository with hundreds of configs (huly has 498) pays
  only for the ones it opens.
- Nothing on the request path changes: `overlay_for_source` already picks the deepest workspace
  containing the document, so an overlay rooted at its config's directory is selected for free.
[...]

PR #4616, opened 2026-09-14
fix(lsp): an inlay-hint viewport that starts at `0:0` is clamped, not dropped

baseballyama:
Refs #4464 (§2 only — §1 is not finished; see the last section).

## What

`textDocument/inlayHint` carries the range the **editor is painting**, not a
position the user pointed at. Upstream treats it that way:
`convertToTargetTextSpan` (`InlayHintProvider.ts:104-113`) maps the two endpoints
independently and substitutes offset `0` for an unmappable start and the
snapshot's length for an unmappable end, so the span always exists.

rsvelte mapped the range as a unit through the generic request mapper, which

1. fails the whole request if either endpoint has no shadow position, and
2. rejects the mapped range when it intersects generated code.

An editor asks for the visible viewport, which starts at `0:0` — before the
script tag — so **(2)** fired on every real request and a `.svelte` file got no
inlay hints at all.

`TsgoResponseMapper` gains a `clamp_request_ranges` mode, set for
`textDocument/inlayHint` alone; in that mode `map_range`'s source→shadow arm
calls the new `TsgoOverlay::clamp_source_range` instead of the
map-then-filter pair.

## Measurement

Protocol probe, both servers over stdio, `TSGO_BIN` pinned to the
`@typescript/native-preview` under `submodules/language-tools`. Ten-line
component with `import Comp from './Comp.svelte'`, a `take(howMany: number)`
call as the positive control, and two `<Comp/>` tags. Hint counts, one endpoint
moved at a time:

| range | official | rsvelte before | rsvelte after |
|---|---:|---:|---:|
| `0:0 - 9:0` | 2 | **null** | 3 |
| `0:0 - 8:8` | 2 | **null** | 3 |
| `0:0 - 4:10` | 2 | **null** | 3 |
| `0:1 - 9:0` | 2 | 3 | 3 |
| `0:18 - 9:0` | 2 | 3 | 3 |
| `1:0 - 4:10` | 2 | 3 | 3 |

Three rows move from `null` to a list; no row that already answered changes.

**Both rsvelte columns are the same binary**, switched by an env flag on the new
mode — not two builds. That matters here: the first attempt used two worktrees at
the same commit as the two arms and they disagreed 5 vs 3 on rows the patch
cannot reach. The cause was not the patch. `rsvelte-language-server` resolves
`svelte` from beside itself when the document's workspace has none, and one
worktree had `node_modules/svelte@5.57.0` while the other had no `node_modules`
at all; only the resolving arm wrote `svelte-type-ref/` + `svelte-types.d.ts`
into the workspace, which adds two `options:` hints on the `<Comp/>` lines. What
named it was a `diff -rq` of the two arms' temp workspaces, and a third arm that
bypasses the new branch entirely and still answered 3.

## Guard

`an_inlay_hint_viewport_starting_at_the_first_character_is_clamped_not_dropped`
in `tsgo_response.rs`. It asserts both halves: the same range is still refused
for `textDocument/documentHighlight`, and accepted for `textDocument/inlayHint`;
the clamped range stays inside the shadow and still covers the body the viewport
asked about. Flipping the new mode to `false` turns it red (recorded in the PR
discussion).

## What this does not fix

The `3` column is still not `2`. The extra hint is the render function's own
return type at `0:1`, which upstream drops with
`inlayHint.position !== renderFunctionReturnTypeLocation` — one of the §1 filters
`filter_generated_inlay_hints` does not yet port. #4464 stays open for it.


baseballyama:
## Red control for the guard

Same tree, `clamp_request_ranges: true` → `false` at the single site that sets it
(the test file untouched):

```
test result: FAILED. 0 passed; 1 failed; 406 filtered out
```

and with it restored:

```
test result: ok. 406 passed; 0 failed; 1 ignored
```

The first version of this test used a component with no imports, and it failed on
its **own negative half**: `textDocument/documentHighlight` mapped that range
fine, so the cell it was built on could not distinguish "clamped" from "mapped
normally". Adding `import Comp from './Comp.svelte'` is what puts a prologue in
the shadow and makes `0:0` land in generated code. Worth stating because the
passing half would have looked identical either way — the assertion that the fix
is doing something only exists because the other method still refuses the range.

baseballyama:
## Blocked on the LSP re-baseline, not on the fix

`LSP fixture parity current` is red, and it is supposed to be: the ratchet is
two-sided and this changes `textDocument/inlayHint` responses.

Ran the gate locally at this commit (`fcefe80e5`, `counts.transportTimeouts: 0`,
`node scripts/compat-lsp/verify.mjs --suites fixtures,upstream-features,upstream-testfiles`):

```
423 cases; 1956/1972 compared, 1193 divergent requests / 1998 divergent fields
174 NEW divergence(s)
208 stale ratchet entry/entries
```

Both sides are **entirely** `textDocument/inlayHint` — 174/174 added and 208/208
removed — so the blast radius is the method the change is scoped to and nothing
else. Broken down:

| | n | verdicts | files |
|---|---:|---|---:|
| removed | 208 | `value-mismatch` 130, `extra-rsvelte` 78 | 104 |
| added | 174 | `extra-rsvelte` 152, `missing-rsvelte` 22 | 76 |

Every added key's file also has a removed key (**0 files newly enter the
ratchet**), and **28 files leave it entirely** — their inlay hints now agree with
official. The 130 `value-mismatch` entries are the `rsvelte=74234e98afe7` shape,
i.e. the digest of `null`: those are exactly the requests this fix stops
dropping. What they re-key to is §1's residue — the generated-hint filters
`filter_generated_inlay_hints` does not yet port, which #4464 stays open for.

So the re-baseline is a shrink, 208 → 174, and it cannot be taken here:
`mergeCurrentArtifacts` requires all **17** artifacts (1 fixture + 16 corpus
shards) at one `projectRevision`, and the corpus half is `lsp-corpus`, which runs
on dispatch only and currently records `transportTimeouts > 0` on nearly every
shard (#4614). Same wall as #4597.

Moving this to draft behind #4614 / #4617. The fix itself and its guard stand;
what is missing is a clean 17-artifact run to re-baseline from.

baseballyama:
## レビュー(コードオーナー視点)

**総合判定: 要修正**

### 移植の妥当性 — OK
上流 `InlayHintProvider.ts:104-113` の `convertToTargetTextSpan`
(`generatedStartOffset.line < 0 ? 0 : …` / `generatedEndOffset.line < 0 ? snapshot.getLength() : …`)
とアルゴリズム・境界の取り方とも一致。`clamp_source_range` が読む `entry.document.text` が
shadow テキストであることは `tsgo_overlay.rs:906-910` の `is_generated_range` が同じフィールドに
shadow 側 Range を当てていることから確認済みで、`getLength()` の対応として正しい。
上流にはリクエスト範囲の拒否そのものが無いので、clamp 経路で `is_generated_range` を飛ばすのも正しい。
適用を `textDocument/inlayHint` 一メソッドに限定している点も上流どおり。

### 【重大】ラチェット再ベースラインが無く CI レッド
実測 `96 NEW / 130 stale`。stale 側のハッシュは `official=4f53cda18c2b` = sha256(`"[]"`)、
`rsvelte=74234e98afe7` = sha256(`"null"`) — 本 PR が直す欠陥そのもので、**修正は効いている**。
ただし追加 96 クラスにはそれぞれ `lsp-mechanisms.json` の attribution と
`KNOWN-FAILURES.md` の正当化が必要で、`compatibility/` は未変更。

再ベースラインは **ジョブログから起こさないこと**。`verify.mjs` は NEW/stale 双方を `SHOW = 30`
で打ち切り、残数マーカーを出さない(#4484)。アップロードされた artifact の `added`/`removed` を使う。

### 【中】本文の機構帰属が、ツリーと矛盾する
> the `inlayHint.position !== renderFunctionReturnTypeLocation` filter … `filter_generated_inlay_hints` does not yet port

これは移植済み:
- `crates/rsvelte_language_server/src/tsgo_response.rs:742` — `!overlay.is_render_return_type_position(shadow_path, position)`
- `crates/rsvelte_language_server/src/tsgo_overlay.rs:881-894` — `is_render_return_type_position`
- `crates/rsvelte_language_server/src/server.rs:3200-3208` — inlayHint 時の配線

「presence is a grep; effectiveness is a run」に照らすと **存在するが効いていない** 別欠陥で、
本 PR が初めてこの経路を到達可能にして露出させた形。`render_return_type_offset`
(`tsgo_overlay.rs:1875`) が当該入力で `None` を返していないかを**実行で**確かめたうえで本文を訂正してください。

### 【中】「a `.svelte` file got no inlay hints at all」は母集団が過大
ゲートは `scripts/compat-lsp/suites.mjs:120-123` でまさに `0:0` 始まりを送るが、既存ラチェットには
rsvelte が答えている inlayHint エントリ(`component-handler … extra-rsvelte[count=14]`)が存在する。
母集団を限定した表現に。

### 【軽微】
- `map_request` 内の `Self { … }` 手組み再構築。現状フィールド 4 つで漏れは無いが、将来の追加フィールドが
  clamp 経路だけ黙って欠落する。`derive(Clone)` + `..self.clone()` か `clamp: bool` を下ろす形が安全。
- `tsgo_overlay.rs:1921-1927` の `ordered_range` は端点を入れ替えるが、上流は `end - start` をそのまま返す
  (負長あり)。挙動の逸脱だが実害は小。

### 他 PR との干渉
**#4609 と NEW キーが重複しています**(`outgoing-component.svelte|textDocument/inlayHint|:extra-rsvelte[count=1,hash=6712cb870407]` 等、ハッシュまで同一)。
2 本が同じ inlayHint 経路を独立に開通させているため、同時に再ベースラインすると互いを打ち消します。
片方を先に landing させ、もう片方はその後に rebase → 再測定してください。

PR #4662, opened 2026-09-20
fix(sourcemap): a segment's original column stays inside the line it names

### LINKED ISSUE #4610: The client and server source maps still point past the end of a source line on 87 corpus segments
### What

With the byte-vs-UTF-16 column mechanism closed (#4599 / #4600) and the corpus
arm of the source-map gate wired up (#4454), `compatibility/pattern-corpus` still
produces **87 source-map segments whose original position lies past the end of
its source line**, over 1,269 components the compiler accepts:

| target | segments | file/target pairs |
|---:|---:|---:|
| client | 77 | 62 |
| server | 10 | 7 |
| **total** | **87** | **69** |

263,879 segments are compared, so this is 0.03%. They are the residue of the
class the two issues above shrank: the same population read **6,016** as #4454
was filed, **137** after #4585, and **77** (client) after #4600.

These are the 69 entries of `compatibility/sourcemap-known-failures.json`, and
the reason `sourcemap-known-failures.json` is now on
`compatibility/attribution-pending.json`: no shared cause has been located, so
none of them can name a target yet.

### Shape of the residue, measured on the client arm

Bucketed by how far past the end of the line the position lands, at the tree
that landed #4600:

| bucket | n | example |
|---|---:|---|
| `ascii-off-by-one` (overshoot `+1`, all-ASCII line) | 18 | `adversarial/bindings/component-bind-to-member-paths.svelte` src 6:53, line `  let box = $state({ deep: { key: 1, list: [2] } });` |
| overflow `+3` | 16 | `adversarial/bindings/bind-computed-member.svelte` src 3:29 |
| overflow `+6` | 12 | `adversarial/events-actions-transitions/attachments-on-components.svelte` src 4:46 |
| overflow `+2` | 10 | `adversarial/bindings/bind-group-nested-each.svelte` src 4:48 |
| overflow `+4` | 9 | `adversarial/expressions/destructuring-declaration-zoo.svelte` src 16:19 |
| overflow `+5` | 5 | `adversarial/expressions/import-forms.svelte` src 2:42 |
| `+7` / `+8` / `+11` / `+14` / `+15` / `+18` | 1 each | two of them on `issues/2599-reactive-else-next-line.svelte` src 9, an **empty** line |

The two `2599` cells and the two `legacy/reactive-ordering.svelte` cells anchor on
a line with no characters at all, so any non-zero column there is out of range —
that sub-family is likely a different mechanism from the `+1` … `+6` ones and
should be split before either is attributed.

### Controls that already exist

- The official compiler emits **0 / 92,698** out-of-range segments on the same
  pattern corpus and **0 / 62,244** on the real-world sample, so the predicate is
  not measuring a property of Source Map v3.
- The upstream `packages/svelte/tests/sourcemaps` samples read **0 / 1,622**;
  they cannot see this and are not a counter-example.
- The server arm has its own 10, so this is not a client-only mechanism even
  though the two ports are separate implementations
  (`GATES.md#two-ports-inventory`).

### Done when

`sourcemap-known-failures.json` is empty again and
`sourcemap-known-failures.json` leaves `attribution-pending.json`, or each
surviving bucket names a target.

baseballyama:
Closes #4610.

## Mechanism

Both ports derive a source column by **arithmetic on a resolved one**, and neither re-checks which line the result landed on. An offset the driver resolves itself cannot escape its line (`offset - line_starts[line - 1]` is bounded by construction); a derived column can.

| port | site | why it overshoots |
|---|---|---|
| client (77 segments) | `rsvelte_esrap`'s keyword anchors — `ctx.location(line, column + keyword.len())` | the keyword is generated text; when it is longer than the source token it stands for, the end anchor runs off the end of the line the start anchor resolved to |
| server (10 segments) | `generate_token_mappings_inner`'s fast path | it read `text.is_ascii()` as "this token stays on one line". A string literal with a line continuation is ASCII **and** carries its own newline, so both derived columns named a position past the end of the line they claimed |

`is_ascii` answers how *wide* a character is, not whether a token stays on one line — that is the whole of the server bug, and the fast path now falls through to `advance_token_utf16` for a token containing a newline.

The client guard sits on `EventKind::Location`, the one event that carries a pre-resolved `(line, column)` pair; `EventKind::LocationOffset` resolves its own offset and is left alone.

## Measurement

Tree `296fb9461`, oracle `submodules/svelte` `7bc0a70fe` / VERSION 5.57.0, 1,390 `.svelte` files × 2 targets = 2,538 live compiles over `compatibility/pattern-corpus`.

| arm | out-of-range segments |
|---|---:|
| neither fix | 87 (client 77, server 10) |
| client guard only | 10 (all server) |
| both | **0** |

The two-step ladder is the live control: each fix owns exactly its family.

Segment census over the same 2,538 compiles, base vs. fixed:

| | base | fixed |
|---|---:|---:|
| rsvelte segments | 264,377 | 264,300 |
| agreeing with official | 121,035 | **121,036** |
| disagreeing | 12,910 | 12,880 |
| rsvelte-only | 96,044 | 96,023 |

All 77 removed segments were already wrong: none of the 87 was ever `official-identical`. Exact agreements went **up** by one, so no correct segment was dropped.

## `map-parity` carve-out

`attached-sourcemap` anchors `let` as `kind + ' '` where the source has a newline after `let`, so the official map itself names source column 5 on a 4-character line. `out-of-range` already declines to call such a position an rsvelte regression; `map-parity` demanding rsvelte *reproduce* one is the same claim from the other side, so `parity()` now counts those segments as `unreachable` and leaves them out of `bad()`.

Blast radius, measured: with **both fixes reverted** the gate reports 69 NEW `corpus-out-of-range` and **0** `map-parity`, so the carve-out hides nothing that exists without the client guard.

## Ratchet

`compatibility/sourcemap-known-failures.json` 69 → **0** entries — every arm (`anchor`, `map-parity`, `out-of-range`, `corpus-out-of-range`) is now empty — and the file leaves `compatibility/attribution-pending.json`.

## Tests

- `rsvelte_esrap`: `a_location_past_the_end_of_its_source_line_emits_no_mapping` — the line terminator stays addressable, one past it does not.
- `rsvelte_core`: `server_token_end_anchors_stay_inside_their_source_line` — a string literal with a line continuation anchors its end on the line it ends on.
- Both fail with their guard ablated (`if false && …`, `single_line || true`), verified.
- `sourcemaps_gate` green with the empty ratchet; the ten sourcemap suites plus `rsvelte_core --lib` (1,975 tests) pass.


baseballyama:
The first cut guarded the mapping event in `rsvelte_esrap::command`, which also refused the anchor upstream emits **on purpose**: `keyword_anchor_fidelity`'s `a_keyword_end_anchor_may_sit_past_the_end_of_its_source_line` and the `attached-sourcemap` fixture both pin `column + keyword.length` landing one past the terminator when the source spells the keyword's trailing separator as a newline (`import\n{a}`, `let\n\tx`). That is why `Test (ubuntu-latest, 2)` was red.

Moved the decision to the three sites that *derive* a column — the keyword cursor, `write_source_keyword`, `write_keyword` — where the keyword is still in hand, and quoted the line to separate the two cases:

- end anchor: kept when the line really holds `keyword.trim_end()` at that column, dropped when it does not;
- derived start column (`ce_col + 1`, the previous fragment's end): plain range check — an empty line holds no column but `0`, which is the `2599` / `reactive-ordering` pair the issue flagged as a probable separate sub-family.

Consequences:

- `map-parity` needs **no** carve-out any more, so `sourcemaps_gate.rs` is back to `main`'s version and this PR no longer touches the gate.
- Corpus out-of-range 87 → **0**; 77 segments removed, exact agreements 121,035 → **121,036** (no correct segment dropped).
- Ablating both predicates turns `sourcemaps_gate` red with 62 NEW entries and fails both `map_offset_past_source_4466` tests, so the pins are live.
- `map_offset_past_source_4466`'s recorded account of `3071`'s last carrier ("the exclusive end of the last node") was wrong on its own numbers — an exclusive end lands at column `len`, which that predicate allows, and the survivor sat at `len + 1`. Its corpus ceiling is now an equality.

PR #4666, opened 2026-09-21
fix(client): hoist imports written without a space after `import`

khromov:
Bun's TS transpiler prints side-effect imports as `import"pkg";`. The client import extractor only recognised `import ` and `import{`, so the statement stayed in the component function body and the output didn't parse. Same for `import*as ns from"pkg"` and semicolon-free `import{x}from"pkg"`.

Detect the declaration by the token after the keyword instead, which also stops `import ("pkg")` from being treated as a declaration.

### Where the shape comes from

Bun (tested on 1.4.2) drops the space when it transpiles a side-effect import, so any `<script lang="ts">` that is run through Bun before `compile()` ends up in this form:

```sh
$ bun -e 'console.log(new Bun.Transpiler({ loader: "ts" }).transformSync(`import "pkg";`))'
import"pkg";
```

Named, default and namespace imports keep their spaces; only the bare `import "pkg"` form is affected.


baseballyama (approved):
Thank you! This is a well-made fix — the defect, the shape it comes from, and the cells around it are all named, and the expected values in `tests/import_without_separator.rs` are the official compiler's own output rather than ours (I re-derived the client cell against `submodules/svelte` at `7bc0a70fe` / 5.57.0 and it matches byte for byte).

What I liked in particular:

- `starts_import_declaration` answers with the **token** after the keyword instead of a literal `"import "` / `"import{"` prefix, and `after_keyword` already rejects `imports` / `import$`, so the widening cannot swallow an identifier that merely starts with the keyword. `import(`, `import.meta` and an `import:` property key are each a listed negative.
- `following` is the raw tail rather than the next trimmed line, so a block comment opened after the keyword and closed on a later line is still read as trivia.
- The cells cover client / client-dev / server / `compileModule`, which is the sibling set that matters here — the same shape is a different code path in each.

I rebased it onto `main` myself (merge commit `b38ee445e`) rather than ask you to: `ImportLineScan::ends_at_specifier` lost its `line_len` argument in #4668 and grew a `last_string_end == last_code_end` condition, and `is_complete_side_effect_import` — which this PR deletes — had gained a comment-aware tail check there in the meantime. The resolution keeps your `starts_import_declaration` and calls `main`'s argument-free `ends_at_specifier()`; nothing else in the PR changed. Locally that is `rsvelte_core --lib` 1975 passed and `import_without_separator` 18/18, and CI is 48/48 green.

One thing left deliberately out of scope, just so it is written down somewhere: the same baked-in `"import "` needle still sits at `3_transform/mod.rs:1401` and `:1411`, `2_analyze/types.rs:447`, `shared/async_body.rs:2201` and `client/mod.rs:5045` (`js_ast/codegen.rs:2460` already accepts a tab). Your tests show the reported shapes do not reach those, so this PR is complete as it stands — they are a separate sweep.

